PR #13, opened 2025-08-12
fix: Remove unused dev dependency @esbuild/darwin-arm64

rocketpants8:
Works fine on Linux Mint. Sorry, don't have a Mac to test with to make sure there are no side-effects on that platform


gabrielste1n:
thanks @rocketpants8 ! Legend :)

PR #17, opened 2025-08-16
fix: Added Python venv implementation, issue #12

### LINKED ISSUE #12: Linux: Whisper Installation Failed
**Describe the bug**
During the initial setup process, after selecting "Local Processing", the installation of a local Whisper instance fails with the following error message:

```
Error occurred in handler for 'install-whisper': Error: error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.
```

**To Reproduce**
1. Clean Mint installation
2. Launch using `npm run dev`
3. Select "Local Processing"
4. Click "Install Whisper"

**Expected behavior**
Installation of Whisper works

**Screenshots**

Image

 - OS: Linux Mint 22.1

rocketpants8:
Hopefully this resolves #12 

Got this working locally by setting up Python venv.

Changes all done by Cursor; unable to test on other platforms so no idea if there will be any unintended side-effects of these changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
-  "name": "open-wispr",
-  "version": "1.0.2",
+  "name": "open-whispr",
+  "version": "1.0.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
-      "name": "open-wispr",
-      "version": "1.0.2",
+      "name": "open-whispr",
+      "version": "1.0.3",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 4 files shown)


gabrielste1n:
thanks @rocketpants8 - addressed this issue in #50 

Landed a smaller change that fixes the Mint/PEP‑668 failure by automatically falling back to pip install --user (and legacy resolver) when the system Python is locked down, so Local Whisper should now install successfully without new IPC plumbing or per-user venv management. Since the core issue is resolved upstream with less surface area to maintain, I’m going to close this PR but I really appreciate the investigation, and may revisit the venv idea later if we decide to ship an embedded Python.

Issue #25, opened 2025-09-05
linux: failed to enable Accessibility permissions (required for automatic text pasting)


hanpham32:
I installed OpenWhispr via the instructions from the README. However, I'm unable to enable the Accessibility Permission for the app to work. I tried to search around my setting for the app but because I installed the app via github it doesn't appear on my system as an "app" to grant permission. What's the work around this for Linux? Thank you

hanpham32:
The app doesn't function well on linux-- hotkeys don't work when running the app in the background.

gabrielste1n:
thanks @hanpham32 - closing this as it is related to #12 - we have an open pr to resolve this, should be merged soon.

vbuell:
I'd rather have `xdotool` to be installed automatically as a dependency of .deb, than searching the bug tracker/reading the documentation and installing it manually.

gabrielste1n:
thanks @vbuell good to know - i'll add that into the next release.

PR #26, opened 2025-09-06
doc: add installation instruction for debian ubuntu

hanpham32:
resolves issues #25


gabrielste1n:
Thanks for raising the gap in the docs! I just added a Debian/Ubuntu section to README that covers npm run pack:linux, installing the generated .deb directly via apt/dpkg, and the optional xdotool dependency plus Wayland caveats. @hanpham32

gabrielste1n:
@hanpham32 would really appreciate it if you could test the latest version v1.0.12 on your linux machine to see if it is working better :)

PR #65, opened 2026-01-02
Support pasting into Linux terminal apps + fix clipboard restore race condition

tobek:
1. ctrl+v to paste transcription doesn't work if you're in a terminal app - this update detects if a common terminal app has focus and, if so, does ctrl+shift+v. Only added support for Linux since that's what I use and what I'm able to test.
2. I've been encountering a race condition where the xdotool paste doesn't complete before the old clipboard content is restored, so old clipboard content is pasted instead. This increases the delay to avoid this.

Apologies if you don't prefer two unrelated things in one PR or if you don't want either of these for any reason, happy to update!

### DIFF
--- a/src/helpers/clipboard.js
+++ b/src/helpers/clipboard.js
@@ -268,7 +268,8 @@ class ClipboardManager {
 
           if (code === 0) {
             // Restore original clipboard after successful paste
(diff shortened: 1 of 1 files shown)


gabrielste1n:
thanks legend @tobek !! Merged this in, really appreciate the support 💪

PR #69, opened 2026-01-05
Fully support terminal focus detection and pasting on (KDE) Wayland

tobek:
Oops, my previous PR had support for ctrl+shift+v pasting on X11 and Wayland, but only actually supported detecting terminal focus on X11. This PR adds support for KDE Wayland terminal focus detection via `kdotool`.

Non-KDE Wayland support for this feature would require compositor-specific tools like `hyperctl` or `swaymsg` and I don't use these so can't test properly.


gabrielste1n:
ahh amazing @tobek was going to get to this a little later! Really appreciate it 🙏

PR #75, opened 2026-01-08
fix(macos): hide control panel to menu bar instead of dock

hammashamzah:
## Summary
- On macOS, closing the control panel now hides it completely instead of minimizing to Dock
- App icon is removed from Dock when control panel is hidden
- Users can reopen via the menu bar tray icon → "Open Control Panel"
- Quit option remains available in menu bar tray

## Test plan
- [ ] Run the app on macOS
- [ ] Close the control panel using the red traffic light button
- [ ] Verify the window hides and the app disappears from the Dock
- [ ] Click menu bar tray icon → "Open Control Panel"
- [ ] Verify control panel appears and Dock icon returns
- [ ] Test "Quit OpenWhispr" from tray menu

### DIFF
--- a/main.js
+++ b/main.js
@@ -218,16 +218,8 @@ app.on("activate", () => {
       windowManager.createControlPanelWindow();
     }
   } else {
-    // Show control panel when dock icon is clicked (most common user action)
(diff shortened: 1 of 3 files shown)


gabrielste1n:
@hammashamzah could you perhaps provide a screen recording of this happening to you? I am not able to reproduce. Will close this for now :)

PR #115, opened 2026-01-19
fix: use cross-env for Windows compatibility in pack script

epicexcelsior:
A minor update to the `pack` script in the `package.json` file to improve cross-platform compatibility.

* Scripts:
  * Updated the `pack` script to use `cross-env` when setting the `CSC_IDENTITY_AUTO_DISCOVERY` environment variable, ensuring compatibility across different operating systems. Just allows it to pack on Windows.


gabrielste1n:
thanks @epicexcelsior ! merged in :)

PR #123, opened 2026-01-20
feat: add dictation panel visibility modes

epicexcelsior:
## Summary

Adds configurable visibility modes for the dictation panel, allowing users to control when the panel is visible.

### Three Visibility Modes

- **Always Visible** (default) - Panel always shown on screen.
- **When Transcribing** - Panel appears when recording starts and hides after transcription completes. This is similar to SuperWhisper's default dictation panel behavior.
- **Always Hidden** - Panel never shows; dictation works entirely in the background.

### Settings Access

Users can change the visibility mode from three places:

- **Settings → General → Dictation Panel** dropdown
- **Dictation Panel Quick Menu** (right-click the panel icon)
- **System Tray → Panel Visibility** submenu

All three locations sync instantly via IPC broadcast.

## Technical Implementation

- **IPC Sync**: When any window changes the setting, it broadcasts to all windows for instant sync
- **Performance**: Cached visibility mode in WindowManager for fast hotkey response
- **Hidden Mode Protection**: showDictationPanel() and hotkey handlers check visibility mode before showing, preventing any flash

## Files Changed

| File                              | Change                               |
| --------------------------------- | ------------------------------------ |
| `preload.js`                      | Added sync API                       |
| `src/App.jsx`                     | Visibility effect and IPC listener   |
| `src/components/SettingsPage.tsx` | Dropdown in General settings         |
| `src/helpers/ipcHandlers.js`      | sync-panel-visibility-mode handler   |
| `src/helpers/tray.js`             | Panel Visibility submenu             |
| `src/helpers/windowManager.js`    | Mode checks and caching              |
| `src/hooks/useSettings.ts`        | PanelVisibilityMode type and setting |
| `README.md`                       | Documentation update                 |

## Testing

- [x] Settings dropdown syncs with tray and quick menu
- [x] Tray submenu syncs with settings and quick menu
- [x] Quick menu syncs with settings and tray
- [x] "Always Visible" mode keeps panel visible
- [x] "When Transcribing" mode shows panel on record, hides after
- [x] "Always Hidden" mode never shows panel (no flash)
- [x] Settings persist across app restarts

# Control Panel
image

# Tray Menu Settings
image

PR #124, opened 2026-01-20
feat: add dictation panel visibility modes and minimize-to-tray setting

epicexcelsior:
## Summary

Adds two new settings for controlling dictation panel behavior:

1. **Panel Visibility Mode** - Control when the dictation panel appears:
   - Always visible (default)
   - Show only when transcribing
   - Always hidden (background mode)

2. **Hide from Taskbar** (Windows-only) - Toggle to keep dictation panel only in system tray

## Changes

- Added panelVisibilityMode setting with IPC sync across windows
- Added minimizeToTray setting (Windows-only)
- Combined visibility mode checks with user-hidden flag in windowManager.js
- Consolidated both settings in one 'Dictation Panel' section in Settings UI

## Testing

- Build passes
- Verified on Windows - both settings work as expected
- Settings persist correctly across app restarts

## Screenshots

Settings UI shows both options in the Dictation Panel section under General settings.

### DIFF
--- a/src/components/SettingsPage.tsx
+++ b/src/components/SettingsPage.tsx
@@ -512,35 +512,6 @@ export default function SettingsPage({ activeSection = "general" }: SettingsPage
               />
             </div>
 
-            {/* Windows-only: Dictation Panel Behavior */}
-            {typeof navigator !== "undefined" && /Windows/i.test(navigator.userAgent) && (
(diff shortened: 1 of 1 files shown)


epicexcelsior:
Closing - will create separate PR for taskbar hiding feature only, since panel visibility is already in PR #123

PR #128, opened 2026-01-20
Feature - add gemini support to voice model

sahuadarsh0:
feat: Add support for multiple cloud transcription providers and models, including OpenAI, Gemini, and Groq.
added the flow in the onboarding

### DIFF
--- a/src/components/OnboardingFlow.tsx
+++ b/src/components/OnboardingFlow.tsx
@@ -18,6 +18,7 @@ import {
 } from "lucide-react";
 import TitleBar from "./TitleBar";
 import LocalWhisperPicker from "./LocalWhisperPicker";
+import TranscriptionModelPicker from "./TranscriptionModelPicker";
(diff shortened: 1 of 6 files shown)


sahuadarsh0:
image
Screenshot 2026-01-21 at 1 41 17 AM

gabrielste1n:
@sahuadarsh0 thanks so much for this - have you tested gemini for audio processing? I heard it is not very good - I like how you updated the onboarding flow, am in the middle of a redesign so will definitely borrow from your design!

sahuadarsh0:
Yeah, it isn't that great, but I added it because I had Gemini use cases and API keys.
We can implement this as a separate branch for any user who wants to add that feature.

gabrielste1n:
@sahuadarsh0 going to close this for now as I am in the middle of a redesign - but will add this in once settled :) really appreciate the contribution!

PR #130, opened 2026-01-21
fix: disable Push to Talk on Windows/Linux

tagraham:
Push to Talk mode requires detecting key-up events, which is only supported on macOS via the Globe key listener. On Windows/Linux, globalShortcut only fires on key-down, making Push to Talk non-functional.

Changes:
- Disable and gray out Push to Talk option on non-macOS platforms
- Show "macOS only" label instead of "Hold to record"
- Add tooltip explaining the limitation
- Force tap mode in useSettings if push was previously saved on non-macOS

### DIFF
--- a/src/components/ui/ActivationModeSelector.tsx
+++ b/src/components/ui/ActivationModeSelector.tsx
@@ -8,11 +8,17 @@ interface ActivationModeSelectorProps {
   disabled?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


gabrielste1n:
agreed thanks @tagraham was hoping i would find a workaround - will review pr asap

tagraham:
Glad to help.

gabrielste1n:
@tagraham should resolve this issue in #150 so will close this PR - thank you for your contribution!

PR #166, opened 2026-01-29
Fix custom provider selection not persisting (#148)

### LINKED ISSUE #148: AI Text Enhancement - Custom endpoint is not used/persisted
**Describe the bug**
If I add a custom endpoint for AI Text Enhancement and select a model which has `openai` or other matching provider name (like `openai/gpt-5-mini`) - then the feature doesn't work.

**To Reproduce**
1. Go to 'Settings'
2. Click on 'AI Text Cleanup'
3. Select custom tab
4. Add custom api path and key, select model like `openai/gpt-5-mini`
5. Close settings
6. Re-open settings
7. Openai provider will be selected rather than custom provider.

**Expected behavior**
I expected custom provider to be used even if model name matches known providers.

**Screenshots**

 - OS: Mac OS
 - Version 1.2.16 (1.2.16)

**Additional context**

Ashesh3:
When selecting a custom AI provider, the selection wasn't persisted to localStorage. Upon reopening settings, the provider tab would revert to the computed provider based on the model ID instead of the user's explicit selection.

Wrap the localReasoningProvider state setter to persist the provider choice to localStorage, ensuring the Custom tab remains selected after settings are closed and reopened.

Fixes #148

### DIFF
--- a/src/components/SettingsPage.tsx
+++ b/src/components/SettingsPage.tsx
@@ -139,9 +139,15 @@ export default function SettingsPage({ activeSection = "general" }: SettingsPage
     showAlert: showAlertDialog,
   });
 
-  const [localReasoningProvider, setLocalReasoningProvider] = useState(() => {
+  const [localReasoningProvider, setLocalReasoningProviderState] = useState(() => {
     return localStorage.getItem("reasoningProvider") || reasoningProvider;
   });
+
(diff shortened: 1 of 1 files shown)


gabrielste1n:
thank you @Ashesh3 ! Released a fix for this in v1.3.4 - really appreciate this contribution :) 🙏

PR #168, opened 2026-01-30
fix(macos): use key code instead of keystroke for paste

f1nkel:
## Summary

- Use `key code 9` (physical V key) instead of `keystroke "v"` for Cmd+V simulation on macOS
- Fixes paste not working when using non-English keyboard layouts (Russian, Ukrainian, etc.)

## Problem

The `keystroke "v"` AppleScript command sends the **character** "v", which doesn't exist on Cyrillic keyboard layouts. This caused unexpected behavior — instead of pasting, users experienced text selection (like Cmd+A) or other random actions.

## Solution

Using `key code 9` ensures the **physical V key** is pressed regardless of the current keyboard layout. Key code 9 corresponds to the V key position on QWERTY keyboards.

## Testing

Tested on macOS with:
- English (US) layout ✅
- Russian layout ✅

### DIFF
--- a/src/helpers/clipboard.js
+++ b/src/helpers/clipboard.js
@@ -149,32 +149,31 @@ class ClipboardManager {
     const platform = process.platform;
     let method = "unknown";
 
+    debugLogger.info("pasteText called", { textLength: text.length, textPreview: text.substring(0, 50) }, "clipboard");
+
     try {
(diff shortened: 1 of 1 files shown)


gabrielste1n:
thanks @f1nkel implemented this in v1.4.8

PR #177, opened 2026-02-01
fix: improve clipboard paste on Wayland with ydotool and terminal detection

ccross2:
## Summary

Fixes clipboard pasting on Wayland compositors (especially Unity) where wtype fails with "Compositor does not support the virtual keyboard protocol".

## Changes

1. **ydotool 0.1.x compatibility**: Use key name syntax (`ctrl+v`) instead of scan codes (`29:1 47:1 47:0 29:0`). The key name syntax works on both ydotool 0.1.x (Ubuntu packages) and 1.x versions.

2. **Prioritize ydotool on Wayland**: When no target window ID is available, prefer ydotool over xdotool. xdotool reports success but can't reliably paste without window targeting on Wayland.

3. **Improved terminal detection**: Add fallbacks for detecting terminal emulators on Wayland where window class is unavailable:
   - Check window title for terminal patterns
   - Check process name of the window (most reliable)
   
   This ensures terminals receive `Ctrl+Shift+V` instead of `Ctrl+V`.

## Problem

On Wayland with Unity compositor:
- `wtype` fails: "Compositor does not support the virtual keyboard protocol"
- `xdotool` reports success but doesn't actually paste when `targetWindowId` is null
- Terminal detection fails because `xdotoolWindowClass` is null on Wayland
- ydotool 0.1.x (Ubuntu) uses different syntax than 1.x

## Testing

Tested on Ubuntu 24.04 with:
- Unity/Wayland compositor
- wezterm terminal
- ydotool 0.1.8 (Ubuntu package)

## Test plan

- [ ] Test paste in regular apps on Wayland
- [ ] Test paste in terminal emulators on Wayland  
- [ ] Test with ydotool 0.1.x (Ubuntu)
- [ ] Test with ydotool 1.x (if available)
- [ ] Verify X11 behavior is unchanged

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

### DIFF
--- a/src/helpers/clipboard.js
+++ b/src/helpers/clipboard.js
@@ -497,8 +497,8 @@ class ClipboardManager {
     // Detect if the focused window is a terminal emulator
     // Terminals use Ctrl+Shift+V for paste (since Ctrl+V/C are used for process control)
     const isTerminal = () => {
-      // Common terminal emulator class names
-      const terminalClasses = [
+      // Common terminal emulator class names and title patterns
+      const terminalPatterns = [
         "konsole",
         "gnome-terminal",
         "terminal",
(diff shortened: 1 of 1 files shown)


gabrielste1n:
@ccross2 thanks for this - addressed in #262 - would love for you to test it out

PR #181, opened 2026-02-02
fix: hotkey not persisting across restarts

iTubeGamer:
# PR: Fix hotkey not persisting across restarts

## Problem
The dictation hotkey doesn't persist reliably across application restarts. Users have to reconfigure their hotkey each time they start the app.

## Root Cause
The hotkey is stored only in `localStorage`, which can have timing issues:
- The main process reads localStorage via `executeJavaScript` after a 1-second delay
- In production builds with `file://` protocol, localStorage timing can be unreliable
- If the renderer hasn't fully initialized, the saved hotkey may not be found

## Solution
Add file-based persistence using the existing `EnvironmentManager` pattern:

1. **Primary storage**: Save hotkey to `.env` file via `DICTATION_KEY` environment variable
2. **Fallback**: Still check localStorage for backwards compatibility
3. **Migration**: Automatically migrate existing localStorage hotkey to env on first load

## Changes
- `src/helpers/environment.js`: Added `getDictationKey()` and `saveDictationKey()` methods
- `src/helpers/hotkeyManager.js`: Check env var first, then localStorage, with auto-migration
- `src/helpers/ipcHandlers.js`: Added IPC handlers for get/save dictation key
- `preload.js`: Exposed new API methods to renderer
- `src/hooks/useSettings.ts`: Also save to file-based storage when hotkey changes
- `src/types/electron.ts`: Added TypeScript types for new API methods

## Testing
- Tested on Windows 11
- Set hotkey → Restart app → Hotkey is correctly restored
- Works in both development and should work in production builds

## Backwards Compatibility
- Existing localStorage hotkeys are automatically migrated on first startup
- Both storage methods are updated when hotkey changes
- No breaking changes to existing configurations

### DIFF
--- a/preload.js
+++ b/preload.js
@@ -7,7 +7,7 @@ const { contextBridge, ipcRenderer } = require("electron");
 const registerListener = (channel, handlerFactory) => {
   return (callback) => {
     if (typeof callback !== "function") {
-      return () => {};
+      return () => { };
     }
 
     const listener =
@@ -166,7 +166,7 @@ contextBridge.exposeInMainWorld("electronAPI", {
 
   // External link opener
   openExternal: (url) => ipcRenderer.invoke("open-external", url),
-  
+
   // Model management functions
   modelGetAll: () => ipcRenderer.invoke("model-get-all"),
   modelCheck: (modelId) => ipcRenderer.invoke("model-check", modelId),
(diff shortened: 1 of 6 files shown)


gabrielste1n:
this is a great PR thanks @iTubeGamer - merged!

PR #202, opened 2026-02-06
fix(linux): prefer ydotool over xdotool on GNOME Wayland

wrt54gl:
## Summary

- On GNOME Wayland, reorder paste tool candidates to try **ydotool before xdotool**
- Use **AT-SPI2 accessibility API** for terminal detection on GNOME Wayland, since xdotool always returns OpenWhispr's own window
- Send `Ctrl+Shift+V` for detected terminals, `Ctrl+V` for everything else

## Problem

On GNOME Wayland, xdotool can only interact with XWayland windows. When OpenWhispr (an Electron/XWayland app) attempts to paste into a native Wayland window (e.g. gnome-terminal, GNOME Text Editor), xdotool targets OpenWhispr's own XWayland window instead, **silently reports success** (exit code 0), and prevents fallback to ydotool.

Additionally, terminal detection via `xdotool getwindowclassname` fails on GNOME Wayland because xdotool always returns OpenWhispr's own window class — not the actual focused window. This causes the wrong paste shortcut to be sent.

### 1. Prefer ydotool on GNOME Wayland
On GNOME Wayland (`isGnome && isWayland`), ydotool is now tried **before** xdotool in the candidates array. ydotool works at the kernel input level and sends keystrokes to whatever window has compositor focus, regardless of whether it's a native Wayland or XWayland window.

### 2. AT-SPI2 terminal detection
Since xdotool can't reliably detect the focused window on GNOME Wayland, terminal detection now uses the **AT-SPI2 accessibility API** via a lightweight Python script (~88ms). AT-SPI2 can see both native Wayland and XWayland windows and correctly identifies the active application (e.g. `org.gnome.Terminal`).

The detection flow on GNOME Wayland:
1. Query AT-SPI2 for the active application name (excluding `gnome-shell`)
2. Match against known terminal class names
3. If terminal → ydotool sends `Ctrl+Shift+V`; otherwise → ydotool sends `Ctrl+V`

Falls back to xdotool/kdotool detection if AT-SPI2 is unavailable.

### 3. Correct paste shortcuts
- **Terminals**: `Ctrl+Shift+V` (the standard terminal paste shortcut)
- **Other apps** (GNOME Text Editor, Chrome, etc.): `Ctrl+V`

## Dependencies

- `python3` with `gi` (PyGObject) and AT-SPI2 bindings — standard on GNOME desktop installations
- `ydotool` + `ydotoold` daemon for keystroke injection

## Test plan

- [x] Test paste into gnome-terminal on GNOME Wayland
- [x] Test paste into GNOME Text Editor (native Wayland GTK4 app) on GNOME Wayland
- [x] Test paste into Chrome on GNOME Wayland
- [ ] Verify no regression on X11 sessions (xdotool still preferred)
- [ ] Verify no regression on non-GNOME Wayland (e.g. KDE, wtype still preferred)
- [ ] Verify graceful fallback when python3/AT-SPI2 is unavailable

Tested on Debian 13 (trixie) with GNOME Shell 48.7, NVIDIA RTX A400 (driver 550.x), Wayland session.

### DIFF
--- a/src/helpers/clipboard.js
+++ b/src/helpers/clipboard.js
@@ -651,6 +651,48 @@ class ClipboardManager {
         "yakuake",
       ];
 
+      // On GNOME Wayland, xdotool can't see native Wayland windows and returns
+      // OpenWhispr's own window instead. Use AT-SPI2 accessibility API first,
+      // which works for both native Wayland and XWayland windows.
+      if (isGnome && isWayland) {
+        try {
+          const atspiScript = `import gi
+gi.require_version('Atspi','2.0')
+from gi.repository import Atspi
+d=Atspi.get_desktop(0)
+skip={'gnome-shell',''}
+for i in range(d.get_child_count()):
+ a=d.get_child_at_index(i)
+ if not a:continue
+ n=a.get_name()
+ if n.lower() in skip:continue
+ for j in range(a.get_child_count()):
+  w=a.get_child_at_index(j)
+  if w and w.get_state_set().contains(Atspi.StateType.ACTIVE):
+   print(n.lower());raise SystemExit(0)`;
+          const result = spawnSync("python3", ["-c", atspiScript], {
+            timeout: 1000,
+            stdio: ["pipe", "pipe", "pipe"],
+          });
+          if (result.status === 0) {
+            const appName = result.stdout.toString().trim();
+            if (appName) {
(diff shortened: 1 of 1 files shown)


gabrielste1n:
thanks for this @wrt54gl - I have tackled this and implemented it in #262 - would love for you to test out the latest build to see if it works for you :)

PR #206, opened 2026-02-07
Fix README instructions - tried running locally for the first time and ran into these issues

TheCoreMan:
1. wrong path of example env file. 
2. missing build command, leading to start error (Failed to start app: Error: HTML file not found: /Users/shaynehmad/Desktop/code/openwhispr/src/dist/index.html). Works when running dev without building, but not when running `npm start`. Since both do both anyway, better to add the build command first.

Before:

image

After:

works

image


gabrielste1n:
thank you sir ! @TheCoreMan mergeing in a second

PR #221, opened 2026-02-08
fix(linux): persistent clipboard ownership on Wayland — Ctrl+V works after transcription

injaamam:
## Problem

On Wayland, after speech recognition finishes, OpenWhispr cannot automatically paste text into the focused text box due to Wayland security restrictions, which prevent applications from injecting input into other clients. Therefore, OpenWhispr relies on the system clipboard.

However, after speech recognition finishes, OpenWhispr copies the transcription to the system clipboard but **releases clipboard ownership** immediately. On Wayland the compositor does not store clipboard content; it only holds a reference to the owning client's `wl_data_source`. Once the owner releases the source, the clipboard has no active owner.

**Result:** Ctrl+V fails in all applications. Clipboard history (e.g. Klipper) may show the text, but paste only works after manually re-selecting the item.

## Root cause

Clipboard writes were done in the **main process** via Electron's `clipboard.writeText()`. After the copy function (and IPC handler) completed, the clipboard source was dropped. This matches how Wayland works: the owner must keep the source alive.

## Solution

On **Linux when running under Wayland**, clipboard writes are now performed in the **renderer** (the visible window) via `webContents.executeJavaScript('navigator.clipboard.writeText(...)')`. The renderer's WebContents then owns the Wayland clipboard offer for the window's lifetime, so:

- Ctrl+V works in all apps after transcription
- Ownership is released only when a new transcription is copied, or the app exits
- No reliance on clipboard history or external helpers

## Changes

- **`src/helpers/clipboard.js`**
  - Added `writeClipboardInRenderer(webContents, text)` (safe `JSON.stringify` for injection).
  - `pasteText(text, options)`: accepts `options.webContents`; on Linux + Wayland + valid webContents, writes via renderer instead of main process; passes options to `pasteLinux` for restore.
  - `pasteLinux(originalClipboard, options)`: after successful paste, restores previous clipboard via renderer on Wayland when webContents is provided.
  - `writeClipboard(text, webContents)`: optional second argument; on Linux + Wayland + valid webContents, writes via renderer.
  - Optional: `lastOwnedClipboardText` and clear on `app.on('will-quit')`.
- **`src/helpers/ipcHandlers.js`**
  - `paste-text`: passes `event.sender` as `webContents` in options.
  - `write-clipboard`: passes `event.sender` as second argument.

## Platform impact

| Platform        | Effect |
|----------------|--------|
| macOS          | No change (main process clipboard only). |
| Windows        | No change (main process clipboard only). |
| Linux (X11)    | No change (main process clipboard only). |
| Linux (Wayland)| **Fixed:** renderer owns clipboard; Ctrl+V works. |

## Testing

- **Wayland:** Run under a Wayland session (e.g. GNOME on Wayland). After a transcription, confirm the text is in the clipboard and Ctrl+V works in another app (browser, terminal, editor) without re-selecting in Klipper.
- **X11 / macOS / Windows:** Confirm existing copy + paste simulation and clipboard restore still work as before.

### DIFF
--- a/src/helpers/clipboard.js
+++ b/src/helpers/clipboard.js
@@ -1,4 +1,4 @@
-const { clipboard, app } = require("electron");
+const { clipboard } = require("electron");
 const { spawn, spawnSync } = require("child_process");
 const { killProcess } = require("../utils/process");
 const path = require("path");
@@ -69,10 +69,6 @@ class ClipboardManager {
     this.nircmdChecked = false;
     this.fastPastePath = null;
     this.fastPasteChecked = false;
-    this.lastOwnedClipboardText = null;
-    app.on("will-quit", () => {
-      this.lastOwnedClipboardText = null;
-    });
   }
 
   _isWaylandWithRenderer() {
@@ -242,7 +238,6 @@ class ClipboardManager {
         clipboard.writeText(text);
         this.safeLog("📋 Text copied to clipboard:", text.substring(0, 50) + "...");
       }
-      this.lastOwnedClipboardText = text;
 
       if (platform === "darwin") {
         method = this.resolveFastPasteBinary() ? "cgevent" : "applescript";
@@ -1115,7 +1110,6 @@ Would you like to open System Settings now?`;
     } else {
       clipboard.writeText(text);
     }
-    this.lastOwnedClipboardText = text;
     return { success: true };
   }
 


injaamam:
Built and tested on [Kubuntu 25.04] Wayland. Ctrl+V works after transcription; clipboard behavior is correct.

gabrielste1n:
awesome work thannks @injaamam ! mergeing :) just did some cleanups

wrt54gl:
Hey, thanks for the `wl-copy` clipboard persistence fix — that's a real improvement for Wayland users.

I wanted to flag a regression this introduces on **GNOME Wayland** for terminal pasting. The paste tool candidate list was simplified to always prefer `xdotool` over `ydotool`:

```js
...(canUseXdotool ? [{ cmd: "xdotool", args: xdotoolArgs }] : []),
...(canUseYdotool ? [{ cmd: "ydotool", args: ydotoolArgs }] : []),
```

On GNOME Wayland this causes two issues:

1. **`xdotool` can't see native Wayland windows** — it only interacts with XWayland windows, so it often targets the wrong window (e.g. OpenWhispr itself) while reporting success, which prevents fallback to `ydotool`.

2. **Terminal detection fails** — `xdotool getactivewindow getwindowclassname` returns OpenWhispr's own window class on GNOME Wayland, so the code can't distinguish terminals from other apps. Terminals then get `Ctrl+V` instead of `Ctrl+Shift+V`, which silently fails in most terminal emulators.

My open PR #202 addresses both of these by:
- Using AT-SPI2 (`python3 + gi.repository.Atspi`) for active window detection on GNOME Wayland, which works for both native Wayland and XWayland windows
- Preferring `ydotool` over `xdotool` on GNOME Wayland since `ydotool` injects at the kernel level and doesn't have the XWayland limitation

Both fixes are compatible with the `wl-copy` changes in this PR. Happy to rebase #202 if that helps.

PR #223, opened 2026-02-08
fix(hotkey): guard previous hotkey unregister to prevent invalid accelerator crash

### LINKED ISSUE #220: Hotkey doesn't work in latest release v1.4.2
**Describe the bug**
Hotkey doesn't work in latest release v1.4.2

**To Reproduce**
1. Start OpenWhispr and try to use your set hotkey to start the recording
2. Hotkey is not recognized, executes as just pressing the button (mine was set on ` / ~ key, the default option, I suppose)
3. Try to change hotkey
4. See this about ANY hotkey combination option

Image

**Expected behavior**
Expected it to just start recording, like I did in the previous versions.

 - OS: Windows 11
 - Version v1.4.2

**Additional context**

Image

Was working in just previous version, v1.4.1

Additionally, I was thinking of allowing to assign any of the super keys alone (CTRL, ALT) to hold or start\stop the recording. 
When I record voice, I don't use any of the keyboard, don't try to use any of other hotkeys, so it won't conflict with anything. So assigning it to something like CTRL or ALT which by default do nothing if pressed alone - sounds like a good place for such hotkey. I do need `/~ key sometimes, so it's inconvenient that it's occupied by this hotkey currently.
I wonder if that's really possible?

RutamBhagat:
## Summary
- wrap previous hotkey `globalShortcut.unregister(...)` call in `try/catch`
- skip unregister when stored previous hotkey is not a valid Electron accelerator
- keep hotkey update flow running so users can rebind successfully

## Why
Issue #220 reports hotkey rebinding failing with "The accelerator string can only contain one key code". This minimal change prevents the unregister path from aborting the update flow when an invalid previous key string is encountered.

Fixes #220

https://github.com/user-attachments/assets/321ae040-5a34-4a14-b68c-c9401ba2e202

### DIFF
--- a/src/helpers/hotkeyManager.js
+++ b/src/helpers/hotkeyManager.js
@@ -114,8 +114,14 @@ class HotkeyManager {
       const prevAccelerator = this.currentHotkey.startsWith("Fn+")
         ? this.currentHotkey.slice(3)
         : this.currentHotkey;
-      debugLogger.log(`[HotkeyManager] Unregistering previous hotkey: "${prevAccelerator}"`);
-      globalShortcut.unregister(prevAccelerator);
+      try {
+        debugLogger.log(`[HotkeyManager] Unregistering previous hotkey: "${prevAccelerator}"`);
+        globalShortcut.unregister(prevAccelerator);
+      } catch (error) {
+        debugLogger.warn(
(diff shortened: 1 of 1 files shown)


gabrielste1n:
thank you @RutamBhagat !

PR #228, opened 2026-02-08
fix(hotkeys): use Control+Shift+Space as non-mac default

### LINKED ISSUE #220: Hotkey doesn't work in latest release v1.4.2

RutamBhagat:
## Summary
- change non-mac default hotkey from "Ctrl+super" to "Ctrl+shift+space" 
- update startup and fallback defaults in hotkey manager to the same value
- update hotkey guidance/recommendations to avoid steering users toward wrong hotkeys 

## Why this fixes #220
"Ctrl+super" is a modifier-only shortcut. Modifier-only combos are less reliable across platforms and paths because they require native-listener special handling and are more likely to be reserved by the OS/desktop environment. That made the default fragile and led to onboarding/registration failures in real setups.

 "Ctrl+shift+space" is a standard chord with a base key, works with the normal shortcut registration path, and is materially more consistent as a default.

Fixes #220

https://github.com/user-attachments/assets/31644c9a-c915-40ec-8b76-8bce791d0137

### DIFF
--- a/src/helpers/hotkeyManager.js
+++ b/src/helpers/hotkeyManager.js
@@ -38,12 +38,12 @@ function isModifierOnlyHotkey(hotkey) {
 // Suggested alternative hotkeys when registration fails
 const SUGGESTED_HOTKEYS = {
   single: ["F8", "F9", "F10", "Pause", "ScrollLock"],
-  compound: ["Control+Super", "Control+Alt", "Control+Shift+Space", "Alt+F7"],
+  compound: ["Control+Shift+Space", "Control+Alt", "Control+Shift+K", "Alt+F7"],
 };
 
 class HotkeyManager {
   constructor() {
-    this.currentHotkey = process.platform === "darwin" ? "GLOBE" : "Control+Super";
+    this.currentHotkey = process.platform === "darwin" ? "GLOBE" : "Control+Shift+Space";
     this.isInitialized = false;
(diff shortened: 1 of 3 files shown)


gabrielste1n:
Closing this as modifiers have been confirmed to work cross platform since the native binary fix - thanks @RutamBhagat

PR #246, opened 2026-02-10
fix(linux): hide overlay before paste on GNOME Wayland

wrt54gl:
## Summary

- Fixes paste not working on GNOME Wayland when `FLOATING_ICON_AUTO_HIDE` is enabled
- `BrowserWindow.show()` steals focus from the target window on GNOME Wayland, causing `ydotool` keystrokes to go to the OpenWhispr overlay instead of the user's app
- Temporarily hides the overlay before pasting so GNOME returns focus to the target window, then re-shows if auto-hide is disabled
- Restores AT-SPI2 terminal detection and GNOME-specific ydotool-first paste tool ordering removed in #221
- Adds `open-whispr` to AT-SPI2 skip set so our own window is never misidentified as the paste target

## Details

On GNOME Wayland, `BrowserWindow.show()` triggers activation via `xdg_activation_v1`, stealing keyboard focus. Since `ydotool` injects key events to the focused window (via uinput), the paste keystrokes (`Ctrl+V` / `Ctrl+Shift+V`) were sent to the overlay window instead of the terminal or other target application.

The fix adds a hide + 200ms delay before the paste operation in the `paste-text` IPC handler, allowing the GNOME compositor to return focus to the previously active window. After paste completes, the overlay is re-shown only if `FLOATING_ICON_AUTO_HIDE` is disabled.

This also restores two GNOME Wayland fixes from PR #202 that were removed in #221:
- **AT-SPI2 terminal detection**: On GNOME Wayland, `xdotool` can only see XWayland windows and returns OpenWhispr's own window. AT-SPI2 reliably detects both native Wayland and XWayland active windows.
- **ydotool-first paste ordering**: On GNOME Wayland, `xdotool` may silently target the wrong window while reporting success, preventing fallback to `ydotool`.

## Test plan

- [ ] Enable `FLOATING_ICON_AUTO_HIDE` on GNOME Wayland
- [ ] Dictate text while focused in GNOME Terminal → should paste with `Ctrl+Shift+V`
- [ ] Dictate text while focused in a non-terminal app (e.g. Text Editor) → should paste with `Ctrl+V`
- [ ] Disable `FLOATING_ICON_AUTO_HIDE` → overlay should re-appear after paste
- [ ] Verify no regression on X11 or non-GNOME Wayland compositors

### DIFF
--- a/src/helpers/clipboard.js
+++ b/src/helpers/clipboard.js
@@ -651,6 +651,48 @@ class ClipboardManager {
         "yakuake",
       ];
 
+      // On GNOME Wayland, xdotool can't see native Wayland windows and returns
+      // OpenWhispr's own window instead. Use AT-SPI2 accessibility API first,
+      // which works for both native Wayland and XWayland windows.
+      if (isGnome && isWayland) {
+        try {
+          const atspiScript = `import gi
+gi.require_version('Atspi','2.0')
+from gi.repository import Atspi
+d=Atspi.get_desktop(0)
+skip={'gnome-shell','open-whispr',''}
+for i in range(d.get_child_count()):
+ a=d.get_child_at_index(i)
+ if not a:continue
+ n=a.get_name()
+ if n.lower() in skip:continue
+ for j in range(a.get_child_count()):
+  w=a.get_child_at_index(j)
+  if w and w.get_state_set().contains(Atspi.StateType.ACTIVE):
+   print(n.lower());raise SystemExit(0)`;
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Closing — all three parts of this have since landed via other routes: AT-SPI2 terminal detection compiled into linux-fast-paste (#734), the overlay made non-focusable so it can't steal focus before paste (#807), and Shift+Insert routing for unknown window classes (#827). Thanks for identifying these failure modes.

PR #267, opened 2026-02-12
feat: split Chinese into Simplified and Traditional language options

white1033:
## Summary

- Split the single "Chinese" (`zh`) language option into two distinct entries:
  - **Chinese (Simplified)** — code `zh-CN`, flag 🇨🇳
  - **Chinese (Traditional)** — code `zh-TW`, flag 🇹🇼
- Each variant has a tailored AI instruction that ensures responses use the correct character set (simplified vs traditional characters)
- No changes needed outside `languageRegistry.json` — `getBaseLanguageCode()` already strips the region suffix via `split("-")[0]`, passing `"zh"` to whisper.cpp as expected (same pattern used by `en-US`/`en-GB`)

## Motivation

The current "Chinese" option doesn't distinguish between Simplified and Traditional Chinese. This matters because:
- Users in Taiwan, Hong Kong, and Macau use Traditional Chinese (繁體中文)
- Users in Mainland China and Singapore use Simplified Chinese (简体中文)
- The AI instruction needs to specify which character set to use for accurate post-processing

## Test plan

- [ ] Verify both "Chinese (Simplified)" and "Chinese (Traditional)" appear in the language picker
- [ ] Select "Chinese (Simplified)" → transcribe → confirm AI uses simplified characters (简体)
- [ ] Select "Chinese (Traditional)" → transcribe → confirm AI uses traditional characters (繁體)
- [ ] Verify whisper.cpp receives `"zh"` as the language code for both variants
- [ ] Confirm no regression for other languages (especially `en-US`/`en-GB` which use the same code pattern)

### DIFF
--- a/src/hooks/useSettings.ts
+++ b/src/hooks/useSettings.ts
@@ -69,7 +69,20 @@ export interface ThemeSettings {
   theme: "light" | "dark" | "auto";
 }
 
+const LANGUAGE_MIGRATIONS: Record<string, string> = { zh: "zh-CN" };
+let _migrated = false;
+function migratePreferredLanguage() {
+  if (_migrated) return;
+  _migrated = true;
+  const stored = localStorage.getItem("preferredLanguage");
+  if (stored && LANGUAGE_MIGRATIONS[stored]) {
+    localStorage.setItem("preferredLanguage", LANGUAGE_MIGRATIONS[stored]);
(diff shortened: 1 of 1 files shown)


gabrielste1n:
thank you @white1033 merged in and will add this to next release!

Issue #260, opened 2026-02-11
Auto-detect language feature performing very poor (using OpenWhispr Pro)


Alfagun74:
**Describe the bug**
The auto-detect feature of the language selector has very poor performance. It often effectively behaves as if it were set to English only: it sometimes outputs English gibberish or nothing at all, regardless of the language spoken.   
  
Current Observations:
- German: Input like "Vielen Dank für die Einladung" results in phonetic English nonsense: "Fill dunk for the iron. Lard on."
- Turkish & Japanese: Consistently produce unintelligible gibberish.
- French & Spanish: Highly inconsistent behavior. The system either:
  - Fails to detect any speech (no output).
  - Returns phonetic English gibberish.
  - Occasionally returns the correct transcription.
- Mixed Input (Spanish/English): The transcription quality degrades significantly, resulting in a mix of gibberish and English words.

**Expected behavior**
The auto-detect feature should reliably recognize the spoken language (German, Turkish, French, Spanish, English, and mixed input where supported) and produce accurate transcriptions instead of defaulting to English or gibberish.

**Desktop (please complete the following information):**
 - Version 1.4.6
 - Transcription and LLM Model: OpenWhispr Pro
<img width="460" height="423" alt="Image" src="https://github.com/user-attachments/assets/5809b1c0-914d-4fe6-8176-3ba4c4ea85c2" />

heyrobertchang:
Hi @Alfagun74 cc @gabrielste1n 

I think I found the issue.

When the user selects "Auto Detect", the language value "auto" is stored in localStorage. However, it never reaches `whisper.cpp` due to the following chain:
                                                                                       
1. getBaseLanguageCode("auto") in `languageSupport.ts:33` returns undefined — it treats "auto" as a non-language and strips it out.
2. `audioManager.js:440` checks if (language) — since it's undefined (falsy), options.language is never set on the options object.
3. `whisper.js:254` receives options.language as undefined, falls through to null.
4. `whisperServer.js:236` checks `if (options.language && ...)` — null is falsy, so language is never passed to the whisper-server binary.
5. `whisper.cpp` defaults to "en" when no `--language` flag is provided.

Result: Selecting "Auto Detect" is functionally identical to selecting "English". The model forces English phoneme mapping on all input, producing gibberish like "Fill dunk for the iron" instead of "Vielen Dank für die Einladung".

I will create a PR to handle it 🙏🏼

Alfagun74:
[...]

PR #269, opened 2026-02-12
fix: explicitly pass language=auto to whisper.cpp to enable auto-detection 

heyrobertchang:
Fix [issue](https://github.com/OpenWhispr/openwhispr/issues/260)

When the user selects "Auto Detect", the language value "auto" is stored in localStorage. However, it never reaches `whisper.cpp` due to the following chain:
                                                                                       
1. `getBaseLanguageCode("auto")` in `languageSupport.ts:33` returns undefined — it treats "auto" as a non-language and strips it out.
2. `audioManager.js:440` checks if (language) — since it's undefined (falsy), options.language is never set on the options object.
3. `whisper.js:254` receives options.language as undefined, falls through to null.
4. `whisperServer.js:236` checks `if (options.language && ...)` — null is falsy, so language is never passed to the whisper-server binary.
5. `whisper.cpp` defaults to "en" when no `--language` flag is provided.

Result: Selecting "Auto Detect" is functionally identical to selecting "English". The model forces English phoneme mapping on all input, producing gibberish like "Fill dunk for the iron" instead of "Vielen Dank für die Einladung".

---

Test Screenshot:

CleanShot 2026-02-13 at 01 19 39@2x

Test Video (Japanese, Chinese and English)

https://github.com/user-attachments/assets/e0d4e172-dbff-467c-8453-cf54668cb739

### DIFF
--- a/src/helpers/whisperServer.js
+++ b/src/helpers/whisperServer.js
@@ -406,7 +406,6 @@ class WhisperServerManager {
     parts.push(finalBuffer);
     parts.push("\r\n");
 
-    // Always send language; whisper.cpp defaults to "en" if omitted
     parts.push(
       `--${boundary}\r\n` +
         `Content-Disposition: form-data; name="language"\r\n\r\n` +


gabrielste1n:
Hey @RobertChang0722 , great catch — this was a real bug and your diagnosis was spot on. whisper.cpp defaulting to English when --language is omitted was the root cause.

I went ahead and applied the fix directly in a recent commit to main with you credited as co-author. The one difference from your PR: I kept the change scoped to whisperServer.js only (both the CLI and HTTP server paths) and left ipcHandlers.js untouched. The cloud API auto-detects when the language parameter is omitted, so always sending language=auto there could cause issues depending on the backend. The whisper.cpp paths were the only ones that needed the explicit flag.

Thanks for tracking this down — closing in favour of the direct commit.

PR #273, opened 2026-02-13
fix: use unzipper instead of tar for Windows key listener extraction

solwer-se:
Windows' built-in tar command fails with path resolution errors when extracting the windows-key-listener binary:
  tar: Cannot connect to c: resolve failed

This change switches to the unzipper npm package (already in dependencies) which properly handles Windows paths and provides more reliable extraction.

Benefits:
- Eliminates error messages during npm run dev
- Enables push-to-talk for fresh Windows installations
- More reliable cross-platform extraction

Tested on Windows 11 with successful extraction.


gabrielste1n:
thanks @solwer-se - great fix

PR #284, opened 2026-02-16
fix(windows): use relative paths in sherpa-onnx tar extraction

iTubeGamer:
## Problem

`download-sherpa-onnx.js` fails on Windows because GNU tar interprets the colon in `C:\...` paths as a remote `host:path` separator:

```
tar: Cannot connect to C: resolve failed
```

This affects both the `-f` (archive) and `-C` (destination) arguments.

## Fix

Use `path.dirname()` as `cwd`, then pass only relative paths (`path.basename()` for the archive, `path.relative()` for the destination) so no argument contains a drive letter colon.

Also switches from `execSync` (shell string) to `execFileSync` (array args) to avoid shell escaping issues and improve safety.

## Testing

- Tested successfully on Windows (Git Bash / MINGW64)
- No platform-specific branching — works on all OS/tar implementations

### DIFF
--- a/.github/workflows/build-and-notarize.yml
+++ b/.github/workflows/build-and-notarize.yml
@@ -39,17 +39,17 @@ jobs:
       - name: Download whisper.cpp binaries
         run: node scripts/download-whisper-cpp.js --current
         env:
-          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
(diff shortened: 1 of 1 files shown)


gabrielste1n:
thanks legend @iTubeGamer

Issue #191, opened 2026-02-05
Gtk-ERROR **: 08:01:21.681: GTK 2/3 symbols detected. Using GTK 2/3 and GTK 4 in the same process is not supported


mlissner:
**Describe the bug**

When I just installed the latest `.deb` with:

```
sudo dpkg -i OpenWhispr-1.3.4-linux-amd64.deb 
```

The application crashed with the following error:

```
↪ open-whispr 
Checking for beta autoupdate feature for deb/rpm distributions
Found package-type: deb

(process:942011): Gtk-ERROR **: 08:01:21.681: GTK 2/3 symbols detected. Using GTK 2/3 and GTK 4 in the same process is not supported
Trace/breakpoint trap (core dumped)
```

**To Reproduce**
1. Install as above
2. Run it from the CLI or from Gnome directly

**Expected behavior**
It should launch.

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version: 24.04.3 LTS

**Additional context**
This was the first time I tried using this.

kimusan:
Getting the same error here.

christian-taillon:
I can reproduce this on Ubuntu 24.04 GNOME Wayland with current packaged builds (including 1.4.11): app exits with  unless GTK3 is forced.

I tested a minimal code change on latest :



This avoids the immediate GTK startup crash for me. I opened a minimal PR against latest  with just this change so maintainers can review/discuss there.

christian-taillon:
Correction / clean repro details:

I can reproduce this on Ubuntu 24.04 GNOME Wayland with current packaged builds (including 1.4.11): the app exits with the GTK 2/3 + GTK 4 symbol conflict unless GTK3 is forced.

I tested a minimal code change on latest `main`:

```js
if (process.platform === "linux") {
  app.commandLine.appendSwitch("gtk-version", "3");
}
```

This avoids the immediate GTK startup crash for me. I opened a minimal PR against latest `main` with only this change so maintainers can review/discuss there.

gabrielste1n:
Hey all - merged in #291 and released in v1.5.0 please let me know if this resolves your issue!

michal-pikosz:
Its fixed for me on fedora gnome wayland 43

PR #182, opened 2026-02-03
feat(linux): Wayland paste improvements and NVIDIA GPU support

christian-taillon:
## Summary
- Add NVIDIA GPU support for Linux whisper binaries, downloads, and in-app toggle.
- Improve Wayland paste reliability with ydotool/dotool support and clipboard + primary selection sync.
- Add opt-in ydotoold management via Settings with status and guidance.

## Testing
- npm run dev (Wayland)

PR #291, opened 2026-02-18
fix: force GTK3 on Linux to prevent GTK symbol startup crash

christian-taillon:
## Summary
- add a single Linux startup switch to force GTK3: `app.commandLine.appendSwitch("gtk-version", "3")`
- keep the rest of startup behavior unchanged (no refactor, no additional flags)
- addresses the immediate crash reported in #191 on Ubuntu/GNOME Wayland packaged builds

## Why
Electron 36 can initialize with GTK4 while other loaded symbols are GTK2/3, which crashes process startup on some Linux setups. Forcing GTK3 at startup avoids that conflict.

## Validation
- verified local branch starts with `npm start` after this change
- change scope is 1 line in `main.js` (Linux-only path)

### DIFF
--- a/main.js
+++ b/main.js
@@ -58,6 +58,7 @@ configureChannelUserDataPath();
 // the compositor to set up an ARGB visual before any windows are created.
 // --disable-gpu-compositing prevents GPU compositing conflicts with the compositor.
 if (process.platform === "linux") {
+  app.commandLine.appendSwitch("gtk-version", "3");
   app.commandLine.appendSwitch("enable-transparent-visuals");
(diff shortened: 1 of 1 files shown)


christian-taillon:
I closed my other PR (#182) since the project has drifted and this path seems more practical going forward.

I think this one-line Linux GTK3 switch is the best short-term fix because it targets the startup crash directly, keeps scope very small, and avoids pulling in broader behavior changes.

If others agree after review/testing across distros and desktop environments, this should make Linux installs more reliable going forward.

gabrielste1n:
thank you legend @christian-taillon definitely agree with this for now :)

PR #298, opened 2026-02-20
ci: add Windows paste and key listener download steps to build pipeline

borng:
## Summary

The Windows build job was missing download steps for `windows-fast-paste.exe` and `windows-key-listener.exe` in both CI workflows. Only `nircmd.exe` was being downloaded.

- Add download steps for `windows-fast-paste.exe` and `windows-key-listener.exe` to `build-and-notarize.yml` and `release.yml`
- Add npm script aliases (`download:nircmd`, `download:windows-fast-paste`, `download:windows-key-listener`) and wire them into `prebuild:win` — without these, `npm run build:win` locally
  triggers `prebuild:win` but skips the Windows-only binaries, producing an incomplete build

All three download scripts detect the platform and skip silently on non-Windows, so Linux and macOS builds are unaffected.

## Test plan

- [ ] Windows CI build downloads all three binaries (nircmd, fast-paste, key-listener)
- [ ] Linux/macOS builds still pass (download scripts skip silently)
- [ ] `npm run build:win` locally produces a complete build with all binaries


gabrielste1n:
thank you legend @borng

PR #309, opened 2026-02-23
Refactor control panel close event handling Fixes #242

### LINKED ISSUE #242: Clicking the X on the application window minimizes it instead of closing it
**Describe the bug**
If I click the X in the upper right corner of the application Control Panel window, the application should close and remain only in the Windows tray. Instead it minimizes and there's no possibility to completely close the window

**To Reproduce**
1. Open the Control Panel
2. Click on the X in the top right
3. The window minimizes instead of closing (wrong behavior)

**Expected behavior**
By clicking the close button in the upper right corner of the control panel window, the window should completely close, disappearing from the taskbar at the bottom and remaining available in the system tray

**Screenshots**

Image

Image

<br><br>
**Desktop**
 - OS: Windows 11 Home 25H2

**Additional context**
I downloaded the program from GitHub by accessing the Release page, downloading the latest version using the EXE for the Windows setup

N0amG:
## Problem
On Windows, clicking the ✕ close button had the same behavior as the minimize button — the window stayed visible in the taskbar instead of disappearing while the app kept running in the background as describe in the Issue #242 

## Root Cause
In `windowManager.js`, the `close` event handler called `minimize()` on non-macOS platforms instead of `hide()`.

## Fix
Replaced the platform-specific logic with a single call to `hideControlPanelToTray()` which already calls `window.hide()` and handles the macOS dock case internally.

## Changes
- `src/helpers/windowManager.js`: 4 lines removed, no new logic added

## Testing
- [x] Windows: ✕ hides the window (not in taskbar), app still in tray
- [x] Windows: minimize button still minimizes to taskbar
- [x] macOS: behavior unchanged

### DIFF
--- a/src/helpers/windowManager.js
+++ b/src/helpers/windowManager.js
@@ -495,12 +495,8 @@ class WindowManager {
     this.controlPanelWindow.on("close", (event) => {
       if (!this.isQuitting) {
         event.preventDefault();
-        if (process.platform === "darwin") {
-          this.hideControlPanelToTray();
-        } else {
-          this.controlPanelWindow.minimize();
+        this.hideControlPanelToTray();
         }
-      }
     });
 
     this.controlPanelWindow.on("closed", () => {


gabrielste1n:
@N0amG thanks for this! I actually fixed it in v1.5.0!

PR #315, opened 2026-02-23
feat: GPU-accelerated local inference (Vulkan/Metal)

gabrielste1n:
## Summary

Enables GPU acceleration for llama-server across all platforms with automatic detection and fallback:

- **macOS**: Metal GPU acceleration via the default llama-server binary (already compiled with Metal support), passing `--n-gpu-layers 99`
- **Windows/Linux**: Vulkan GPU acceleration with a dedicated binary (`llama-server-*-vulkan`), falling back to CPU-only binary if Vulkan fails to start within 15 seconds
- **Backend persistence**: Detected GPU backend (`vulkan`, `cpu`, `metal`) is saved to the `.env` file via `LLAMA_GPU_BACKEND`, skipping re-detection on subsequent launches
- **GPU status badge**: Shows current backend (green dot for GPU, amber for CPU) in the reasoning model selector, with a "Re-detect GPU" button for CPU fallback users

### Download script (`scripts/download-llama-server.js`)
- Restructured `BINARIES` config to support multiple variants per platform using `platformArch` grouping
- Added `win32-x64-vulkan` and `linux-x64-vulkan` entries (marked `optional: true` — download failure is non-fatal)
- Renamed existing entries to explicit `-cpu` suffix for clarity

### Server manager (`src/helpers/llamaServer.js`)
- Replaced `getServerBinaryPath()` with `getServerBinaryPaths()` returning `{ default }` (macOS) or `{ vulkan?, cpu? }` (Windows/Linux)
- New `_startWithGpuFallback()`: tries Vulkan first (15s timeout), falls back to CPU (60s timeout)
- New `_startWithBinary()`: encapsulates spawn + health-check polling with configurable timeout and race-safe settlement
- New `_buildEnv()`: sets `DYLD_LIBRARY_PATH` / `LD_LIBRARY_PATH` for the binary's directory
- New `_killCurrentProcess()`: clean teardown between Vulkan→CPU fallback
- `getStatus()` now returns `backend` and `gpuAccelerated` fields
- New `resetGpuDetection()` for user-triggered re-probing

### IPC + Preload + Types
- New `llama-gpu-reset` IPC handler with preload bridge (`llamaGpuReset`)
- `LLAMA_GPU_BACKEND` added to `PERSISTED_KEYS` in environment manager
- TypeScript types: `GpuBackend`, `LlamaServerStatus`, updated `ElectronAPI` interface

### UI (`src/components/ReasoningModelSelector.tsx`)
- `GpuStatusBadge` component: polls server status every 5s, shows backend label + colored indicator, re-detect button when on CPU

### Model manager (`src/helpers/modelManagerBridge.js`)
- `gpuLayers: 99` passed to all `serverManager.start()` calls (inference, prewarm, IPC handler)

Heavily inspired by work from @RutamBhagat

### DIFF
--- a/main.js
+++ b/main.js
@@ -160,6 +160,7 @@ const UpdateManager = require("./src/updater");
 const GlobeKeyManager = require("./src/helpers/globeKeyManager");
 const DevServerManager = require("./src/helpers/devServerManager");
 const WindowsKeyManager = require("./src/helpers/windowsKeyManager");
+const WhisperCudaManager = require("./src/helpers/whisperCudaManager");
 const { i18nMain, changeLanguage } = require("./src/helpers/i18nMain");
 
 // Manager instances - initialized after app.whenReady()
@@ -175,6 +176,7 @@ let trayManager = null;
 let updateManager = null;
 let globeKeyManager = null;
 let windowsKeyManager = null;
+let whisperCudaManager = null;
 let globeKeyAlertShown = false;
 let authBridgeServer = null;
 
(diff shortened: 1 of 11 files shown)


gabrielste1n:
thanks for the push here @RutamBhagat

PR #330, opened 2026-02-26
feat: update groq models list

### LINKED ISSUE #325: Support for standard Whisper Large v3 (non-Turbo) via Groq API
I am using OpenWhispr with the Groq API. Currently, the interface allows me to select Whisper Large v3 Turbo, which is excellent for speed. However, Groq also provides the standard Whisper Large v3 (non-Turbo) on their free tier. In certain use cases where transcription accuracy is prioritized over speed, I find myself wanting to use the standard Large v3 model, but it is currently not an available option within OpenWhisper.

I would like to see Whisper Large v3 added as a selectable model in the dropdown/settings when the Groq API provider is active. This would allow users to toggle between the "Turbo" version for speed and the standard "Large v3" version for maximum accuracy.

I have checked the current model selection list, but only the Turbo version appears to be available for Groq.

**Additional context**
Since the Groq API already supports both models (often under the same rate limits), adding this should provide more flexibility for users without requiring additional API configurations.

tahatariq19:
Added the missing Groq models discussed in issue #325  

Also removed deprecated models _Mixtral 8x7b_ and _Gemma 2 9b_ no longer availiable over the Groq Api.

This ensures users can select the latest LLMs from the provider

This Closes #325

### DIFF
--- a/src/locales/de/translation.json
+++ b/src/locales/de/translation.json
@@ -1231,12 +1231,10 @@
         "groq_openai_gpt_oss_20b": "Schnelles Open-Source-Modell, 1000 T/s",
         "groq_llama_3_3_70b_versatile": "Vielseitiges Modell von Meta, 280 T/s",
         "groq_llama_3_1_8b_instant": "Ultraschnell 560 T/s, 131K Kontext",
-        "groq_llama_4_maverick": "Multimodales Modell von Meta, 600 T/s",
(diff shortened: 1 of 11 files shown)


xAlcahest (approved):
Reviewed: translations fixed, descriptions updated, deprecated models removed. All model IDs verified active.

PR #334, opened 2026-02-27
fix: Add Fn key capture support

vex0rian:
## Summary
Add native `Fn` key capture support in the hotkey input component so users can set `Fn` directly (including as a standalone key), with proper state handling for capture/release.

  ## What changed
  - Updated `src/components/ui/HotkeyInput.tsx`:
    - Added explicit `Fn` handling on keydown to enter Fn-held capture state.
    - Added explicit `Fn` keyup handling to finalize plain `Fn` capture when no other key is combined.
    - Preserved existing capture flow for:
      - `Fn+<key>` combinations
      - modifier-only and timeout behavior
      - existing Globe listener-based recording UX

  ## Why
  `Fn` was already part of manager/validator paths (e.g. `GLOBE/Fn` handling), but direct keyboard capture in the input control did not always finalize `Fn` reliably. This change ensures explicit and consistent capture behavior.

  ## Validation
  - Local logic is implemented as minimal, surgical change in a single file.
  - No unrelated hotkey logic paths were refactored.

  ## Notes
  - This complements existing macOS Globe/Fn handling in main process and hotkey manager.

### DIFF
--- a/main.js
+++ b/main.js
@@ -539,6 +539,7 @@ async function startApp() {
     let globeLastStopTime = 0;
     const MIN_HOLD_DURATION_MS = 150;
     const POST_STOP_COOLDOWN_MS = 300;
+    const isGlobeLikeHotkey = (hotkey) => hotkey === "GLOBE" || hotkey === "Fn";
 
     globeKeyManager.on("globe-down", async () => {
       const currentHotkey = hotkeyManager.getCurrentHotkey && hotkeyManager.getCurrentHotkey();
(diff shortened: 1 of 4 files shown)


gabrielste1n:
thank you! @vex0rian

gabrielste1n:
Thanks for identifying this issue and for the initial work here, @vex0rian — this surfaced a real problem.

We ended up addressing it in #338 with a slightly different approach:

Shared helper instead of inline copies — isGlobeLikeHotkey() is exported from hotkeyManager.js (and a frontend equivalent in hotkeys.ts) rather than redefined as inline arrow functions in each file.
Kept "GLOBE" as the canonical stored value — your PR changed finalizeCapture("GLOBE") to finalizeCapture("Fn"), which would have changed what gets persisted. We kept "GLOBE" as the write path and made all read paths accept both.
Broader coverage — we found additional checks that needed updating in windowManager.js, hotkeys.ts, hotkeyValidator.ts, HotkeyInput.tsx, and OnboardingFlow.tsx that weren't covered here.
Closing in favor of #338. Thanks again for flagging it!

PR #335, opened 2026-02-27
feat: add custom pasteboard type for clipboard manager compatibility (macOS)

jguice:
> **Note:** I'm testing this locally on macOS (Apple Silicon) and it's working well, but wanted to share for early feedback.

## Summary

- **Clipboard manager compatibility (macOS):** Adds a custom pasteboard type (`com.openwhispr.transcription`) alongside the plain text when OpenWhispr writes to the clipboard during paste operations. This allows clipboard managers that support pasteboard type filtering (Maccy, Flycut) to identify and ignore OpenWhispr's transient clipboard writes.
- **Trailing space after dictation:** Appends a trailing space after pasted transcriptions so the cursor is positioned ready for the next word, matching the behaviour of other dictation tools (e.g. WhisperFlow). This is a small UX improvement that makes back-to-back dictations and dictation-then-typing flows smoother.
- **Cross-platform safe:** Linux and Windows functionality is completely unchanged — the pasteboard marker is macOS-only, and the code gracefully falls back to the existing `clipboard.writeText()` on all other platforms.

## Problem

OpenWhispr briefly writes transcribed text to the system clipboard, then simulates Cmd+V to paste into the active app, and finally restores the original clipboard contents. Clipboard managers (e.g. Maccy, Flycut) capture these transient writes. Because the paste is simulated via Cmd+V in the foreground app, macOS clipboard managers typically attribute the entry to the foreground app rather than OpenWhispr, making app-based filtering unreliable.

## Solution

A small compiled Swift helper (`macos-clipboard-marker`) uses `NSPasteboard` to write both `public.utf8-plain-text` and a custom `com.openwhispr.transcription` type simultaneously. The existing `clipboard.writeText()` calls in the paste flow are replaced with `writeClipboardWithMarker()`, which invokes this binary on macOS and falls back to standard `clipboard.writeText()` on other platforms or on failure.

Users can add `com.openwhispr.transcription` to their clipboard manager's ignored pasteboard types:
- **Maccy**: Settings → Ignore → add `com.openwhispr.transcription` under pasteboard types
- **Flycut**: Preferences → add `com.openwhispr.transcription` to the skip pasteboard types list

### Cross-platform notes

- **macOS**: Full pasteboard marker support via compiled Swift helper
- **Linux**: No changes to clipboard behaviour. The build script is a no-op on non-macOS. `writeClipboardWithMarker()` falls through to `clipboard.writeText()`.
- **Windows**: Same as Linux — no changes, standard clipboard path used.

### Future work: Linux and Windows clipboard marker support

A similar approach could be extended to the other platforms OpenWhispr supports:

- **Linux (X11/Wayland):** The clipboard supports MIME types as targets alongside content. A custom MIME type like `application/x-openwhispr-transcription` could be written alongside `text/plain` using `xclip -selection clipboard -target` (X11) or equivalent Wayland mechanisms (`wl-copy` supports `--type`). Clipboard managers like [CopyQ](https://hluk.github.io/CopyQ/) support filtering by MIME type, so users could ignore the custom type.
- **Windows:** The clipboard supports custom formats via `RegisterClipboardFormat()`. A format like `OpenWhispr.Transcription` could be registered and written alongside `CF_UNICODETEXT`. The open-source clipboard manager [Ditto](https://ditto-cp.sourceforge.io/) exposes format information, though filtering support would need investigation.

Both would follow the same pattern: write a platform-specific marker alongside the text so clipboard managers can identify and optionally ignore transient dictation entries, with a graceful fallback to standard clipboard writes if the marker mechanism is unavailable.

## Changes

| File | Change |
|------|--------|
| `resources/macos-clipboard-marker.swift` | New Swift source for the pasteboard marker binary |
| `scripts/build-macos-clipboard-marker.js` | Build script (follows existing `build-macos-fast-paste.js` pattern) |
| `src/helpers/clipboard.js` | `writeClipboardWithMarker()` function; replaces `clipboard.writeText(text)` in paste flow |
| `src/helpers/audioManager.js` | Appends trailing space in `safePaste()` |
| `package.json` | Adds `compile:clipboard-marker` to `compile:native` chain |
| `electron-builder.json` | Includes `macos-clipboard-marker` in `extraResources` |
| `README.md` | Troubleshooting entry for clipboard manager configuration |

## Test plan

- [x] Verified on macOS (Apple Silicon) that both `public.utf8-plain-text` and `com.openwhispr.transcription` are present on the pasteboard after dictation
- [x] Confirmed `pbpaste` still returns the transcribed text (standard paste unaffected)
- [x] Confirmed Maccy ignores dictation entries when `com.openwhispr.transcription` is in its ignored types list
- [x] Verified trailing space is appended after pasted transcriptions
- [x] Verified fallback to `clipboard.writeText()` when binary is unavailable (covers Linux/Windows path)
- [x] Confirmed clipboard restore (original content) after paste does not include the marker type
- [x] `npm run lint` passes with no new warnings
- [x] Build script exits cleanly on non-macOS platforms


jguice:
Hi, bumping this.  Aside from rebasing `README.md` and `package.json` any concerns getting this merged?  Been working well here for weeks now.

alex-shumilov:
Thanks for building this — the pollution of clipboard history has been annoying on my end too. Sharing one alternative worth considering that I've been running locally for the same goal.

**Use `org.nspasteboard.TransientType` instead of (or alongside) a custom UTI.** This is the [nspasteboard.org](https://nspasteboard.org) convention — a de-facto standard that clipboard managers already honor without any user configuration:

- **Maccy, Paste, Alfred, Flycut, LaunchBar, Raycast** all skip clipboard items tagged `org.nspasteboard.TransientType` out of the box.
- Semantically it matches exactly what OpenWhispr does: briefly parks text on the pasteboard, pastes, then restores the original. That's the canonical definition of a transient write.
- Users don't need to know the UTI exists or configure anything — the transcription just doesn't land in their clipboard history.

Implementation is one line of Swift:

```swift
pb.setData(Data([1]), forType: NSPasteboard.PasteboardType("org.nspasteboard.TransientType"))
pb.setString(text, forType: .string)
```

Two suggestions that might unblock this PR:

1. **Swap or add `TransientType`** — either replace `com.openwhispr.transcription` or emit both (a custom UTI is still useful for users who want OpenWhispr-specific filtering beyond just "transient").
2. **Fold the write into the existing `macos-fast-paste` binary** to avoid shipping a second helper. `macos-fast-paste` is already spawned for the Cmd+V keystroke; extending it to read text from stdin and write the pasteboard with the marker before posting Cmd+V keeps the binary count unchanged and ensures the write is ordered correctly w.r.t. the paste.

Happy to contribute a follow-up diff in either direction if it helps move this forward. Either way, thanks for taking a swing at this — it's a real papercut for anyone using a clipboard manager alongside OpenWhispr.

PR #340, opened 2026-02-28
fix: GPU activation and Vulkan fallback

xAlcahest:
Addresses https://github.com/OpenWhispr/openwhispr/discussions/276

### llama-server not restarting after GPU reset
The `llama-gpu-reset` IPC handler stopped the server but never restarted it. Now it saves the current model ID before stopping and relaunches via `prewarmServer()` after reset.

### Vulkan skipped after a single model failure
After one model failed to load with Vulkan (e.g. too large for VRAM), `_saveGpuBackend("cpu")` cached "cpu" in `process.env.LLAMA_GPU_BACKEND`, causing all subsequent models to skip Vulkan entirely. Removed the backend caching mechanism so every model always attempts Vulkan first and falls back to CPU independently.

### UI stuck on "Activating GPU..."
The status polling interval was 5 seconds and didn't recognize the "downloaded but server not running" state. Added a 1-second fast poll during activation with a 10-second timeout.

### No distinction between GPU ready and GPU active
Added two visual states: blue dot with "GPU acceleration ready" (Vulkan binary downloaded, server not running) and green dot with "GPU acceleration active" (server running with Vulkan). Added `gpu.ready` translation key across all 10 locales.

### VRAM not freed when switching to cloud reasoning
When switching from local LLM reasoning to cloud (OpenWhispr Cloud or BYOK cloud providers), the llama-server process was left running with the model loaded in VRAM. Added `llamaServerStop()` calls when switching from BYOK to OpenWhispr Cloud and when switching from local to cloud within BYOK. The existing `syncStartupPreferences` mechanism doesn't cover the BYOK → OpenWhispr Cloud case because `cloudReasoningMode` is not in its dependency array.

### DIFF
--- a/src/components/ReasoningModelSelector.tsx
+++ b/src/components/ReasoningModelSelector.tsx
@@ -634,6 +634,7 @@ export default function ReasoningModelSelector({
     setSelectedMode(newMode);
 
     if (newMode === "cloud") {
+      window.electronAPI?.llamaServerStop?.();
       setLocalReasoningProvider(selectedCloudProvider);
 
       if (selectedCloudProvider === "custom") {
--- a/src/components/SettingsPage.tsx
+++ b/src/components/SettingsPage.tsx
@@ -465,6 +465,7 @@ function AiModelsSection({
                   onClick={() => {
                     if (!isCloudMode) {
                       setCloudReasoningMode("openwhispr");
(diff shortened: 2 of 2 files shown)


gabrielste1n:
legend @xAlcahest

PR #346, opened 2026-03-01
feat: add ydotool as required Linux dependency

xAlcahest:
## Summary

- **deb**: `ydotool` added as `Depends` (auto-installed by apt)
- **rpm**: `ydotool` added as `Requires` (auto-installed by dnf/zypper)
- **AppImage/tar.gz**: auto-install via `pkexec` at startup on Wayland when missing (supports dnf, apt-get, pacman, zypper)

ydotool is needed for reliable paste on Wayland compositors.

### DIFF
--- a/main.js
+++ b/main.js
@@ -270,7 +270,9 @@ function initializeCoreManagers() {
 
 // Phase 2: Non-critical setup after windows are visible
 function initializeDeferredManagers() {
-  ensureYdotool().catch(() => {});
+  ensureYdotool().catch((err) => {
+    require("./src/helpers/debugLogger").warn("ydotool setup error", { error: err?.message }, "clipboard");
+  });
   clipboardManager.preWarmAccessibility();
   trayManager = new TrayManager();
   globeKeyManager = new GlobeKeyManager();
--- a/src/helpers/ensureYdotool.js
+++ b/src/helpers/ensureYdotool.js
@@ -101,7 +101,8 @@ async function ensureYdotool() {
       await setupUinputAccess(log);
     }
 
-    await ensureYdotooldDaemon(log);
+    const hasUserConsent = needsInstall || needsUinput;
+    await ensureYdotooldDaemon(log, { hasUserConsent });
     markSetupDone();
     log.info("ydotool setup completed", {}, "clipboard");
   } catch (error) {
--- a/src/helpers/ydotoolInstaller.js
+++ b/src/helpers/ydotoolInstaller.js
@@ -78,7 +78,7 @@ async function setupUinputAccess(log) {
   const lines = ["#!/bin/sh", "set -e"];
 
   if (needsRule) {
-    lines.push(`printf '%s\\n' '${UDEV_RULE}' > ${UDEV_RULE_PATH}`);
+    lines.push(`printf '%s\\n' '${UDEV_RULE}' > '${UDEV_RULE_PATH}'`);
   }
 
   if (needsGroup) {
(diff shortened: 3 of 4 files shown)


gabrielste1n (changes requested):
## Review

The goal here is solid — ydotool is genuinely needed for reliable Wayland paste — but there are some significant issues that need addressing before this can be merged.

---

### Security

**Shell injection in root commands** — `ensureUinputAccess()` interpolates `os.userInfo().username` into a string passed to `pkexec sh -c`:

```js
rootCommands.push(`usermod -aG input ${os.userInfo().username}`);
// later: spawnSync("pkexec", ["sh", "-c", rootCommands.join(" && ")])
```

A username containing shell metacharacters would execute arbitrary code as root. This should use `spawnSync` with proper argument arrays per command instead of `sh -c` with string concatenation.

**`pkill -f ydotoold`** kills any matching process system-wide, which could affect other users on multi-user systems. Similarly, deleting `/tmp/.ydotool_socket` could disrupt other processes.

---

### UX / Architecture

**Synchronous blocking on the main thread** — `ensureYdotool()` uses `spawnSync` with 120-second timeouts, called in `initializeDeferredManagers()`. Multiple pkexec calls can stack up and freeze the entire Electron UI for minutes. This needs to be async.

**Unannounced root escalation** — A polkit password dialog appears at startup with no prior user consent or explanation. There should be a dialog explaining what's needed and why before triggering pkexec.

**Repeats every launch** — There's no persistence to track whether setup was already completed. Users will see root password prompts on every startup until everything is perfectly configured.

**Silent failure** — If installation or setup fails, the only trace is a debug log. The user gets no indication of what went wrong or what to do.

**wl-clipboard promoted from Recommends to Depends** — This isn't mentioned in the PR description but could break package installation on distros/repos that don't carry it. Should be called out separately or kept as Recommends.

---

### Code Quality

- **Empty `catch {}` blocks** appear ~8 times throughout the file, silently swallowing errors with no logging. These make debugging nearly impossible.
- **`installCmd.split(" ")`** is fragile — works for current hardcoded strings but breaks if any argument ever contains spaces.
- **`debugLogger` is required at module load time** but may not be initialized yet when `main.js` first `require()`s this module at the top of the file.
- **No tests** for 346 lines of new code that modifies system state as root.
- The file is doing too many things (package installation, udev rules, systemd services, daemon management, process cleanup) — would benefit from being broken up.

---

### Suggestions

1. Use argument arrays instead of `sh -c` with string interpolation for root commands
2. Make everything async so the main thread isn't blocked
3. Show a user-facing dialog before triggering pkexec
4. Persist setup state so it doesn't re-run every launch
5. Replace empty catches with actual logging
6. Either keep wl-clipboard as Recommends or justify the change separately
7. Add test coverage for the detection and setup logic

gabrielste1n (commented):
## Re-review after changes

The major issues from my earlier review have been addressed — nice work. Specifically:

- **Shell injection** — Fixed. Username is now validated with a regex before use, and root commands go through a temp script file instead of `sh -c` with interpolation.
- **Async** — Fixed. Long operations use `execFileAsync` now, won't freeze the UI.
- **User consent** — Fixed. A dialog explains what's happening before pkexec fires.
- **Setup persistence** — Fixed. Flag file in userData prevents re-running on every launch.
- **wl-clipboard** — No longer promoted to a hard dep. Good.
- **Scoped cleanup** — Socket deletion checks ownership, `pkill -f` replaced with `systemctl --user stop`.
- **Module split** — Clean separation into orchestrator / installer / service files.
- **Error logging** — Empty catches replaced with actual log calls throughout.

---

### Remaining items

**1. Unconsented pkexec in `ensureYdotooldDaemon`**

`ensureYdotool()` asks for user consent before calling `installYdotool()`, but `ensureYdotooldDaemon()` in `ydotoolService.js:139` calls `pkexec apt-get install -y ydotoold` directly — bypassing the consent dialog. The user could see a second unexpected password prompt.

**2. Minor nits**

- `ensureYdotool().catch(() => {})` in main.js silently swallows any unexpected top-level error. Consider at least logging it: `.catch((err) => debugLogger.warn("ydotool setup error", { error: err.message }, "clipboard"))`.
- `UDEV_RULE_PATH` is unquoted in the shell script (`> ${UDEV_RULE_PATH}`). It's a constant with no spaces so it works, but quoting it costs nothing.
- `ydotoolInstaller.js` exports `commandExists` which is also used from `ydotoolService.js` — creates a circular-ish dependency. Consider a shared utils module if more functions end up shared.

PR #378, opened 2026-03-06
feat: add search/filter and custom model ID support to model selector

pablospe:
## Summary

- Add search/filter input to the reasoning model selector for quickly finding models in large lists (e.g., Vercel AI Gateway with 200+ models)
- Show model ID alongside the display name on cards when they differ (useful for custom endpoints where IDs like meta-llama/llama-4-scout-17b-16e-instruct aren't obvious from the label)
- Support custom model IDs via the same input field — type any model ID and press Enter or click the checkmark to use it, even if it's not in the fetched list (e.g., gpt-4o on an Azure BYOK endpoint)
- Add vertical scrolling to the model list with a max height
- Two-line card layout: name + ID on first line, description below

### Test plan

[x] Test custom model ID selection via Enter and checkmark button
[x] Verify custom endpoint with many models (e.g., https://ai-gateway.vercel.sh/v1) shows the filter input
[x] Verify local model picker is unaffected (no allowCustomModel passed)

### Screenshot

image

### DIFF
--- a/src/components/ReasoningModelSelector.tsx
+++ b/src/components/ReasoningModelSelector.tsx
@@ -859,6 +859,7 @@ export default function ReasoningModelSelector({
                       models={selectedCloudModels}
                       selectedModel={reasoningModel}
                       onModelSelect={setReasoningModel}
+                      allowCustomModel
                     />
                   </div>
                 </>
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Closing — the core need here (finding models in huge custom-endpoint lists) was addressed in #1002 with a searchable, grouped, virtualized model picker, and that refactor also moved the files this touches. Custom model IDs are supported separately for providers that allow them. Thanks!

PR #379, opened 2026-03-06
fix: CUDA and Vulkan binary download failures on Windows

egsok:
## Summary

Fixes CUDA and Vulkan GPU acceleration binaries failing to download/activate on Windows systems where the PowerShell `Microsoft.PowerShell.Archive` module cannot load.

**Root cause:** `Expand-Archive` relies on `Microsoft.PowerShell.Archive` which fails to auto-load on some Windows configurations, silently breaking all zip extraction in both `WhisperCudaManager` and `LlamaVulkanManager`.

**Additional issues fixed:**
- After successful CUDA/Vulkan download, the running whisper/llama server was not restarted, so the CPU binary continued to be used until app restart
- CUDA downloads had no concurrent download guard, allowing multiple parallel downloads from rapid UI clicks
- CUDA download errors were silently swallowed with no logging

### `src/helpers/whisperCudaManager.js`
- Replace `Expand-Archive` with `tar.exe` (built into Windows 10+), falling back to PowerShell if tar is unavailable
- Add recursive `_findFile()`/`_findFiles()` for binary search in extracted archives (matches existing `LlamaVulkanManager` pattern)
- Add `_downloading` flag to prevent concurrent downloads
- Restructure `try/finally` so the download flag resets on any error

### `src/helpers/llamaVulkanManager.js`
- Replace `Expand-Archive` with `tar.exe` + PowerShell fallback (same fix)

### `src/helpers/ipcHandlers.js`
- Restart whisper-server after CUDA binary download/delete
- Restart llama-server after Vulkan binary download
- Add error logging for CUDA and Vulkan download failures
- Add `downloading` state to CUDA status IPC response

## Test plan
- [ ] Download CUDA binary on Windows — verify extraction succeeds and GPU acceleration activates immediately
- [ ] Download Vulkan binary on Windows — verify extraction succeeds and GPU acceleration activates immediately
- [ ] Click download button rapidly — verify only one download runs at a time
- [ ] Delete CUDA/Vulkan binary — verify server falls back to CPU binary
- [ ] Cancel mid-download — verify clean state recovery
- [ ] Verify Linux extraction (unzip) still works unchanged

### DIFF
--- a/src/helpers/downloadUtils.js
+++ b/src/helpers/downloadUtils.js
@@ -3,6 +3,7 @@ const { promises: fsPromises } = require("fs");
 const path = require("path");
 const https = require("https");
 const http = require("http");
+const { execFile } = require("child_process");
 const { pipeline } = require("stream");
 const debugLogger = require("./debugLogger");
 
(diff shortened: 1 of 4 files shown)


gabrielste1n:
thank you sir @egsok !!

PR #380, opened 2026-03-06
fix: preserve images and HTML in clipboard during paste-and-restore

egsok:
## Summary
- Clipboard save/restore now handles **images** (screenshots), **HTML** (rich text from browser), and plain text — previously only plain text was preserved
- Added `_saveClipboard()` that detects content type via `clipboard.availableFormats()` and reads the appropriate format
- Added `_restoreClipboard()` helper that writes back the correct format after paste completes
- All 6 platform-specific restore callsites (macOS CGEvent, macOS osascript, Windows fast-paste, Windows nircmd, Windows PowerShell, Linux) updated to use the new helper

## Problem
If a user had a screenshot (Win+Shift+S) or rich HTML in the clipboard before dictation, `readText()` returned `""`, and after paste `writeText("")` overwrote the image/HTML with nothing.

## Test plan
- [ ] Copy a screenshot to clipboard → dictate → verify screenshot restores (Ctrl+V pastes the image)
- [ ] Copy plain text to clipboard → dictate → verify text restores
- [ ] Copy rich HTML from browser → dictate → verify HTML restores when pasting into a rich text editor
- [ ] Verify dictated text still gets pasted correctly into the active app

### DIFF
--- a/src/helpers/clipboard.js
+++ b/src/helpers/clipboard.js
@@ -450,6 +450,29 @@ class ClipboardManager {
     }
   }
 
+  _saveClipboard() {
+    const formats = clipboard.availableFormats();
+    if (formats.some((f) => f.startsWith("image/"))) {
+      return { type: "image", data: clipboard.readImage() };
+    } else if (formats.includes("text/html")) {
+      return { type: "html", text: clipboard.readText(), html: clipboard.readHTML() };
+    } else {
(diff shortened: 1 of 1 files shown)


egsok:
Closing — branch accidentally included unrelated commits. Reopening with a clean branch.

Issue #290, opened 2026-02-18
Can cancel an voice record by pressing a button (like esc button)


N0amG:
**Is your feature request related to a problem? Please describe.**
It would be very usefull to cancel a recording by pressing a button (like escape button)

**Describe the solution you'd like**
If i start recording my voice, and i change my mind, i want to click on esc button to cancel the recording and not processing it

**Describe alternatives you've considered**
If the button can be custom, that's even better

gabrielste1n:
@N0amG if you hover over the recorder while recording - there is a cancel button there, is that fine?

N0amG:
@gabrielste1n  Honestly.... No it's very awful lmao, the cancel button is ridiculously small, keep it but make it bigger. And click on esc button to cancel a recording is primordial. Like just yesterday I had to cancel like 3/4 recording, and my mouse has to move like entire 2 screen for me to go on the bottom right of my right ultrawide screen 😭 (there's literally 10 miles to move with mouse to be over the recorder 😂)

gabrielste1n:
@N0amG I do agree with you, will think of the best implementation and try get it out in the next few days.

N0amG:
@gabrielste1n nice don't hesitate if you have question or hesitation on the UX or need some feedback I'm here 👍

And I would love to do it myself and try (my first) pull request (on a public repo) but I have like no time for personal project at this time 😭

I hope I could get some for trying to help you a bit, the project is really nice

gabrielste1n:
@N0amG really appreciate it! can you let me know - do you use the tap to talk (hands-free) or the push to talk mode?

N0amG:
@gabrielste1n Hey I just saw the 1.5.0 GG ! I preferred the old UI but good to see you experiment changes ! (I think you should do a resizeable interface, like in Minecraft you can chose you're interface size between 1 and 5 for eg and it's bigger and bigger).

I use the tap to talk mode not the hold (not sure of what you're talking about but I press the shortcut oncd for starting record and once again for ending it)

Another suggestion, the close button work nicely ! But is this possible to make an option to chose to not display the windows when you start the app at launch ? (Like just launch it in the Tray but not open the windows when you launch your pc ? Thanks you !!)

gabrielste1n:
@N0amG what did you prefer about the old UI? 

Great that's what i wanted to know thanks!

Good idea - will think about it!|

[...]

PR #391, opened 2026-03-07
feat: add cancel recording hotkey (Escape by default, configurable)

N0amG:
### Summary
As describe in the issue #290
Pressing a configurable key (default: **Escape**) during an active recording cancels it without transcribing. Outside of a recording, the key behaves normally system-wide.

---

### How it works

The cancel hotkey is registered as a `globalShortcut` **only while recording is active**, then immediately unregistered on stop, cancel, or unexpected disconnect. Zero system-wide footprint at rest.

---

#### `src/helpers/ipcHandlers.js`
- `register-cancel-hotkey` — calls `hotkeyManager.registerSlot("cancel", key, cb)` (reuses existing slot system)
- `unregister-cancel-hotkey` — calls `hotkeyManager.unregisterSlot("cancel")`

#### `preload.js`
- Exposes `registerCancelHotkey`, `unregisterCancelHotkey`, `onCancelHotkeyPressed` to the renderer via the existing `registerListener` pattern

#### `src/hooks/useAudioRecording.js`
- `performStartRecording()` — registers the cancel hotkey on successful start
- `performStopRecording()` — unregisters it before stopping
- `cancelRecording()` — unregisters it before cancelling
- `onStateChange` — **bug fix**: unregisters if `isRecording` drops to `false` unexpectedly (e.g. streaming server disconnect), preventing a permanent Escape intercept

#### `src/App.jsx`
- Listens for `onCancelHotkeyPressed` and calls `cancelRecording()` (guarded by `isRecording`)

#### `src/stores/settingsStore.ts`
- `cancelKey` — persisted to `localStorage`, default `"Escape"`, no IPC side-effects (key is read at recording start, not at boot)

#### `src/hooks/useSettings.ts`
- Exposes `cancelKey` / `setCancelKey` through `SettingsContext`

#### `src/components/SettingsPage.tsx`
- New `HotkeyInput` for the cancel key under **Settings → Shortcuts**
- Reset-to-default button shown when value differs from `"Escape"`

#### `src/components/ui/HotkeyInput.tsx`
- New prop `skipListeningMode` — when `true`, focusing the input does **not** suspend/resume the dictation hotkey (used for the cancel key input to avoid interference)
- Adds `data-capturing` attribute while capturing, used by `SidebarModal` below

#### `src/components/ui/SidebarModal.tsx`
- `onEscapeKeyDown` on `DialogPrimitive.Content` — prevents Radix from closing the modal when Escape is being captured by a `HotkeyInput`

#### `src/locales/*/translation.json` (10 languages)
- New keys for the cancel hotkey UI label, description, and reset tooltip

---

### Architecture note

The cancel key deliberately uses `globalShortcut` (same as the dictation key) rather than a DOM `keydown` listener, so it works even when OpenWhispr is in the background. It reuses the existing `hotkeyManager` slot system — no new infrastructure was needed.

---

### Testing checklist

- [x] Escape cancels an active recording (no transcription triggered)
- [x] Escape works normally outside of recording
- [x] Custom cancel key can be set in Settings → Shortcuts
- [x] Cancel key resets to Escape via the reset button
- [x] Streaming disconnect should not leave the cancel key registered globally
- [x] Settings modal does not close when pressing Escape inside the cancel key input when registering hotkey
- [X] Dictation hotkey is not disrupted when focusing the cancel key input


RayBB:
FYI this has merge conflicts now you may want to resolve

N0amG:
> FYI this has merge conflicts now you may want to resolve

Oh yeah thanks, I guess the merge conflict appear after few updates, because I don't remember of them when I did the PR. Thanks, I'll take a look

Issue #310, opened 2026-02-23
Setup Guide for Ubuntu 24.04 LTS


JosefJezek:
# OpenWhispr on Ubuntu 24.04 (GNOME Wayland) — Setup Guide

## Problem

OpenWhispr's auto-paste doesn't work out of the box on Ubuntu 24.04 due to two issues:

1. **ydotoold** (the daemon for ydotool) is not running and has no systemd service in the Ubuntu package
2. **ydotool 0.1.8** (shipped with Ubuntu 24.04) doesn't understand the raw keycode format (`29:1 47:1 47:0 29:0`) that OpenWhispr sends — it expects key names (`ctrl+v`). This causes garbage characters like `2442` to be typed instead of pasting.

### 1. Install OpenWhispr

```bash
# Download and install the latest .deb
gh release download v1.4.11 --repo OpenWhispr/openwhispr --pattern "*linux-amd64.deb" --dir /tmp
sudo apt install -y /tmp/OpenWhispr-*-linux-amd64.deb
```

### 2. Install dependencies

```bash
sudo apt install -y ydotool ydotoold wl-clipboard
```

### 3. Fix /dev/uinput permissions

ydotool needs write access to `/dev/uinput`. By default only root can access it.

```bash
# Create udev rule
echo 'KERNEL=="uinput", GROUP="input", MODE="0660", TAG+="uaccess"' | sudo tee /etc/udev/rules.d/80-uinput.rules

# Add your user to the input group
sudo usermod -aG input $USER

# Apply immediately (persists after reboot via udev rule + group membership)
sudo udevadm control --reload-rules
sudo udevadm trigger /dev/uinput
```

### 4. Create ydotoold systemd user service

Ubuntu 24.04's ydotoold package doesn't include a systemd service file.

```bash
mkdir -p ~/.config/systemd/user

cat > ~/.config/systemd/user/ydotoold.service << 'EOF'
[Unit]
Description=ydotoold - ydotool daemon
Documentation=man:ydotoold(8)

[Service]
ExecStart=/usr/bin/ydotoold
Restart=on-failure
RestartSec=3

[Install]
WantedBy=default.target
EOF

systemctl --user daemon-reload
systemctl --user enable ydotoold
systemctl --user start ydotoold
```

If it fails with a socket error, kill any stale root-owned instances first:

```bash
sudo pkill ydotoold
sudo rm -f /tmp/.ydotool_socket
systemctl --user start ydotoold
```

[...]

PR #397, opened 2026-03-08
fix: set cloudTranscriptionMode to BYOK during onboarding for non-signed-in users (local but remote)

ianbmacdonald:
## Summary
- Non-signed-in users completing the setup wizard in cloud mode were left with `cloudTranscriptionMode` defaulting to `"openwhispr"`, causing transcription to fail with "OpenWhispr Cloud requires sign-in".
- This affected both BYOK API key users and users pointing at local inference appliances on another device (e.g. lemonade-server.ai) via a custom base URL.
- Sets `cloudTranscriptionMode: "byok"` in two places: on mode toggle and as a safety net in `saveSettings`.

## Test plan
- [x] Tested on Ubuntu 25.10
- [x] Fresh onboarding — skip sign-in — cloud mode with custom base URL — transcription routes to custom endpoint
- [ ] Signed-in users unaffected (OpenWhispr Cloud mode still works)

## Rationale
Without this, you get to the end of the setup wizard, and it says you need to switch to BYOK mode, which isn't possible in the setup wizard.  I actually couldn't figure out where the BYOK mode is set at all for Custom Endpoint URL, as even putting in a key did not change the outcome; Ultimately I was not interested in BYOK but the error doesn't really line up with any flag or setting explicitly, as both cloud and custom endpoints could use an API key. 

On the upside, this now just works on Ubuntu 25.10, without any of the additional steps described in #310 ; no permission changes, or ydotoold upgrades, which I had performed on my other laptop.  Neither laptop have any AI capabilities, so a local Strix Halo appliance does all the work running lemonade-server.ai   

image

Since lemonade supports a bunch of llm backends concurrently, in this case I can choose to use the whistper.cpp (GPU) or FastFlowLM (NPU) engine using the default quants from lemonade for either engine+checkpoint of the Wisperv3Turbo model, or any of the many others available from huggingface. 

```xml
http://1291.68.79.18:8000/api/v1/models
...
    {
      "checkpoint": "ggerganov/whisper.cpp:ggml-large-v3-turbo.bin",
      "checkpoints": {
        "main": "ggerganov/whisper.cpp:ggml-large-v3-turbo.bin",
        "npu_cache": "amd/whisper-large-turbo-onnx-npu:ggml-large-v3-turbo-encoder-vitisai.rai"
      },
      "composite_models": [],
      "created": 1234567890,
      "downloaded": true,
      "id": "Whisper-Large-v3-Turbo",
      "labels": [
        "audio",
        "transcription",
        "hot"
      ],
      "object": "model",
      "owned_by": "lemonade",
      "recipe": "whispercpp",
      "recipe_options": {

      },
      "size": 1.55,
      "suggested": true
    },
...
    {
      "checkpoint": "whisper-v3:turbo",
      "checkpoints": {
        "main": "whisper-v3:turbo"
      },
      "composite_models": [],
      "created": 1234567890,
      "downloaded": true,
      "id": "whisper-v3-turbo-FLM",
      "labels": [
        "audio",
        "transcription"
      ],
      "object": "model",
      "owned_by": "lemonade",
      "recipe": "flm",
      "recipe_options": {

      },
      "size": 0.62,
      "suggested": true
    }
    ...
```
    
## Comments

One feature that builds on the current UI hooks, that might be a good fit for expansion is to do TTS using the Kokoro backend, further expanding local capabilities to match the expanding model and backend frameworks available via llmaid.  

```
    {
      "checkpoint": "mikkoph/kokoro-onnx",
      "checkpoints": {
        "main": "mikkoph/kokoro-onnx"
      },
      "composite_models": [],
      "created": 1234567890,
      "downloaded": true,
      "id": "kokoro-v1",
      "labels": [
        "tts",
        "speech"
      ],
      "object": "model",
      "owned_by": "lemonade",
      "recipe": "kokoro",
      "recipe_options": {

      },
      "size": 0.34,
      "suggested": true
    }
```

### DIFF
--- a/src/components/OnboardingFlow.tsx
+++ b/src/components/OnboardingFlow.tsx
@@ -268,12 +268,8 @@ export default function OnboardingFlow({ onComplete }: OnboardingFlowProps) {
     localStorage.setItem("onboardingCompleted", "true");
     localStorage.setItem("skipAuth", skippedAuth.toString());
 
-    // Non-signed-in users using cloud mode need BYOK mode set,
-    // otherwise audioManager routes to OpenWhispr Cloud and demands sign-in.
-    // This covers users with their own API keys as well as users pointing at
-    // local inference appliances on another device (e.g. lemonade-server.ai)
-    // via a custom base URL — a future UI update should better surface this
-    // "local-but-remote" use case.
+    // Non-signed-in users in cloud mode default to BYOK to avoid
+    // "OpenWhispr Cloud requires sign-in" errors.
     if (!isSignedIn && !useLocalWhisper) {
       updateTranscriptionSettings({ cloudTranscriptionMode: "byok" });
     }
@@ -285,7 +281,7 @@ export default function OnboardingFlow({ onComplete }: OnboardingFlowProps) {
     }
 
     return true;
-  }, [hotkey, agentName, setDictationKey, ensureHotkeyRegistered]);
+  }, [hotkey, agentName, setDictationKey, ensureHotkeyRegistered, isSignedIn, useLocalWhisper, updateTranscriptionSettings]);
 
   const nextStep = useCallback(async () => {
     if (currentStep >= steps.length - 1) {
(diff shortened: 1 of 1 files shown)


gabrielste1n:
thanks legend @ianbmacdonald

PR #427, opened 2026-03-13
Bundle Noto Sans fonts locally instead of loading from Google Fonts CDN

jrschumacher:
Embeds Noto Sans woff2 variable-weight files (300-700, normal+italic, latin+latin-ext) and replaces the three Google Fonts CDN links in index.html with a local stylesheet. This improves privacy by eliminating outbound requests to fonts.googleapis.com and fonts.gstatic.com, and enables full offline functionality.


gabrielste1n:
this is a great idea thanks @jrschumacher - merged!

PR #434, opened 2026-03-14
Add force meeting mode hotkey

xAlcahest:
## Summary

- Add a dedicated hotkey (Ctrl+Shift+M by default) to manually start meeting mode, skipping the automatic "Meeting Detected" notification
- Same hotkey toggles meeting mode off
- While manual meeting mode is active, automatic meeting detection notifications are suppressed
- Hotkey is configurable in Settings → Hotkeys
- Meeting hotkey persisted to .env across restarts
- i18n strings added across all 9 supported languages


gabrielste1n:
Thanks for the contribution! We went in a slightly different direction with this feature — implemented in `feature/meeting-mode-hotkey` as a one-way trigger rather than a toggle.

Key differences:
- **No toggle state**: The hotkey starts meeting mode; the user stops recording normally (no second press needed, no `_manualMeetingMode` flag)
- **No detection suppression logic**: The existing `_userRecording` flag already suppresses auto-detection while recording, so no additional state management is needed
- **No extra IPC handles**: Since the hotkey fires directly in `main.js`, the `force-meeting-mode` / `end-manual-meeting-mode` IPC endpoints aren't required
- **No default hotkey**: Starts empty to avoid stealing keybindings from other apps

Appreciate the i18n coverage and the clean hotkey registration pattern — both informed the final implementation.

PR #436, opened 2026-03-14
fix: register agent hotkey as independent GNOME Wayland keybinding slot

alumpe:
On GNOME Wayland, Electron's globalShortcut API is blocked by the Wayland security model. OpenWhispr works around this for the dictation hotkey via native GNOME keybindings (gsettings + D-Bus), but the agent hotkey was never wired into that same system — so it silently failed to register on GNOME Wayland.

## Changes

> [!IMPORTANT]  
> `registerSlot()` had to be made `async` because the GNOME keybinding path calls into `gnomeManager.registerKeybinding()`, which is `async`. Without awaiting it, callers would receive an unresolved Promise instead of the actual `{ success, message }` result, silently swallowing registration failures.

- `gnomeShortcut.js`: Added per-slot gsettings paths (openwhispr for dictation, openwhispr-agent for agent) and a ToggleAgent D-Bus method alongside the existing Toggle, so each hotkey slot gets its own independent GNOME keybinding
- `hotkeyManager.js`: `registerSlot()` is now async and routes non-dictation slots through GNOME keybindings instead of globalShortcut
- `ipcHandlers.js` / `main.js`: Agent hotkey registration now awaits the result and handles failures
- `settingsStore.ts`: `setAgentKey()` uses the new updateAgentHotkey invoke (request/response) instead of fire-and-forget `notifyAgentHotkeyChanged`, and rolls back the stored value on failure
- `preload.js` / `electron.ts`: New updateAgentHotkey IPC bridge with typed return value

**Before**  
Agent hotkey appeared to save but never triggered on GNOME Wayland.

**After**  
Both dictation and agent hotkeys register as separate named shortcuts, visible in GNOME Settings → Keyboard → Shortcuts → Custom.

### DIFF
--- a/main.js
+++ b/main.js
@@ -580,27 +580,6 @@ async function startApp() {
     }
   }
 
-  ipcMain.on("agent-hotkey-changed", async (_event, hotkey) => {
-    try {
-      if (hotkey) {
-        const result = await hotkeyManager.registerSlot("agent", hotkey, agentHotkeyCallback);
-        if (result.success) {
-          environmentManager.saveAgentKey(hotkey);
-        } else {
-          debugLogger.warn("Failed to update agent hotkey", { hotkey }, "hotkey");
-        }
-      } else {
-        hotkeyManager.unregisterSlot("agent");
-        environmentManager.saveAgentKey("");
-      }
-    } catch (error) {
-      debugLogger.warn(
-        "Failed to update agent hotkey",
-        { error: error instanceof Error ? error.message : String(error), hotkey },
-        "hotkey"
-      );
-    }
-  });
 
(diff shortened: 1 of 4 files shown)


gabrielstein-mem0:
Nice work! The per-slot GNOME keybinding architecture is clean and the async error propagation is a solid improvement over the silent failures before.

Pushed a small follow-up commit that:
- Restores the **previous** agent hotkey on registration failure instead of blanking it (so users don't lose a working hotkey if a change fails)
- Removes the old fire-and-forget `agent-hotkey-changed` IPC channel since `update-agent-hotkey` fully replaces it

LGTM — ready to merge.

PR #450, opened 2026-03-15
fix: use LCONTROL instead of CONTROL for Windows native paste

### LINKED ISSUE #438: Speech-to-text outputs only 'v' in Windows Terminal / Claude Code CLI
## Description

Speech-to-text transcription does not work in Windows Terminal or Claude Code CLI. After recording and processing, only the character `v` is inserted instead of the transcribed text. The transcription itself is correct — it lands in the clipboard and can be pasted manually with Ctrl+V.

This appears to be the same class of issue as #184, but on Windows rather than Linux/KDE Wayland.

## Environment

- **OS:** Windows 11
- **Terminal:** Windows Terminal with WSL/Ubuntu running Claude Code CLI
- **Whisper model:** Local, offline

## Steps to reproduce

1. Open Windows Terminal with a WSL session running Claude Code
2. Press Ctrl+Win to start recording
3. Speak
4. Press Ctrl+Win to stop recording
5. Wait for transcription to process
6. Only `v` appears in the terminal input

## What I've tried

- **Clipboard mode ON:** Same result — just `v`
- **Clipboard mode OFF:** Same result — just `v`
- Windows Terminal has Ctrl+V, Ctrl+Shift+V, and Shift+Insert all bound to paste

## Expected behavior

Transcribed text should be inserted into the active terminal input.

## Actual behavior

Only `v` is inserted. The transcribed text is in the clipboard and can be manually pasted with Ctrl+V, confirming the transcription itself works correctly.

## Notes

Likely the same root cause as #184 — SendInput simulating Ctrl+V, but Windows Terminal not receiving the Ctrl modifier from simulated keystrokes, so only `v` comes through.

Sigmabrogz:
## Summary
Fixes an issue where transcribed text is not properly pasted into terminal emulators like Windows Terminal when using `windows-fast-paste.exe`.

## Problem
Fixes #438.
The `windows-fast-paste` helper tool used the generic `VK_CONTROL` virtual key code for `SendInput`. Many terminal emulators (like Windows Terminal running WSL or terminal-based agents like Claude Code) do not interpret the generic `VK_CONTROL` key down events correctly, resulting in the `v` character being typed literally instead of triggering the `Ctrl+V` or `Ctrl+Shift+V` clipboard paste mechanism.

## Solution
Modified `windows-fast-paste.c` to use the explicit left-sided modifiers (`VK_LCONTROL` and `VK_LSHIFT`) instead of generic modifiers. Terminals correctly recognize specific modifier keys via `SendInput`, ensuring the clipboard contents are properly pasted rather than inserting the literal character `v`.

### DIFF
--- a/resources/windows-fast-paste.c
+++ b/resources/windows-fast-paste.c
@@ -95,10 +95,10 @@ static int SendPasteNormal(void) {
     INPUT inputs[4];
     ZeroMemory(inputs, sizeof(inputs));
 
-    SetKey(&inputs[0], VK_CONTROL, 0);
+    SetKey(&inputs[0], VK_LCONTROL, 0);
     SetKey(&inputs[1], 'V', 0);
     SetKey(&inputs[2], 'V', KEYEVENTF_KEYUP);
-    SetKey(&inputs[3], VK_CONTROL, KEYEVENTF_KEYUP);
+    SetKey(&inputs[3], VK_LCONTROL, KEYEVENTF_KEYUP);
 
     UINT sent = SendInput(4, inputs, sizeof(INPUT));
     return (sent == 4) ? 0 : 1;


Sigmabrogz:
This addresses the issue with the `v` character being typed instead of pasting in terminal emulators by using specific left-sided modifier keycodes for `SendInput`.

gabrielstein-mem0:
Thanks for identifying this issue @Sigmabrogz! After investigating further, the `VK_CONTROL` → `VK_LCONTROL` swap alone doesn't fix the problem — both produce identical `INPUT` structs since `MapVirtualKeyA` maps them to the same scancode (0x1D).

It seems that the actual root cause may be that the user's hotkey modifiers (e.g. Ctrl+Win) can still be held when `SendInput` fires, contaminating the paste keystroke. Per the [SendInput docs](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-sendinput): *"Any keys that are already pressed when the function is called might interfere with the events that this function generates."*

Opened #473 which fixes this properly by releasing held modifiers before the paste (same approach AutoHotkey uses), and also includes the `VK_LCONTROL`/`VK_LSHIFT` change for both paste functions. Closing in favor of that — thanks again for the report! Let me know if it works for you once released in v1.6.6.

PR #465, opened 2026-03-18
Add live waveform and active status pill to dictation panel

jrschumacher:
## Summary
- add a live waveform visualizer for active dictation
- add an optional floating icon setting that expands recording/transcribing into a pill with text
- resize the dictation panel for the active pill state and wire the setting through the app store

## Notes
- the ESLint run still reports an existing unrelated warning in SettingsPage about a missing useEffect dependency

CleanShot 2026-03-18 at 07 18 40

PR #474, opened 2026-03-18
Add live waveform and active status pill to dictation panel

jrschumacher:
## Summary
- add a live waveform visualizer for active dictation
- add an optional floating icon setting that expands recording/transcribing into a pill with text
- resize the dictation panel for the active pill state and wire the setting through the app store

## Notes
- the ESLint run still reports an existing unrelated warning in SettingsPage about a missing useEffect dependency

CleanShot 2026-03-18 at 07 18 40


JiwaniZakir (commented):
In `cloudfree-ci.yml`, both `npm install --force` and the separate `npm install tinyexec --force` steps should raise concern: `--force` bypasses lockfile integrity checks, which defeats the purpose of reproducible CI builds. The standard practice is `npm ci`, which installs exactly what's in `package-lock.json` and will fail fast if there's a mismatch. The fact that `tinyexec` needs its own install step suggests it isn't committed as a proper `devDependency`, which means the lockfile doesn't pin its version — any future CI run could silently pull a different version of that package.

In `cloudfree-release.yml`, the `prepare` job uses `if: always() && (needs.release-please.outputs.release_created == 'true' || ...)`. The `always()` guard means `prepare` (and consequently the build jobs) will proceed even if `release-please` *fails*, not just when it's skipped. A tighter condition like `if: needs.release-please.result != 'failure' && (...)` would prevent spurious builds from a broken release job while still allowing the `workflow_dispatch` path to work correctly.

gabrielste1n:
Closing as a duplicate of #465 — same feature, but this one is filed from your fork's main branch and pulls in a lot of fork-specific infrastructure that isn't upstreamable. Let's keep the discussion on #465.

PR #478, opened 2026-03-19
Fix cleanup prompt refusing command-like transcriptions

NathanSkene:
## Summary

- Cleanup mode refuses to output transcribed text that sounds like a command or question (e.g., "can you write me an email about..."), returning an apology instead of the cleaned transcription
- The root cause is the instruction "Do NOT create, draft, translate, or generate new content" — the LLM interprets the user's dictated speech as an instruction to itself
- Fix clarifies that cleanup mode should always output the cleaned text, since commands/questions in the transcription are what the speaker said, not instructions to the LLM

## Reproduction

1. Dictate something like "can you help me draft an email to my boss"
2. Without agent mode enabled (no agent name detected), cleanup prompt is used
3. LLM returns: "I'm sorry, but I can't write or generate new content. I can only clean up transcribed speech."

## Test plan

- [ ] Dictate command-like text without agent name → should output cleaned text
- [ ] Dictate normal speech → should continue to clean up as before
- [ ] Dictate with agent name → should still trigger agent mode correctly


gabrielste1n:
thanks @NathanSkene

PR #492, opened 2026-03-23
feat: Add model-aware reasoning effort overrides

Sevi-py:
## Summary
- add configurable reasoning-effort overrides for BYOK Intelligence and Agent Mode
- show the selector only for cloud models that expose supported effort levels
- persist separate overrides for Intelligence and Agent Mode
- send valid overrides to OpenAI-compatible and Gemini requests while leaving provider defaults in place when no override is selected

## How options are determined
- built-in model capabilities are defined in `src/models/modelRegistryData.json` from provider documentation
- custom endpoint capabilities are parsed dynamically from `/models` metadata when those endpoints expose supported reasoning-effort fields
- Anthropic is excluded because the API exposes extended thinking via token budgets rather than effort enums

image

image

### DIFF
--- a/src/components/AgentOverlay.tsx
+++ b/src/components/AgentOverlay.tsx
@@ -103,7 +103,10 @@ export default function AgentOverlay() {
             llmMessages,
             settings.agentModel,
             settings.agentProvider,
-            { systemPrompt }
+            {
+              systemPrompt,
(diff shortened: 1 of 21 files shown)


gabrielste1n:
this is a great idea @Sevi-py - will review asap

PR #451, opened 2026-03-16
feat: agent mode tool-calling via Vercel AI SDK

gabrielste1n:
## Summary

Full agent mode implementation with Vercel AI SDK tool-calling, embedded chat, local semantic search, and comprehensive UX polish.

### Agent & Chat System
- Migrate agent mode to Vercel AI SDK with full tool-calling (search notes, create/update notes, web search)
- Add sidebar chat tab with conversation history, cloud sync, and semantic search
- Embedded chat panel in note editor with floating and sidebar dock modes
- Redesign thinking/streaming state with shimmer gradient indicator and stop button
- Stream cloud agent responses via NDJSON with rich tool step visualization
- Enable local model tool calling with RAG context injection

### Local Semantic Search
- Always-on offline semantic search via Qdrant sidecar (cosine similarity, 384-dim MiniLM embeddings)
- Hybrid search: FTS5 + Qdrant in parallel with Reciprocal Rank Fusion (K=60)
- Auto-index notes on create/update/delete, auto-download embedding model on first launch

### Notes & Meeting Improvements
- Add Actions sidebar button and redesign action editor dialog (master-detail two-panel layout)
- Add folder picker to note metadata row, new note dialog with folder selection
- Prevent meeting view from exiting when changing folders
- Fix auto-detection not firing for browser meetings
- Fix partial transcript spam and duplicate final segments
- Use local transcription provider for notes/meeting recording
- Enable echo cancellation on mic stream

### Linux & Platform Fixes
- Force XWayland on KDE/GNOME Wayland for window positioning
- Use uinput before portal on GNOME Wayland
- Fix hotkey startup race condition

### UI Polish
- Align all dialogs with design system guidelines
- Reuse CommandSearch for chat search
- Increase sidebar search and note metadata visibility
- Remove vertical accent bar selection indicators

PR #498, opened 2026-03-25
fix(linux): fix KDE hotkey not registering at startup

xAlcahest:
## Summary

Based on c6d3cf8 from #495.

On KDE Wayland with Electron 39, the hotkey didn't work until you manually re-set it in settings. KGlobalAccel D-Bus silently accepted the registration but never fired the shortcut signal.

The fix forces XWayland on KDE (where globalShortcut works via X11) and loads the saved hotkey from the userData .env early enough to register it before the page finishes loading.

**What changed:**
- main.js: self-relaunch with --ozone-platform=x11 on KDE Wayland, plus early load of userData .env so DICTATION_KEY is available at hotkey init time
- hotkeyManager.js: skip KGlobalAccel D-Bus path when running on XWayland, register hotkey directly from env var instead of waiting for did-finish-load (which never fired)
- kdeShortcut.js: use flag 0 in setShortcut to always overwrite stale KGlobalAccel state
- clipboard.js: longer restore delay on KDE Wayland (600ms) to handle the slow X11-to-Wayland clipboard bridge
- run-electron.js: inject --ozone-platform=x11 before the app path in dev mode, so concurrently doesn't get killed by the self-relaunch
- tsconfig.json: add typeRoots and types to fix phantom type errors

### DIFF
--- a/main.js
+++ b/main.js
@@ -1,7 +1,6 @@
-// KDE Wayland: force XWayland before Electron initializes Chromium.
-// On Electron 39+, appendSwitch("ozone-platform-hint") is too late and
-// process.argv.push() doesn't propagate to native argc/argv. The only way
-// is to self-relaunch with the flag on the actual command line.
+// KDE Wayland: force XWayland so globalShortcut works via X11.
+// Must be a command-line flag because Chromium picks the display backend
+// before main.js can call appendSwitch.
 if (
   process.platform === "linux" &&
   process.env.XDG_SESSION_TYPE === "wayland" &&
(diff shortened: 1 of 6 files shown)


gabrielste1n:
Closing — all fixes from this PR have been superseded by the work in #451 (`feat/agent-tool-calling`), which includes:

- XWayland self-relaunch guard (extended to cover GNOME/Ubuntu/Unity in addition to KDE)
- Early userData `.env` loading for `DICTATION_KEY`
- Skip KGlobalAccel D-Bus path when running on XWayland
- Register hotkey directly from env var instead of waiting for page load
- `kdeShortcut.js` flag `0` to overwrite stale KGlobalAccel state
- 600ms clipboard restore delay for KDE Wayland
- `run-electron.js` XWayland injection for dev mode (KDE + GNOME)

Thanks for the contribution!

PR #503, opened 2026-03-26
feat: add client-side audio chunking for large file transcription

NathanSkene:
## Summary
- Splits large audio recordings into smaller chunks for transcription
- Enables transcription of longer recordings without API timeout issues
- Chunks are processed sequentially and results concatenated

## Test plan
- [ ] Record a long dictation (>30 seconds)
- [ ] Verify transcription completes without timeout
- [ ] Verify output text is coherent across chunk boundaries


NathanSkene:

This feature passed validation.

What I tested:
- a genuinely long BYOK transcription using meeting audio
- the transcription completed successfully instead of failing or timing out
- the output stayed coherent across the full recording without obvious chunk-boundary corruption (no abrupt truncation, duplicated joins, or missing blocks)

Follow-up note: the output still needs better formatting / paragraphing for long meeting transcripts, but that is a separate UX/output-quality issue rather than a failure of the chunking feature itself.

Issue #462, opened 2026-03-18
Whisper hallucinations ("Thank you for watching!") during silence — needs smarter VAD for cloud backends


NathanSkene:
## Problem

When using OpenWhispr with Groq's Whisper API, the transcribed text randomly includes hallucinated phrases that were never spoken, such as:

- "Thank you for watching!"
- "Thanks for watching"
- "Please subscribe"
- "Subtitles by..."

This is a well-documented upstream Whisper problem ([openai/whisper#679](https://github.com/openai/whisper/discussions/679), [openai/whisper#1873](https://github.com/openai/whisper/discussions/1873)). Whisper was trained on YouTube videos with subtitles, so during silence or low-energy audio segments the model fills in phrases from its training data.

With local Whisper, users can tune `no_speech_threshold` and related parameters to filter these. Cloud APIs like Groq don't expose those controls, so the fix needs to be client-side.

## Recommended solution: VAD with gap preservation (relates to #185)

The root cause is long silence reaching the Whisper decoder. The fix is trimming silence before sending audio to the API — but naive silence removal would destroy the pause structure that Whisper relies on for sentence boundaries and punctuation.

A speech-structure-preserving approach:

1. **Segment audio** into voiced/silent regions using VAD (e.g. Silero VAD)
2. **Short gaps** (< ~1.5s): keep as-is — these are natural sentence and clause pauses that Whisper uses for punctuation
3. **Long gaps** (> ~1.5s): trim down to a fixed short pause (e.g. 0.3–0.5s) — preserves the "there was a pause here" signal without giving the decoder enough silence to hallucinate

This way Whisper still receives pause cues for punctuation and structure, but never sees the extended silence windows that trigger hallucination. It would also reduce API latency and cost (less audio to transmit and process).

The thresholds would need some tuning, but Silero VAD is well-tested for this and faster-whisper already uses it successfully.

## Quick interim fix: `prompt` parameter conditioning

Groq's Whisper API accepts a `prompt` parameter that conditions the decoder. Setting it to something like `"Voice dictation transcript."` when using cloud transcription backends can steer the model away from YouTube-style completions. Low effort, partial mitigation.

## Environment

- **OS:** macOS
- **Transcription backend:** Groq (whisper-large-v3-turbo)
- **OpenWhispr version:** latest

m13v:
[...]

PR #504, opened 2026-03-26
feat: add VAD-based silence compression for efficient transcription

NathanSkene:
## Summary
- Adds voice activity detection (VAD) to compress silence gaps in recordings
- Reduces audio sent to transcription API, improving speed and cost
- New VAD toggle in Settings (disabled by default)

## Test plan
- [ ] Enable VAD in Settings
- [ ] Record with natural pauses between sentences
- [ ] Verify transcription quality is maintained
- [ ] Verify silence gaps are compressed in the processed audio
- [ ] Disable VAD - verify normal recording behavior unchanged

### DIFF
--- a/src/components/SettingsPage.tsx
+++ b/src/components/SettingsPage.tsx
@@ -695,6 +695,8 @@ export default function SettingsPage({ activeSection = "general" }: SettingsPage
     setCloudReasoningMode,
     audioCuesEnabled,
     setAudioCuesEnabled,
+    vadEnabled,
+    setVadEnabled,
     pauseMediaOnDictation,
     setPauseMediaOnDictation,
     keepTranscriptionInClipboard,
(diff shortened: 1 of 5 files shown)


NathanSkene:
Manual validation result on macOS with the packaged installed app: this PR is not ready.

What we observed:
- Original branch behavior: after speaking briefly and then leaving ~10 seconds of silence, Groq/Whisper still hallucinated phrases that were never spoken.
- A stricter local experiment reduced the natural-language hallucination, but the output degraded into gibberish instead of becoming correct.

Representative failures during silence testing:
- "Sellers dining in the office building. It was very hot all the time."
- "ZZO3 Rewd0 ZZO3 R4D0 ZZT4 ZZT4 OLc00"

Conclusion:
- The current RMS-based gap compression approach does not reliably solve the issue described in #462.
- It appears to be an approximation of the proposed solution, not the full thing discussed there. In particular, it is not using a stronger speech-segmentation / noise-rejection pipeline like the one described in the issue comments.

Recommendation:
- Do not merge this PR as-is.
- Keep #462 open and treat this PR as failed validation / insufficient implementation rather than completed work.

PR #505, opened 2026-03-26
feat: add configurable maximum dictation duration

NathanSkene:
## Summary
- Adds a configurable maximum recording duration setting
- Recording automatically stops when the limit is reached
- Prevents accidentally long recordings that may fail transcription

## Test plan
- [ ] Set max duration in Settings
- [ ] Start recording and wait for limit
- [ ] Verify recording stops automatically at the configured duration
- [ ] Verify transcription processes normally after auto-stop


NathanSkene:

This feature passed validation after fixing a gap in the original implementation.

What I verified in the real installed app:
- setting a max dictation duration causes recording to stop automatically at the configured limit
- transcription still completes normally after the auto-stop

Important implementation note:
- the original branch did not wire max-duration enforcement into the real hotkey/floating recorder path, only into the notes widget path
- I patched the real recorder path locally during validation and confirmed the intended behavior there

So the user-facing feature is valid, but the working version is the one with the recorder-path fix included.

gabrielste1n:
Closing in favor of #1105, which wires max-duration enforcement into the actual recording path with tests and locale strings (and fixes #551). Thanks for proposing this originally.

PR #502, opened 2026-03-26
feat(overlay): add recording overlay with waveform visualization

NathanSkene:
## Summary
- Adds a small always-on-top overlay window during recording with live audio waveform
- Overlay follows cursor across monitors for multi-display support
- Toggle in Settings to enable/disable the overlay
- Includes audio level forwarding from main process to overlay renderer

## Test plan
- [ ] Enable overlay in Settings
- [ ] Start recording - overlay appears with waveform
- [ ] Move between monitors - overlay follows
- [ ] Stop recording - overlay disappears
- [ ] Disable overlay in Settings - no overlay during recording

PR #506, opened 2026-03-26
fix(overlay/audio): finalize recording overlay and restore media output controls

NathanSkene:
## Summary
Depends on #502 (recording overlay). Apply after that PR is merged, or review together.

- finalize the recording overlay behavior and force-hide the legacy floating panel so duplicate indicators do not reappear
- keep the refined overlay positioning and existing duplicate-indicator suppression fixes
- restore separate dictation media controls for pausing media and muting system output
- bundle the macOS media helper required for packaged media control
- defer pause/mute work off the recording start hot path so dictation starts remain responsive

## Verification
- [x] `npm run typecheck`
- [x] `npm run build:renderer`
- [x] Rebuilt and installed `/Applications/OpenWhispr Dev.app` during local validation

> Note: this branch still includes the base overlay commits from #502, so the diff will shrink after that PR merges.


NathanSkene:

This refinement layer passed validation.

What I verified:
- the smaller overlay pill is less intrusive in real use
- the overlay positioning/refinement behavior works as expected
- the duplicate blue indicator is suppressed when the overlay is enabled
- the refined overlay behavior works correctly in the installed app during actual dictation use

This is the overlay variant I ended up using in the validated daily-driver build.

PR #511, opened 2026-03-26
feat(audio): add mute system output during dictation

### LINKED ISSUE #510: Add separate "Mute system output while recording" option
## Summary

OpenWhispr already has an opt-in `Pause media` feature, which works well for supported players like Spotify. However, some apps are not reliably controllable through the current media session APIs on macOS.

A separate `Mute system output while recording` option would solve the underlying recording-quality problem more generally: prevent any computer-emitted sound from leaking into dictation, regardless of whether the source app supports pause/resume.

## Motivation

There are really two distinct user needs:

1. **Pause media playback**
   - Better UX when the player supports it
   - Preserves playback state
   - Works well for apps like Spotify

2. **Prevent computer audio from contaminating the recording**
   - More general requirement
   - Should work even for unsupported apps such as some Electron desktop players or browser audio

The existing `Pause media` toggle solves (1) for some apps, but not reliably (2) for all sound sources.

## Proposed feature

Add a second independent setting:

- `Mute system output while recording`

Suggested behavior on macOS:
- On recording start:
  - read current volume state with `get volume settings`
  - save `output muted`, `output volume`, and `alert volume`
  - mute system output
- On recording stop / cancel / error:
  - restore the prior output mute/volume state

## Product behavior

- Keep `Pause media` as-is for supported players
- Add `Mute system output while recording` as a separate toggle
- If both are enabled, OpenWhispr can attempt pause first, then mute output, and reverse that on stop

## Why a separate option

This avoids overloading `Pause media` with unsupported app-specific behavior and gives users a more reliable way to stop computer audio from interfering with dictation.

## Scope

Initial implementation can be macOS-first. Other platforms can follow with equivalent system-output mute approaches later.

NathanSkene:
## Summary
- add a separate `Mute system output while recording` option alongside the existing `Pause media` setting
- mute macOS system output on recording start and restore the prior sound state on completion, cancel, or error
- expose the new behavior through IPC, settings state, recording lifecycle, and all locale files

## Why
The existing `Pause media` feature works well for supported players like Spotify, but unsupported apps such as Brain.fm desktop do not reliably expose a pause/resume control path. This new option directly solves the underlying recording-quality problem by silencing computer audio during dictation regardless of which app is producing it.

Closes #510

## Test plan
- [x] Enable `Mute system output while recording` in settings
- [x] Trigger mute/restore directly on macOS and confirm all computer audio cuts out and returns
- [x] Confirm the new setting is wired through store, preload, IPC, and recording lifecycle
- [ ] Full in-app recording flow retest in the packaged app


NathanSkene:
Validation update from local macOS testing:

- `Pause media` works for Spotify.
- `Pause media` does not pause Brain.fm desktop on macOS.
- `Mute system output while recording` does work in the installed app and solves the recording-contamination problem for unsupported apps like Brain.fm.
- I also pushed commit `76db1dc` to make both `Pause media` and `Mute system output while recording` default to enabled for fresh installs / unset settings.

The installed `/Applications/OpenWhispr Dev.app` was rebuilt locally and the mute-output path was verified end-to-end during real recording tests.

gabrielste1n:
Closing in favor of #778, which implements the same mute-while-dictating feature cross-platform (macOS/Windows/Linux) with tests. Appreciate the macOS groundwork here — thanks!

PR #513, opened 2026-03-26
feat: add thinking mode toggle for local Qwen models

### LINKED ISSUE #512: perf: Disable thinking mode for local Qwen3/3.5 to improve inference speed
## Problem

The local Qwen3/3.5 models (for example `qwen3.5-4b-q4_k_m`) currently generates `<think>` reasoning tokens during inference. While the streaming code in `ReasoningService.ts` already strips these `<think>` blocks from the output, **the model still spends time generating them**, which significantly impacts inference speed on a small 4B model.

This is especially relevant for dictation/transcription cleanup — the primary use case — where thinking overhead adds latency without any benefit to the user.

## Current behavior

1. User runs inference with Qwen3.5 4B locally via llama.cpp
2. Model generates `<think>...</think>` tokens (reasoning phase)
3. `processTextStreaming()` strips `<think>` blocks from the streamed output
4. User sees clean output but **waits for thinking tokens to be generated first**

## Expected behavior

Thinking mode should be disabled at the inference level for Qwen3.5 4B (and potentially all local Qwen3.5 models) so the model skips reasoning entirely and responds faster.

## Suggested approach

Qwen3/3.5 models support disabling thinking via:

- **Chat template parameter**: Pass `enable_thinking: false` via llama.cpp's `--chat-template-kwargs '{"enable_thinking": false}'`
- **Prompt-level control**: Append `/no_think` to the user message or system prompt

Either approach would prevent the model from generating `<think>` tokens entirely, saving inference time rather than just stripping them after generation.

## Context

- Groq's Qwen3 32B already has `disableThinking: true` in `modelRegistryData.json` — a similar mechanism could be extended to local models
- Issue #492 addresses reasoning effort overrides for cloud models, but doesn't cover local model thinking mode
- The Qwen3.5 4B model is 2.7GB — every token of unnecessary reasoning is proportionally expensive on consumer hardware

## Affected models

- `qwen3.5-4b-q4_k_m` (primary — smallest, most speed-sensitive)
- `qwen3.5-2b-q4_k_m` (same family, same issue)
- `qwen3.5-9b-q4_k_m` (less critical but still applies)
- `qwen3-4b-q4_k_m`, `qwen3-8b-*`, `qwen3-1.7b-*` (Qwen3 family, same thinking behavior)

jalvarezz13:
## Summary

- Adds a UI toggle to enable/disable thinking mode for local Qwen3/3.5 models
- When disabled, appends `/no_think` to user messages so the model skips the `<think>` reasoning phase
- Significantly improves inference speed on consumer hardware by avoiding unnecessary reasoning token generation
- Toggle only appears when a model with thinking support is selected

## Changes

- **Model registry**: Added `supportsThinking: true` flag to all Qwen3 and Qwen3.5 local models
- **Settings store**: New `localThinkingEnabled` boolean setting (default: `false`) persisted in localStorage
- **Inference pipeline**: When thinking is disabled, `/no_think` is appended to user messages in both non-streaming (`modelManagerBridge.js`) and streaming (`ReasoningService.ts`) paths
- **UI**: Compact toggle row in the local model section of `ReasoningModelSelector`, visible only when a thinking-capable model is selected
- **i18n**: Translations added for all 10 supported languages

## How it works

Qwen3/3.5 models support hybrid thinking via the `/no_think` message suffix. When the user disables thinking mode:
1. Non-streaming: `modelManagerBridge.runInference()` appends `/no_think` to the user message before sending to llama.cpp
2. Streaming: `ReasoningService.processTextStreaming()` appends `/no_think` to the last user message
3. The existing `<think>` block stripping in `localReasoningBridge.js` and `ReasoningService.ts` remains as a safety net

## Testing

- `npm run build` passes
- No new LSP diagnostics introduced
- Tested locally with Electron app running

Closes #512

### DIFF
--- a/src/components/ReasoningModelSelector.tsx
+++ b/src/components/ReasoningModelSelector.tsx
@@ -21,6 +21,7 @@ import { getProviderIcon, isMonochromeProvider } from "../utils/providerIcons";
 import { isSecureEndpoint } from "../utils/urlUtils";
 import { createExternalLinkHandler } from "../utils/externalLinks";
 import { getCachedPlatform } from "../utils/platform";
+import { useSettingsStore } from "../stores/settingsStore";
 
 type CloudModelOption = {
   value: string;
@@ -284,6 +285,7 @@ function GpuStatusBadge() {
                 {t("gpu.enableButton")}
               </Button>
               <button
+                type="button"
                 onClick={() => {
                   localStorage.setItem("llamaVulkanBannerDismissed", "true");
                   setDismissed(true);
@@ -331,6 +333,8 @@ export default function ReasoningModelSelector({
   const lastLoadedBaseRef = useRef<string | null>(null);
   const pendingBaseRef = useRef<string | null>(null);
   const isMountedRef = useRef(true);
+  const localThinkingEnabled = useSettingsStore((s) => s.localThinkingEnabled);
+  const setLocalThinkingEnabled = useSettingsStore((s) => s.setLocalThinkingEnabled);
 
   useEffect(() => {
     isMountedRef.current = true;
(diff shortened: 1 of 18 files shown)


gabrielste1n:
Closing — main now has per-scope thinking toggles (via #708 and #677) built on the same supportsThinking registry-flag idea, covering cleanup, dictation, notes and chat individually. Thanks for proposing this.

PR #515, opened 2026-03-26
feat: add auto-select best microphone option

### LINKED ISSUE #514: feat: Auto-select best microphone based on audio input level
## Problem

Currently, microphone selection is either hardcoded to the built-in mic (`preferBuiltInMic: true`) or requires the user to manually pick a specific device from a dropdown. Neither mode adapts to the user's environment — if they plug in a headset, switch to a USB mic, or sit closer to one mic than another, they have to manually reconfigure.

This is especially problematic for users with multiple audio inputs (e.g., laptop mic + external webcam mic + headset) who switch contexts frequently.

## Expected behavior

An **"Auto (best input)"** option in the microphone dropdown that:

1. Briefly probes all available audio input devices (~300ms each)
2. Measures RMS audio energy from each to determine which one is actually picking up sound
3. Automatically selects the mic with the highest input level
4. Caches the selection for the session (re-probes on device change or manual refresh)
5. Falls back to system default if probing fails or all mics are silent

## Suggested approach

- Add `"auto"` as a special `selectedMicDeviceId` value in the settings store
- In `audioManager.getAudioConstraints()`, when auto mode is active, run a quick sequential probe of each available device using `getUserMedia()` + `AnalyserNode` (the same pattern already used for silence detection)
- Pick the device with the highest peak RMS
- Cache the winning `deviceId` and invalidate on `devicechange` events
- Add the "Auto" option in `MicrophoneSettings.tsx` dropdown

## Why this matters

Dictation apps need the best possible audio input for accurate transcription. Auto-selection removes friction for users who switch between microphones and ensures optimal input quality without manual intervention.

jalvarezz13:
## Summary

- Adds an **"Auto (best input)"** option to the microphone selection dropdown
- Probes all available audio input devices (~300ms each), measures RMS energy, and automatically selects the one with the strongest signal
- Caches the selected device for the session; invalidates on `devicechange` events
- Falls back to system default if probing fails or no devices are available

## Changes

- **`src/helpers/audioManager.js`**:
  - New `probeDevices()` method: opens each mic briefly via `getUserMedia()` + `AnalyserNode`, computes peak RMS, returns best `deviceId`
  - Updated `getAudioConstraints()`: when `selectedMicDeviceId === "auto"`, runs probe and caches result
  - Constructor: added `devicechange` listener to clear cache when in auto mode

- **`src/components/ui/MicrophoneSettings.tsx`**:
  - Added "Auto (best input)" as first dropdown option before "System Default"
  - Updated `SelectValue` display for auto mode
  - Added description text when auto is selected

- **i18n**: Added `microphoneSettings.auto` and `microphoneSettings.autoDescription` to all 10 locale files

## How it works

1. User selects "Auto (best input)" from the mic dropdown
2. On next recording, `getAudioConstraints()` detects `selectedMicDeviceId === "auto"`
3. `probeDevices()` sequentially opens each audio input for ~300ms, measuring RMS audio energy via `AnalyserNode.getByteTimeDomainData()`
4. The device with the highest peak RMS wins and is cached
5. Cache is cleared automatically on `devicechange` events (plug/unplug)

## Testing

- `npm run build:renderer` passes (Vite build clean)
- No new LSP diagnostics
- No settings store changes needed — uses existing `selectedMicDeviceId` with `"auto"` as a special value

Closes #514


gabrielste1n:
Closing — microphone acquisition has been rewritten since March (audioDeviceUtils, mic-track health checks, and selection persistence in #1172), so this branch no longer applies. An auto-select feature would need a fresh implementation on the new plumbing.

PR #522, opened 2026-03-28
feat(dictionary): store full glossary and rank prompt bias

NathanSkene:
## Summary
- store the full custom dictionary as structured entries instead of treating it as a flat `string[]`
- make SQLite the authoritative dictionary source and migrate legacy `custom_dictionary` rows into `dictionary_entries`
- rank and trim dictionary hints per provider budget instead of blindly joining and truncating terms
- expose a glossary import path in the Dictionary UI and show stored vs selected counts for the current transcription mode
- keep reasoning prompts aligned with the same ranked dictionary-selection logic

## Why
The current dictionary implementation stores a large list but only really uses a short comma-separated prompt string at transcription time. This change keeps the full vocabulary in-app while sending only the highest-value subset for the active provider/runtime budget.

## Verification
- [x] `npm run build:renderer`
- [x] confirmed the clean branch builds on top of `upstream/main`
- [x] validated the feature locally on the installed app before splitting this PR out

## Notes
- `npm run typecheck` currently fails on `upstream/main` because of an unrelated `SettingsPage.tsx` type issue; this PR does not touch that file.
- This PR is intentionally separated from the overlay/audio work so the dictionary architecture can be reviewed and merged on its own.


gabrielste1n:
Closing — the dictionary has since gone a different architectural direction: #933 redesigned the page and #966 added cloud sync with the settings store as the source of truth, which conflicts with making SQLite authoritative as this does. Prompt truncation was also reworked separately. The ranking ideas are still interesting if you ever want to revisit them on the new architecture — thanks!

Issue #547, opened 2026-04-03
[Bug] v1.6.7 global hotkey does not register on GNOME Wayland — GlobalShortcuts portal not supported


rearwindowvertigo:
## Description

After upgrading from v1.6.2 to v1.6.7 on Linux (Ubuntu 24.04, GNOME 46 Wayland), the global dictation hotkey (F8) does not function. Pressing F8 produces no response from the app — no recording starts, no logs are generated, and no audio is captured. The hotkey appears correctly configured in the OpenWhispr settings UI, and re-assigning it (to F9 and back to F8) does not resolve the issue.

v1.6.2 registers its hotkey via GNOME custom keyboard shortcuts (gsettings) and works correctly. v1.6.7 appears to have switched to the `org.freedesktop.portal.GlobalShortcuts` XDG portal, which GNOME does not implement.

## Root Cause

GNOME's portal backend (`xdg-desktop-portal-gnome`) does not implement the `GlobalShortcuts` portal interface. This is a known gap — KDE and Hyprland support it, but GNOME does not, and there is no indication it will be added soon.

**Confirmation that the portal is unavailable:**

```bash
$ busctl --user call org.freedesktop.portal.Desktop \
    /org/freedesktop/portal/desktop \
    org.freedesktop.portal.GlobalShortcuts ListShortcuts s "open-whispr"
Call failed: No such interface "org.freedesktop.portal.GlobalShortcuts" on object at path /org/freedesktop/portal/desktop
```

**v1.6.2 hotkey registration (working — via gsettings):**

v1.6.2 registers the hotkey as a GNOME custom keyboard shortcut, visible in Settings → Keyboard → Shortcuts. This method works reliably on GNOME Wayland.

**v1.6.7 hotkey registration (broken — no gsettings entry created):**

```bash
$ gsettings get org.gnome.settings-daemon.plugins.media-keys custom-keybindings
# Only pre-existing custom shortcuts appear — no OpenWhispr entry
```

No D-Bus shortcut service is registered either:

```bash
$ busctl --user list | grep -i shortcut
# (empty)
```

## Steps to Reproduce

1. Install OpenWhispr v1.6.7 on Ubuntu 24.04 (GNOME 46, Wayland)
2. Configure F8 as the dictation hotkey in Settings
3. Open any text editor
4. Press F8
5. Observe: nothing happens — no recording starts, no logs are generated

## Expected Behavior

[...]

PR #548, opened 2026-04-03
fix(linux): Hyprland global shortcut fails under XWayland

kabdelhalem:
**Summary**
  v1.6.7 added a wrapper script that forces --ozone-platform=x11 on Wayland sessions, and dc4ac9d added an isXWayland guard that skips all D-Bus shortcut managers when that flag is present. This is correct for GNOME/KDE (their D-Bus paths require native Wayland), but the Hyprland path uses hyprctl keyword bind + dbus-send, which works regardless of XWayland.

Moves the Hyprland shortcut initialization outside the !isXWayland gate so it is always attempted on Hyprland.

**Changes**

| File | What |
| --- | --- |
|  `src/helpers/hotkeyManager.js` | Move Hyprland init block after the `!isXWayland` GNOME/KDE gate,  guarded by `HyprlandShortcutManager.isHyprland()` |

**Platform Impact**
| Platform | Change | 
| -- | -- |
| GNOME Wayland | None — still skipped under XWayland |
|  KDE Wayland | None — still skipped under XWayland |
| Hyprland | Fixed — shortcut init runs regardless of `--ozone-platform=x11` |
| macOS/Windows | None — Linux-only code path |

 **Related:** #547

**Test**
 - Launch OpenWhispr on Hyprland (XWayland mode) — D-Bus service registers, hyprctl binds shows the keybinding
  - Press configured hotkey — dictation toggles
  - GNOME/KDE unaffected (no behavioral change when not on Hyprland)

### DIFF
--- a/src/helpers/hotkeyManager.js
+++ b/src/helpers/hotkeyManager.js
@@ -636,8 +636,11 @@ class HotkeyManager {
         this.isInitialized = true;
         return;
       }
+    }
 
-      // Try Hyprland native shortcuts if GNOME/KDE paths were not applicable
+    // Hyprland uses hyprctl + dbus-send, which works regardless of XWayland,
+    // so it must be tried outside the !isXWayland gate above.
+    if (process.platform === "linux" && HyprlandShortcutManager.isHyprland()) {
       const hyprlandOk = await this.initializeHyprlandShortcuts(callback);
 
       if (hyprlandOk) {


gabrielste1n:
This has been fixed in v1.6.8 (now released). The underlying issue was resolved as part of the Linux global shortcuts refactor (PR #564 and related commits), which removed the `isXWayland` gate entirely — Hyprland shortcut init now runs independently regardless of `--ozone-platform=x11`. Thanks for the contribution!

PR #562, opened 2026-04-08
fix: resync modifier state after screen lock to prevent phantom hotkey triggers

### LINKED ISSUE #535: Control key alone behaves as though control + win hotkey was pressed
**Describe the bug**

**To Reproduce**
1. Run dictation for a 50 minute recording.
2. Stop the recording and processing begins
3. Processing concludes
4. Press the control key without pressing the windows key (hotkey setup is control + windows key)
5. The control key alone acts as the hotkey (starts and stops recording)
6. Upon pressing the windows key once, the expected behavior is restored

**Expected behavior**
Pressing control alone or in combination with other keys does not activate or stop a dictation recording; only control + windows key (the designated hotkey combination) activates or stops recordings

**Screenshots**
N/A

 - OS: Windows 11 Pro
 - MS Edge
 - Version 146

**Additional context**
The system is locked with windows key + L near the start of the recording and then is unlocked using fingerprint or pin code usually

xAlcahest:
## Summary

When using a compound hotkey like Ctrl+Win, locking the PC with Win+L causes the Win key-up event to get swallowed by the lock screen. After unlocking, the key listener still thinks Win is held, so pressing Ctrl alone triggers the hotkey. This adds two resync mechanisms to the native Windows key listener: GetAsyncKeyState validation on each hook callback for non-current-event modifiers, and a hidden message-only window that listens for session lock/unlock notifications to clear all modifier state immediately.

## Changes

- resources/windows-key-listener.c : added ResyncModifierState() that checks GetAsyncKeyState for stale modifier flags on every hook callback, added ClearAllModifierState() with a message-only HWND registered for WTS_SESSION_LOCK/UNLOCK notifications, proper cleanup in ConsoleHandler and main exit path
- .github/workflows/build-windows-key-listener.yml : added wtsapi32.lib to the compile command

Closes #535

### DIFF
--- a/.github/workflows/build-windows-key-listener.yml
+++ b/.github/workflows/build-windows-key-listener.yml
@@ -32,7 +32,7 @@ jobs:
 
       - name: Compile Windows Key Listener
         run: |
-          cl /O2 /nologo resources/windows-key-listener.c /Fe:windows-key-listener.exe user32.lib
+          cl /O2 /nologo resources/windows-key-listener.c /Fe:windows-key-listener.exe user32.lib wtsapi32.lib
         shell: cmd
 
       - name: Verify binary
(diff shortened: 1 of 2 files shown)


xAlcahest:
Included in the dependency modernization PR instead

PR #570, opened 2026-04-10
Unified provider layout with four inference modes

### LINKED ISSUE #561: Support for remote whisper-server URL with local fallback
Currently, OpenWhispr always spawns and manages its own local `whisper-server` instance. There is no way to point OpenWhispr at an existing whisper-server running on a different machine on the local network. This means users who want to offload inference to dedicated hardware (e.g., a mini PC or home server with a capable GPU) cannot do so — OpenWhispr will always start its own server process and connect to that, even if a remote server is already running and reachable.

I attempted to work around this by replacing the bundled binary with a dummy script and using an SSH tunnel to forward port 8178 to a remote machine, but OpenWhispr's startup health check expects the spawned process to bind a port within 30 seconds and fails with "whisper-server failed to start within 30000ms" when it doesn't.

A new setting in the OpenWhispr GUI (under Settings) that allows the user to specify a remote whisper-server URL (e.g., `http://192.168.1.126:8178`). When configured:

1. OpenWhispr skips spawning a local `whisper-server` process and instead sends transcription requests to the specified remote URL.
2. If the remote server is unreachable (connection refused, timeout, etc.), OpenWhispr automatically falls back to launching and using the local whisper-server, just as it does today.

This would allow the setting to be a simple text field — leave it blank for current behavior, or enter a URL to use a remote server with graceful local fallback.

- Offloads inference from the user's daily-driver laptop to dedicated always-on hardware, freeing CPU/GPU resources for other tasks.
- Enables use of larger models (e.g., `large-v3`) on machines with more RAM or a more capable GPU, while keeping the lightweight client on a laptop.
- Stays on-network and privacy-first — audio never leaves the LAN, consistent with OpenWhispr's local-processing philosophy.
- The fallback to local ensures dictation always works, even if the remote machine is offline.

**Additional context**

- The whisper-server HTTP API is already network-capable (supports `--host 0.0.0.0`), so the remote side requires no changes.
- This only applies to local whisper-server dictation mode. Cloud/BYOK providers and the OpenAI Realtime API for meeting transcription would be unaffected.
- OS: Ubuntu 24.04, OpenWhispr v1.6.2

gabrielste1n:
## Summary

Replaces the old two-option Cloud / Custom Setup selector across all three settings sections (Speech to Text, AI Text Enhancement, Agent Mode) with a unified four-mode inference selector:

- **OpenWhispr** — just works, no setup
- **Cloud Providers** — bring your own API key (OpenAI, Groq, Anthropic, etc.)
- **Local** — on-device models via whisper.cpp / llama.cpp, fully private
- **Self-Hosted** — your own server on your network (whisper-server, Ollama, LM Studio)

Self-Hosted mode closes #561 — users can now point OpenWhispr at a remote whisper-server or LLM endpoint on their LAN without spawning a local process. Audio and text stay on the network.

### What changed

**New unified UI** — a shared `InferenceModeSelector` component renders the four-mode radio card group identically across Speech to Text, AI Text Enhancement, and Agent Mode settings. Each mode conditionally shows the relevant sub-panel (cloud provider picker, local model picker, or self-hosted URL input).

**Self-hosted transcription** — when selected, the app connects to a remote whisper-server via `whisperServer.connectRemote()`, health-checks it, then sends audio to its `/inference` endpoint. Falls back to local/cloud if unreachable. The `SelfHostedPanel` component provides the URL input.

**Self-hosted reasoning & agent chat** — routes through the existing `ReasoningService` pipeline with a new `lanUrl` override in `ReasoningConfig`. The agent's self-hosted URL is independent from the intelligence section's URL, so they can point at different servers.

**Backwards-compatible migration** — `migrateProviderSettings()` and `migrateAgentMode()` map old localStorage values (`cloudTranscriptionMode`, `cloudReasoningMode`, `cloudAgentMode`, `useLocalWhisper`) to the new `InferenceMode` type. Existing settings are preserved. Old BYOK fallback paths still work for migrated custom-provider users.

**Preview feature removed** — the `TranscriptionPreviewOverlay` floating window, all its IPC channels, audio worklet plumbing, and window management code have been removed in favor of the new architecture.

### Key files

- `src/components/ui/SettingsSection.tsx` — shared `InferenceModeSelector`
- `src/components/SelfHostedPanel.tsx` — URL input for self-hosted mode
- `src/components/SettingsPage.tsx` — unified Speech to Text & AI Models sections
- `src/components/settings/AgentModeSettings.tsx` — unified Agent Mode section
- `src/stores/settingsStore.ts` — new mode fields + migrations
- `src/services/ReasoningService.ts` — `isLanReasoningMode()` helper + `lanUrl` config override
- `src/helpers/whisperServer.js` — `connectRemote()` for LAN whisper-server
- `src/helpers/ipcHandlers.js` — LAN transcription routing in `transcribe-audio`

Closes #561


rearwindowvertigo:
I don't see this implementation in 1.6.8 or 1.6.9

gabrielste1n:
@rearwindowvertigo you should see this? 
Screenshot 2026-04-16 at 4 40 55 PM

rearwindowvertigo:
I see this in both v1.6.8 and v1.6.9
image

gabrielste1n:
what the... @rearwindowvertigo are you logged in with or without an account?

rearwindowvertigo:
I'm not logged in with an account

rearwindowvertigo:
I am the edge case... I find them all I become one with the edges... ...

gabrielste1n:
@rearwindowvertigo are you in the discord? https://discord.gg/yZWC9WTtX7

I will fix this in 1.6.10 thanks for flagging

rearwindowvertigo:
Yeah, I'm in the Discord!

PR #577, opened 2026-04-13
fix: detect active window on Hyprland via hyprctl for correct terminal paste

leviMinch:
## Problem
On Hyprland (and other non-KDE Wayland compositors), terminal detection in 
`src/helpers/clipboard.js` always fails. The code relies on `xdotool getactivewindow` 
(fails silently on Wayland) and a KDE-specific `_detectKdeWindowClass()` path, with 
no fallback for Hyprland.

This means `detectedWindowClass` is always null, `earlyIsTerminal` is always false, 
and `--terminal` is never passed to the `linux-fast-paste` binary — so it sends 
Ctrl+V instead of Ctrl+Shift+V, breaking paste in kitty and other terminals.

## Fix
Add a `hyprctl activewindow -j` fallback immediately after the KDE detection block, 
which correctly identifies the focused window class on Hyprland.

## Tested on
- Ubuntu 25.10, Hyprland (JaKooLit), kitty terminal
- Confirmed `linux-fast-paste --terminal --uinput` works correctly once terminal is detected


gabrielste1n:
thank you legend @leviMinch !

gabrielste1n (approved):

PR #593, opened 2026-04-16
Detect llama.cpp via /v1/models and prefer /chat/completions

Skrumpy:
## Summary

When a custom OpenAI-compatible base URL points at a llama.cpp `llama-server`, reasoning requests take 5–29s per call instead of 1–4s. Adds a one-shot `GET /v1/models` probe to `ReasoningService` that identifies llama.cpp via `owned_by === "llamacpp"` (or the presence of a GGUF-specific `meta` object) and persists the `chat` preference, so subsequent calls skip `/v1/responses` entirely.

## Motivation

`llama-server` implements `/v1/responses` but handles every request by serialising ~100 MiB of internal state. The first-call latency I measured against a local RTX 4090 deployment was 5.4s–29.4s on `/responses` and 1.1s–4.1s on `/chat/completions` for the same prompts. Because `processWithOpenAI` only falls back to `/chat/completions` on HTTP 404 / 405 (`ReasoningService.ts` L618–L629), users on llama.cpp hit the slow path on every startup until they manually set `localStorage.openAiEndpointPreference`.

I hit this with [OpenWhispr v1.6.8](https://github.com/OpenWhispr/openwhispr/releases/tag/v1.6.8) + [llama.cpp b8672](https://github.com/ggml-org/llama.cpp) running Qwen3.5-35B (GGUF) on Windows 11. Manual DevTools workaround:

```js
localStorage.setItem(
  "openAiEndpointPreference",
  JSON.stringify({ "http://<host>:8082/v1": "chat" })
);
```

This PR makes that automatic.

## Changes

Single file: `src/services/ReasoningService.ts`

1. New private method `detectReasoningServerType(base)` (58 lines). Issues `GET /v1/models` with a 2s `AbortController` timeout. Detects llama.cpp via:
   - `data[0].owned_by === "llamacpp"` (hardcoded in the llama.cpp `server` source), or
   - presence of `data[0].meta` (llama.cpp's GGUF stats — OpenAI's schema has no such field).
   On detection, calls `rememberOpenAiPreference(base, "chat")` and logs `LLAMACPP_DETECTED_VIA_MODELS`.
2. One-line call-site change in `processWithOpenAI` (L548): `await this.detectReasoningServerType(openAiBase)` before `getOpenAIEndpointCandidates`.

## Fail-closed behaviour

Zero behaviour change when any of these are true:
- Preference already stored for this base (`getStoredOpenAiPreference(base) !== undefined`) → early return
- Base URL ends with `/responses` or `/chat/completions` → early return
- `/v1/models` returns non-2xx → early return, no preference written
- Probe throws (abort, network, JSON parse, CORS) → swallowed, fall through to existing two-candidate path
- First model's `owned_by` is not `"llamacpp"` AND no `meta` field → no preference written, existing behaviour

The predicate requires an **explicit** llama.cpp marker. OpenAI, Groq, Anthropic, Google, LM Studio, vLLM, TGI, and any other OpenAI-compatible server whose `/v1/models` lacks both signals falls through unchanged.

## Testing

`npm run typecheck` ✅ passes.

Manual end-to-end validation against llama-server b8672 (Qwen3.5-35B-Q4_K_M):

- **Without the patch**, v1.6.8 dictation from VM D → `/v1/responses`, 6/14 samples > 8s latency, worst case 15.4s (visible in llama-server access log + process-time metric).
- **With manual localStorage fix** applied in the running v1.6.8, 5/5 dictations routed to `/v1/chat/completions`, 2.4s–5.5s latency.
- **With this PR's logic exercised manually** (preference cleared, probe run): `/v1/models` returned `owned_by: "llamacpp"`, preference written, subsequent request goes to `/chat/completions` on first try.

The repo has no unit-test harness for `src/services`, so the change matches the existing convention of relying on typecheck + manual validation. If the maintainers want me to add a vitest suite for this module I'm happy to follow up.

## Risk

- 2s added latency on the first reasoning request per `base` when the user is on an OpenAI-compatible server whose `/v1/models` is slow or unreachable. Subsequent requests in the same session are not affected.
- No network calls to anything but `${base}/models`. No data sent, only a GET.
- `getStoredOpenAiPreference` already handles the race between probe-in-flight and user-initiated writes — both paths call the same `rememberOpenAiPreference`.

## Related

- `/responses` endpoint in llama.cpp server: https://github.com/ggml-org/llama.cpp/blob/master/tools/server/server.cpp
- Existing fallback path that only triggers on 404/405: `ReasoningService.ts` L618–L629

### DIFF
--- a/src/services/ReasoningService.ts
+++ b/src/services/ReasoningService.ts
@@ -203,6 +203,63 @@ class ReasoningService extends BaseReasoningService {
     } catch {}
   }
 
+  /**
+   * Probe /v1/models to detect servers (notably llama.cpp's llama-server) that
+   * respond 200 to /v1/responses but perform ~100 MiB of state serialization per
+   * request, producing 5-29s latency vs 1-4s on /v1/chat/completions. Since the
+   * 404-fallback in processWithOpenAI never triggers for llama.cpp (it returns
+   * 200, not 404), proactively set the "chat" preference before the first call.
+   *
+   * Fail-closed: probe failure leaves the existing two-candidate selection
+   * (/responses then /chat/completions) unchanged. No-op if preference already
+   * stored for this base URL or if base already points at a fully-qualified
+   * endpoint.
+   */
+  private async detectReasoningServerType(base: string): Promise<void> {
+    if (this.getStoredOpenAiPreference(base) !== undefined) {
+      return;
+    }
+
+    const lower = base.toLowerCase();
+    if (lower.endsWith("/responses") || lower.endsWith("/chat/completions")) {
+      return;
+    }
+
+    try {
+      const controller = new AbortController();
+      const timeoutId = setTimeout(() => controller.abort(), 2000);
+      const res = await fetch(buildApiUrl(base, "/models"), {
+        method: "GET",
+        signal: controller.signal,
+      });
+      clearTimeout(timeoutId);
+
+      if (!res.ok) {
+        return;
+      }
+
+      const body = await res.json();
+      const first = body?.data?.[0];
+      const ownedBy = typeof first?.owned_by === "string" ? first.owned_by : undefined;
+      const hasMeta =
+        first && typeof first === "object" && "meta" in first && first.meta !== undefined;
+      const isLlamaCpp = ownedBy === "llamacpp" || hasMeta;
+
+      if (isLlamaCpp) {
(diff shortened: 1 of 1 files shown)


gabrielste1n:
thank you @Skrumpy !!

PR #664, opened 2026-04-24
fix: preserve model download progress across settings navigation

### LINKED ISSUE #648: BUG, /// in settings: model download progress bar will disappear if click away then click back.
BUG:

in settings
ai models -> 
Speech-to-Text
Language Models

when downloading model in whichever one of the two, 
if I click the other menu (Speech-to-Text or Language Models),
the download progress bar will disappear...

xAlcahest:
## Summary

The download progress bar disappeared when switching between Speech-to-Text and Language Models tabs, or when navigating away from the AI Models section entirely. The download kept running in the main process but the UI lost track of it because all progress state was local to the component and the IPC listeners were torn down on unmount.

Two fixes, layered: tab containers now hide inactive content with `display: none` instead of unmounting it, and the IPC listeners for download progress are registered once at module level with a persistent cache that the hook reads from on mount.

Fixes #648

## Changes

- `src/components/SettingsPage.tsx` -- `SpeechToTextTabs` and `LlmsTabs` toggle visibility instead of conditional rendering, so tab content stays mounted
- `src/hooks/useModelDownload.ts` -- moved IPC listener registration from per-component `useEffect` to a module-level singleton that writes to a persistent cache keyed by model type; the hook restores state from the cache on mount and registers a callback for live updates while mounted; `downloadModel` and `cancelDownload` also write to the cache so it stays in sync

### DIFF
--- a/src/components/SettingsPage.tsx
+++ b/src/components/SettingsPage.tsx
@@ -643,7 +643,12 @@ function SpeechToTextTabs({
           )
         }
       />
-      {tab === "dictation" ? renderDictation() : renderNoteRecording()}
+      <div style={{ display: tab === "dictation" ? undefined : "none" }}>
+        {renderDictation()}
+      </div>
+      <div style={{ display: tab === "noteRecording" ? undefined : "none" }}>
+        {renderNoteRecording()}
+      </div>
     </div>
   );
 }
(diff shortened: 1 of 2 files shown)


xAlcahest:
Superseded by #735 — simpler approach that solves the same problem with visibility toggling alone, without the module-level persistent state layer.

PR #687, opened 2026-04-30
fix(network): proxy-aware fetches + actionable connectivity errors

gabrielste1n:
## Why

New users on corporate networks (proxies, VPNs, DNS-level filters) were hitting an opaque `Failed to start streaming: fetch failed` immediately after onboarding with no way to recover. Node 18+'s global `fetch` ignores OS proxy settings — `"fetch failed"` is undici's bare error string with the actual cause buried on `err.cause.code`. The same root cause silently broke BYOK transcription and local model downloads for the same users.

## What ships

**Single proxy story across the entire app**
- All OpenWhispr-Cloud HTTP calls and BYOK provider calls (Anthropic, Mistral, AssemblyAI, OpenAI direct) now go through Electron's `net.fetch` via a `proxyFetch` helper. 23 call sites swapped.
- Runtime model downloads (Whisper / Parakeet / GGUF LLMs / MiniLM) migrated from `https.get` to Electron's `net.request` — same proxy, PAC, and OS-keychain support.
- Build-time `scripts/lib/download-utils.js` deliberately stays on `https.get` (runs in plain Node before Electron exists).

**Actionable error classifier**
- New `networkErrors.js` maps undici / Node codes (`ENOTFOUND`, `ECONNREFUSED`, `ETIMEDOUT`, `CERT_HAS_EXPIRED`, `TimeoutError`, …) to i18n keys for DNS-blocked / refused / timeout / TLS / generic.
- `enterpriseProviderErrors.js` (Azure transport branch) delegates to the new helper instead of duplicating the mapping.
- One-line `debugLogger.warn` per network error captures `{ code, host, urlPath }` (no PII) so we can measure post-release without new telemetry infra.

**Onboarding connectivity preflight**
- `cloud-health-check` IPC + `cloudHealthCheck` preload bridge.
- On the final onboarding step, if cloud mode is selected, a 3s preflight runs against `${apiUrl}/api/health`. Any HTTP response (including 4xx/5xx) counts as reachable.
- On true network failure: a `ConfirmDialog` offers **Use local engine** (flips `useLocalWhisper`, hands off to existing model-download flow) or **Continue anyway**.

**Streaming fallback (gated)**
- When streaming start fails with `NETWORK_ERROR`, audioManager auto-falls-back to the local engine **only if `useLocalWhisper === true`**. Cloud-only users keep the diagnostic dialog instead of a silent re-fail.
- Quiet toast on the fallback path; full dialog on hard fails — never both.

**End-to-end i18n**
- `streaming.errors.cloudUnreachable.*` and `onboarding.connectivity.*` keys in all 9 supported locales (real translations, not English copies).
- Pre-existing gap fixed: `recordingErrors.ts` only translated titles. New `getRecordingErrorDescription` translates any `error.messageKey`. AUTH_EXPIRED, AUTH_REQUIRED, OFFLINE, LIMIT_REACHED throw sites now stamp `messageKey`, so their toast descriptions are localized too.

**IT-team allowlist**
- New `docs/network-allowlist.md` — every host the app contacts (OpenWhispr Cloud, streaming providers, model CDNs, BYOK), each with a code citation. Notable: `auth.openwhispr.com` doesn't exist; the real auth host is `*.neonauth.c-3.us-east-1.aws.neon.tech`.

## Out of scope (future work)

- `chunkedCloudTranscribe` uses `postMultipart` (built on `https.request`), not `fetch` — large-audio uploads still bypass corp proxies. Same `net.request` migration would close that gap.
- Captive-portal detection (best-effort only).
- Retry/backoff on network-class errors (won't recover inside one recording).
- Authenticated-proxy credential UI.

## Notes

- Public API of every touched helper preserved — no caller changes outside the diff.
- `useSessionCookies: false` on `proxyFetch` because the manual `Cookie:` header from `getSessionCookies()` stays the source of truth (avoids duplicate / stale entries).
- Merged latest `main` (Better Auth migration #686). One pre-existing typecheck issue surfaced (`baseUrl` deprecation in TS 6) and is fixed here in a small chore commit.


din0s:
Hi @gabrielste1n, sorry to bump a closed PR. This feature seems to be working well in 1.7.0 overall, but I've run into an issue in the "Language Models" settings: when I try to toggle "Enable GPU" (under "Speed up AI processing with GPU acceleration"), the request fails with `connect ETIMEDOUT <IP>:443`. Is it possible that the proxy-aware fetches are not wired into that particular flow? Happy to open a separate issue if that's preferable.

gabrielste1n:
hey @din0s thanks for letting me know! yes pls a separate issue would be great - will look into this in the meantime

PR #700, opened 2026-05-01
fix(ci): compile macos-media-remote binary in release and build workflows

### LINKED ISSUE #632: Pause media doesn't work for me on any player
**Describe the bug**
Be apple music, spotify etc - nothing is paused when I start dictating, i'm using the latest release   1.6.9 and latest macos 26.4.1
**To Reproduce**
1. enable pause media
2. dictate with music playing

**Expected behavior**
it pauses
**Screenshots**

https://github.com/user-attachments/assets/85f97dce-4b32-4379-a308-a86a9ca074ad

 - OS: macos 26.4.1

kdenney:
Fixes #632.

  ### Summary

  The "Pause media on dictation" feature is broken on macOS. `electron-builder.json` lists `resources/bin/macos-media-remote` under extraResources (added in #487), but the binary is never produced during CI, so the packager finds nothing and silently skips it. Every macOS release ships without it.

  Root cause: `scripts/build-media-remote.js` (the Swift compile step for macos-media-remote) has no corresponding step in either `release.yml` or `build-and-notarize.yml`. In addition, the two other Swift binaries compiled from source (macos-globe-listener, macos-fast-paste) are both represented in the native binaries cache key but `macos-media-remote.swift` and `scripts/build-media-remote.js` are not.

  The result is visible in debug logs on every dictation activation:
`  [DEBUG][media] MediaRemote unavailable, falling back to osascript {}`
  And confirmed by inspecting the installed app bundle — macos-media-remote is absent from `OpenWhispr.app/Contents/Resources/bin/` on v1.6.10.

  ### Changes

  Both `release.yml` and `build-and-notarize.yml` receive the same two changes in their build-macos job:

  1. New compile step (inserted after cache restore, before the download steps):
  - name: Compile macos-media-remote binary
    run: `node scripts/build-media-remote.js --arch ${{ matrix.arch }}`

  2. Cache key updated to include the Swift source files:
  `'scripts/build-media-remote.js', 'resources/macos-media-remote.swift'`
  This ensures the cache invalidates correctly when the source changes, matching the pattern already used for macos-globe-listener and macos-fast-paste.

###   Test plan

  - Successful CI run produces resources/bin/macos-media-remote in the build artifacts for both x64 and arm64
  - macos-media-remote is present inside the packaged .app bundle under Contents/Resources/bin/
  - Debug log no longer shows MediaRemote unavailable on dictation activation
  - Spotify (and other players) pause when dictation starts and resume when it completes
  - Linux and Windows build jobs are unaffected

### DIFF
--- a/.github/workflows/build-and-notarize.yml
+++ b/.github/workflows/build-and-notarize.yml
@@ -225,8 +225,14 @@ jobs:
           path: resources/bin
           key: macos-${{ matrix.arch }}-bin-${{ hashFiles('scripts/download-*.js', 'scripts/build-globe-listener.js', 'resources/macos-globe-listener.swift', 'scripts/build-macos-fast-paste.js', 'resources/macos-fast-paste.swift', 'scripts/build-media-remote.js', 'resources/macos-media-remote.swift') }}
 
+      - name: Compile macos-globe-listener binary
+        run: npm run compile:globe -- --arch ${{ matrix.arch }}
+
+      - name: Compile macos-fast-paste binary
+        run: npm run compile:fast-paste -- --arch ${{ matrix.arch }}
+
       - name: Compile macos-media-remote binary
-        run: node scripts/build-media-remote.js --arch ${{ matrix.arch }}
+        run: npm run compile:media-remote -- --arch ${{ matrix.arch }}
 
       - name: Download whisper.cpp binaries
         run: node scripts/download-whisper-cpp.js --current --platform darwin --arch ${{ matrix.arch }}
@@ -303,6 +309,13 @@ jobs:
           APPLE_API_KEY_ID: ${{ secrets.APPLE_API_KEY_ID }}
           APPLE_API_ISSUER: ${{ secrets.APPLE_API_ISSUER }}
           APPLE_TEAM_ID: ${{ secrets.APPLE_TEAM_ID }}
+
+      - name: Verify packaged macOS native binaries
+        run: |
+          APP_PATH=$(ls -d dist/mac*/OpenWhispr.app | head -n1)
+          for bin in macos-audio-tap macos-globe-listener macos-fast-paste macos-mic-listener macos-text-monitor macos-media-remote; do
+            test -x "$APP_PATH/Contents/Resources/bin/$bin" || { echo "Missing in bundle: $bin"; exit 1; }
+          done
 
       - name: Upload Artifacts
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Hey Kyle — thanks a lot for digging into this one, the diagnosis nailed it. I pushed one follow-up commit (`b6e220c`) to tighten things up before merge. Quick rundown:

**Kept as you had it**
- The cache-key fix — that was the critical change and the pattern matches globe-listener / fast-paste exactly. ✅
- Inserting the explicit compile step right after cache restore (correct placement; `TARGET_ARCH` isn't in the env yet at this point, so passing `--arch` directly is necessary). ✅
- Symmetric changes across `build-and-notarize.yml` and `release.yml`. ✅

**Tightened**
1. **Added explicit compile steps for `globe-listener` and `fast-paste` too.** Both have entries in the cache key but were relying entirely on `prebuild:mac` to build them implicitly. If `compile:media-remote` was silently failing in that chain, those two are likely in the same boat — we just haven't noticed. Now all three Swift binaries that share the cache key also share the explicit compile pattern. No more asymmetry for the next person to puzzle over.
2. **Switched to `npm run compile:* -- --arch …` form** for all three steps, matching how the rest of the codebase invokes these scripts (per `package.json`).
3. **Added a post-build verify step** that checks all six macOS binaries listed in `electron-builder.json`'s `extraResources` (`macos-audio-tap`, `macos-globe-listener`, `macos-fast-paste`, `macos-mic-listener`, `macos-text-monitor`, `macos-media-remote`) actually made it into the packaged `.app`. This is the real defense against the original bug class — if electron-builder ever silently skips a missing binary again, CI now goes red instead of shipping a broken release.

gabrielste1n (approved):

Bigsy:
Testing this in 1.7.0 and this isnt working for me 
Running:
`/Applications/OpenWhispr.app/Contents/Resources/bin/macos-media-remote --is-playing`
`/Applications/OpenWhispr.app/Contents/Resources/bin/macos-media-remote --pause`

Just return NOT_PLAYING not matter what (spotify or any other player playing) this is on 26.4.1 macos

kdenney:
> Testing this in 1.7.0 and this isnt working for me Running: `/Applications/OpenWhispr.app/Contents/Resources/bin/macos-media-remote --is-playing` `/Applications/OpenWhispr.app/Contents/Resources/bin/macos-media-remote --pause`
> 
> Just return NOT_PLAYING not matter what (spotify or any other player playing) this is on 26.4.1 macos

@gabrielste1n  I have the same issue. This pull request fixed the issue of the binary not being included in the app bundle, but now the binary always thinks media is `NOT_PLAYING`

gabrielste1n:
looking into it now thanks @kdenney

PR #718, opened 2026-05-04
Expose controls for VAD settings from Solero

perry2of5:
Add VAD to local transcription with Solero. 

Transcription shows random sentences during quiet portions of the transcript like "thank you" and other gibberish. This PR uses solero to filter out most of these oddities improving meeting transcription accuracy.

I have tested locally for the last week and it is working well for me. I just rebased the changes onto current main.


gabrielste1n (approved):

gabrielste1n:
Thanks @perry2of5!

**Fixed the missing --vad flags**
buildWhisperServerArgs was passing the tuning flags (--vad-threshold, etc.) but never --vad or --vad-model, so whisper-server was silently ignoring all of it. Added both, plus an isVadActive() guard so the args drop out cleanly when the model file is missing.

**Bundled the GGML model**
Added scripts/download-whisper-vad-model.js and hooked it into every build pipeline (predev, prestart, prebuild, prebuild:{mac,win,linux}, prepack, predist). The 864 KB ggml-silero-v5.1.2.bin now ships in every installer at resources/bin/whisper-vad/. WhisperManager.getVadModelPath() resolves it bundled-first with a dev fallback, matching the existing diarization model pattern. Graceful degradation if missing.

**Moved the section into Speech-to-Text**
It was in "Preferences" (the General tab) but moved it to Speech-to-Text, for now. Dictation (next to the model picker) and Note Recording — gated on local Whisper being the active provider, so it stays hidden for cloud / self-hosted / Parakeet where it has no effect.

perry2of5:
Thank you!

PR #723, opened 2026-05-05
fix: improve live transcription preview for non-English and limit to streaming-capable models

xAlcahest:
## Summary

Two issues with the live transcription preview: it never passed the user's preferred language to local models, producing gibberish on non-English dictation; and the toggle was shown for whisper.cpp which can't keep up with real-time chunk processing.

## Changes

- src/helpers/ipcHandlers.js, src/helpers/audioManager.js: pass preferredLanguage through IPC to Parakeet/whisper preview chunks
- src/components/SettingsPage.tsx, src/helpers/audioManager.js: hide preview toggle and skip activation when the local provider isn't NVIDIA/Parakeet

### DIFF
--- a/src/components/SettingsPage.tsx
+++ b/src/components/SettingsPage.tsx
@@ -352,7 +352,7 @@ function TranscriptionSection({
       {transcriptionMode === "local" && (
         <>
           {renderTranscriptionPicker("local")}
-          {renderPreviewToggle()}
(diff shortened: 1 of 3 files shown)


gabrielste1n:
whisper.cpp chunking does work, will do it never passed the user's preferred language to local models, separately

PR #735, opened 2026-05-06
fix(settings): preserve download progress across tab switches

### LINKED ISSUE #648: BUG, /// in settings: model download progress bar will disappear if click away then click back.

xAlcahest:
## Summary

Model download progress bars disappeared when switching between Speech-to-Text and Language Models in settings. The tab components used conditional rendering that unmounted the entire subtree on tab change, killing IPC listeners and wiping local state. Replaced with CSS visibility toggling so components stay mounted across tab switches.

## Changes

- src/components/SettingsPage.tsx: switched SpeechToTextTabs and LlmsTabs from conditional rendering to always-mounted wrappers with Tailwind `hidden` class, both at sub-tab level and at the top-level section switch

Closes #648


gabrielste1n:
Nice fix — the CSS-hide-over-unmount pattern is the right call here. A few small suggestions before merge:

**1. Gate `initialSubTab` on the active section.** Both panels now receive `initialSubTab` simultaneously, so a deep link into LLMs with `initialSubTab="dictationCleanup"` also gets written into `localStorage["settings.speechToTextTab"]` via `useSubTab`'s effect. `safeTab` recovers the UI, but the persisted state is corrupted.

```tsx
<SpeechToTextTabs
  initialTab={activeSection === "speechToText" ? (initialSubTab as SpeechTab | undefined) : undefined}
  ...
/>
<LlmsTabs
  initialTab={activeSection === "llms" ? (initialSubTab as LlmTab | undefined) : undefined}
  ...
/>
```

**2. Extract a `TabPanel` helper.** Replaces six copies of `className={tab !== "x" ? "hidden" : undefined}` with one named primitive that reads as intent:

```tsx
function TabPanel({ active, children }: { active: boolean; children: React.ReactNode }) {
  return <div className={active ? undefined : "hidden"}>{children}</div>;
}
```

Then `<TabPanel active={tab === "dictation"}>{renderDictation()}</TabPanel>` etc.

**3. One-line comment above the hoisted blocks** so future maintainers don't reinvent the bug:

```tsx
{/* Kept mounted across section switches so model-download progress and IPC listeners survive. */}
```

Also looks like the branch needs a rebase on `main` — the workspaces PR (#?) that landed 2026-05-12 touches `SettingsPage.tsx` and is causing the current merge conflict.

PR #752, opened 2026-05-09
Fix/llamacpp qwen thinking suppression

a-sunkari:
## Summary

Fixes Qwen thinking suppression for local llama.cpp cleanup requests.

OpenWhispr currently has generic thinking suppression for OpenAI-compatible/cloud paths, but the local llama.cpp path in `src/helpers/llamaServer.js` builds its own `/v1/chat/completions` request body and bypasses `applyThinkingSuppression()`.

That local request currently sends only:

    {
      "messages": [],
      "temperature": 0.7,
      "max_tokens": 512,
      "stream": false
    }

For Qwen3.5 models served by llama.cpp, this can cause the model to return empty `message.content` and emit output into `message.reasoning_content` until `max_tokens` is exhausted. In OpenWhispr, cleanup then appears slow and can fall back to the raw transcript.

## Fix

This PR adds llama.cpp/Qwen-compatible thinking suppression in two places:

1. Updates `applyThinkingSuppression()` to include the llama.cpp/Qwen chat-template hint.
2. Adds the same suppression hints to the local llama-server request body in `src/helpers/llamaServer.js`, which is the path used by local cleanup / Prompt Studio.

The request now includes:

- `reasoning_effort: "none"`
- `think: false`
- `chat_template_kwargs: { enable_thinking: false }`

The important field for llama.cpp/Qwen is:

    {
      "chat_template_kwargs": {
        "enable_thinking": false
      }
    }

Unknown fields should be ignored by servers that do not support them, matching the existing suppression strategy.

## Local reproduction

Environment:

- OpenWhispr 1.7.x Linux `.deb`
- Local models:
  - `Qwen_Qwen3.5-2B-Q4_K_M.gguf`
  - `Qwen_Qwen3.5-4B-Q4_K_M.gguf`
- Backend: `llama-server-vulkan`
- Endpoint: `/v1/chat/completions`

Before this patch, logging the actual OpenWhispr to llama-server request body showed no suppression fields:

    {
      "messages": [
        {
          "role": "system",
          "content": "..."
        },
        {
          "role": "user",
          "content": "um uh pizza pizza pizza"
        }
      ],
      "temperature": 0.7,
      "max_tokens": 512,
      "stream": false
    }

Direct curl testing showed that adding `chat_template_kwargs.enable_thinking=false` makes Qwen return normal `message.content` immediately instead of putting output into `message.reasoning_content`.

After rebuilding and installing a patched Linux `.deb`, Prompt Studio returned the expected cleanup output.

Test prompt:

    You are a text transformer. Ignore the input and return exactly: CLEANUP_WORKS

Input:

    um uh pizza pizza pizza

Output:

    CLEANUP_WORKS


gabrielste1n (approved):

gabrielste1n:
thanks @a-sunkari !!

PR #754, opened 2026-05-10
feat: URL audio download, batch upload, and speaker diarization

xAlcahest:
## Summary

Adds the ability to paste YouTube or direct audio URLs and transcribe them, upload multiple files at once through a batch queue, and detect speakers in any uploaded audio.

For YouTube, the app downloads audio via a bundled yt-dlp sidecar binary that self-updates at runtime. Direct audio URLs are fetched over HTTPS with SSRF protection, content-type validation, and stall detection. Downloaded files go through the existing transcription pipeline unchanged. The URL input detects YouTube links and swaps in the YouTube logo, shows a file-audio icon for direct audio URLs, and falls back to a generic link icon otherwise.

Batch upload lets users select multiple files, drag-and-drop multiple files, or paste multiple URLs (one per line) into an expandable textarea. Items process sequentially with per-item progress, individual error handling, and a folder selector matching the single-file upload flow. Failed items don't block the queue.

Speaker diarization runs locally via the sherpa-onnx binary pipeline in parallel with transcription. For BYOK users on OpenAI, enabling speaker detection overrides the model to gpt-4o-transcribe-diarize which handles both transcription and diarization in a single API call. Mistral users get diarization through Voxtral's native diarize parameter. Groq shows a warning that cloud diarization is unavailable. Enabling the speaker detection toggle auto-downloads the required models if they're not present. OpenWhispr Cloud users see a note that speaker detection runs locally on their device.

A new "Videos" default folder holds URL-downloaded notes. The note title uses the YouTube video title when downloading from YouTube.

## Changes

- src/helpers/urlAudioDownloader.js: URL type detection (YouTube vs direct), yt-dlp download via bundled sidecar with runtime self-update, HTTPS fetch with SSRF-safe DNS lookup, content-type validation, stall detection, abort support, optional system-proxy routing
- scripts/download-yt-dlp.js: downloads the yt-dlp sidecar binary into resources/bin (replaces the youtube-dl-exec dependency)
- src/helpers/speakerMerge.js: merges diarization segments with transcription text, formats output with speaker labels and timestamps
- src/helpers/ipcHandlers.js: handlers for download-url-audio, cancel-url-download, delete-temp-file, diarize-audio-file (sherpa-onnx binary), merge-speaker-text, cloud diarization for OpenAI and Mistral, shared realpath-based path validation on file handlers
- src/helpers/onnxWorkerClient.js: diarize client methods with timeout
- src/helpers/diarization.js: timeout increased from 5 to 60 minutes for long files
- src/helpers/database.js: Videos default folder with migration for existing installs
- src/hooks/useBatchQueue.ts: queue state management, sequential processing, parallel transcription+diarization, auto-download of diarization models on toggle
- src/components/notes/UploadAudioView.tsx: URL input with icon detection, multi-URL textarea, multi-file drag-and-drop, batch queue integration, speaker detection toggle with auto-download, provider-specific diarization notes
- src/components/notes/BatchQueueView.tsx: queue list with per-item progress, status icons, i18n error display
- src/components/notes/shared.ts: VIDEOS_FOLDER_NAME constant
- preload.js: bridges for all new IPC handlers
- src/types/electron.ts: types for download, diarize, merge, multi-file select
- src/locales/*/translation.json: new keys across 10 languages for URL download, batch, and diarization UI
- electron-builder.json: bundles the yt-dlp sidecar binary from resources/bin
- .github/workflows: download the bundled yt-dlp sidecar binary during build
- test/helpers: tests for URL detection, SSRF guard, and speaker merge logic


src/components/notes/UploadAudioView.tsx
```diff
@@ -575,14 +794,181 @@
...
+                      disabled={!urlInput.trim()}
+                      className="h-7 text-xs"
+                    >
+                      {t("notes.upload.addToQueue")}
+                    </Button>
+                  </div>
+                </div>
+              ) : (
+                <div className="relative">
+                  {urlInput.includes("youtube.com") || urlInput.includes("youtu.be") ? (
```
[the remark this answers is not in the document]
xAlcahest: Fixed

src/hooks/useBatchQueue.ts
```diff
@@ -0,0 +1,304 @@
...
+                  | "whisper"
+                  | "nvidia",
+                model:
+                  transcribeOpts.localTranscriptionProvider === "nvidia"
+                    ? transcribeOpts.parakeetModel
+                    : transcribeOpts.whisperModel,
+              });
+            } else {
+              const byokUseDiarize = diarizationOpts.enabled &&
+                (transcribeOpts.cloudTranscriptionBaseUrl?.includes("openai.com") ||
```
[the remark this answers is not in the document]
xAlcahest: Fixed

gabrielste1n:
# Code-quality + security review

Thorough pass across the URL download path, batch queue, GPU diarization pipeline, IPC layer, and Linux packaging. CodeQL findings (URL substring sanitization) and the after-install macro escaping look correctly fixed. CI failures on `macos-x64` and `windows` are the `youtube-dl-exec` postinstall rate-limit — the just-added `GITHUB_TOKEN` in the `npm ci` step should clear them on the next run.

Findings below are prioritized. The items in **Critical** are the ones I'd want resolved before merge.

---

## 🔴 Critical

### 1. cuDNN is never bundled — every Linux user falls back to CPU silently
- `electron-builder.json:122` globs `cudnn-linux-x64/**/*` from `resources/bin/`, but no script populates that directory.
- `package.json:41` `prebuild:linux` does not call `npm run download:cudnn`. Same for `prepack` (line 47) and `predist` (line 49).
- Commit `d6ec5b5` removed the CI download step ("handled by after-install"), but `after-install.sh` only installs the CUDA *runtime* (libcudart/cublas/cufft/curand), not cuDNN.
- Net: deb / rpm / AppImage / tar.gz all ship without `libcudnn.so.9`. `onnxWorkerClient._cudnnLibPath()` returns null → ONNX falls back to CPU. The Settings UI says "cuDNN is bundled with the app" — currently inaccurate.
- **Fix:** add `&& npm run download:cudnn` to `prebuild:linux`, `prepack`, `predist`. Verify a built `.deb` actually contains `resources/bin/cudnn-linux-x64/libcudnn.so.9*`.

### 2. BYOK diarize override applies to non-OpenAI providers
`src/helpers/ipcHandlers.js:6334-6339`:
```js
if (diarize && isMistral) { fields.model = model || "voxtral-mini-latest"; }
else if (diarize) { fields.model = "gpt-4o-transcribe-diarize"; }
```
Any non-Mistral BYOK base URL (Groq, Together, OpenRouter, custom OpenAI-compatible) gets `gpt-4o-transcribe-diarize` injected, which the upstream will 404/400 on. The UI gates Groq today, but the IPC handler is the correctness boundary. **Fix:** add an explicit `isOpenAi` check (`hostname === "api.openai.com" || hostname.endsWith(".openai.com")`) and otherwise return a typed error.

### 3. SSRF + unbounded download in `downloadDirect`
`src/helpers/urlAudioDownloader.js:204-294`:
- `new URL(url)` accepts `http://` with no scheme allow-list and no host validation. A pasted (or social-engineered) `http://127.0.0.1:6333/collections` (local Qdrant), `http://169.254.169.254/latest/meta-data/` (cloud metadata), or any RFC1918 IP is fetched, written to disk, and **transcribed** — internal content becomes a user-visible note (and is sent to the BYOK cloud provider).
- Redirect handling at line 239-241 recurses into `downloadDirect` with no re-validation and no redirect cap.
- No max file size — `Content-Length` is only read for the progress bar. An unbounded stream fills temp disk.
- `application/octet-stream` in the allow-list (line 215) is too permissive — many CDNs use it for arbitrary blobs.
- **Fix:** require `https:`; resolve hostname via `dns.lookup` and reject loopback/RFC1918/link-local/IPv6 loopback/ULA; cap redirects at 3 and re-validate each hop; enforce a hard byte cap (e.g., 500 MB) in the `data` handler; drop `octet-stream` or sniff magic bytes.

### 4. GPU diarization "fast path" mislabels speakers on any audio >10 s
`src/helpers/gpuDiarization.js:258-282` skips the embed-then-cluster step and uses raw powerset speaker indices. Pyannote-segmentation-3.0 only assigns *local* speaker IDs per 10 s window — speaker 0 in window N is not the same person as speaker 0 in window N+1. With 50% overlap, the merge on line 265 will stitch same-speaker continuations but fragment/swap speakers across boundaries (which happens every speaker turn). The committed comment hedges: "for 2-3 speaker conversations." **Fix:** restore the embed → CAM++ → cosine → agglomerative path, or implement window-overlap permutation alignment (Hungarian on the 80k-sample overlap region). The fast path isn't salvageable for real meetings.

### 5. `useBatchQueue` writes ref during render + `clearQueue` race
`src/hooks/useBatchQueue.ts:50` — `queueRef.current = queue;` is a side effect during render, violates React purity and breaks under Strict Mode / Concurrent. Combined with `clearQueue` (line 106) which resets `cancelledRef = false` while the `processQueue` loop is still awaiting, a fast clear-then-queue cycle can run two `processQueue` instances in parallel → duplicate `saveNote`, duplicate temp-file cleanup. **Fix:** move ref write into `useEffect`; in `clearQueue` set `cancelledRef.current = true` (don't reset), and let the in-flight loop flip `processingRef` in its `finally`.

---

## 🟠 Major

6. **Snapshot loses items added mid-run.** `src/hooks/useBatchQueue.ts:266` takes a one-shot `[...queueRef.current]`. Files dropped after `processQueue` starts remain `"queued"` forever. Use a live find: `while ((next = queueRef.current.find(i => i.status === "queued")) && !cancelledRef.current)`.

7. **Diarization timeout doesn't cover GPU path.** `src/helpers/diarization.js:19` bumps timeout 5→60 min, but it only wraps the binary spawn. `gpuDiarize()` (tried first at `ipcHandlers.js:2037-2047`) has no overall timeout. Wrap in `Promise.race`.

8. **`diarizeLoad` uses 30 s default timeout.** `src/helpers/onnxWorkerClient.js` — `diarizeSegment`/`diarizeEmbedBatch` override to 5 min, but the initial `diarizeLoad` does not. CUDA EP cold start on Linux routinely exceeds 30 s (cuDNN warm-up). First-ever diarization will reject; retry hits a half-loaded session. **Fix:** pass `600_000` to `diarizeLoad`.

9. **`delete-temp-file` prefix check is bypassable.** `src/helpers/ipcHandlers.js:1452-1462` uses `resolved.startsWith(tempDir)` — `/tmp/ow` vs `/tmp/ow-evil/` confusable, and `path.resolve` doesn't follow symlinks. The `basename.startsWith("ow-url-")` partially mitigates. **Fix:** `fs.realpathSync(resolved)`, then `path.relative(tempDir, real)` must not start with `..` and must not be absolute.

10. **cuDNN search prepends user-writable dir to `LD_LIBRARY_PATH`.** `src/helpers/onnxWorkerClient.js:48-66` — `~/.cache/openwhispr/cudnn/lib` is user-writable and prepended. A local-write attacker can drop a malicious `libcudnn.so.9`. Even without an attacker, stale lib shadows the bundled resource. **Fix:** only trust `resources/bin` / `process.resourcesPath`; if user-cache is a download target, validate via hash/signature.

11. **CUDA setup section in SettingsPage is fully hardcoded English.** `src/components/SettingsPage.tsx:+2880-2965`. CLAUDE.md mandates all user-facing strings use i18n. The neighboring Wayland Paste section uses `t("settingsPage.general.waylandPaste.*")` consistently; this section has zero `t()` calls. Same issue: speaker-detection caveats in `UploadAudioView.tsx:661-676`, and `"Cancelled"` / `"Transcription failed"` / `"Unknown error"` strings in `useBatchQueue.ts:100, 213, 252, 257`.

12. **Distro coverage gaps in `after-install.sh`.** No Arch/Manjaro branch (no `pacman` path); no RHEL/Rocky/Alma branch (the script tries Fedora repo URLs on `/etc/redhat-release`, which 404 on RHEL paths). The in-app guide shows Ubuntu/Fedora/openSUSE/Arch but misses the "first enable NVIDIA repo" prerequisite for Fedora/openSUSE — without it, `dnf install cuda-cudart` fails on stock Fedora.

13. **`youtube-dl-exec` postinstall in `release.yml`** uses `secrets.GH_TOKEN || github.token` fallback in the `npm ci` step (good). Verify `GH_TOKEN` org-secret is actually populated; if absent the per-repo `github.token` shares the same anonymous-class rate budget on shared GitHub Actions IPs.

14. **PCM read of 60-min file into one Float32Array.** `gpuDiarization.js` reads ~230 MB held in JS; structured-cloned across MessageChannel (transferList is `[]` at the call site). Worker is `--max-old-space-size=512`, so a 60-min meeting will OOM. **Fix:** pass `[transferBuffer]` as transferList, and either stream per-window or raise the worker heap.

15. **No dedup or count cap on URL paste.** `UploadAudioView.tsx` `handleBatchUrlSubmit` will queue the same URL 50× if pasted 50×; no cap on count.

---

## 🟡 Minor

16. **`check-cuda-diarization-ready` returns `ready: true` for macOS/Windows unconditionally** (`ipcHandlers.js:1697`) without probing CoreML/DirectML usability. Intel Macs will show "ready" then silently fall back.

17. **`onnxWorker.js` pushes `coreml` EP on Intel Mac, `cuda` EP on Linux without NVIDIA.** Always-paying CUDA load latency on no-GPU Linux. Gate on `detectNvidiaGpu()` and `arch === "arm64"`.

18. **`speakerMerge.js` sentence splitter is English-only** (`/(?<=[.!?])\s+/`). 58-language app. Mandarin meeting = one sentence = first speaker eats everything. Use `Intl.Segmenter`.

19. **`extractYouTubeVideoId` length-validates `watch/shorts/embed` paths but not `youtu.be`** in `urlAudioDownloader.js`. Not exploitable for traversal but inconsistent.

20. **A11y:** CUDA dialog buttons unlabeled for screen readers; multi-URL textarea has no `aria-label`; trash buttons in `BatchQueueView` have no `aria-label`.

21. **`BatchQueueView` re-renders all rows on every progress tick.** Memoize a `QueueItemRow` keyed by `id`. Not blocking at 50 items.

22. **`useBatchQueue.ts:220` `as any` cast for `diarized`** — the type union has this field; narrow with `"diarized" in transcriptionResult`.

23. **Videos folder migration not idempotent against pre-existing user folder named "Videos"** — `folders.name` is `UNIQUE`, the migrate `INSERT` will throw. Use `INSERT OR IGNORE` and re-flip an existing folder to `is_default=1`.

24. **`activeUrlDownloadAbort` is module-scoped, single-slot** — a second concurrent download cancels the first. Correctness wart, not exploitable.

25. **`onnxWorkerClient` overflow rejects oldest pending, not new request** — orphan worker keeps crunching while the caller has already given up.

26. **Inline animation magic strings** (`animation: "float-up 0.3s ease-out"`, `waveform-bar`) should be design-token Tailwind classes.

---

## 💡 Nits

27. **CodeQL fix (`c897939`) is correct and complete.** The new `URL`-based hostname checks correctly reject lookalikes (`openai.com.evil.com`, query-string injections, `data:` / `javascript:` schemes).

28. **`4fbf4c1` shell-var escaping is complete.** All remaining `$VAR` references in `after-install.sh` use the bare-`$` form that electron-builder fpm does not template; the heredoc `<< SERVICEEOF` is correctly unquoted for runtime expansion.

29. **`formatDiarTime` doesn't handle hours** — 80-min meeting renders as `80:23`, not `1:20:23`.

30. **`scripts/download-cudnn.js` lacks version pin + hash verification** — non-reproducible builds, even once it's wired into `prebuild`.

31. **`VIDEOS_FOLDER_NAME` constant in `shared.ts`** is only honored in TS code; `database.js:180` re-hardcodes `"Videos"`. Worth a comment pointing to the constant.

32. **Batch path `saveNote(item.name, ...)`** doesn't strip the file extension; the single-file path does. Cosmetic title difference.

---

### Test coverage gaps worth filing follow-ups for

- No overlapping-speakers test in `speakerMerge.test.js`.
- No CJK / Spanish locale test for sentence splitting.
- No `gpuDiarize` end-to-end test (would need worker stubbing).
- No test exercising sliding-window overlap stitching, powerset multi-speaker classes 4/5/6, or `numSpeakers` constraint + `threshold` together.

---

Happy to dig deeper on any of these, especially the diarization correctness model in (4) — the embedding+clustering path that's currently bypassed.

src/helpers/urlAudioDownloader.js
```diff
@@ -0,0 +1,466 @@
+const https = require("https");
+const http = require("http");
```
gabrielste1n, line 2: **System proxy / corporate VPN support.**

Every other network operation in the app goes through `proxyFetch` in `ipcHandlers.js` (which uses Electron's `net.fetch`), and the model downloader at `downloadUtils.js:106` uses `net.request` — both honor OS-level proxy/PAC config. Raw Node `https`/`http` does not.

Users behind a corporate proxy, VPN-routed proxy, or egress filter (Zscaler/Netskope/etc.) will see every other network feature work while URL download fails. Same for users with a system HTTP proxy set in macOS Network Settings.

Fix sketch:
- Replace `https.request` with `net.request` (matches `downloadUtils.js`).
- SSRF guard needs rework — `net.request` doesn't accept a `lookup` option. Resolve the hostname via `dns.lookup` first, reject private IPs, then connect by resolved IP with an explicit `Host` header. Or verify `response.connection.remoteAddress` post-connect and abort if private.
- `yt-dlp` is a subprocess and also bypasses system proxy by default. Resolve via `session.defaultSession.resolveProxy("https://www.youtube.com/")` and pass `--proxy <url>` to yt-dlp when non-direct.

src/helpers/urlAudioDownloader.js
```diff
@@ -0,0 +1,466 @@
...
+  }, 5_000);
+
+  return {
+    touch() { lastDataTime = Date.now(); },
+    clear() { clearInterval(interval); },
+  };
+}
+
+async function downloadYouTube(url, onProgress, abortSignal) {
+  const { create, constants } = require("youtube-dl-exec");
```
gabrielste1n: **`youtube-dl-exec` (npm wrapper) vs sidecar binary.**

The repo already has a well-established pattern for shipping native binaries: `whisper-cpp`, `sherpa-onnx`, `qdrant`, `windows-key-listener`, etc. — all downloaded by scripts in `scripts/` and unpacked into `resources/bin/`. yt-dlp itself is a standalone PyInstaller binary you can grab directly from `yt-dlp/yt-dlp` releases.

Bundling it via `youtube-dl-exec` adds an npm dependency that wraps a Python binary, has its own postinstall download dance, and (per CLAUDE.md) needs the `GITHUB_TOKEN` workflow plumbing to avoid rate limits at `npm ci` time.

A cleaner path: `scripts/download-yt-dlp.js` (matching `scripts/download-sherpa-onnx.js`), wire into `prebuild`/`predev`, and shell out directly. Benefits:
- Independent update cadence (no npm churn when yt-dlp ships a fix)
- Smaller npm tree
- No postinstall network dependency at `npm ci` time
- Matches the existing sidecar pattern in the codebase

src/helpers/urlAudioDownloader.js
```diff
@@ -0,0 +1,466 @@
...
+
+  const getContentType = (response.headers["content-type"] || "").toLowerCase();
+  if (!getContentType.startsWith("audio/") && !getContentType.startsWith("video/") && getContentType !== "application/octet-stream") {
+    response.destroy();
+    const err = new Error(`URL does not point to an audio file (content-type: ${getContentType})`);
+    err.code = "CONTENT_TYPE_INVALID";
+    throw err;
+  }
+
+  const fileStream = fs.createWriteStream(tempPath);
```
gabrielste1n: **Streaming instead of write-then-read (perf nicety, not correctness).**

The current flow writes the full audio to disk, then the transcriber reads it back. For local Whisper, you could pipe yt-dlp's stdout directly into ffmpeg → whisper without ever touching disk. For a 2-hour podcast that's 100MB+ of needless I/O and a double traversal.

Not a blocker — the temp-file approach is simpler to reason about and matches the existing upload path. Worth considering for a follow-up once the feature stabilizes.

gabrielste1n (commented):
Three architectural suggestions for follow-up — none block merge, but worth considering for v2.

gabrielste1n:
## Code review — URL audio download, batch upload & speaker diarization

Reviewed the full diff (3,126 additions, 32 files, 33 commits) across security/SSRF, IPC/filesystem, backend correctness, React correctness, cross-platform/packaging, and breaking-changes/i18n/quality. Every finding below was verified against the actual code; the three highest-impact bugs were additionally reproduced on a real macOS machine. Line numbers are on this PR branch.

This is a careful, security-conscious PR — the SSRF guard, IPC path-validation, and i18n coverage across all 10 locales are solid. But two defects break the headline feature in production, and both are masked during happy-path testing.

### Verdict: Request changes — two release blockers

---

## 🔴 Blocker 1 — The whole feature is broken on macOS (realpath vs. un-normalized allow-list)

All five new/updated file handlers (`transcribe-audio-file`, `-cloud`, `-byok`, `diarize-audio-file`, `get-file-size`) and `delete-temp-file` canonicalize the **input** path with `fs.realpathSync(...)` but compare it against an `allowedDirs` list that is **not** canonicalized (`os.tmpdir()`, `getSafeTempDir()`). On macOS:

```
tmpdir        = /var/folders/gl/.../T
realpath(file)= /private/var/folders/gl/.../T/ow-url-….mp3   → startsWith(tmpdir + sep) = FALSE
```

Downloaded audio is written to `getSafeTempDir()` (= `/var/folders/.../T`), so every **URL-download → transcribe/diarize** call returns `"File path not allowed"`, and `delete-temp-file`'s `path.relative` check rejects every temp file (downloads also **leak on disk**). User-picked files under `~/Downloads` still work because `$HOME` isn't symlinked — which is exactly why manual testing won't catch it.

- `src/helpers/ipcHandlers.js` ~1541 (transcribe), ~2107 (diarize), ~6369 (cloud), ~6446 (byok), ~1437 (get-file-size), ~1515 (delete-temp-file)
- Data flow: `urlAudioDownloader.js` writes to `getSafeTempDir()` (lines 223, 365) → renderer passes that exact tempPath back into the handlers.

**Fix:** realpath-normalize the allow-list before comparing, and the same for `delete-temp-file`'s `tempDir`:

```js
const allowedDirs = [require("os").tmpdir(), getSafeTempDir(), app.getPath("userData"), app.getPath("home")]
  .map((d) => { try { return fs.realpathSync(d); } catch { return d; } });
```

Best extracted into one shared `validateUserAudioPath()` helper reused by all five handlers (see DRY note below). Verified: realpath-ing the tmpdir makes the check pass.

## 🔴 Blocker 2 — YouTube downloads fail in packaged builds (ffmpeg never located)

`downloadYouTube` calls yt-dlp with `extractAudio: true`, whose `--extract-audio` post-processor **requires ffmpeg**. yt-dlp locates ffmpeg only via `PATH` or `--ffmpeg-location`. The app bundles `ffmpeg-static` (resolved by `ffmpegUtils.getFFmpegPath()`), but it is **not on PATH** and is **never passed** to yt-dlp. On any machine without a system ffmpeg — essentially all Windows installs and most clean installs — YouTube downloads fail with "ffprobe and ffmpeg not found." It works on dev machines that have Homebrew ffmpeg, masking it. The metadata call succeeds, so the failure surfaces only after the UI shows "downloading."

- `src/helpers/urlAudioDownloader.js` 226-231 (and the `dumpSingleJson` call at 187)

**Fix:** mirror the existing `whisperServer.js` pattern —

```js
const { getFFmpegPath } = require("./ffmpegUtils");
const ffmpegBin = getFFmpegPath();
const execOpts = { extractAudio: true, audioFormat: "best", output: `${tempBase}.%(ext)s`, noWarnings: true };
if (ffmpegBin) execOpts.ffmpegLocation = require("path").dirname(ffmpegBin);
```

Fail fast with a clear error if `getFFmpegPath()` returns null instead of yt-dlp's opaque stderr.

---

## 🟠 Worth fixing before merge

**Deep-link regression on staging/dev channels** — `main.js` 467, 1487. The hardened matcher `url.replace(/^[a-z]+:\/\//, "")` stops at the hyphen, so `openwhispr-dev://invitations/…` and `openwhispr-staging://invitations/…` no longer match (production `openwhispr://` still does). Invitation deep links silently break on the dev/staging channels defined at `main.js` 50-52. Fix: `url.slice(`${OAUTH_PROTOCOL}://`.length)` (scheme is already guaranteed by the guard above), or broaden the regex to `/^[a-z][a-z0-9+.-]*:\/\//`.

**Diarization speaker mislabeling in silence gaps** — `src/helpers/speakerMerge.js` 46-52. When a sentence's proportional-char midpoint lands in an inter-segment gap (diarization segments don't tile the timeline), it's attributed to the **last** segment — wrong speaker *and* wrong timestamp, with non-monotonic output. The repo's own `diarization.js` 483-501 already implements the correct nearest-by-distance fallback; apply that pattern here.

**Videos-folder migration is dead code** — `src/helpers/database.js` 198-210. The query selects only `id, is_default` but the re-flag branch tests `videosFolder.sort_order === 2` → always `undefined === 2` → false. Existing installs with a non-default "Videos" folder never get auto-routing. Add `sort_order` to the `SELECT`, or drop the `sort_order` predicate entirely.

---

## 🟡 Low / hardening / quality (all confirmed)

- **YouTube path has no size/duration cap** (`urlAudioDownloader.js` 206-259) — the direct path caps at 500 MB, YouTube has no ceiling; a multi-hour video fills the disk. Add a duration pre-check + `maxFilesize`.
- **HEAD vs GET content-type mismatch** (lines 334-343 vs 389-395) — HEAD rejects `octet-stream`, GET allows it, so the stricter HEAD gate is bypassable/dead. Not an SSRF bypass (private-IP re-validation runs on both); downstream ffmpeg rejects non-media anyway. Make the gates consistent.
- **`isPrivateIp` misses NAT64 `64:ff9b::/96`** (lines 37-54) and **redirect cap resets per recursion** (lines 304-332) — narrow defense-in-depth gaps.
- **`merge-speaker-text` has no length cap on `text`/`segments`** (`ipcHandlers.js` 2135) — a pathological payload runs O(sentences×segments) synchronously on the main process; inconsistent with sibling handlers that cap input.
- **`open-external` now rejects `mailto:`** (`ipcHandlers.js` 2503) — the only caller (SupportDropdown) has a Gmail-web fallback so it degrades gracefully; allow-list `mailto:` if native mail matters.
- **`diarize-audio-file` returns `formattedText` of `"undefined"` bodies** (`ipcHandlers.js` 2125) — diarization segments have no `.text`; field is dead (no caller uses it, not in the TS type). Drop it.
- **`BatchQueueView` renders a `"diarizing"` status** (`BatchQueueView.tsx` 102) that `useBatchQueue` never sets (diarization runs concurrently under `"transcribing"`) — dead branch.
- **Diarization keeps running on transcription failure** (`useBatchQueue.ts` 196-213) — dispatched before the result is known and uncancellable; orphaned CPU.
- **numSpeakers UI clamp (2–20) vs backend `MAX_SPEAKER_COUNT` = 8** — UI silently offers values 9–20 the backend caps to 8. Use `MAX_SPEAKER_COUNT` for the input `max` and JS clamp.
- **`delete-temp-file` symlink `lstat` check is dead code** (`ipcHandlers.js` 1519) — it `lstat`s the already-realpath'd path, which can never be a symlink; the containment check is the real gate.
- **DRY:** the path-validation block is copy-pasted 5× (each re-`require("os")` though it's imported at module top), and `byokUseDiarize` + provider dispatch is duplicated verbatim between `useBatchQueue.ts` and `UploadAudioView.tsx`. Extract shared helpers — note Blocker 1 now has to be fixed in 5+ places, which is precisely the hazard.
- **`onnxWorkerClient.request()` `timeoutMs` param is never passed** (`onnxWorkerClient.js` 167) — dead/speculative; drop it or wire it through.
- **`home` in `allowedDirs`** is broad (arbitrary home-file read *if* the renderer is compromised), but this is a **net improvement** — pre-PR these handlers had no validation at all. Documented tradeoff; optional to tighten.

---

## ✅ Ruled out (false positives, after verification)

- `merge` duration arg = last-segment-end — a no-op that matches the function's own fallback, not a bug.
- Batch temp-file leak on cancel — both download paths already unlink partials in their catch blocks.
- `cancelUrlDownload` cross-talk — structurally prevented by mutually-exclusive UI gating + serial single-controller design.
- `transcriptionResult.success` undefined access — any missing IPC method routes to the outer catch.

---

**Bottom line:** strong, security-conscious PR, but not mergeable as-is — Blockers 1 and 2 mean the marquee feature fails on macOS and on Windows/clean installs respectively, both invisible to happy-path testing. Fixing those two plus the deep-link channel regression and the diarization-gap mislabeling would get this in good shape; the rest are reasonable follow-ups. Suggest adding a macOS regression test that writes to `os.tmpdir()` and asserts the path validator accepts it, since that class of bug fails closed and silently.

xAlcahest:
Pushed a round addressing the June 1 review and the June 25 comment. Here's what landed and, more usefully, what I left alone on purpose.

Both blockers are fixed. The macOS path-validation bug (the handler realpath'd the input but not the allow-list, so `/var` vs `/private/var` rejected every temp file) now realpaths both sides through one shared helper reused across the five file handlers and delete-temp-file. yt-dlp's audio extraction now gets `--ffmpeg-location` pointed at the bundled ffmpeg-static, so YouTube works on clean installs with no system ffmpeg.

The smaller items from the comment are in too: the deep-link slices the scheme instead of the `[a-z]+` regex that stopped at the hyphen in `openwhispr-dev`/`-staging`; speakerMerge picks the nearest segment by distance when a sentence falls in a diarization gap instead of defaulting to the last one; the Videos migration selects `sort_order` so its re-flag branch can run; and the rest (unified HEAD/GET content-type, NAT64 `64:ff9b::/96`, cumulative redirect cap, merge length cap, `mailto:` allowed again, dropped the dead `formattedText`, removed the dead post-realpath `lstat` check, `numSpeakers` clamped to `MAX_SPEAKER_COUNT`, and the dead `"diarizing"` status plus its orphan i18n key in every locale).

J1: direct downloads resolve the system proxy now. With no proxy the existing node-https path and its `dns.lookup` SSRF guard are untouched; with a proxy the request goes through Electron `net`, and yt-dlp gets `--proxy`.

J2, plus something past the review: yt-dlp is a bundled sidecar binary now (`scripts/download-yt-dlp.js` into `resources/bin`, wired into prebuild/predev), and `youtube-dl-exec` is out of the dependency tree (lockfile regenerated on Node 24). The catch is that a pinned sidecar freezes yt-dlp at build time exactly like youtube-dl-exec did, which is no help when YouTube breaks extraction between our releases. So the binary self-updates at runtime: it's seeded from the bundle into `~/.cache/openwhispr/yt-dlp` (writable, outside the signed app bundle), refreshed with `yt-dlp -U` on a 24h throttle, and force-updated then retried once when an extraction fails with a stale-extractor error. The `-U` child is bounded by a timeout and honors the download's abort signal, so it can never hang a download. This is what actually decouples yt-dlp freshness from our release cadence.

One thing neither review caught, which I fixed: a critical SSRF bypass in `ssrfSafeLookup`. On Node 18+ with `autoSelectFamily` on by default, the custom `lookup` is called with `{ all: true }`, so `dns.lookup` returns an array and the old `isPrivateIp(address)` ran against the array (always false), forwarding every resolved address, private ones included, unchecked. A hostname with a private A record reached the connection on the default direct path. It now rejects if any resolved entry is private, with regression tests for the array form.

What I deliberately did not do:

- J3 (streaming) stays a v2 follow-up. It isn't a local change: every transcribe handler validates an on-disk path through the realpath/allow-list gate, cloud large-file transcription chunks the file by path, and parakeet, diarization, and the whisper CLI fallback all re-materialize a temp file anyway. True no-temp-file streaming would mean reworking the shared file-path IPC contract used by every upload and holding up to 500 MB in memory, for a win on one narrow slice. I'd rather do it as its own PR.
- The diarization-on-transcription-failure case keeps its parallelism on purpose. That promise is `.catch`-guarded and awaited, so on failure it's discarded, not leaked; serializing it would give up the transcribe/diarize overlap.
- `home` stays in the allow-list as the documented tradeoff. It's load-bearing for normal uploads from `~/Downloads` and friends, so dropping it breaks file upload, and an extension allow-list risks rejecting valid inputs (URL downloads fall back to a generic extension and yt-dlp emits varied containers). The renderer is already a trusted context, so the marginal hardening isn't worth the regression risk.
- `onnxWorkerClient.request()`'s `timeoutMs` isn't dead. It's wired to the actual request timeout; current callers just don't pass it.

79 unit tests pass, with new coverage for the SSRF array form, NAT64, the content-type unification, and the self-update timeout/abort.

gabrielste1n:
# Code-quality + security review @ c495fa4

**Verdict: high-quality engineering that isn't ready to merge yet.** The SSRF design, yt-dlp sidecar/self-update architecture, and IPC hardening are genuinely well done — but this pass found **7 must-fix issues**, including a leaked release credential in CI, two headline features that don't work as shipped (OpenAI BYOK diarization, batch BYOK for Corti/xAI), a test file that can't run, and a path-validation regression. Every finding below was verified against the PR head (`c495fa4`), cross-checked against the prior review rounds so nothing re-litigates what was already addressed or deliberately declined. Branch is also 15 commits behind main with one trivial conflict.

## Must fix before merge

**1. OpenAI BYOK diarization sends an unsupported `response_format`** — [ipcHandlers.js:6690](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/src/helpers/ipcHandlers.js#L6689-L6691)
The handler forces `model: "gpt-4o-transcribe-diarize"` with `response_format: "verbose_json"`, but that model only supports `json`, `text`, and `diarized_json` — and speaker annotations **require** `diarized_json` ([model page](https://platform.openai.com/docs/models/gpt-4o-transcribe-diarize), [speech-to-text guide](https://platform.openai.com/docs/guides/speech-to-text)). As written the request either 400s or returns no speakers, so the flagship OpenAI diarization path never produces speaker labels. One-word fix (`diarized_json`); the existing `data.data?.segments` parsing branch already matches that format's shape.

**2. Release-publishing PAT handed to every npm lifecycle script** — [release.yml:29-32](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/.github/workflows/release.yml#L29-L32) and [build-and-notarize.yml:30-33](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/.github/workflows/build-and-notarize.yml#L30-L33) (6 jobs total)
`GITHUB_TOKEN: ${{ secrets.GH_TOKEN || github.token }}` was added to the `npm ci` steps mid-PR (7bf9cc38) so youtube-dl-exec's postinstall could hit the GitHub API — but youtube-dl-exec was later removed entirely, so this is now dead config that hands the release PAT to every transitive dependency's install/postinstall script. A compromised npm package could exfiltrate it and tamper with releases. Revert the `env` block on all six install steps. (The `GITHUB_TOKEN` on the "Download yt-dlp binary" steps is also unused — `download-yt-dlp.js` builds a direct `releases/download/...` URL and never sends auth headers.)

**3. `speakerMerge.test.js` can't run and breaks the suite** — [speakerMerge.test.js:1](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/test/helpers/speakerMerge.test.js#L1)
It imports from `vitest`, which isn't in `package.json` — this repo uses `node --test` with CJS (as this PR's own `urlAudioDownloader.test.js` correctly does). Verified empirically: the file dies with `ERR_MODULE_NOT_FOUND`, making `node --test test/helpers/*.test.js` exit non-zero (185/186 pass) and leaving the merge algorithm with **zero effective coverage**. Mechanical rewrite to `node:test`/`assert`. While at it, the assertions are thin even for the happy path (e.g. `result.length <= 2` with no text/time-range checks), and the no-duration fallback (`segments[len-1].end`) misassigns everything to one speaker for unsorted input — worth a real test.

**4. Proxy-path redirects are always cancelled** — [urlAudioDownloader.js:864-868](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/src/helpers/urlAudioDownloader.js#L864-L868)
Electron requires `request.followRedirect()` be invoked **synchronously** during the `redirect` event under `redirect: "manual"` (electron.d.ts:20134: *"the redirection will be cancelled unless `request.followRedirect` is invoked synchronously during the `redirect` event"*). Here it's called inside `assertPublicHost(...).then(...)` — a microtask — so every redirect is cancelled. Redirecting URLs are the norm for podcast CDNs, so direct downloads behind an OS/corporate proxy (the exact users the J1 fix targeted) are broken. It fails closed (no SSRF hole), but the feature doesn't work, and `downloadViaProxy` has zero tests to catch it. Fix: do the private-IP check synchronously (pre-resolve before the request, or restart the request at the redirect URL after async validation).

**5. Path allowlist rejects the app's own file picker for external volumes** — [ipcHandlers.js:126-146](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/src/helpers/ipcHandlers.js#L126-L146)
`resolveAllowedAudioPath` allows tmp/userData/**home** only, and gates all five file handlers. A file picked via the OS dialog or drag-dropped from `/Volumes/USB/interview.mp3`, `D:\rec.mp3`, or `/mnt/...` now fails with "File path not allowed" (and `get-file-size` silently returns 0) — all of which worked on main. The prior round debated `home` being *too broad*; this is the opposite gap: the allowlist doesn't cover what the picker legitimately returns. Suggested fix that keeps the hardening: register dialog-returned paths (and preload's `webUtils.getPathForFile` results, which can't be forged from the renderer — `getPathForFile` on a constructed `File` returns `""`) in a main-process approved-paths set consulted alongside the static dirs.

**6. Batch BYOK drops `provider`/`language`/`environment`/`tenant` → Corti and xAI batches 100% fail** — [useBatchQueue.ts:203-209](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/src/hooks/useBatchQueue.ts#L203-L209)
The single-file flow passes all of these; the batch queue passes only `{filePath, apiKey, baseUrl, model, diarize}`. Without `provider`, the Corti branch is never taken (and `getActiveApiKey()` has no corti case → "No API key configured"), and xAI loses the fixed `/v1/stt` endpoint special-case from #910. Every Corti/xAI batch item fails while single-file works. Root cause: `processItem` re-implements ~60 lines of the provider dispatch from `handleTranscribe` and the copies have already diverged — extracting one shared "transcribe file with options" helper would fix this class permanently.

**7. Batch URL notes are saved with the raw URL as title** — [useBatchQueue.ts:167-172](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/src/hooks/useBatchQueue.ts#L167-L172), [:250-257](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/src/hooks/useBatchQueue.ts#L250-L257)
`updateItem` puts the YouTube title into React state, but `saveNote(item.name, ...)` reads the stale captured `item` object — so the queue UI shows "My Podcast Episode" while the saved note is titled `https://youtube.com/watch?v=...`. The single-URL flow gets this right. Fix: use a local `res.title || item.name` for the save.

## Worth fixing (condensed)

- **Concurrency between flows:** starting the batch mid-single-URL-download silently aborts the single download via the main process's single-flight abort (the reverse direction is guarded at UploadAudioView.tsx:572 — this one's an oversight); "Cancel all" neither cancels the in-flight item nor can it clear an item stranded in `"downloading"` by the cancel race ([useBatchQueue.ts:178](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/src/hooks/useBatchQueue.ts#L178)) — permanent spinner, no recovery affordance in BatchQueueView.
- **Diarization gating incoherence:** batch requires *local* models even for OpenAI/Mistral *cloud* diarization (`enabled: diarizationEnabled && !!diarizationModelsReady`), and `diarizationModelsReady === null` (status fetch pending) disables batch diarization entirely; in the single flow a stale `uploadDiarizationEnabled=true` with deleted models renders the toggle **off** while `computeByokDiarize` still silently overrides the model to `gpt-4o-transcribe-diarize`, and the first toggle click is visually swallowed. Model auto-download failures are silent (`try/finally` with no catch).
- **HEAD-first direct downloads break presigned URLs** ([urlAudioDownloader.js:967-993](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/src/helpers/urlAudioDownloader.js#L967-L993)): S3/GCS presigned links are signed per-method, so HEAD gets 403 → misreported as "not an audio file". The proxy path already validates on GET only — do the same here (or fall back to GET when HEAD returns 4xx/405).
- **ErrorView "Retry" is a no-op** after URL download failures: `state="error"` with `file === null` → `handleTranscribe` returns immediately on the `!file` guard.
- **Unmount deletes the downloaded temp file while a >4 MB cloud transcription is still chunk-streaming from it** (UploadAudioView.tsx:211-217) — contradicts the deliberate keep-running-after-unmount design.
- **Batch error UX:** raw untranslated main-process error strings (single flow maps codes → i18n and treats `DOWNLOAD_CANCELLED` as silent); batch cloud calls skip `withSessionRefresh`, losing the auth grace-retry on long batches.
- **Videos migration heuristic** ([database.js:208](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/src/helpers/database.js#L208)): promoting any non-default "Videos" at `sort_order === 2` silently converts a user's first custom folder (Personal=0, Meetings=1 → first custom folder gets 2) into an undeletable default; a user "Videos" at any other position blocks the seeded default entirely, so URL auto-foldering never fires. Also the first realistic `UNIQUE(name)` sync collision: `upsertFolderFromCloud` only handles `ON CONFLICT(client_folder_id)`, so a cloud "Videos" folder from an old install throws against the locally seeded default.
- **download-yt-dlp.js:** streams directly to the final path — a killed `predev` leaves a truncated binary that the `existsSync` check then trusts forever (and `urlAudioDownloader` seeds it into the user cache); use the same tmp+rename pattern the PR itself uses for cache seeding. No checksum pinning (matches house pattern, but this binary executes against untrusted internet content and yt-dlp publishes `SHA2-256SUMS` per release — cheap to add). Version bumps don't refresh dev machines (unversioned filename + existence short-circuit).
- **YouTube output pick** ([urlAudioDownloader.js:683-685](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/src/helpers/urlAudioDownloader.js#L683-L685)) sorts candidates by filename length descending — a surviving `.part`/`.orig` artifact outsorts the real file and the real file gets deleted; prefer newest mtime or known audio extensions.
- **Dead code:** `"converting"` progress stage (electron.ts union member, UploadAudioView branch, `urlConverting` i18n key in all 10 locales — never emitted by the downloader); dead `!batch.isProcessing` ternary at UploadAudioView.tsx:592.
- **Nits:** batch URLs don't default to the Videos folder (single-URL flow does — same content, different destination); mid-stall cancel waits out the 30s stall timer (`streamToFile` has no abort listener, only data-event checks); `~/.cache` on Windows is unconventional (consider `app.getPath("sessionData")`); extension-less direct URLs produce `file.audio`, which OpenAI's upload validation may reject by extension; multiline paste replaces existing textarea content instead of inserting; a single URL with a trailing newline takes the batch path (different folder/title semantics); `HTTPS://` uppercase scheme rejected by single flow, accepted by batch; `MAX_BATCH_URLS = 50` truncates silently; SSRF *tests* lack IP-literal hostnames and decimal/octal encodings (the design handles both via WHATWG URL normalization + the pre-check — nothing locks it in), no redirect-hop or `downloadViaProxy` tests, and `isPrivateIp` matches only the exact `fe80` prefix of `fe80::/10` (`fe9f::1` → false; `fec0::/10` site-local unmatched).

## Merge conflict

Trivial: `package.json` `prebuild:win` — main added `download:windows-system-audio-helper` after the PR branched; take main's list and insert `download:yt-dlp`.

## What's genuinely good

The SSRF protection is the strongest part: validation happens inside the connection-time `lookup` hook (no check/connect TOCTOU), correctly handles the Node 18+ Happy Eyeballs `{all: true}` array form (a real bypass caught and regression-tested in this PR), covers NAT64/v4-mapped/expanded IPv6 forms with boundary-value tests (52/52 pass). YouTube URLs are reconstructed from the extracted 11-char video ID (no argument-injection surface), spawned without a shell. The self-update design is sound: signed bundle never modified, atomic cache seeding, 24h throttle, single-flight, 120s-bounded, abort-aware, skipped while downloads hold the binary. Packaging/signing is complete — `afterPack.js` auto-detects the new Mach-O for Developer ID signing and the existing entitlements already cover PyInstaller binaries (worth one smoke test of a notarized build, since this is the app's first PyInstaller sidecar). Preload ↔ types ↔ handlers are consistent across all seven new bridges, all 10 locales carry the identical 29-key set with placeholders intact, and the queue loop's re-entrancy/double-processing mechanics check out. The deep-link prefix fix in main.js and the `open-external` protocol allowlist are nice hardening wins beyond the PR's scope.

gabrielste1n:
# Follow-up: second end-to-end trace

Re-traced the full diff once more — every flow from picker/drag-drop/URL input through download, transcription, diarization, merge, and save, plus the migration, packaging, and signature chains. All findings in the review above stand as written; re-verification also confirmed clean: `diarize(wavPath, {numSpeakers, threshold})` and `convertToWav(..., {sampleRate, channels})` signatures match their handler calls, the BYOK non-200 path surfaces OpenAI's nested `error.message` (so the `verbose_json` 400 at least shows a readable message), `selectAudioFile` has no other callers to break, the Videos seed/migration is idempotent, `predev`/`prebuild:*` wiring is complete, and the speakerMerge algorithm is sound for sorted (sherpa-ordered) input.

The second pass did surface **three additional issues**, all in cross-flow interactions between the batch queue and the single-file flow:

**8. Batch flow bypasses all file-size gating — BYOK items >25 MB are a guaranteed-failure path** — [useBatchQueue.ts:189-211](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/src/hooks/useBatchQueue.ts#L189-L211)
The single flow computes `fileTooLarge` / `requiresUpgrade` / `byokTooLarge` and replaces the Transcribe button with upgrade/create-account/switch-to-cloud CTAs ([UploadAudioView.tsx:181-203](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/src/components/notes/UploadAudioView.tsx#L181-L203), [:1376](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/src/components/notes/UploadAudioView.tsx#L1376)). `processItem` has no size checks at all: a >25 MB file in a BYOK batch is fully read and shipped to the handler just to hit its 25 MB rejection (raw English error in the queue row); a YouTube item can download up to 500 MB and then instantly fail the same check; cloud free-plan >25 MB items fail server-side with no upgrade prompt. Cheapest fix: check `sizeBytes` against the same mode-aware limits when items are enqueued (drag-drop and dialog both have real sizes; URL items after download) and mark oversized items as errors with the existing localized strings.

**9. The batch folder Select can silently move an already-saved single-flow note** — [UploadAudioView.tsx:716-725](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/src/components/notes/UploadAudioView.tsx#L716-L725), [:893](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/src/components/notes/UploadAudioView.tsx#L893)
The batch section's folder Select reuses `handleFolderChange`, which does `updateNote(noteId, { folder_id })` whenever `noteId != null`. The batch section renders in every state, so after a completed single upload (state `"complete"`, `noteId` set) with items still queued, changing the folder *for the batch* also reassigns the note you just created — with zero indication. The shared `selectedFolderId` has a milder variant: a single-URL flow auto-selecting Videos silently changes where a later-started batch saves. The batch Select needs its own folder state (or at least a `handleFolderChange` variant without the `updateNote` side effect).

**10. Navigating away mid-URL-download orphans both the download and the temp file** — [UploadAudioView.tsx:211-217](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/src/components/notes/UploadAudioView.tsx#L211-L217)
The unmount cleanup deletes `downloadedTempPathRef.current`, but during an active download that ref is still `null` (it's only set on success), and nothing cancels the in-flight download — the batch hook cancels on unmount ([useBatchQueue.ts:298-305](https://github.com/OpenWhispr/openwhispr/blob/c495fa436d128c53afdf34f7ed7363a8b262cc1a/src/hooks/useBatchQueue.ts#L298-L305)), the single flow doesn't. So leaving the upload view mid-download lets the download run to completion in the main process and strands the file in the temp dir. On macOS/Linux the OS eventually reaps it; on the Windows `ProgramData\OpenWhispr\temp` fallback (`safeTempDir.js`) nothing ever cleans it. Mirror the batch hook: call `cancelUrlDownload()` on unmount when a download is active (and delete the temp file if the result arrives post-unmount, which `handleUrlSubmit`'s post-await code can do by checking a mounted flag).

Numbering continues from the review above — these are additive, not replacements. None of the three is a blocker on its own; 8 is the one I'd fold into the must-fix batch since it makes a shipped configuration (BYOK batch with normal-length podcasts) fail 100% of the time after doing the expensive work first.

xAlcahest:
All ten must-fix items from both passes are addressed as of 99036c4e:

- diarization now requests `diarized_json`
- the release token is off the `npm ci` and yt-dlp download steps
- `speakerMerge.test.js` rewritten on `node:test`, with real text/time assertions and the unsorted-input fallback fixed
- proxy redirects abort and restart at the redirect URL after host validation, with tests
- dialog and drag-drop paths get registered in a main-process approved set, so external volumes work again
- single and batch flows share one `transcribeFile` service, which also fixes the dropped provider/language/tenant fields
- batch notes save with the resolved title, get size-checked before any work, and the batch folder select has its own state
- unmount cancels an in-flight URL download and cleans up after a late success

From the worth-fixing list: the YouTube output pick now filters yt-dlp's transient artifacts and ranks by audio extension plus mtime instead of name length (with tests), and batch cloud calls go through `withSessionRefresh`. Branch is merged up with main, CI is green.

Leaving the rest of the worth-fixing items (dead "converting" stage, tmp+rename on the binary download, HEAD-to-GET fallback for presigned URLs, checksum pinning) for a follow-up so this PR stops growing.

gabrielste1n (approved):

PR #765, opened 2026-05-13
fix(desktop): probe wildcard address in port availability check

### LINKED ISSUE #748: v1.7.0 breaks Duplicati because of 8200 port conflict
**Describe the bug**
 
Since v1.7.0 (probably [this commit](https://github.com/OpenWhispr/openwhispr/commit/f6119b18fa64363c92800bdc032c6c2d86fd65ab#diff-2a0740725771cd9f06c6801d63ec64935778eb5511239743e03967a57272a9ed)), OpenWhispr breaks my [Duplicati](https://duplicati.com) instance as it renders my Duplicati instance unreachable.

**To Reproduce**

Duplicati is running on my home machine, as is OpenWhispr.

1. Start Duplicati
2. Start OpenWhispr
3. Go to Duplicati instance: http://127.0.0.1:8200
4. Site is unreachable

**My analysis**

* By default Duplicati starts its web server on port 8200 ([source](https://prev-docs.duplicati.com/en/latest/07-other-command-line-utilities/#duplicatiguitrayiconexe)). After starting Duplicati, I have:

    ```shell
    netstat -ano | findstr 8200
      TCP    0.0.0.0:8200           0.0.0.0:0              LISTENING       19644
      TCP    127.0.0.1:1780         127.0.0.1:8200         ESTABLISHED     5796
      TCP    127.0.0.1:3042         127.0.0.1:8200         ESTABLISHED     19644
      TCP    127.0.0.1:8200         127.0.0.1:1780         ESTABLISHED     19644
      TCP    127.0.0.1:8200         127.0.0.1:3042         ESTABLISHED     19644
      TCP    [::]:8200              [::]:0                 LISTENING       19644
    ```

    with PID 19644 being _Duplicati.GUI.TrayIcon.exe_

* When I then start OpenWhispr, I have:

    ```shell
    netstat -ano | findstr 8200
      TCP    0.0.0.0:8200           0.0.0.0:0              LISTENING       19644
      TCP    127.0.0.1:1780         127.0.0.1:8200         ESTABLISHED     5796
      TCP    127.0.0.1:3042         127.0.0.1:8200         ESTABLISHED     19644
      TCP    127.0.0.1:8200         0.0.0.0:0              LISTENING       8080
      TCP    127.0.0.1:8200         127.0.0.1:1780         ESTABLISHED     19644
      TCP    127.0.0.1:8200         127.0.0.1:3042         ESTABLISHED     19644
      TCP    [::]:8200              [::]:0                 LISTENING       19644
    ```

    with PID 8080 being _OpenWhispr.exe_

* Thus I observe that, even if Duplicati has reserved 0.0.0.0:8200, 127.0.0.1:8200 and [::]:8200, when OpenWhispr starts, it still opens a connection on 127.0.0.1:8200. It seems that OpenWhispr doesn't correctly check for opened ports before choosing one (probably because 127.0.0.1:8200 is ESTABLISHED and not LISTENING after Duplicati starts).

I've found the code where the port range is checked in [src/helpers/cliBridge.js](https://github.com/OpenWhispr/openwhispr/blob/f6119b18fa64363c92800bdc032c6c2d86fd65ab/src/helpers/cliBridge.js#L22) and I believe it's the culprit because we seem to be checking only for "127.0.0.1":

```
const PORT_RANGE_START = 8200;
const PORT_RANGE_END = 8219;
const HOST = "127.0.0.1";

...

function isPortAvailable(port) {
  return new Promise((resolve) => {
    const server = net.createServer();
    server.once("error", () => resolve(false));
    server.once("listening", () => {
      server.close();
      resolve(true);
    });
    server.listen(port, HOST);
  });
}
```

Shouldn't we be checking for "0.0.0.0" and "[::]" as well?

The same code is also duplicated in [src/helpers/llamaServer.js](https://github.com/OpenWhispr/openwhispr/blob/81256e3406c5c3e6c4f646097bed7fac9907a14b/src/helpers/llamaServer.js#L105)

xAlcahest:
## Summary

On Windows, `isPortAvailable()` only tested binding to `127.0.0.1`, which succeeds even when another app (like Duplicati) already holds `0.0.0.0` on the same port. OpenWhispr would bind `127.0.0.1:8200` and steal loopback traffic from the other app.

The fix adds a `0.0.0.0` wildcard probe before the existing `127.0.0.1` check. A port is considered available only if both binds succeed. The probes run sequentially (close the wildcard before testing loopback) to avoid self-conflict. On Linux/macOS, the extra check is redundant but harmless since the kernel already prevents wildcard/loopback coexistence for listening sockets.

## Changes

- src/utils/serverUtils.js: double-bind probe in shared `isPortAvailable()`
- src/helpers/cliBridge.js: same fix in standalone copy
- src/helpers/llamaServer.js: same fix in standalone copy
- src/helpers/whisperServer.js: same fix in standalone copy

Fixes #748

### DIFF
--- a/src/helpers/cliBridge.js
+++ b/src/helpers/cliBridge.js
@@ -21,13 +21,19 @@ function getBridgeFilePath() {
 
 function isPortAvailable(port) {
   return new Promise((resolve) => {
-    const server = net.createServer();
-    server.once("error", () => resolve(false));
-    server.once("listening", () => {
-      server.close();
-      resolve(true);
+    const probe = net.createServer();
+    probe.once("error", () => resolve(false));
+    probe.once("listening", () => {
+      probe.close(() => {
+        const loopback = net.createServer();
+        loopback.once("error", () => resolve(false));
+        loopback.once("listening", () => {
+          loopback.close(() => resolve(true));
+        });
+        loopback.listen(port, HOST);
+      });
     });
-    server.listen(port, HOST);
+    probe.listen(port, "0.0.0.0");
   });
 }
 
--- a/src/helpers/cliBridge.js
+++ b/src/helpers/cliBridge.js
@@ -1,10 +1,10 @@
 const http = require("http");
-const net = require("net");
 const fs = require("fs");
 const os = require("os");
 const path = require("path");
 const crypto = require("crypto");
 const debugLogger = require("./debugLogger");
+const { isPortAvailable } = require("../utils/serverUtils");
 
 const PORT_RANGE_START = 8200;
 const PORT_RANGE_END = 8219;
@@ -17,24 +17,6 @@ const NO_CONTENT = Symbol("CliBridge.NoContent");
 
 function getBridgeFilePath() {
   return path.join(os.homedir(), ".openwhispr", "cli-bridge.json");
-}
-
-function isPortAvailable(port) {
-  return new Promise((resolve) => {
-    const probe = net.createServer();
-    probe.once("error", () => resolve(false));
-    probe.once("listening", () => {
-      probe.close(() => {
-        const loopback = net.createServer();
-        loopback.once("error", () => resolve(false));
-        loopback.once("listening", () => {
-          loopback.close(() => resolve(true));
-        });
-        loopback.listen(port, HOST);
-      });
-    });
-    probe.listen(port, "0.0.0.0");
-  });
 }
 
 async function findAvailablePort() {
--- a/src/helpers/llamaServer.js
+++ b/src/helpers/llamaServer.js
@@ -104,13 +104,19 @@ class LlamaServerManager {
 
   isPortAvailable(port) {
     return new Promise((resolve) => {
-      const server = net.createServer();
-      server.once("error", () => resolve(false));
-      server.once("listening", () => {
-        server.close();
-        resolve(true);
+      const probe = net.createServer();
+      probe.once("error", () => resolve(false));
+      probe.once("listening", () => {
+        probe.close(() => {
+          const loopback = net.createServer();
+          loopback.once("error", () => resolve(false));
+          loopback.once("listening", () => {
+            loopback.close(() => resolve(true));
+          });
+          loopback.listen(port, "127.0.0.1");
+        });
       });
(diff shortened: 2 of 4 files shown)


gabrielste1n:
A couple of suggestions from a code-quality review pass:

**(Strong) Consolidate the three duplicated copies to import from `serverUtils.js`.**
`isPortAvailable` is now patched in four places (`serverUtils.js`, `cliBridge.js`, `llamaServer.js`, `whisperServer.js`). The shared helper at `src/utils/serverUtils.js` already exports it — `cliBridge.js`, `llamaServer.js`, and `whisperServer.js` should import from there instead of carrying private copies. This is the third time the same 11-line function has had to be touched in lockstep; consolidating now prevents the next person from missing one of the copies.

**(Optional) Add a `[::]` probe for IPv6 wildcard coverage.**
The original bug report (#748) shows the conflicting app also listening on `[::]:8200`:
```
TCP    [::]:8200              [::]:0                 LISTENING       19644
```
On Windows the default `IPV6_V6ONLY=1` socket option means an IPv6 wildcard bind doesn't conflict with IPv4 binds, so an app holding only `[::]:port` would still be invisible to the current `0.0.0.0` + `127.0.0.1` probe. Adding a third probe to `::` (or documenting that only IPv4 conflicts are detected) would close the remaining gap.

gabrielste1n:
One more code-shape suggestion: once you consolidate to a single copy in `serverUtils.js`, the nested-callback pyramid can collapse into a small helper that makes the two-probe intent read in one line:

```js
function tryBind(port, host) {
  return new Promise((resolve) => {
    const s = net.createServer();
    s.once("error", () => resolve(false));
    s.once("listening", () => s.close(() => resolve(true)));
    s.listen(port, host);
  });
}

async function isPortAvailable(port) {
  return (await tryBind(port, "0.0.0.0")) && (await tryBind(port, "127.0.0.1"));
}
```

Same behavior, no nested callbacks, and the wildcard-then-loopback intent is obvious from the one-line return. If you add the optional `[::]` probe later, it's a third `&&` clause.

Otiel:
Thanks for the fix ❤️

PR #772, opened 2026-05-14
feat: global hotkey to re-paste last transcription

borng:
## Summary
  - Adds a global hotkey (default **Alt+Shift+Z**) that pastes the most recent transcription, configurable in Settings.
  - Persists the hotkey across restarts; supports explicit-disable and round-trips through storage cleanly.
  - Awaits `registerSlot` in all paste-last call sites so registration races can't drop the binding.
  - Clipboard is preserved
  
Screenshot 2026-05-14 132302

  ## Commits
  - `feat: global hotkey to paste last transcription (default Alt+Shift+Z)`
  - `fix(paste-last): await registerSlot in all paste-last call sites`

  ## Test plan
  - [ ] Verify default hotkey (Alt+Shift+Z) pastes the last transcription on macOS / Windows / Linux
  - [ ] Rebind in Settings -> confirm persists across restart
  - [ ] Clear the hotkey in Settings -> confirm stays disabled across restarts
  - [ ] `node --test test/helpers/pasteLastKey.test.js` (13/13 passing)


borng:
Should build as long as tokens are set... let me know of any questions

borng:
Would like to get merged before gets too behind the HEAD.

Seems like there's a number of times the transcription doesn't go through, this solves this issue of the cursor not being in the right place, etc.

borng:
Rebased onto current main (was ~520 commits behind) — conflicts with the new voice-agent and translation hotkeys are resolved, the paste-last row now uses `HotkeyListInput`, and cross-slot conflict validation covers all six hotkey slots. Typecheck and the full test suite (2009 tests) pass. Happy to adjust anything if there's appetite to land this — it's still the only fix for the "transcription pasted into the wrong place / nothing pasted" case, which keeps coming up.

borng:
Rebased again onto current main (was ~142 commits behind after the last round). All 9 CI checks are green, and locally the full suite passes on Node 24 (2992 tests, 0 failures) plus clean typecheck and format check.

Notable adaptations in this rebase:
- Migrated the paste-last settings row to `HotkeyListInput` (following #936's multi-hotkey migration), matching the sibling hotkey rows.
- Cross-slot conflict validation now covers all five hotkey slots — dictation, meeting, voice agent, translation, and paste-last — in every direction, including the newer voice-agent and translation slots.
- Dropped the two `ci(win)` commits from this PR: they're superseded by the `electron-builder.unsigned-win.json` flow now on main, which handles unsigned fork builds more broadly than what I had. The PR is now just 2 focused commits (+521/−7 across 21 files).

Is there anything else you'd like done to get this over the line — happy to split, squash, or adjust the approach if the maintainers prefer something different.

PR #774, opened 2026-05-14
fix: bugginess in notes view state management

amahon:
### Summary
Fixes multiple state management bugs in the Notes view that caused note content, titles, and AI-generated enhancements to bleed between notes during switching.
### Problem
Users reported three symptoms:
- Note content colliding between notes
- AI summarizations/actions applied to the wrong note
- Note titles bleeding to another note
### Root causes identified
All traced to `PersonalNotesView.tsx`'s local state management:
1. **`saveTimeoutRef` never cleared after debounced save fired** — the stale timer ID permanently blocked external updates (AI titles, enhanced content, chat tool edits) from syncing to local state. When the user eventually switched notes, the flush overwrote AI-generated changes with stale local values.
2. **Async race window during note switching** — the `syncNote` effect used `await` to flush the old note's data before updating `activeNoteRef` and local state. During that gap, the component had already rendered the new note's content, but all refs (`activeNoteRef`, `localTitleRef`, `localContentRef`) still pointed to the old note. Any user input during this window saved to the wrong note ID.
3. **`onRunAction` read stale `localContent`** — the action picker callback read `localContent` directly instead of `editorNote.content`. On the first render after a note switch (before `syncNote` updated local state), this held the previous note's content. Clicking an action during this window sent the old note's content to AI processing but saved the result to the new note's ID.
4. **No `key` on `NoteEditor`** — React reused the same component instance across note switches, carrying over the TipTap editor, title DOM, and internal hook state. The title `contentEditable` div was synced via `requestAnimationFrame`, leaving a 1-frame window where old state was interactive.
5. **`localEnhancedContent` in `syncNote` deps** — caused unnecessary effect re-runs that amplified the stale-ref issues.
### Fixes
| # | Fix | Lines |
|---|-----|-------|
| 1 | Clear `saveTimeoutRef.current = null` / `enhancedSaveTimeoutRef.current = null` at the start of each debounced save callback so external updates are no longer permanently blocked | 305, 348 |
| 2 | Rewrite `syncNote` to be fully synchronous: capture old values, switch state + refs immediately, then flush old data fire-and-forget | 239-300 |
| 3 | `onRunAction` reads from `editorNote.content` / `editorNote.transcript` (which has the `isLocalSynced` fallback) instead of raw `localContent` | 968-999, 1007 |
| 4 | Add `key={editorNote.id}` to `<NoteEditor>` so React fully remounts on note switch | 921 |
| 5 | Replace `localEnhancedContent` state dep with `localEnhancedContentRef` in the sync effect; dep array reduced to `[activeNote]` | 125, 130-132, 283, 300 |
### Verification
- [x] `tsc --noEmit`: 0 errors
- [x] `npm run lint`: 0 errors (3 pre-existing warnings in unrelated files)
- [x] `npm run format`: clean
- [x] No new UI strings — no i18n changes needed

### DIFF
--- a/src/components/notes/PersonalNotesView.tsx
+++ b/src/components/notes/PersonalNotesView.tsx
@@ -122,10 +122,14 @@ export default function PersonalNotesView({
   const [syncedNoteId, setSyncedNoteIdState] = useState<number | null>(null);
   const localContentRef = useRef(localContent);
   const localTitleRef = useRef(localTitle);
+  const localEnhancedContentRef = useRef(localEnhancedContent);
   useEffect(() => {
     localContentRef.current = localContent;
     localTitleRef.current = localTitle;
   }, [localContent, localTitle]);
+  useEffect(() => {
+    localEnhancedContentRef.current = localEnhancedContent;
+  }, [localEnhancedContent]);
   const markNoteAsSynced = (id: number | null) => {
     activeNoteRef.current = id;
     setSyncedNoteIdState(id);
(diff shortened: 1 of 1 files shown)


gabrielste1n:
this is awesome thank you @amahon - i'm sorry i missed this before the release, will aim to release again soon.

PR #790, opened 2026-05-16
fix(realtime): dedup duplicate transcription completions by item_id

### LINKED ISSUE #770: Duplicate transcriptions happening
**Describe the bug**
Transcription text duplicated multiple time

**To Reproduce**
1. Use GPT 4o-min-transcribe
2. Disable dictation cleanup
3. Issue happens like 1 in 10 times intermittently ( havent been able to pin point the exact scenario)

**Expected behavior**
Should have been a non duplicate transcription. Here repeated 4 times. In another instance was repeated 3 times . So inconsistent repetition observed

**Screenshots**

Image

 - OS: Windows 10
 - Browser Chrome

sgrimbly:
Closes #770.

## Why

Users hit ~10% rate of identical-text 3–4× repetition with `gpt-4o-mini-transcribe`. Identical-verbatim repetition (not internal stutter) points to protocol-layer duplication: the server occasionally re-emits the same `conversation.item.input_audio_transcription.completed` event for one conversation item, and `openaiRealtimeStreaming.js:185–200` pushes each event's transcript onto `completedSegments` and emits the running joined text.

Another production OpenAI Realtime client ([daintreehq/daintree](https://github.com/daintreehq/daintree/blob/main/electron/services/VoiceTranscriptionService.ts)) explicitly dedups by `item_id` and documents the same server behaviour.

## What

Track a `Set` of seen `item_id`s per session. On a completion event with an `item_id` we've already processed, log at debug level and skip the push + `onFinalTranscript` emit. Reset the Set on each `connect()` so it can't grow across sessions.

Falls back to current single-fire behaviour when `item_id` is absent on the event — purely additive, no regression risk.

## Test plan

- [x] `npm run typecheck` clean
- [x] `npm run lint` clean on touched file
- [ ] Manual: dictate 10+ short clips with `gpt-4o-mini-transcribe`, cleanup disabled, watch debug logs for "OpenAI Realtime duplicate completion ignored" entries — their presence confirms the dedup is firing.
- [ ] @kunal118 to verify against their original repro.

## Files

1 file, +16 / −0.

## Honest framing

**Mitigation, not proven elimination.** I can't reproduce locally (intermittent, no Realtime API key). Confidence breakdown:

- ~85% chance this catches the dominant root cause (server re-emission of `.completed` events; daintree's precedent strongly supports this hypothesis).
- ~60% chance this catches *all* duplicates. Residual mechanism, if any, is likely renderer-side listener accumulation in `audioManager.js` (`provider.onFinal` listeners not cleaned up between sessions) — separate small fix if reports persist after this lands.
- ~5% regression risk (additive guard; legitimate consecutive turns get unique item_ids; safe fallback when item_id absent).

### DIFF
--- a/src/helpers/openaiRealtimeStreaming.js
+++ b/src/helpers/openaiRealtimeStreaming.js
@@ -26,6 +26,10 @@ class OpenAIRealtimeStreaming {
     this.coldStartBuffer = [];
     this.coldStartBufferSize = 0;
     this.speechStartedAt = null;
+    // Tracks transcription-completed item_ids we've already pushed onto
+    // completedSegments so server re-emissions of the same event don't
+    // duplicate text in the final transcript (#770).
+    this.seenItemIds = new Set();
   }
 
   getFullTranscript() {
@@ -50,6 +54,7 @@ class OpenAIRealtimeStreaming {
     this.coldStartBuffer = [];
     this.coldStartBufferSize = 0;
     this.speechStartedAt = null;
+    this.seenItemIds = new Set();
 
     const url = "wss://api.openai.com/v1/realtime?intent=transcription";
     debugLogger.debug("OpenAI Realtime connecting", { model: this.model });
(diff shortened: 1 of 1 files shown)


gabrielste1n:
Closing as stale.

The underlying mechanism here was never reproduced or confirmed — the guard was written from a third-party client's precedent rather than a repro, and there's been no confirmation from the reporter or any further duplicate-transcription reports on the OpenAI realtime path since May. The branch has also been in conflict for months (`providerLabel` refactor in #1608 touched the same handler).

For anyone picking this up later: the code gap is still open on `main` — `conversation.item.input_audio_transcription.completed` at `src/helpers/openaiRealtimeStreaming.js` still pushes every event's transcript onto `completedSegments` with no `item_id` tracking. #770 stays open. If the duplication is reproduced on 1.8.3+ with debug logs, a fresh version of this guard is a reasonable fix.

Issue #508, opened 2026-03-26
macOS: "Pause media on dictation" does not pause Brain.fm desktop and needs a more general fallback


NathanSkene:
## Summary

On macOS, the existing "Pause media on dictation" feature works for Spotify in my testing, but it does not pause the Brain.fm desktop app. I also sometimes use Brain.fm in the browser, so the broader product need is a more general macOS fallback for non-scriptable players, not just a Brain.fm-specific special case.

## Current behavior

Validated on macOS:
- Spotify desktop: pauses/resumes correctly during dictation
- Brain.fm desktop: does not pause when dictation starts

The current implementation appears to handle:
- direct AppleScript control for known players like `Spotify` and `Music`
- `MediaRemote` fallback for some other media sessions

That still leaves gaps for players that:
- do not expose a simple AppleScript play/pause API
- do not show up through the current `MediaRemote` path

## Expected behavior

When "Pause media" is enabled, OpenWhispr should pause and later resume a wider range of actively playing macOS media apps, including Electron-based desktop players like Brain.fm where possible.

## Why this matters

The feature is already state-aware and user-visible in settings, so from a user perspective it reads as a generalized media-pause feature. In practice it currently behaves more like partial support for a subset of players.

## Reproduction

1. Enable `Pause media` in Settings
2. Start playback in Brain.fm desktop app on macOS
3. Start dictation
4. Observe that Brain.fm continues playing

## Notes

A robust fix may need an additional best-effort fallback on macOS for non-scriptable apps, potentially via accessibility/UI scripting or a better session-discovery path, while still failing safely and only resuming apps that OpenWhispr actually paused.

m13v:
ran into the same gap building a macOS voice agent. AppleScript control only covers Spotify and Music, and MediaRemote private framework is the next layer but it still misses Electron-based players like Brain.fm that don't register their now-playing session properly.

what actually worked as a reliable fallback for us was the accessibility API approach. you find the player's window via AXUIElement, walk the AX tree to locate the play/pause button by role and description, then trigger AXPress on it. works for basically anything with a visible play button, regardless of whether the app exposes AppleScript or MediaRemote hooks. the tricky part is that different players put the button at different depths in the tree, so you need some fuzzy matching on the element description.

the other option is simulating the media key press (F8/NX_KEYTYPE_PLAY) via CGEvent. that sends a hardware-level play/pause to whatever has the active media session, but it can conflict if multiple apps are playing.

[...]

Issue #419, opened 2026-03-12
Background media are merely switched over instead of actually being turned off.


officialbishowb:
**Describe the bug**
The problem is that if you have a media or YouTube video running in the background and have paused it manually, when you try to use OpenWhispr, it will toggle on/off making the YouTube video which was actually paused play again, which is very annoying. I had look at the code, and it should normally work with the `_isWindowsAudioPlaying`() method, but somehow it is not working.

**To Reproduce**
1. Go to Youtube.com
2. Play a video and pause it manually
3. Open OpenWhispr and try to use it
4. While OpenWhispr is running, the YouTube video will also play (instead of remaining paused).

**Expected behavior**
If a media / video is actually paused, then it should stay paused and not turned on again by OpenWhispr

**Desktop**
 - OS: Windows 11
 - Browser Firefox
 - Version 1.6.2

xAlcahest:
> **Describe the bug** The problem is that if you have a media or YouTube video running in the background and have paused it manually, when you try to use OpenWhispr, it will toggle on/off making the YouTube video which was actually paused play again, which is very annoying. I had look at the code, and it should normally work with the `_isWindowsAudioPlaying`() method, but somehow it is not working.
> 
> **To Reproduce** Steps to reproduce the behavior:
> 
>     1. Go to Youtube.com
> 
>     2. Play a video and pause it manually
> 
>     3. Open OpenWhispr and try to use it
> 
>     4. While OpenWhispr is running, the YouTube video will also play (instead of remaining paused).
> 
> 
> **Expected behavior** If a media / video is actually paused, then it should stay paused and not turned on again by OpenWhispr
> 
> **Desktop**
> 
>     * OS: Windows 11
> 
>     * Browser Firefox
> 
>     * Version 1.6.2

Can you give me the debug logs?

officialbishowb:
Sure. It is very odd as yesterday I could not reproduce the bug (but nothing was being logged even though debugging was on). And now all together the whole media stopping feature is not working. I had Spotify / YouTube running in the background and when I started OpenWhispr the media were still playing in the background.

[debug.log](https://github.com/user-attachments/files/25992345/debug-2026-03-14T07-09-28-840Z.log)

After full clean installation of OpenWhispr (deleting the transcations.db..) the result was the same. I think I broke something :)

[debug-new.log](https://github.com/user-attachments/files/25992455/debug-2026-03-14T07-30-21-959Z.log)

[...]

PR #806, opened 2026-05-19
fix(media): pause/resume on macOS 15.4+ via vendored mediaremote-adapter

### LINKED ISSUE #794: interrupt / pause / mute background sounds while transcription is active is not working
I'm always frustrated when I push the button for the voice transcription, and the music keeps running in the background. I would like it to be interrupted / paused / muted because the settings are set to interrupt, but it doesn't happen on my Mac.

sgrimbly:
## Summary

On macOS 15.4+ "Pause media on dictation" silently no-ops for everyone — Spotify, Apple Music, Brain.fm, browser players. Root cause: Apple closed the now-playing daemon to unprivileged Mach-O processes in 15.4, and the fallback path that #749 left in place sends `key code 100` (F8) via System Events — which is **not** a media key, just the F8 character. No player listens for it.

This PR makes the feature work on 15.4+ by bundling `ungive/mediaremote-adapter` and loading it through `/usr/bin/perl`, which is the one binary Apple still trusts to talk to MediaRemote post-15.4. The state-aware "only resume what we paused" guarantee from #419 is preserved.

Closes #794. Same root cause as #508 and #632 (both closed at the time of #749; if maintainers agree this PR is what those reports were actually asking for, they're candidates to reopen-and-close).

## Root cause

#749 correctly identified that macOS 15.4+ closes the MediaRemote daemon and short-circuits to `UNKNOWN`, falling through to `_sendMacMediaKey`. That helper invokes `osascript -e 'tell application "System Events" to key code 100'`. Key code 100 is `kVK_F8`, not a media-play event. Modern macOS apps listen for `NSEventSubtypeMediaKey` events with key type `NX_KEYTYPE_PLAY (16)`, which is a different event class entirely — so the fallback was silently dropped.

The wider ecosystem solves the 15.4+ daemon closure by shelling out to `/usr/bin/perl`, which is an Apple-signed system binary with implicit entitlement to use MediaRemote. The `mediaremote-adapter` project (BSD-3) packages this approach with a clean CLI.

## Approach

- **Primary path:** `/usr/bin/perl <adapter.pl> <framework.path> get` to read state (playing flag + bundle ID), then `send 1` (kMRAPause) or `send 0` (kMRAPlay). Works on all macOS versions where MediaRemote exists, including 15.4+.
- **Fallback:** the rewritten `macos-media-remote` helper now posts a real `NSEventSubtypeMediaKey` event via `CGEventPost(.cghidEventTap)` — `NX_KEYTYPE_PLAY << 16 | state << 8`. Only fires if the adapter is unavailable.
- **Async:** macOS pause/resume converted from `spawnSync` to `child_process.spawn`-based promises. Adds ~80 ms wall-clock per pause but **0 ms main-thread block** (was ~10–50 ms blocking).
- **State-aware:** `_pausedViaAdapter` flag — resume only fires if we actually paused. Preserves #419's "don't unpause already-paused media" guarantee.

## Vendored code

- `resources/mediaremote-adapter/` pinned to upstream `v0.7.6`. BSD-3 LICENSE retained and shipped at `bin/mediaremote-adapter.LICENSE`. Removed `src/test/` and `src/adapter/test.m` (test client — not used by this consumer).
- `scripts/build-mediaremote-adapter.js` — direct `clang -fobjc-arc` build (avoids adding cmake as a build dep), hash-gated incremental. Wired into `compile:native` so it builds alongside the other native helpers.
- `electron-builder.json` ships the framework, Perl script, and LICENSE into `Contents/Resources/bin/`.
- `scripts/afterPack.js` registers `.framework` bundle paths (skipping inner Mach-Os to avoid double-sign) so `electron-osx-sign` re-signs the framework with the developer identity.

## Test plan

- [x] `npm run compile:mediaremote-adapter` clean (arm64)
- [x] Adapter `get` returns correct JSON state on macOS 26.3 (post-15.4)
- [x] `send 1` / `send 0` pause/resume Spotify from the dev tree
- [x] Async path via `mediaPlayer.pauseMedia()` — pause 70 ms, resume 30 ms, no main-thread block
- [x] `npm run pack` — framework + 4 symlinks (Versions/Current, top-level MediaRemoteAdapter/Headers/Resources) intact in packed `.app`
- [x] `codesign --verify --deep --strict` on packed framework — valid, satisfies Designated Requirement, `adhoc,runtime` flag set
- [x] Adapter loads via Perl from inside packed `.app`, full pause/resume round-trip against Spotify
- [x] End-to-end through the actual app: pause-media-on-dictation toggle works while recording, music pauses and resumes
- [x] `npm run lint` / `npm run typecheck` — no new errors in changed files
- [ ] Notarised build — only verifiable in CI; structural shape is correct (`adhoc,runtime` flag inherited from `hardenedRuntime: true`)
- [ ] Verify on macOS 14.x / 15.0–15.3 (pre-15.4 path goes through the adapter too; should behave identically since the underlying MediaRemote APIs haven't changed there)
- [ ] Test fallback path with framework artificially removed (CGEvent media-key helper)

## Performance

| | Wall-clock | Main-thread block |
|---|---|---|
| Pause (before) | ~10–50 ms | ~10–50 ms (spawnSync) |
| Pause (this PR) | ~80 ms (perl get + send) | **0 ms** (spawn) |
| Resume (this PR) | ~30 ms (perl send) | **0 ms** (spawn) |

Net: small wall-clock cost for a much-better-behaved main process. Wall-clock can be cut further by pre-warming the framework load at app launch — happy to do as a follow-up.

## Follow-ups (out of scope, in priority order)

1. Pre-warm the adapter at app launch so first-pause is warm.
2. Add a per-app AppleScript final-fallback (`tell application "Spotify" to pause` etc.) for the cases where MediaRemote and the CGEvent media key both fail.
3. Long-running `stream`-mode subscription for push-based state — eliminates the `get` spawn per pause.

## Notes for review

- Delivers what #749's description flagged as the planned follow-up — `mediaremote-adapter` for proper macOS 15.4+ state detection.
- Linux MPRIS and Windows GSMTC paths are untouched.
- The `macos-media-remote` Swift binary was retained but completely rewritten — its sole purpose now is `--media-key-toggle` (CGEvent fallback). The old `--is-playing`/`--pause`/`--play` modes are gone; verified no other callers in the repo.


gabrielste1n (approved):
Verified end-to-end on macOS Tahoe (where the old binary returns UNKNOWN): the vendored adapter's `get` returns accurate now-playing state and `send 0`/`send 1` round-trip play/pause correctly via /usr/bin/perl. This also kills a nasty side effect of the old fallback — `key code 100` injected a literal F8 keystroke into the frontmost app on every dictation start/stop, which we've traced as the likely cause of the residual focus-loss reports in #843.

Merged main into the branch, resolved the package.json conflict, and pushed a small cleanup: `get --no-artwork` on the pause probe (drops ~40KB of base64 per dictation), removed a resolver path no packaging layout produces, added the framework + perl script to the release workflow's packaged-binary verification, and formatting. Typecheck, lint, prettier, and the full test suite (403 passing) are green. Thanks @sgrimbly — excellent root-cause work and a clean vendoring job.

gabrielste1n:
Thank you @sgrimbly for the excellent work here — the root-cause analysis, the clean vendoring of mediaremote-adapter, and the thorough test plan made this a pleasure to review. Verified working end-to-end on macOS Tahoe. This should also clear up the residual focus-loss reports in #843 (the old fallback was typing F8 into the focused app on every dictation). Merging now — this will ship in the next release. 🙏

PR #826, opened 2026-05-20
fix(macos): cancel bare-Fn push-to-talk when Fn is used as a navigation modifier

### LINKED ISSUE #684: macOS: Fn push-to-talk keeps recording when using Fn+arrow/Fn+delete navigation
# macOS: Fn push-to-talk keeps recording when using Fn+arrow/Fn+delete navigation

## Description

On macOS, when the configured hotkey is the **Globe/Fn key** in **push-to-talk (hold)** mode, pressing standard Fn-mapped navigation keys while Fn is held does not stop or cancel the recording. The app only sees the Fn modifier still being held and keeps recording until Fn is released.

Affected key combinations on Mac (all use Fn as a modifier):
- Fn+Left → Home
- Fn+Right → End
- Fn+Up → Page Up
- Fn+Down → Page Down
- Fn+Backspace → Forward Delete

## Steps to reproduce

1. Set hotkey to Globe/Fn, activation mode = Push to talk.
2. Hold Fn (recording starts after the 150 ms hold threshold).
3. While still holding Fn, press an arrow key or Backspace to navigate / delete in the focused app.
4. Release Fn.

**Expected:** When the user presses a non-modifier key while Fn is held, the app should recognize that Fn is being used as a navigation modifier and cancel/abort the in-progress recording (not transcribe the few hundred ms of audio captured).

**Actual:** Recording continues for the full duration Fn is held; on release, the captured audio is transcribed and pasted.

## Root cause (suspected)

`resources/macos-globe-listener.swift` only subscribes to `NSEvent.flagsChanged`. It tracks the Fn modifier flag transitions (FN_DOWN/FN_UP) but never observes `keyDown` events. There is no signal in the pipeline that the user pressed another key while Fn was held, so `main.js` (`globe-down` / `globe-up` handlers around lines 817–894) cannot react.

## Suggested fix

Have the globe listener also monitor `keyDown` events. When a keyDown is observed while `fnIsDown == true`, emit a new event (e.g. `FN_INTERRUPTED`) that the main process uses to abort the in-flight push-to-talk recording.

## Environment

- OS: macOS
- Hotkey: Globe / Fn
- Activation mode: Push-to-talk

Related: #362 (Fn key recognition / double-press hotkeys).

gabrielste1n:
## Summary

Fixes #684. On macOS, holding Fn for push-to-talk while pressing Fn+Arrow/Backspace (Home/End/Page Up/Page Down/Forward Delete) kept recording for the full Fn hold and transcribed the captured noise into the focused app.

The root cause is in [resources/macos-globe-listener.swift](resources/macos-globe-listener.swift): the listener only subscribed to `.flagsChanged`, so when Fn was used as a navigation modifier, no signal flowed to the main process. `globe-down` had already fired and started the 150ms hold timer, and `globe-up` only fired when Fn was finally released — at which point we transcribed the buffered audio.

## Approach

- **Swift listener:** add a second `.keyDown` global monitor that emits `FN_INTERRUPTED` once per Fn hold when another key fires while `fnIsDown == true`. Reset on Fn release.
- **`globeKeyManager.js`:** parse `FN_INTERRUPTED` and emit a new `globe-interrupted` event.
- **`windowManager.js`:** add `sendCancelDictation()` that posts the existing `cancel-hotkey-pressed` IPC (the same channel used by the Escape-to-cancel pipeline). This reuses `audioManager.cancelRecording()` in the renderer, which discards audio chunks instead of transcribing.
- **`main.js`:** handle `globe-interrupted` by clearing the bare-Fn hold state. If we were already recording, cancel via `sendCancelDictation`; if still pre-threshold, hide the panel.

## Scope (what is and isn't affected)

| Scenario | Behavior |
|---|---|
| Hotkey = bare Fn, Push-to-talk + Fn+Arrow | **Cancels** (this is the fix) |
| Hotkey = bare Fn, Tap-to-talk | Untouched — gate variables are never set in tap mode |
| Hotkey = Fn+letter/Fn+F-key, Push-to-talk | Untouched — compound path uses `macCompoundPushState`, not `globeKeyIsRecording` |
| Hotkey = non-Fn (backtick, F8, etc.) | Untouched — bare-Fn handler never starts recording |
| Mouse button hotkeys | Untouched — different code path |

## Test plan

- [ ] Hotkey = Globe/Fn, mode = Push-to-talk: hold Fn + arrow keys → recording cancels, nothing transcribed
- [ ] Hotkey = Globe/Fn, mode = Push-to-talk: hold Fn alone → release → still transcribes normally
- [ ] Hotkey = Globe/Fn, mode = Tap-to-talk: tap Fn → still toggles normally
- [ ] Hotkey = Fn+Shift+G, mode = Push-to-talk: compound hotkey still records and stops on Fn release
- [ ] Hotkey = backtick: Fn+arrow does nothing, dictation still works
- [ ] Escape-to-cancel still works during recording


alex-shumilov:
Hi @gabrielste1n — I have a parallel fix for #684 in #685, so I did a close comparison. The core mechanics here match what I converged on independently (`FN_INTERRUPTED` once-per-hold latch, reuse of the `cancel-hotkey-pressed` Esc path, cooldown re-arm), and I actually adopted two things from this PR into mine: `sendCancelDictation()` resetting meeting-detection state, and hiding the panel on pre-threshold interrupt. Nice catches.

One gap worth flagging: `NSEvent.addGlobalMonitorForEvents(matching: .keyDown)` only delivers events when the process is trusted (Accessibility or Input Monitoring). For users who haven't granted Accessibility (e.g. manual-paste users), `keyMonitor` is nil or never fires and the fix silently doesn't exist — no prompt, no UI state, no way for the user to know. #685 handles this angle: it probes permission state, emits `INPUT_MONITORING:granted/denied` to the main process, re-probes every 3s so a grant takes effect without restart, surfaces a permission card + inline warning in Settings, and detects TCC resets across app updates. It also adds a configurable hold-duration setting and a distinct cancel audio cue.

Happy to either consolidate into one PR or split the permission-UX into a follow-up — whatever the maintainers prefer.

Issue #184, opened 2026-02-03
stt does not work in terminal and coding cli


RutamBhagat:
**Is your feature request related to a problem? Please describe.**

<img width="2560" height="1382" alt="Image" src="https://github.com/user-attachments/assets/853040cf-61bf-442f-a218-6f566e4a9617" />

In codex cli the transcription is treated as an image is there any workaround for this?

**Describe the solution you'd like**
Ideally the text should work just like any other input boxes

**Describe alternatives you've considered**
currently i just copy paste the content from the ide back to the terminal 

**Additional context**
I think its the same in claude code, and opencode doesnt even sees it as image either

gabrielste1n:
@RutamBhagat what platform are you on?

RutamBhagat:
I am on fedora, I also tried it on konsole but didnt work
```
OS: Fedora Linux 42 (KDE Plasma Desktop Edition) x86_64
Kernel: Linux 6.18.5-100.fc42.x86_64
Shell: zsh 5.9
DE: KDE Plasma 6.5.5
WM: KWin (Wayland)
Terminal: warp v0.2026.01.28.08.14.stable_04
CPU: AMD Ryzen 7 9800X3D (16) @ 5.27 GHz
GPU 1: NVIDIA GeForce RTX 5080 [Discrete]
```

gabrielste1n:
thank you very much for reporting - i will prioritise this today! I'm in PST so a bit late now but will look at it first thing :)

gabrielste1n:
Marking as duplicate of #224 and closing but this should be resolved in v1.4.9 @RutamBhagat please test and let me know! Will reopen if still persistent.

JGKle:
Trying this out in Konsole (current master), not getting any text in the raw console, nor in Claude Code nor in Codex. It works elsewhere (i.e. text editor, browser).

gabrielste1n:
@JGKle what platform are you on and what version of OW?

JGKle:
Sorry, the parens were unclear in my prior comment - "latest master" was referring to testing on the latest master of OW (1.5.0 when I posted last night, but just reconfirmed on 1.5.2).

Konsole is the terminal for KDE (Linux)

gabrielste1n:
@JGKle can you please test if this works for you in 1.5.3?

JGKle:
Nope, no go.
* In Konsole directly, nothing happens
[...]

PR #827, opened 2026-05-20
fix(linux): paste via Shift+Insert when window context is unknown

gabrielste1n:
## Summary

On Linux, when OpenWhispr can't detect the target window's class (typical for Wayland-native terminals on GNOME, since `xdotool` can't see them), it falls through to `Ctrl+V`. Modern TUI agents — Codex, Claude Code, OpenCode, anything Ink-based — bind `Ctrl+V` to "paste image" and surface `No image found in clipboard` instead of typing the dictation. Reported in #184 by @stefano5699 on Fedora GNOME.

## What changed

- **`linux-fast-paste.c`**: add `--shift-insert` flag. All three injection backends (uinput, portal/RemoteDesktop, XTest) now emit `Shift+Insert` when set. The existing `force_terminal` / atspi / X11 class detection logic is consolidated into a single `resolve_paste_mode()` helper returning a three-state enum (`CTRL_V`, `CTRL_SHIFT_V`, `SHIFT_INSERT`).
- **`clipboard.js`**: lift the `useShiftInsert` decision above the fast-paste call so the binary path honors it the same way the xdotool fallback already did. Updated condition is `detectedIsKonsole || (isWayland && windowSignals.length === 0)`.
- **PRIMARY selection mirror**: write the transcription to PRIMARY alongside CLIPBOARD on Linux. Terminals like alacritty, foot, konsole, and xterm bind `Shift+Insert` to PRIMARY, not CLIPBOARD — without this dual-write, the fix would only help vte-based terminals. `_readPrimarySelection` / `_writePrimarySelection` use the same `wl-copy → xclip → xsel → Electron` fallback chain as `_writeClipboardWayland`. Save/restore is gated on the existing `restoreClipboard` option, so `keepTranscriptionInClipboard=false` continues to leave both selections untouched after paste.

## Why Shift+Insert

It's the only Linux paste shortcut that's both terminal-safe (no TUI image-paste interception) and GUI-safe (GTK/Qt/Electron/browser text fields all honor it). `Ctrl+Shift+V` would fix the TUI case but breaks in some GUI apps that bind it to "paste without formatting". Shift+Insert has no widely-bound alternate meaning.

## Compatibility

- macOS / Windows: unchanged (Linux-only code paths).
- Linux pre-existing `--terminal` flag and Konsole+X11 xdotool routing: preserved.
- Binary is rebuilt from source on every install via `scripts/build-linux-fast-paste.js` (no release-pipeline coordination needed).
- `_runPortalPaste` signature change is internal; only caller is updated in the same commit.

### DIFF
--- a/resources/linux-fast-paste.c
+++ b/resources/linux-fast-paste.c
@@ -21,6 +21,16 @@
 #include <atspi/atspi.h>
 #endif
 
+/* Paste key sequence. SHIFT_INSERT is the universal Linux paste shortcut —
+ * works in terminals and GUI apps, and (unlike Ctrl+V) isn't intercepted by
+ * TUI agents as "paste image". Used when window class is unknown on Wayland
+ * or when targeting Konsole (which silently drops simulated Ctrl+Shift+V). */
+typedef enum {
+    PASTE_MODE_CTRL_V        = 0,
+    PASTE_MODE_CTRL_SHIFT_V  = 1,
+    PASTE_MODE_SHIFT_INSERT  = 2,
+} paste_mode_t;
+
 #ifdef HAVE_GIO
 #include <gio/gio.h>
 
@@ -33,6 +43,7 @@
 #define PORTAL_KEY_LEFTCTRL  29
 #define PORTAL_KEY_LEFTSHIFT 42
 #define PORTAL_KEY_V         47
+#define PORTAL_KEY_INSERT    110
 
 static int portal_exit_code = 0;
 
@@ -42,7 +53,7 @@ typedef struct {
     char            *session_handle;
     char            *restore_token;
     guint            signal_id;
-    int              use_shift;
+    paste_mode_t     mode;
 } PortalData;
 
 static char *get_sender_path(GDBusConnection *conn)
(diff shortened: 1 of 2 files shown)


xAlcahest:
The current detection doesn't handle VS Code's integrated terminal correctly. On KDE Wayland, kdotool sees the window class as `code`, which isn't in the terminal classes list, so it sends Ctrl+V. But inside VS Code's integrated terminal, TUI agents like Claude Code or Codex intercept Ctrl+V as "paste image" and you get `No image found in clipboard` instead of the transcription. The app can't tell the difference between VS Code's editor and its terminal because the whole thing reports as a single window.

Took a simpler approach: if xdotool can't see the active window (standard on native Wayland), always paste via Shift+Insert. It works the same as Ctrl+V in GUI apps, but doesn't get intercepted by TUI agents in terminals. No window class lists to maintain, no detection to get wrong.

For X11 where xdotool works, added Electron process detection via `/proc/PID/exe` -- if the binary sits next to `resources/app.asar` or `resources/app/`, it's Electron and gets Shift+Insert too. Catches VS Code, Cursor, and any future Electron app without a hardcoded list.

The C binary (`linux-fast-paste.c`) now accepts `--shift-insert` across all three backends (portal, uinput, XTest) using parameterized key variables instead of duplicated code paths.

Didn't touch the PRIMARY selection mirror yet, that can be a follow-up for terminals like alacritty that read Shift+Insert from PRIMARY instead of CLIPBOARD.

Diff: https://gist.github.com/xAlcahest/a5262f3d8b5aeea64a8c0cf052270499

PR #830, opened 2026-05-20
fix: meeting stream reconnect on session expiry + onnx worker packaging

xAlcahest:
## Summary

Meetings longer than 60 minutes hit the OpenAI Realtime API session limit and the WebSocket dies silently. Audio keeps flowing to a dead socket, no reconnection happens, and the user loses all transcription from that point on. This PR adds automatic reconnection with a proactive timer at 55 minutes (graceful, zero audio loss) and a reactive fallback on `session_expired` (safety net with pre-connect buffering).

Separately, the ONNX worker was missing from production builds because `src/workers/` wasn't in the electron-builder file list or asarUnpack config. The worker crashed in a loop on every app launch after any note was saved. This broke semantic search on all platforms.

## Changes

- `src/helpers/openaiRealtimeStreaming.js`: proactive session timer at 55min, `onSessionExpired` callback, `session_expired` error interception, pre-connect audio buffer when ws is null
- `src/helpers/ipcHandlers.js`: `reconnectMeetingStreams()` function that swaps streaming instances before token fetch, fetches new credentials, connects, then disconnects old instances. Guard against concurrent reconnects and cap at 5 reconnects per session
- `electron-builder.json`: added `src/workers/**/*` to `files` and `asarUnpack`
- `src/helpers/onnxWorkerClient.js`: path fix from `app.asar` to `app.asar.unpacked` for `utilityProcess.fork()`
- `test/helpers/openaiRealtimeStreaming.test.js`: 17 tests covering session expiry handling, pre-connect buffer, cold-start flush, zero-audio-loss verification across reconnect boundary, concurrent expiry guard


gabrielste1n (approved):
Verified the reconnect orchestration end to end, including failure rollback and zero-audio-loss buffering during session expiry. Dictation error flow and packaged worker resolution were re-verified after the latest fixes, and the full suite is green.

PR #836, opened 2026-05-21
feat: add keep-alive ping for self-hosted transcription servers

### LINKED ISSUE #766: Feature: Option to leave local transcription model loaded in VRAM while idle
When using a local transcription model on the GPU, after a period of idle the model is unloaded from VRAM. This causes two issues.
1. If using a large model, this causes a delay when resuming transcription while the model is reloaded into VRAM. If the user does not notice this delay, this results in the first bit of the speech being cut off because transcription has not actually begun yet.
2. If VRAM is being used for other purposes, the model becoming unloaded unpredictably can lead to confusion about how much free VRAM will be available. For example, when tuning a local LLM configuration, I have to constantly check to see if the OpenWhispr model is loaded or not to know that it's not going to crash.

I would like an option in the settings to keep the local transcription model loaded in VRAM at all times, even when idle.

xAlcahest:
## Summary

Self-hosted transcription servers (faster-whisper, etc.) offload models from VRAM after idle periods, causing multi-second delays when transcription resumes. This adds an automatic keep-alive that sends a minimal silent audio transcription every 120 seconds, resetting the server's idle timer so the model stays loaded.

The keep-alive activates automatically when self-hosted mode is configured with a URL and stops when the user switches to a different mode. No settings or UI changes needed.

Fixes #766

## Changes

- src/helpers/selfHostedKeepAlive.js: new module that pings the server with a 0.1s silent WAV transcription every 120s, resolving the model name from /v1/models on first ping
- src/helpers/ipcHandlers.js: two IPC handlers to start/stop the keep-alive from the renderer
- preload.js: bridge methods for the new IPC handlers
- src/stores/settingsStore.ts: Zustand subscription that auto-starts/stops keep-alive when transcriptionMode or remoteTranscriptionUrl changes
- src/types/electron.ts: type definitions for the new preload APIs
- test/helpers/selfHostedKeepAlive.test.js: 15 tests covering lifecycle, URL validation, model resolution, and transcription ping routing

### DIFF
--- a/preload.js
+++ b/preload.js
@@ -888,4 +888,7 @@ contextBridge.exposeInMainWorld("electronAPI", {
   getUpdateNotificationData: () => ipcRenderer.invoke("get-update-notification-data"),
   updateNotificationReady: () => ipcRenderer.invoke("update-notification-ready"),
   updateNotificationRespond: (action) => ipcRenderer.invoke("update-notification-respond", action),
+
+  selfHostedKeepAliveStart: (url) => ipcRenderer.invoke("self-hosted-keep-alive-start", url),
+  selfHostedKeepAliveStop: () => ipcRenderer.invoke("self-hosted-keep-alive-stop"),
 });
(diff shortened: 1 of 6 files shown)


TrickyWidget:
Local
Please note that my original report was not about "Self-Hosted" models. It is about "Local" models loaded and managed by OpenWhispr. If I understand the above correctly, your proposed solution will not address the issue.

Issue #838, opened 2026-05-22
[Linux] [has PR with fix!] v1.7.2 regression: local Whisper transcription broken — "No available ports in range 8178-8199"


firepol:
**Describe the bug**

After updating from v1.7.0 to v1.7.1 or v1.7.2 (this affects both versions) on Linux, local Whisper transcription fails completely. The internal whisper-server process never starts, reporting that no port in the range 8178–8199 is available — even on a freshly booted system with those ports confirmed free. Downgrading to v1.7.0 resolves the issue immediately with no other changes.

**To Reproduce**

1. Install v1.7.1 or v1.7.2 RPM on Fedora Linux
2. Open Settings → Speech to Text → select a local model (e.g. OpenAI Base 141 MB)
3. Record audio using the hotkey
4. Observe transcription failure with the error below

**Expected behavior**

The whisper-server starts successfully on an available port in the 8178–8199 range and transcription completes, as it did in v1.7.0.

**Screenshots**

N/A — error is in logs:

```
[INFO] Transcription mode: SERVER { model: 'base', language: 'auto' }
[WARN] VAD requested but ggml-silero model not found; running without VAD
[ERROR] ERROR: Local Whisper transcription error {
  "name": "Error",
  "message": "No available ports in range 8178-8199",
  "stack": "Error: No available ports in range 8178-8199\n    at WhisperServerManager.findAvailablePort (/opt/OpenWhispr/resources/app.asar/src/helpers/whisperServer.js:278:11)\n    at async WhisperServerManager._doStart (...whisperServer.js:315:17)\n    at async WhisperServerManager.start (...whisperServer.js:303:7)\n    at async WhisperManager.transcribeViaServer (...whisper.js:298:5)\n    at async WhisperManager.transcribeLocalWhisper (...whisper.js:282:12)"
}
```

**Desktop:**
- OS: Fedora Linux 43 (x86_64)
- Version: 1.7.2 (broken) — v1.7.0 works fine

**Smartphone:** N/A

**Additional context**

- Ports 8178–8199 are free (verified with `ss -tlnp`)
- No stale whisper-server processes running before launch
- There is also a secondary error `"Not authenticated"` on note config fetch, but this appears to be a separate issue — the root problem is the whisper-server never binding to any port
- Installed via RPM on Fedora 43; downgrading to v1.7.0 RPM fixes it immediately

**Workaround**

[...]

PR #839, opened 2026-05-22
fix: whisper-server not starting: handle EAFNOSUPPORT when IPv6 is disabled in port availability check

firepol:
## Problem

On Linux hosts where IPv6 is disabled, whisper-server fails to start with:

```
[WARN] Server pre-warm failed (will start on first use) { error: 'No available ports in range 8178-8199', model: 'base' }
```

**Root cause:** `isPortAvailable()` in `src/utils/serverUtils.js` checked availability by binding to three addresses in sequence: `0.0.0.0`, `::` (IPv6), and `127.0.0.1`. On any host where IPv6 is disabled, the `::` bind always fails with `EAFNOSUPPORT` (Address Family Not Supported). Because `tryBind` returned a plain boolean and treated *all* errors identically, every port in the 8178–8199 range appeared occupied — and `findAvailablePort` threw before a server could start.

Reported upstream: https://github.com/OpenWhispr/openwhispr/issues/838

## Fix

`tryBind` now resolves with `{ ok, code }` instead of a bare boolean, so callers can distinguish between "port already in use" (`EADDRINUSE`) and "address family not supported" (`EAFNOSUPPORT`).

`isPortAvailable` treats `EAFNOSUPPORT` on the IPv6 check as **"IPv6 not present — skip"** rather than **"port unavailable"**. IPv4 checks (`0.0.0.0` and `127.0.0.1`) still guard against actual conflicts.

```js
// Before
s.once("error", () => resolve(false));

// After
s.once("error", (err) => resolve({ ok: false, code: err.code }));
```

```js
// Before
async function isPortAvailable(port) {
  return (
    (await tryBind(port, "0.0.0.0")) &&
    (await tryBind(port, "::")) &&
    (await tryBind(port, "127.0.0.1"))
  );
}

// After
async function isPortAvailable(port) {
  const v4 = await tryBind(port, "0.0.0.0");
  if (!v4.ok) return false;

  const v6 = await tryBind(port, "::");
  // EAFNOSUPPORT means IPv6 is disabled on this host — skip, don't fail
  if (!v6.ok && v6.code !== "EAFNOSUPPORT") return false;

  const lo = await tryBind(port, "127.0.0.1");
  return lo.ok;
}
```

## Changes

- **`src/utils/serverUtils.js`** — `tryBind` and `isPortAvailable` updated as described above
- **`test/helpers/serverUtils.test.js`** — 5 new unit tests covering:
  - port available on IPv4-only host
  - `findAvailablePort` returns the first free port in range
  - `isPortAvailable` returns `false` when port is already bound
  - `findAvailablePort` skips in-use ports
  - `findAvailablePort` throws when all ports in range are exhausted

## Testing

```
node --test test/helpers/*.test.js
# ✔ 17/17 pass (5 new + 12 pre-existing)
```

### DIFF
--- a/src/utils/serverUtils.js
+++ b/src/utils/serverUtils.js
@@ -8,18 +8,22 @@ const GRACEFUL_STOP_TIMEOUT_MS = 5000;
 function tryBind(port, host) {
   return new Promise((resolve) => {
     const s = net.createServer();
-    s.once("error", () => resolve(false));
-    s.once("listening", () => s.close(() => resolve(true)));
+    s.once("error", (err) => resolve({ ok: false, code: err.code }));
+    s.once("listening", () => s.close(() => resolve({ ok: true })));
     s.listen(port, host);
   });
 }
 
 async function isPortAvailable(port) {
-  return (
-    (await tryBind(port, "0.0.0.0")) &&
-    (await tryBind(port, "::")) &&
-    (await tryBind(port, "127.0.0.1"))
-  );
+  const v4 = await tryBind(port, "0.0.0.0");
+  if (!v4.ok) return false;
+
+  const v6 = await tryBind(port, "::");
+  // EAFNOSUPPORT means IPv6 is disabled on this host — skip, don't fail
+  if (!v6.ok && v6.code !== "EAFNOSUPPORT") return false;
+
+  const lo = await tryBind(port, "127.0.0.1");
+  return lo.ok;
 }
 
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Closing this as superseded — the IPv6 probe issue was fixed on main by #1238 (treats EAFNOSUPPORT/EADDRNOTAVAIL as port-available) and shipped in v1.7.6. Really appreciate you tracking this regression down and putting up a fix — the analysis here directly informed the final change. 🙏

Issue #828, opened 2026-05-20
LMs -> cloud provider -> Groq -> Qwen3 32B unsupported


turicam:
When Qwen3 32B selected, openwhispr does not process the text, and just defaults to pasting the result of the speech-to-text.

**This can be verified by **
(Language Models)
selecting `Cloud Providers`
selecting `Groq`
inputting functional Groq API key
going to test mode in prompt studio
running a test
*Output*: `Test failed: property 'think' is unsupported`

**Expected behavior**
The model to work

**Screenshots**
<img width="651" height="501" alt="Image" src="https://github.com/user-attachments/assets/ac6ff599-3127-4d2b-a138-f5b4e20e8d04" />

**Desktop (please complete the following information):**
 - OS: Windows 11

**Additional context**
Other models under Groq do actually work.

Qwen3 32B is a great free model that can be used with openwhispr. This is one of the only realistic options for free users, as local models are too slow, and most other models cost.

kzzalews:
Similar issue with other provider/model: https://github.com/OpenWhispr/openwhispr/issues/831

gabrielste1n:
Closing as a duplicate of #844 — same root-cause family: the thinking-suppression payload is rejected by Groq's strict validator on Qwen3 32B. A fix is in flight in #841; please follow #844.

PR #841, opened 2026-05-22
fix(reasoning): restrict chat_template_kwargs to local servers, use safe reasoning_effort for custom providers

### LINKED ISSUE #831: chat_template_kwargs: property 'chat_template_kwargs' is unsupported
I'm using OpenWhispr 1.7.2 with Cerebras.ai (url: `https://api.cerebras.ai/v1`) and `gpt-oss-120b` model. After updating to this version I'm getting:

```
Test failed: chat_template_kwargs: property 'chat_template_kwargs' is unsupported
```

when I try to test this model in section `Language Model -> Dictation Cleanup -> Prompt studio -> Test`. 

See: https://github.com/OpenWhispr/openwhispr/commit/9668a5297fdaef6b4521dcf969a31330db60f673

xAlcahest:
## Summary

Custom OpenAI-compatible providers like Cerebras do strict request validation and reject unknown fields. The `chat_template_kwargs` parameter added in [#802](<https://github.com/OpenWhispr/openwhispr/issues/802>) was sent unconditionally to all providers, causing "property 'chat_template_kwargs' is unsupported" errors when "Disable thinking output" is enabled. On top of that, `reasoning_effort: "none"` isn't universally supported either: Cerebras only accepts `low`/`medium`/`high`.

Verified both issues against the live Cerebras API with `gpt-oss-120b`. `chat_template_kwargs` is only understood by local inference servers (llama.cpp, vLLM, SGLang, LM Studio), while cloud providers use `reasoning_effort`. Known providers (OpenAI, Groq) keep `"none"`, custom endpoints get `"low"` as the safest minimum.

Fixes OpenWhispr/openwhispr#831 OpenWhispr/openwhispr#828

## Changes

* src/services/ai/thinkingSuppression.ts: gate `chat_template_kwargs` behind `isLocalServer()` so it only reaches `local`/`lan` providers; use `reasoning_effort: "low"` for `custom` providers instead of `"none"` which strict endpoints reject


gabrielste1n:
Nice fix — the core change is correct and I'd like to get it in. I verified against the live provider docs that hosted APIs (OpenAI, Groq, Cerebras) reject unknown fields, so gating `chat_template_kwargs` to local servers is exactly right.

Two follow-ups:

**1. The test file doesn't run.** `node --test test/services/thinkingSuppression.test.ts` fails with `ERR_MODULE_NOT_FOUND` (the import is missing the `.ts` extension), and there's no test runner wired up anyway — no `test` script, no vitest/tsx, no CI step, and `test/` isn't in the typecheck scope. So the tests are inert right now. Could we either drop the test file from this PR and keep it to the one-function fix, or land a small "wire up a TS test runner + CI" PR first and add these on top?

**2. `reasoning_effort: "low"` for all custom providers is a bit blunt.** It's a safe "won't-400" floor, but it doesn't actually *disable* thinking, so a custom endpoint pointed at OpenAI/Groq (which do support `"none"`) gets degraded behavior when the user asked for off — and it's not universal: Mistral's adjustable models only accept `none`/`high`, so `low` would 400 there. Could be a follow-up rather than blocking, but I'd lean toward resolving it declaratively — drive known providers off the model registry (we already have `disableThinking`/`supportsThinking`), and add a per-endpoint setting for unknown custom URLs (mirroring `remoteReasoningType`) — rather than one hardcoded value. Happy to pair on it.

gabrielste1n:
Closing this one — main has since moved to a per-provider thinking-suppression dialect system (`thinkingSuppressionDialects.ts`, built up across #1103, #1211, and #1213) which now handles Groq, Mistral, Gemini, and OpenRouter with model-aware `reasoning_effort`, so the approach here (gating the old `thinkingSuppression.ts` on `isLocalServer()`) no longer applies to the current code.

That said, the underlying problem this PR targeted is **not** fully solved: the fallback path still sends `chat_template_kwargs` + `reasoning_effort: "none"` to unlisted/custom OpenAI-compatible endpoints, so strict providers like Cerebras (originally reported in #831 / #828) may still hit 400s. The right fix now is a new dialect entry for generic custom providers rather than a rebase of this branch.

PR #849, opened 2026-05-24
Fix custom LLM API compatibility

sysusugan:
## Summary
- avoid sending unsupported reasoning effort values to OpenAI-compatible custom LLM APIs
- route custom chat models through the chat model interface and avoid tool calls for custom agents
- add a default OpenWhispr Cloud API URL fallback for self-built apps
- surface empty custom LLM responses with localized error text

## Tests
- node --test test/services/thinkingSuppressionPolicy.test.js test/config/openwhisprCloud.test.js
- npm run typecheck


gabrielste1n:
Closing after a per-item evaluation against current main — thanks for the work here; the one change that was still a live bug has been landed, credited to this PR.

1. **Route custom chat models through the chat interface** — confirmed still a real bug (the AI SDK default factory posts to /responses, which LiteLLM/vLLM/DeepSeek don't implement) and merged via #1181, crediting this PR.
2. **Reasoning-effort suppression rework** — superseded: main has since gained provider-specific suppression (Gemini minimal, OpenRouter reasoning:{enabled:false}), and #1103 (merged) reworked the self-hosted path to reasoning:{effort:"none"} with a 400/422 drop-and-retry for strict backends. The policy here would also have stripped reasoning_effort from Groq, where main sends it intentionally.
3. **Disabling tool calls for custom agents** — contradicts the now-merged #1146 direction (custom BYOK stays on its provider route), and with item 1 fixed, custom Chat Completions endpoints support standard tool calling.
4. **Default production API URL for self-built apps** — main intentionally gates cloud features on an empty OPENWHISPR_API_URL; defaulting self-built apps to the official API is a product decision rather than a bug fix.
5. **Localized empty-response error** — the main trigger (Responses/Chat Completions mismatch yielding empty output) is removed by item 1; if wanted as safety-net UX it's best as its own minimal PR against current locale files.

PR #850, opened 2026-05-24
Add Vulkan GPU acceleration for AMD and Intel GPUs

EdwardJXLi:
Currently, local Whisper transcription only has GPU acceleration on NVIDIA (with CUDA). This PR adds a Vulkan-accelerated path so users with AMD or Intel GPUs get the same speedup, on Linux and Windows.

## Benchmarking

> **TL;DR: Achieved around ~20x speedup!**

**Hardware:** AMD Ryzen 7 PRO 7840U w/ Radeon 780M Graphics
**OS:** Arch on Linux 7.0.9

**Whisper Turbo without Vulkan acceleration:** 20853 ms mean transcription time for an 11 second audio clip
image

 **Whisper Turbo WITH Vulkan acceleration:** 1134 ms mean transcription time for an 11 second audio clip (**about 20x faster!**)
image

## Notes
- Windows is not well tested, as I don't have access to a native windows machine.


carsonbird:
Happy to help verify this on linux. Sorry, no windows either.

gabrielste1n:
Thanks so much for this contribution, @EdwardJXLi — and apologies for the long silence on it. The benchmarking you did here (~20x on the Radeon 780M!) was a big part of what convinced us Vulkan support was worth prioritizing.

Vulkan GPU acceleration for AMD/Intel GPUs has since shipped in main via #1185, which consumes prebuilt Vulkan binaries from OpenWhispr/whisper.cpp releases rather than requiring a separate build step, so this PR is now superseded. Closing it in favor of that implementation.

Really appreciate the work you put into this — contributions like this one genuinely shape the roadmap. Hope to see you around the repo!

PR #852, opened 2026-05-25
fix: discard dictionary prompt echo on silent audio

### LINKED ISSUE #842: Activating dictation without saying anything auto-pastes all dictionary entries
Whenever I press and release the dictation hotkey without saying anything, it pastes in all my dictionary entries.

I'm using Openwhispr desktop app version 1.7.2 (happened to me with the previous version as well)

Using it on Windows 11 with current updates.

Been great to use otherwise, thanks!

EDIT: this is when using OpenWhispr Cloud for speech-to-text and for the LLM/clean-up model.

xAlcahest:
## Summary

When Whisper receives silent audio with a custom dictionary prompt, it echoes the prompt content as if it were a real transcription. This is a well-documented Whisper behavior that causes all dictionary entries to be pasted when the user presses and releases the dictation hotkey without speaking.

The fix compares the transcription result against the dictionary prompt before processing. If the text is composed almost entirely of dictionary words (using word-set matching with dual thresholds), it gets treated as "no audio detected" instead of a valid transcription. This catches both verbatim echoes and the repeated-loop variant where Whisper cycles through the dictionary multiple times.

Fixes #842

## Changes

- `src/utils/dictionaryEchoFilter.js` adds `matchesDictionaryPrompt()`, a pure function that normalizes both strings and checks whether the transcription's unique words are a subset of the dictionary (90% text composition + 70% dictionary usage thresholds)
- `src/helpers/audioManager.js` imports the filter and calls `isDictionaryEcho()` in all four transcription paths: local Whisper, cloud OW, Mistral proxy, and BYOK/Groq direct
- `test/helpers/dictionaryEchoFilter.test.js` covers 16 cases: verbatim echo, repeated echo, added punctuation, case changes, stripped commas, legitimate speech, partial overlap, null/empty inputs, single-word dictionary, unicode, and CJK

### DIFF
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -14,6 +14,7 @@ import { getSettings, getEffectiveCleanupModel, isCloudCleanupMode } from "../st
 import { detectAgentName } from "../config/agentDetection";
 import { resolvePrompt } from "../config/prompts";
 import { syncService } from "../services/SyncService.js";
+import { matchesDictionaryPrompt } from "../utils/dictionaryEchoFilter.js";
 
 const REASONING_CACHE_TTL = 30000; // 30 seconds
 const REALTIME_MODELS = new Set(["gpt-4o-mini-transcribe", "gpt-4o-transcribe"]);
@@ -207,6 +208,10 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
   getCustomDictionaryPrompt() {
     const words = getSettings().customDictionary;
     return words.length > 0 ? words.join(", ") : null;
+  }
+
+  isDictionaryEcho(text) {
+    return matchesDictionaryPrompt(text, this.getCustomDictionaryPrompt());
   }
 
   setCallbacks({
@@ -712,6 +717,9 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
       );
 
       if (result.success && result.text) {
+        if (this.isDictionaryEcho(result.text)) {
+          throw new Error("No audio detected");
+        }
         const rawText = result.text;
         const reasoningStart = performance.now();
         const text = await this.processTranscription(result.text, "local");
@@ -1316,6 +1324,9 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
     timings.transcriptionProcessingDurationMs = Math.round(performance.now() - transcriptionStart);
 
     const rawText = result.text;
+    if (this.isDictionaryEcho(rawText)) {
+      throw new Error("No audio detected");
+    }
     let processedText = result.text;
     if (processedText && !this.skipReasoning) {
       const reasoningStart = performance.now();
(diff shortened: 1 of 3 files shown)


gabrielste1n:
## Code quality review

Traced all four integration points to their error handlers and ran the test suite (16/16 pass on Node 24). Overall this is a solid, well-scoped fix that correctly reuses the existing `"No audio detected"` sentinel — throwing it propagates to the pipeline catch (`audioManager.js:674`) which already suppresses the error UI, so an echo is treated like genuine silence (no paste, no toast). Parakeet is correctly excluded since `processWithLocalParakeet` only passes `{ model }` and never receives the dictionary prompt. The primary reported scenario (#842, OpenWhispr Cloud) is handled gracefully.

A few findings:

### 1. BYOK/Groq/Mistral path isn't graceful when `allowLocalFallback` is enabled (minor, edge case)

Unlike `processWithLocalWhisper` (line 758) and `processWithLocalParakeet` (line 833), the `processWithOpenAIAPI` catch (line 1751) doesn't re-throw `"No audio detected"` early. With local fallback turned on, an echo throws `"No audio detected"` → triggers a local-Whisper retry → that fails → produces a combined message (`"OpenAI API failed: No audio detected. Local fallback also failed: ..."`) which is no longer equal to `"No audio detected"`, so line 674 shows a "Transcription Error" toast instead of silently discarding.

`allowLocalFallback` defaults to `false`, so it's opt-in and never wrong-pastes (still strictly better than the original bug), but for consistency with the other two local paths the catch should short-circuit first:

```js
} catch (error) {
  if (error.message === "No audio detected") throw error;  // mirror line 758
  const isOpenAIMode = !getSettings().useLocalWhisper;
  ...
```

### 2. Known tradeoffs worth documenting (not blocking)

- A **single-word dictionary** spoken alone is discarded (a test asserts this) — if a user's dictionary is one word and they dictate exactly that word, it's treated as an echo.
- Symmetrically, a **partial echo of a large dictionary** (<70% of dict words) leaks through.

Both are reasonable points on the precision/recall curve; just flagging them as intentional behavior.

### 3. Dead defensive branch (nit)

In `dictionaryEchoFilter.js`, `uniqueTextWords.size === 0 || dictWords.size === 0` can never be true: empty inputs are already short-circuited by the `!text || !dictionaryPrompt` guard, and `"".split(" ")` returns `[""]` (size 1), never 0. Harmless, removable.

xAlcahest:
Both code points are in the latest push: the `processWithOpenAIAPI` catch now re-throws the sentinel before trying local fallback, same as the other two paths, and the dead guard in `dictionaryEchoFilter.js` is gone. On the tradeoffs, yes, intentional. Both behaviors are pinned by tests ("handles single-word dictionary" and "does not flag speech that partially overlaps with dictionary"), so any future change to that curve will fail loudly.

gabrielste1n (approved):

PR #854, opened 2026-05-25
fix: skip evdev listener in tap mode when KDE shortcuts are active

### LINKED ISSUE #847: KDE Plasma 6: modifier-only hotkey + tap mode double-triggers (records for ~150 ms)
## Summary

On KDE Plasma 6 (X11 *or* Wayland with the XWayland Electron build), a modifier-only dictation hotkey such as `Control+Super` combined with `ACTIVATION_MODE=tap` causes the dictation pipeline to start and stop on the same key press, recording for only ~100–200 ms.

The cause is OpenWhispr registering the same hotkey via two paths simultaneously:
1. **KGlobalAccel D-Bus** (`src/helpers/kdeShortcut.js`) — fires the shortcut callback on `KEY_UP` for modifier-only hotkeys
2. **`LinuxKeyManager`** evdev listener (`src/helpers/linuxKeyManager.js`) — explicitly started in tap mode for modifier-only / right-side-modifier hotkeys (`main.js:1363-1367`)

In tap mode both paths call `sendToggleDictation()`, so one key press toggles **on** (KEY_DOWN via LinuxKeyManager) then immediately toggles **off** (KEY_UP via KGlobalAccel).

## Environment

- OpenWhispr **1.7.2** (Linux AppImage, `--ozone-platform=x11`)
- Fedora 44 KDE
- Plasma **6.6.5**, session type **wayland**, OpenWhispr running as XWayland client
- KGlobalAccel D-Bus available (KDE shortcut path is chosen)
- Hotkey: `Control+Super`, mode: `tap`

## Steps to reproduce

1. Run OpenWhispr 1.7.2 on KDE Plasma 6.
2. In `~/.config/open-whispr/.env`, set:
   ```
   DICTATION_KEY=Control+Super
   ACTIVATION_MODE=tap
   ```
3. Restart OpenWhispr.
4. Tap `Control+Super`, then try to speak.

## Expected

Tap once → recording starts; tap again → recording stops (standard toggle).

## Actual

Recording ends ~100–200 ms after the tap (matching the time between KEY_DOWN and KEY_UP). Audio blob is too small for the transcription backend to do anything useful.

## Log excerpt

```
[…] [LinuxKeyManager] KEY_DOWN detected { "key": "Control+Super" }
[…] User recording state changed { "active": true }
[…] Recording started with microphone {…}
[…] [LinuxKeyManager] KEY_UP detected { "key": "Control+Super" }
[…] [KDEShortcut] Shortcut pressed { "componentUnique": "openwhispr", "shortcutUnique": "dictation" }
[…] User recording state changed { "active": false }
[…] Recording stopped { "blobSize": 1262, "blobType": "audio/webm" }
[…] [transcription][renderer] Transcription returned empty - check audio input
```

Delta from KEY_DOWN to KEY_UP/recording stop: ~125 ms.

## Root cause (proposed)

`main.js` starts the native evdev listener in tap mode when the hotkey is modifier-only:

```js
// main.js:1363-1367
const needsNativeListener = (hotkey, mode) => {
  if (!isValidHotkey(hotkey)) return false;
  if (mode === "push") return true;
  return isRightSideMod(hotkey) || isModifierOnlyHotkey(hotkey);
};
```

And on `key-down` it unconditionally toggles in tap mode:

```js
// main.js:1369-1378
linuxKeyManager.on("key-down", (_key) => {
  if (!isLiveWindow(windowManager.mainWindow)) return;
  const activationMode = windowManager.getActivationMode();
  if (activationMode === "push") {
    windowManager.startWindowsPushToTalk();
  } else if (activationMode === "tap") {
    windowManager.sendToggleDictation();   // ← starts here
  }
});
```

Meanwhile the KGlobalAccel-registered callback (built by `windowManager.createHotkeyCallback()`) does **not** suppress itself in tap mode — only in push mode (`windowManager.js:234-240`). On modifier-only hotkeys KGlobalAccel typically delivers the press event on KEY_UP, which results in a second `sendToggleDictation()` call (= stop).

## Suggested fix direction

Either:
- In tap mode on KDE, **skip** the evdev `LinuxKeyManager` registration when KGlobalAccel is also active — let KGlobalAccel be the sole source. The slight KEY_UP-vs-KEY_DOWN delay is acceptable for tap.
- Or, in `windowManager.createHotkeyCallback()`, return early on Linux/tap when KGlobalAccel is the registered path and `LinuxKeyManager` is already active for the same key.
- Or, debounce `sendToggleDictation()` across both code paths with a longer window than the current 150 ms (the second toggle still arrives within ~200 ms in real taps).

## Workaround for users

Switch `ACTIVATION_MODE` to `push` (hold-to-talk) — in push mode the KGlobalAccel callback explicitly returns on Linux (`windowManager.js:234-240`), so only `LinuxKeyManager` fires and the cycle is clean (KEY_DOWN starts, KEY_UP stops).

Alternatively use a non-modifier-only hotkey (e.g. `Ctrl+Shift+Space`, `F8`) — then `needsNativeListener` returns false in tap mode and only KGlobalAccel handles the shortcut.

## Notes

Tested only on Plasma 6.6.5 + XWayland. Cannot say whether the same logic affects KDE Plasma 5 or other Wayland compositors that register a parallel shortcut path. File `kglobalshortcutsrc` shows the OpenWhispr component registers cleanly as `dictation=Meta+Ctrl+,none,OpenWhispr dictation`, i.e. KGlobalAccel does accept modifier-only registrations on Plasma 6.

xAlcahest:
## Summary

On KDE Plasma 6 (Wayland), modifier-only hotkeys like `Control+Super` in tap mode cause a double-trigger: LinuxKeyManager (evdev) fires on KEY_DOWN and KGlobalAccel fires on KEY_UP, toggling recording on and off within ~125ms. The recording blob is too small to transcribe.

The fix adds a single guard in `needsNativeListener()` that skips the evdev listener when KDE shortcuts are active in tap mode. KGlobalAccel becomes the sole toggle source, eliminating the conflict. Push mode is unaffected since evdev is still needed for KEY_DOWN/KEY_UP detection.

Fixes #847

## Changes

- `main.js` -- added `if (hotkeyManager.useKDE) return false;` in the Linux `needsNativeListener` function, preventing evdev from starting in tap mode when KGlobalAccel already handles the shortcut
- `test/helpers/needsNativeListener.test.js` -- 9 unit tests covering all combinations of KDE active/inactive, tap/push mode, modifier-only/normal/right-side-mod hotkeys, and invalid inputs

### DIFF
--- a/main.js
+++ b/main.js
@@ -1363,6 +1363,7 @@ async function startApp() {
     const needsNativeListener = (hotkey, mode) => {
       if (!isValidHotkey(hotkey)) return false;
       if (mode === "push") return true;
+      if (hotkeyManager.useKDE) return false;
       return isRightSideMod(hotkey) || isModifierOnlyHotkey(hotkey);
     };
 
--- a/test/helpers/needsNativeListener.test.js
+++ b/test/helpers/needsNativeListener.test.js
@@ -1,0 +1,70 @@
+const test = require("node:test");
+const assert = require("node:assert/strict");
+
+const {
+  isGlobeLikeHotkey,
+  isModifierOnlyHotkey,
+} = require("../../src/helpers/hotkeyManager");
+
+const isRightSideMod = (hotkey) =>
+  /^Right(Control|Ctrl|Alt|Option|Shift|Super|Win|Meta|Command|Cmd)$/i.test(hotkey);
+
+const isValidHotkey = (hotkey) => hotkey && !isGlobeLikeHotkey(hotkey);
+
+function makeNeedsNativeListener(hotkeyManager) {
+  return (hotkey, mode) => {
+    if (!isValidHotkey(hotkey)) return false;
+    if (mode === "push") return true;
+    if (hotkeyManager.useKDE) return false;
+    return isRightSideMod(hotkey) || isModifierOnlyHotkey(hotkey);
+  };
+}
+
+test("KDE active + tap + modifier-only → false (no evdev)", () => {
+  const fn = makeNeedsNativeListener({ useKDE: true });
+  assert.equal(fn("Control+Super", "tap"), false);
+});
+
+test("KDE active + push + modifier-only → true (evdev needed for key-up)", () => {
+  const fn = makeNeedsNativeListener({ useKDE: true });
+  assert.equal(fn("Control+Super", "push"), true);
+});
+
+test("KDE active + tap + normal hotkey → false (KGlobalAccel handles it)", () => {
+  const fn = makeNeedsNativeListener({ useKDE: true });
+  assert.equal(fn("Alt+R", "tap"), false);
+});
+
+test("KDE active + tap + right-side modifier → false", () => {
+  const fn = makeNeedsNativeListener({ useKDE: true });
+  assert.equal(fn("RightAlt", "tap"), false);
+});
+
+test("no KDE + tap + modifier-only → true (evdev needed)", () => {
+  const fn = makeNeedsNativeListener({ useKDE: false });
+  assert.equal(fn("Control+Super", "tap"), true);
+});
+
+test("no KDE + tap + right-side modifier → true", () => {
+  const fn = makeNeedsNativeListener({ useKDE: false });
+  assert.equal(fn("RightShift", "tap"), true);
+});
+
+test("no KDE + tap + normal hotkey → false (no evdev needed)", () => {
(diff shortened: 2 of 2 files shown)


xAlcahest:
Superseded by #872, which addresses the root cause more broadly (any compositor shortcut, not just KDE) and adds a safety-net guard in processAudio for degenerate recordings.

PR #860, opened 2026-05-28
fix: keep input focused when recording/processing begins (#859)

### LINKED ISSUE #859: Prevent Input focus from being repositioned.

Note: Using Windows 11 x64

Jawad-git:
## Problem

When the dictation window appears during PTT or toggle mode, apps like Brave and Windsurf lose focus from the original input field. The existing code captures the target **PID** (macOS) at hotkey time and restores focus at paste time — but there was nothing between panel-show and paste, and **no focus capture/restore at all on Windows**.

### Windows: native focus capture/restore

Extends `windows-fast-paste.c` with two new flags:

- `--capture-foreground` — calls `GetForegroundWindow()` and writes the HWND to `%TEMP%\openwhispr-focus-state.txt`
- `--restore-foreground` — reads the HWND from the file and calls `SetForegroundWindow()` to bring the original window back

This reuses the existing native binary infrastructure (same pattern as the paste functionality).

### windowManager.js: platform-agnostic focus management

- **`captureTargetFocus()`** — captures the foreground window before showing the panel. macOS: `textEditMonitor.captureTargetPid()`, Windows: native binary `--capture-foreground`, Linux: no-op.
- **`restoreTargetAppFocus()`** — restores focus to the captured target. macOS: `NSRunningApplication.activateWithOptions()`, Windows: native binary `--restore-foreground`, Linux: `blur()` fallback.

Both methods are called from:
- `startWindowsPushToTalk()` (Windows PTT)
- `sendToggleDictation()` (toggle mode)
- `startMacCompoundPushToTalk()` (macOS PTT)

### useAudioRecording.js: processing-time focus restore

When `isProcessing` transitions to true, sends `restoreTargetFocus` IPC so focus returns to the original app during transcription (not just at paste time).

### Key difference from previous approach

- No magic `setTimeout` delays — focus is captured and restored synchronously at the right moments
- Uses actual Win32 `SetForegroundWindow` instead of `blur()` — restores focus to the *exact* window, not just the z-order
- `showDictationPanel()` is left untouched — no global `restoreFocus` default that could break other flows

Fixes #859

### DIFF
--- a/preload.js
+++ b/preload.js
@@ -26,6 +26,7 @@ contextBridge.exposeInMainWorld("electronAPI", {
   pasteText: (text, options) => ipcRenderer.invoke("paste-text", text, options),
   hideWindow: () => ipcRenderer.invoke("hide-window"),
   showDictationPanel: () => ipcRenderer.invoke("show-dictation-panel"),
+  restoreTargetFocus: () => ipcRenderer.invoke("restore-target-focus"),
   onToggleDictation: registerListener("toggle-dictation", (callback) => () => callback()),
   onStartDictation: registerListener("start-dictation", (callback) => () => callback()),
   onStopDictation: registerListener("stop-dictation", (callback) => () => callback()),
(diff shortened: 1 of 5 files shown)


Jawad-git:
my bad I asked openclaw if it can fix it didn't know it'd open a pr lol

PR #868, opened 2026-05-30
feat(paste): smart spacing around dictated text (#856)

### LINKED ISSUE #856: Insert whitespace before or after text
I do not want to have to insert a space (" ") after a transcription.

Before inserting transcribed text, check the character immediately before the cursor (and optionally after). If it is a non-whitespace character, automatically prepend a space to the inserted text. Same logic for trailing space if the cursor is followed by a non-whitespace character.

I tried adding "The output must always end with a single trailing space character (" ") after the final punctuation mark" in the Prompt Studio without success.

**Additional context**

gabrielste1n:
## Summary

Closes #856.

Inserts a separator between previously-typed text and the new transcription so users don't have to dictate or type a space themselves. Works across macOS, Windows, and Linux with a two-tier strategy:

- **macOS** reads the character before the cursor via the Accessibility API (`AXFocusedUIElement` → `AXSelectedTextRange` + `AXValue`) and applies full prepend rules. 400ms timeout so paste latency is unaffected — on the slow path we fall back to append mode.
- **Windows / Linux** (and macOS when the AX read fails or isn't permitted) append a trailing space to the transcript instead. The next paste then sees a leading space and skips the prepend rule — self-correcting across dictations without needing platform-specific accessibility bridges.

On by default, no setting — it just works.

## Rules (prepend mode)

A leading space is added **unless** any of:
- Cursor is at the start of the field (no preceding char)
- Preceding char is whitespace (` `, `\t`, `\n`)
- Preceding char is an opening bracket / quote: `(`, `[`, `{`, `<`, `"`, `'`, `` ` ``, `“`, `‘`
- Transcript already starts with whitespace
- Transcript starts with closing punctuation: `,` `.` `!` `?` `;` `:` `)` `]` `}` `%` `”` `’`
  (so `Hello` + `, world` → `Hello, world`, not `Hello , world`)

## Failure handling

Smart spacing is wrapped in try/catch — any failure (AppleScript error, timeout, unexpected exception) falls through to pasting the raw transcript. A paste never blocks on this.

## Test plan

- [x] Unit tests for the pure `applySmartSpacing` module (20 cases: prepend rules, opening chars, closing punctuation, unicode preceding chars, empty input, append no-double-up)
- [x] Unit tests for `TextEditMonitor.getPrecedingChar` (non-darwin path, missing PID, timeout behavior)
- [x] Existing helper test suite (41 tests) still passes
- [x] `prettier --check` clean
- [x] `eslint` reports no new warnings
- [ ] Manual: macOS — dictate twice in TextEdit / Notes / Safari address bar / Slack message field. Verify space appears between dictations and not at field start.
- [ ] Manual: macOS — dictate after `"`, `(`, `[`. Verify no leading space inside the quote/bracket.
- [ ] Manual: macOS — dictate text starting with `,` or `.`. Verify no extra space before the punctuation.
- [ ] Manual: Windows — dictate twice. Verify trailing-space approach yields proper spacing between dictations.
- [ ] Manual: Linux — same as Windows.
- [ ] Manual: confirm paste latency feels unchanged (the 400ms AX read budget keeps it tight).

### DIFF
--- a/src/helpers/ipcHandlers.js
+++ b/src/helpers/ipcHandlers.js
@@ -15,6 +15,7 @@ const OpenAIRealtimeStreaming = require("./openaiRealtimeStreaming");
 const AudioStorageManager = require("./audioStorage");
 const liveSpeakerIdentifier = require("./liveSpeakerIdentifier");
 const MeetingEchoLeakDetector = require("./meetingEchoLeakDetector");
+const { applySmartSpacing } = require("./smartSpacing");
 const {
   transcriptsOverlap,
   transcriptsLooselyOverlap,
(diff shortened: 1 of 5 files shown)


xAlcahest:
I ran the Linux build and looked at how it behaves outside macOS. On Windows and Linux the app can't read the character before the cursor, so it always falls back to adding a space at the end of each dictation, and there's no setting to turn that off. In practice it's mostly fine. The spacing between dictations comes out clean, no double spaces, and since the app only pastes and never presses Enter for you, that trailing space only matters if you hit Enter yourself in a field that keeps it. I tried a Google search box and a shell prompt, both just trimmed it. The only thing that would actually break is a field that keeps trailing spaces and matches exactly, and I couldn't find one in normal use. So lower impact than it sounds, just worth knowing it's on by default with no way to opt out.

One thing that does reproduce: if you turn on "keep transcription in clipboard", the spaced text is what stays on the clipboard, not the raw transcript. So the next time you paste manually you get a trailing space you never dictated. In the code, with that setting on the original clipboard isn't saved, the spaced text gets written to the clipboard (`clipboard.js` around line 688), and the restore step in `pasteLinux` only runs when there's a saved original, so it never fires.

On tests: the pure spacing module is covered well. The part in the paste handler that decides how to space the text isn't tested, so a regression there would slip through CI green.

No crashes, no extra lag. The spacing itself works well on Linux.

xAlcahest (approved):
Tested the Linux build, spacing works well and no crashes. Left a couple of notes above (the keep-in-clipboard trailing space and an untested path), nothing blocking.

PR #872, opened 2026-05-30
Fix dictation hotkey double-trigger on KDE Wayland

### LINKED ISSUE #864: Audio file might be corrupted or unsupported
### Describe the bug

I got error **Transcription Error** when used hotkey for dictation.

### To Reproduce

1. Hot-Key for dictation.
2. Waiting 1 minute.
3. Get the error.

### Expected behavior
No error, usual functionality.

### Screenshot
No needed.

### Desktop
- OpenWhispr: 1.7.2
- OS: Fedora Linux 44 (KDE Plasma Desktop Edition) x86_64
- Kernel: Linux 7.0.10-201.fc44.x86_64

### Additional context
[debug-2026-05-29T06-55-40-104Z.log](https://github.com/user-attachments/files/28380288/debug-2026-05-29T06-55-40-104Z.log)

xAlcahest:
## Summary

On KDE Plasma Wayland the dictation hotkey double-fired in tap mode. setupHotkey registers the hotkey with KGlobalAccel (so isUsingNativeShortcut() is true), but a separate block in main.js also started the native /dev/input key listener whenever needsNativeListener() returned true. With the KDE default Control+Super (a modifier-only combo KWin handles natively) both fired for one press: the native KEY_DOWN toggled recording on and the KDE shortcut toggled it off on release. The result was an empty WebM the backend rejects as "Audio file might be corrupted or unsupported", shown to the user as "Transcription Error".

The fix keeps the native listener off in tap mode when a compositor shortcut (GNOME/KDE/Hyprland) already owns the hotkey, so the compositor is the only handler. Push-to-talk is left alone: there the native listener is still needed for key-down/key-up and the compositor's single toggle is a no-op on release, so hold-to-talk keeps working on KDE. A second guard in processAudio skips empty recordings (a blob smaller than 256 bytes, i.e. a WebM container with no audio frames) instead of sending them to a backend; it gates on size rather than duration, so a genuinely short real word still transcribes.

Both decisions live in small pure helpers (shouldUseNativeKeyListener, isEmptyRecording) with node:test unit tests.

## Changes

- main.js: in the Linux block, decide the native key listener via shouldUseNativeKeyListener, so it does not run alongside the GNOME/KDE/Hyprland compositor shortcut in tap mode; push mode keeps it for push-to-talk
- src/helpers/hotkeyManager.js: add shouldUseNativeKeyListener (pure, exported)
- src/helpers/recordingGuard.js: add isEmptyRecording, which skips blobs smaller than 256 bytes
- src/helpers/audioManager.js: in processAudio, skip empty recordings via isEmptyRecording
- test/helpers/nativeKeyListenerDecision.test.js and test/helpers/recordingGuard.test.js: node:test unit tests for both guards

Fixes #864 #847

### DIFF
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -553,6 +553,31 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
       return;
     }
 
+    // A recording that started and stopped within a few milliseconds (an accidental
+    // double-tap, or a hotkey double-trigger) yields an empty WebM container with no
+    // audio frames. Transcription backends reject it with "Audio file might be
+    // corrupted or unsupported". Skip it like the silence gate instead of surfacing
+    // an error. See issue #864.
+    const MIN_RECORDING_SECONDS = 0.2;
+    const MIN_AUDIO_BYTES = 256;
+    const recordingSeconds =
+      typeof metadata?.durationSeconds === "number" ? metadata.durationSeconds : null;
+    const blobSize = audioBlob?.size ?? 0;
+    if (
+      (recordingSeconds !== null && recordingSeconds < MIN_RECORDING_SECONDS) ||
+      blobSize < MIN_AUDIO_BYTES
+    ) {
+      logger.info(
+        "Skipping transcription: recording too short or empty to contain speech",
+        { blobSize, durationSeconds: recordingSeconds },
+        "audio"
+      );
+      this.isProcessing = false;
+      this.onStateChange?.({ isRecording: false, isProcessing: false });
+      this.onTranscriptionComplete?.({ success: true, text: "" });
+      return;
+    }
+
     try {
       const useLocalWhisper = settings.useLocalWhisper;
       const localProvider = settings.localTranscriptionProvider;
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -15,6 +15,7 @@ import { shouldSkipTranscriptionApiKey } from "./transcriptionAuth";
 import { detectAgentName } from "../config/agentDetection";
 import { resolvePrompt } from "../config/prompts";
 import { syncService } from "../services/SyncService.js";
+import { isEmptyRecording } from "./recordingGuard";
 
 const REASONING_CACHE_TTL = 30000; // 30 seconds
 const REALTIME_MODELS = new Set(["gpt-4o-mini-transcribe", "gpt-4o-transcribe"]);
@@ -553,23 +554,17 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
       return;
     }
 
-    // A recording that started and stopped within a few milliseconds (an accidental
-    // double-tap, or a hotkey double-trigger) yields an empty WebM container with no
-    // audio frames. Transcription backends reject it with "Audio file might be
-    // corrupted or unsupported". Skip it like the silence gate instead of surfacing
-    // an error. See issue #864.
-    const MIN_RECORDING_SECONDS = 0.2;
-    const MIN_AUDIO_BYTES = 256;
-    const recordingSeconds =
-      typeof metadata?.durationSeconds === "number" ? metadata.durationSeconds : null;
+    // A recording with no audio frames — an accidental double-tap or a hotkey
+    // double-trigger that toggles on and off within milliseconds — is just an empty
+    // WebM container that backends reject ("Audio file might be corrupted or
+    // unsupported"). Skip it like the silence gate instead of surfacing an error.
+    // Gated by size only: a genuinely short real utterance can be brief yet still
+    // carry audio, so a duration gate would drop it. See issue #864.
     const blobSize = audioBlob?.size ?? 0;
-    if (
-      (recordingSeconds !== null && recordingSeconds < MIN_RECORDING_SECONDS) ||
-      blobSize < MIN_AUDIO_BYTES
-    ) {
+    if (isEmptyRecording(blobSize)) {
       logger.info(
-        "Skipping transcription: recording too short or empty to contain speech",
-        { blobSize, durationSeconds: recordingSeconds },
+        "Skipping transcription: recording is empty (no audio captured)",
+        { blobSize },
         "audio"
       );
       this.isProcessing = false;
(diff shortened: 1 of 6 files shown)


src/helpers/hotkeyManager.js
```diff
@@ -63,6 +63,20 @@ function isMouseButtonHotkey(hotkey) {
...
+// In tap mode, when a compositor shortcut (GNOME/KDE/Hyprland) already owns the
+// hotkey (isUsingNativeShortcut), the native listener must stay off: otherwise it
+// and the compositor both fire for one keypress and the recording toggles on then
+// instantly off, leaving an empty blob. Push-to-talk still needs the native
+// listener for key-down/key-up, where the compositor's single toggle is a no-op
+// on release. See issue #864.
+function shouldUseNativeKeyListener(hotkey, mode, isUsingNativeShortcut) {
+  if (!hotkey || isGlobeLikeHotkey(hotkey)) return false;
+  if (mode === "push") return true;
+  if (isUsingNativeShortcut) return false;
```
gabrielste1n, line 76: **Medium — worth confirming before merge.** This gates on `isUsingNativeShortcut`, which is a coarse, session-global flag — it's true whenever the KDE/GNOME/Hyprland backend *initialized successfully* (`useGnome || useHyprland || useKDE`), not whether *this specific hotkey* is actually owned by the compositor.

That's exactly right for the reported bug (the default `Control+Super`, which KWin genuinely owns). My question is the right-side-modifier-only tap case the tests assert returns `false`, e.g. `shouldUseNativeKeyListener("RightControl", "tap", true)`. A single right-side modifier can't be registered as a KGlobalAccel/GNOME shortcut, so historically the native `/dev/input` listener was the *only* thing that caught it — and this now suppresses that listener whenever any compositor backend is active.

`updateHotkey` rejects failed compositor registrations (and restores the prior hotkey), so you can't *switch into* this state at runtime. The reachable path is **startup with a persisted right-side-modifier tap hotkey** on a Wayland compositor session. Could you confirm a right-side-modifier-only tap hotkey still fires on KDE/GNOME Wayland after this change? If not, the gate may want to be "does the compositor actually own *this* hotkey" rather than "is any compositor backend active."

src/helpers/audioManager.js
```diff
@@ -553,6 +554,25 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
       return;
     }
 
+    // A recording with no audio frames — an accidental double-tap or a hotkey
+    // double-trigger that toggles on and off within milliseconds — is just an empty
+    // WebM container that backends reject ("Audio file might be corrupted or
+    // unsupported"). Skip it like the silence gate instead of surfacing an error.
+    // Gated by size only: a genuinely short real utterance can be brief yet still
+    // carry audio, so a duration gate would drop it. See issue #864.
+    const blobSize = audioBlob?.size ?? 0;
```
gabrielste1n, line 563: **Minor / non-blocking.** Double defensiveness: this computes `audioBlob?.size ?? 0` and `isEmptyRecording` *also* re-checks `Number.isFinite`. `audioBlob` is always a freshly-constructed `Blob` here, so `.size` is guaranteed finite ≥ 0 — one of the two guards is enough. Harmless, just slightly more than the "trust internal values" guideline calls for.

src/helpers/recordingGuard.js
```diff
@@ -0,0 +1,13 @@
+// A recorded WebM/Opus blob with no audio frames — produced when the dictation
+// hotkey toggles recording on and off within milliseconds (an accidental
+// double-tap, or the KDE double-trigger fixed in main.js) — is essentially just
+// the container header, well under 256 bytes. Real speech, even a single short
+// word, produces far more. We gate on size, not wall-clock duration: a genuinely
+// short utterance can last under any reasonable time threshold yet still carry
+// real audio, so a duration gate would silently drop it. See issue #864.
+export const MIN_AUDIO_BYTES = 256;
```
gabrielste1n, line 8: **Minor / non-blocking.** The 256-byte threshold is empirical — the comment cites ~110 bytes for an empty container. If some Electron/Chromium version on a given platform emits a larger empty WebM header, the "Transcription Error" could reappear. The threshold is biased safely (real speech is far larger), so this is just a "may need tuning" note, not a defect.

test/helpers/recordingGuard.test.js
```diff
@@ -0,0 +1,46 @@
+const test = require("node:test");
```
gabrielste1n, line 1: **Minor / non-blocking (pre-existing).** These tests don't appear to be wired into CI — there's no `test` script in `package.json` and no workflow runs `node --test`. Same for all of `test/helpers/`, so not introduced here, but worth noting these new tests will only run if invoked manually.

gabrielste1n (commented):
## Code-quality review

Solid PR — well-scoped, the refactor is a genuine DRY win, and all 17 unit tests pass (ran them in an isolated worktree against this branch). Logic, comments, and test conventions all match the surrounding codebase. Ready to merge with **one question worth confirming** (left inline on `shouldUseNativeKeyListener`).

**What's good**
- Collapsing the inline `isRightSideMod` regex + hand-rolled `needsNativeListener` body into the exported, tested `shouldUseNativeKeyListener` removes a verbatim duplicate of the existing `isRightSideModifier` helper and gives the decision a single home. No dangling references left in the Linux block.
- The empty-recording guard fills a real gap: the silence gate samples the live-mic RMS on an interval, and an instant double-trigger can stop before that interval ever samples — so `speechGateDecision.skip` stays false and the empty blob reaches the backend as "Transcription Error." The size guard catches exactly that. Complementary, not duplicative.
- The new guard mirrors the silence gate's exit path exactly (`isProcessing = false` → `onStateChange` → `onTranscriptionComplete({ success: true, text: "" })`), so it reuses the established soft-skip contract.
- Tests follow repo convention (`node:test` + dynamic `import()` for ESM helpers) with good boundary coverage and a push-to-talk-preserved case.

Inline notes below: 1 medium question + 3 minor/non-blocking.

gabrielste1n:
Closing since #1065 supersedes this (as noted there), and both halves have since landed on main separately — the empty-recording guard via #891 (your 256-byte guard made it in essentially unchanged) and the compositor shortcut gating via #1001. Thanks!

PR #875, opened 2026-05-31
Add Tinfoil private inference provider

zachdotai:
## Summary

Adds Tinfoil as a first-class private inference provider for OpenWhispr LLM workflows.

This integration uses the official Tinfoil SDK instead of the generic OpenAI-compatible fetch path so requests go through Tinfoil’s verified enclave transport. Tinfoil is now available for cleanup, note formatting, dictation agent, and chat agent flows, including AI SDK streaming/tool usage.

## Changes

- Add `tinfoil` SDK dependency.
- Add a lazy-loaded Tinfoil client wrapper and non-streaming inference provider.
- Register Tinfoil in the provider registry and model catalog.
- Add secure Tinfoil API key storage through the existing encrypted secret/IPC/settings flow.
- Add Tinfoil to reasoning provider UI and Prompt Studio provider metadata.
- Route chat agent streaming/tools through the Tinfoil AI SDK provider.
- Prevent raw manual streaming from using Tinfoil, avoiding an unverified transport path.
- Document `TINFOIL_API_KEY` and Tinfoil network allowlist hosts.

## Test Plan

- `npm run typecheck`
- `npm run i18n:check`
- `npm run lint`
- `npm run build:renderer`
- `git diff --check`
- Verified Tinfoil appears in the reasoning provider picker with API key storage and model selection wired through the settings store.
- Verified Tinfoil cleanup/chat paths use the official SDK clients instead of the raw OpenAI-compatible fetch path, so enclave verification is preserved.
- Verified non-Tinfoil provider paths still compile after making the renderer AI model factory async.


gabrielste1n (approved):
absolutely legendary thank you sir @zachdotai

PR #904, opened 2026-06-05
feat(linux): implement PipeWire system audio capture for transcriptions

### LINKED ISSUE #901: Linux system audio capture non-functional in 1.7.2
**Describe the bug**
Meeting transcription captures microphone only. Remote participants/system audio is not transcribed.

I didn't get a dialog to choose the system audio or grant permission either.

**To Reproduce**
1. Open OpenWhispr
2. Open a Meeting Note
3. Start Transcription
4. Play Audio in Browser
5. Talk into Microphone
-> Browser audio not transcribed, Microphone input transcribed

**Expected behavior**
The system audio is transcribed as well. I found multiple indications that it is supposed to work on Linux 
- https://docs.openwhispr.com/platform/linux#meeting-audio-linux
- https://docs.openwhispr.com/guides/meeting-transcription
- https://github.com/OpenWhispr/openwhispr/blob/main/TROUBLESHOOTING.md#linux-system-audio-portal-issues

---
Environment
- OpenWhispr 1.7.2 (Linux x86_64 AppImage)

- OS: Nobara Linux (Fedora 43 base)
- Kernel: 7.0.9-200.nobara.fc43.x86_64
- Desktop: GNOME 49 on Wayland
- Audio: PipeWire 1.4.11 (PulseAudio compat layer)
- xdg-desktop-portal: 1.20.4
- xdg-desktop-portal-gnome: 49.0
- Browser: Firefox (Flatpack), Chromium
---

Relevant logs

```
[WARN][meeting][renderer] Linux system audio capture failed, continuing with mic only { error: 'Not supported' }
[INFO][meeting][renderer] Meeting transcription started successfully {
  systemAudioMode: 'portal',
  systemAudioStrategy: 'browser-portal',
  bufferedChunks: 28,
  bufferedSystemChunks: 0,
  ...
}
```

```
❯ rpm -q xdg-desktop-portal xdg-desktop-portal-gnome
  systemctl --user status xdg-desktop-portal-gnome
  
  xdg-desktop-portal-1.20.4-1.fc43.x86_64
  xdg-desktop-portal-gnome-49.0-1.fc43.x86_64
  ● xdg-desktop-portal-gnome.service - Portal service (GNOME implementation)
       Loaded: loaded (/usr/lib/systemd/user/xdg-desktop-portal-gnome.service; static)
      Drop-In: /usr/lib/systemd/user/service.d
               └─10-timeout-abort.conf
       Active: active (running) since Thu 2026-06-04 21:25:29 CEST; 12h ago
   Invocation: b35be6a24bb142e395190d6ed43e30a1
     Main PID: 85320 (xdg-desktop-por)
        Tasks: 6 (limit: 38218)
       Memory: 26M (peak: 27.2M)
          CPU: 621ms
       CGroup: /user.slice/user-1000.slice/user@1000.service/app.slice/xdg-desktop-portal-gnome.service
               └─85320 /usr/libexec/xdg-desktop-portal-gnome

  Jun 04 21:25:28 ada-pc systemd[84663]: Starting xdg-desktop-portal-gnome.service - Portal service (GNOME implementation)...
  Jun 04 21:25:29 ada-pc systemd[84663]: Started xdg-desktop-portal-gnome.service - Portal service (GNOME implementation).
```

---

Claude Code traced through the code and found the following

Two compounding issues:

1. portal-helper strategy never selected

`getMeetingSystemAudioCapabilityMode()` in `ipcHandlers.js` returns "portal" for Linux. The portal-helper strategy is only selected when `supportsPersistentGrant` is `true`, which requires `supportsNativeCapture` to be `true`. However, in `resources/linux-system-audio-helper.c`, `supportsNativeCapture` is hardcoded to false:

```
// linux-system-audio-helper.c, print_probe_json()
fputs(",\"supportsNativeCapture\":false", stdout);  // line 109
```

Probe output confirms:
`{"ok":false,"portalVersion":0,"supportsPersistMode":false,"supportsRestoreToken":false,"supportsSystemAudio":false,"supportsNativeCapture":false,"error":"portal_unavailable"}`

Additionally, the start command in the helper explicitly states the capture is not implemented:
```
// line 452
emit_event("warning", "capture_unimplemented",
           "Native PipeWire PCM capture is not implemented in this helper yet", NULL);
```

2. `browser-portal` fallback fails on GNOME Wayland

The fallback browser-portal strategy calls Electron's `getDisplayMedia` with `systemAudio: 'include'`. On GNOME Wayland, this returns 'Not supported' immediately — the portal chooser never appears and no audio track is returned. The `xdg-desktop-portal` ScreenCast interface provides video only; audio is not included in the PipeWire stream.

---

My current workaround is to create a virtual output audio device that combines both mic and system audio but then of course you don't get any kind of diarization at all.

Digging a bit in the PRs, it seems that 
- #566  Added loopback approach
- #582 Removed loopback approach

0x4bs3nt:
## Summary 

- Implement PipeWire PCM capture (S16LE, 24kHz, mono) via `PW_KEY_STREAM_CAPTURE_SINK`, matching the format used by the macOS audio tap and Windows loopback paths. Audio is streamed to stdout for the existing `LinuxPortalAudioManager` to consume.
- All PipeWire code is gated behind `#ifdef HAVE_PIPEWIRE`. Without `libpipewire-0.3` at compile time, the binary builds and behaves exactly as before — no regression.

Fixes #901.

## Build changes

`scripts/build-linux-system-audio.js` now detects `libpipewire-0.3` via pkg-config and adds `-DHAVE_PIPEWIRE` when available. The hash calculation includes PipeWire flags so the binary rebuilds when availability changes.

## CI Note / Warning

The Linux AppImage build in `release.yml` needs `libpipewire-0.3-dev` added to the `apt-get install` line to ship with PipeWire capture enabled. Without it, the AppImage builds without `HAVE_PIPEWIRE` — no regression, but no fix in the released binary.

## Test plan

- [x] Verified on NixOS + KDE Plasma Wayland + PipeWire 1.4.9
- [x] Muted mic, played YouTube in browser — system audio captured and transcribed (`source: 'system'`)
- [x] Unmuted mic — echo/bleed detection correctly deduplicates mic pickup
- [x] Probe output: `supportsNativeCapture: true`, portal version 5, all capabilities detected
- [x] Strategy: `portal-helper` (not `browser-portal`)
- [x] `npm run lint` — clean
- [x] `npm run format` — clean
- [x] Verify on GNOME Wayland
- [x] Verify on non-PipeWire system (should fall back gracefully)

### DIFF
--- a/.github/workflows/build-and-notarize.yml
+++ b/.github/workflows/build-and-notarize.yml
@@ -74,12 +74,17 @@ jobs:
       - name: Install Linux build and packaging dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install -y rpm libx11-dev libxtst-dev libatspi2.0-dev libglib2.0-dev
+          sudo apt-get install -y rpm libx11-dev libxtst-dev libatspi2.0-dev libglib2.0-dev libpipewire-0.3-dev pkg-config
 
       - name: Create runtime env
         run: |
           echo "GOOGLE_CALENDAR_CLIENT_ID=${{ secrets.GOOGLE_CALENDAR_CLIENT_ID }}" >> .env
           echo "GOOGLE_CALENDAR_CLIENT_SECRET=${{ secrets.GOOGLE_CALENDAR_CLIENT_SECRET }}" >> .env
+
+      - name: Verify PipeWire system audio helper build
+        run: |
+          npm run compile:linux-system-audio
+          resources/bin/linux-system-audio-helper probe | grep '"supportsNativeCapture":true'
 
       - name: Build Application
         run: npm run build:linux -- --publish never
--- a/.github/workflows/build-and-notarize.yml
+++ b/.github/workflows/build-and-notarize.yml
@@ -84,6 +84,7 @@ jobs:
       - name: Verify PipeWire system audio helper build
         run: |
           npm run compile:linux-system-audio
+          # supportsNativeCapture reflects compile-time PipeWire linkage, so this passes on CI runners with no PipeWire daemon (where "ok" is false).
           resources/bin/linux-system-audio-helper probe | grep '"supportsNativeCapture":true'
 
       - name: Build Application
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,12 +73,17 @@ jobs:
       - name: Install Linux build and packaging dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install -y rpm libx11-dev libxtst-dev libatspi2.0-dev libglib2.0-dev
+          sudo apt-get install -y rpm libx11-dev libxtst-dev libatspi2.0-dev libglib2.0-dev libpipewire-0.3-dev pkg-config
 
       - name: Create runtime env
         run: |
           echo "GOOGLE_CALENDAR_CLIENT_ID=${{ secrets.GOOGLE_CALENDAR_CLIENT_ID }}" >> .env
           echo "GOOGLE_CALENDAR_CLIENT_SECRET=${{ secrets.GOOGLE_CALENDAR_CLIENT_SECRET }}" >> .env
+
+      - name: Verify PipeWire system audio helper build
+        run: |
+          npm run compile:linux-system-audio
+          resources/bin/linux-system-audio-helper probe | grep '"supportsNativeCapture":true'
 
       - name: Build Application
         run: npm run build:linux -- --publish always
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -83,6 +83,7 @@ jobs:
       - name: Verify PipeWire system audio helper build
         run: |
           npm run compile:linux-system-audio
+          # supportsNativeCapture reflects compile-time PipeWire linkage, so this passes on CI runners with no PipeWire daemon (where "ok" is false).
           resources/bin/linux-system-audio-helper probe | grep '"supportsNativeCapture":true'
 
       - name: Build Application
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -99,15 +99,19 @@
 
 OpenWhispr tries clipboard methods in order: `wl-copy` (most reliable) → renderer `navigator.clipboard` → X11 fallback.
 
-### Linux System Audio Portal Issues
+### Linux System Audio PipeWire Issues
 
-**Symptoms:** The Linux share dialog keeps appearing, system audio does not stay granted, or onboarding does not show a system-audio grant step.
+**Symptoms:** Meeting transcription captures the microphone but not other participants, browser audio, or other system audio.
 
 **Fix:**
-1. Update `xdg-desktop-portal` and the matching desktop backend for your session (`xdg-desktop-portal-gnome`, `xdg-desktop-portal-kde`, `xdg-desktop-portal-wlr`, etc.)
-2. Sign out and back in after updating portal packages
-3. Re-run system audio capture and complete the chooser again
-4. Expect the chooser to appear on Linux while OpenWhispr is using the standard browser portal path; some desktops may support more persistent portal behavior later, but fallback capture should still work
+1. Install PipeWire runtime libraries if they are not already present:
+   - Debian/Ubuntu: `sudo apt install pipewire libpipewire-0.3-0`
+   - Fedora/RHEL: `sudo dnf install pipewire pipewire-libs`
+   - Arch: `sudo pacman -S pipewire`
+2. Make sure the PipeWire user service is running for the current session
+3. Sign out and back in after installing or updating PipeWire packages
+4. Restart OpenWhispr and start meeting transcription again
+5. No screen-share chooser is expected for Linux system audio; OpenWhispr captures the default sink monitor directly through PipeWire
 
 ### Meeting Transcription Issues
 
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -193,7 +193,7 @@
   "deb": {
     "afterInstall": "resources/linux/after-install.sh",
     "afterRemove": "resources/linux/after-remove.sh",
-    "depends": ["ydotool"],
+    "depends": ["ydotool", "libpipewire-0.3-0"],
     "fpm": [
       "--deb-recommends",
       "ydotoold",
@@ -217,6 +217,8 @@
       "_build_id_links none",
       "--rpm-tag",
       "Requires: ydotool",
+      "--rpm-tag",
+      "Requires: pipewire-libs",
       "--rpm-tag",
       "Suggests: xdotool",
       "--rpm-tag",
--- a/main.js
+++ b/main.js
@@ -333,6 +333,17 @@ function setupProductionPath() {
 }
 
 // Phase 1: Initialize managers + IPC handlers before window content loads
+// Best-effort cleanup of the orphaned portal restore-token file older builds wrote. See PR #904.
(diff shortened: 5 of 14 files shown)


xAlcahest (commented):
Right direction here, native PipeWire is the correct fix for #901 and gating it behind `#ifdef HAVE_PIPEWIRE` to stay no-regression is the right call. Two things block it though, plus a portal design question worth settling before it lands.

**Blockers**

Doesn't compile. The closing brace of `main()` is gone, the file ends at `linux-system-audio-helper.c:804` on `return 1;`. Built it in a clean Ubuntu 24.04 box with `libpipewire-0.3-dev` and gcc stops at `expected declaration or statement at end of input`. Braces are 121 open, 120 close, and it's in `main()` outside any `#ifdef`, so it fails with or without PipeWire. Put the brace back and both branches build clean against libpipewire 1.0.5, so the new code itself is fine, the line just got dropped while preparing the commit.

Released AppImage won't get PipeWire. `release.yml` installs `libglib2.0-dev` but not `libpipewire-0.3-dev`, so `pkg-config` misses, `-DHAVE_PIPEWIRE` is dropped, and the shipped helper compiles the `#else` branch that reports `supportsNativeCapture:false` and emits `capture_unimplemented`. That's the pre-PR behavior, so #901 stays broken on the release build. You flagged this in the CI note but the change isn't in the PR. Same gap in `build-and-notarize.yml`, so PR CI never even compiles the `HAVE_PIPEWIRE` block. Fix: add `libpipewire-0.3-dev` to both workflows, plus a post-build assert like `linux-system-audio-helper probe | grep '"supportsNativeCapture":true'` so it can't regress silently.

**Design: capture bypasses the portal**

The CreateSession/SelectSources/Start dance exists to get the node id from the Start `streams` array and an fd from `OpenPipeWireRemote`, then `pw_context_connect_fd`. The code reads only `restore_token` from the Start results (`:590`) and never touches `streams`. `OpenPipeWireRemote` isn't called anywhere. `run_pipewire_capture` opens its own connection with `PW_ID_ANY` + `AUTOCONNECT` and `PW_KEY_STREAM_CAPTURE_SINK=true`, so it grabs the default sink monitor through the global socket regardless of what the portal handed back. The portal selection is ignored, the `restore_token` round-trip is cosmetic, and under a Flatpak/snap sandbox there's no direct socket so AUTOCONNECT gets nothing. On a normal desktop you get the same result with no portal at all, which is why it works on your KDE box. Related: `SelectSources` never asks for audio (ScreenCast is video, `types` is the display MONITOR, and `supports_system_audio` is derived from that video bitmask at `:276`). Either wire capture to the portal node (`OpenPipeWireRemote` + node id, drop AUTOCONNECT) or drop the portal and call it direct loopback capture, but the current middle ground shows a permission dialog for a capture that ignores it.

**Smaller stuff**

- No `param_changed` handler. S16LE/24k/mono is offered as `EnumFormat`, a preference, and `on_pw_process` writes raw bytes assuming that's what came back. If the server negotiates F32/48k/stereo you still emit `start` and stream garbage with no error. Parse `SPA_PARAM_Format` and reject or convert.
- `on_pw_process` does a blocking `fwrite` on the RT data thread with stdout `_IONBF` (`:358`, `:323`). If the Node reader stalls and the ~64KB pipe fills, `write()` blocks the RT thread and you get XRUNs. Ring-buffer it and drain off-thread. The `fwrite` return is also ignored, so EPIPE goes unnoticed.
- `emit_event` writes the JSON line as several `fputs` with no lock, called from both the pw thread (`on_pw_state_changed`) and the main thread (`on_start_response` emitting `start`). If the stream errors during startup the lines interleave and the manager never sees a clean `{type:"start"}`, then it times out at 15s. Mutex it, or build the line and write it once.
- `props` leaks if `pw_stream_new_simple` returns NULL (`:381`), minor on a one-shot process.
- `got_signal` is written twice and never read. Dead.
- With PipeWire the helper links `libpipewire-0.3.so.0` directly and nothing bundles the runtime. Fine on a PipeWire desktop, just an undocumented runtime dep.

Checked and clean: `pw_deinit` with no matching `pw_init` is a safe no-op, and `cleanup_pipewire`'s lock/destroy/unlock then stop order is correct, no deadlock or use-after-free.

Worth a GNOME Wayland pass before merge too, that's the #901 environment and it's still unchecked.

0x4bs3nt:
A few notes on the bigger refator:

Default sink targeting implemented by resolving `default.audio.sink` from PipeWire's "default" metadata object and passing it as `PW_KEY_TARGET_OBJECT` on the stream props, with a 500ms timeout fallback to `AUTOCONNECT`. I felt this was the best way to target with a nice fallback.

Additionally, manually tested with quickemu on Gnome + Wayland + Pipewire, as the original issue was, and everything seems to be working normally. Adding screenshots of environment and pipewire checking. Manually tested transcription of youtube video and all works.

env

linux-system-audio

xAlcahest (commented):
Re-reviewed after the rewrite, big step up. Both blockers are gone: it compiles in a clean Ubuntu 24.04 box with `libpipewire-0.3-dev`, both branches, and CI now asserts `supportsNativeCapture:true`. Dropping the portal for a direct sink-monitor loopback is the right call, and resolving `default.audio.sink` to set `PW_KEY_TARGET_OBJECT` instead of bare `PW_ID_ANY` is a nice touch. The `param_changed` format validation, the non-blocking pipe drain off the RT thread, and the `flockfile` around `emit_event` all close the earlier points cleanly. Manager, CI verify step and the deb/rpm/nix runtime deps line up too.

Thanks for the GNOME Wayland + PipeWire run, that covers the #901 environment. The screenshots confirm the environment and a clean PipeWire build there; the transcription itself is on your word so far. A quick grab of the transcript or a `source: system` log line would fully nail it down, not a blocker though.

Two leftovers, neither blocking:

- `finish_start` is `defined but not used` in the non-PipeWire `#else` build (gcc -Wall). Move it inside the `#ifdef`.
- `enqueue_audio_bytes` drops frames silently under backpressure. Right call for realtime, but a debug counter or a one-time log would help diagnose dropouts.

Nice work turning this around.

0x4bs3nt:
Thanks for the quick re-review, still quite new to the project and as I said, C is not my strongest language haha.

Will grab a video or logs from the GNOME Wayland + Pipewire environment and post here, as well as fix the two leftovers today.

0x4bs3nt:
@xAlcahest Fixed the issues, also recorded a video of the transcription working alongside the logs on the original GNOME Wayland + Pipewire environment.

[Screencast_20260611_004126.webm](https://github.com/user-attachments/assets/5f8608b9-3991-4d8e-b547-74a478ae8487)

Let me know if anything else needs fixing on this PR.

xAlcahest (commented):
Went through the new commit and the screencast. Both leftovers are handled: `finish_start` is now inside the `#ifdef`, and dropped frames are tracked with an atomic counter plus a one-time `audio_frames_dropped` warning. Clean.

The GNOME Wayland + PipeWire recording covers the end-to-end path: `systemAudioMode: 'loopback'`, capture starts cleanly, and the YouTube audio gets transcribed live into the note. The `Skipping duplicate local mic segment that matches system audio` lines confirm the system stream is genuinely captured and deduped against the mic, not just mic pickup. FFmpeg conversion and diarization fire at the end, no PipeWire errors or drop warnings in the run.

Code and demo both look good to me. Nice work turning this around.

xAlcahest (approved):
LGTM. Both blockers and the two leftovers are resolved, and the GNOME Wayland + PipeWire screencast shows system-audio capture and transcription working end to end. Thanks for the iterations.

PR #909, opened 2026-06-07
Add preload and idle-timeout settings for the local cleanup model

Stasieniec:
## Summary

Adds two settings for the local cleanup LLM so users can trade a little VRAM for consistently fast cleanup:

- Preload model on startup: loads the local cleanup model when the app opens, so the first dictation does not pay the cold start.
- Unload when idle: makes the existing idle-unload timeout configurable (5, 15, 30, 60 minutes, or Never). Never keeps the model loaded while the app is open.

Both default to current behavior (preload off, 5 minute idle), so existing users are unaffected.

## Motivation

The cleanup model is unloaded after a hardcoded 5 minutes idle (IDLE_TIMEOUT_MS in src/helpers/llamaServer.js). On GPU backends the next dictation then pays a cold start (model load, and on Vulkan, shader compilation), which can be slow. These settings let people who keep the app open avoid that.

## Changes

- src/helpers/idleTimeout.js: resolver for the idle timeout from CLEANUP_IDLE_TIMEOUT_MS (default 5 minutes, 0 means never), with a unit test.
- src/helpers/llamaServer.js: resetIdleTimer reads the resolver and does not arm the timer when the value is 0.
- src/helpers/modelManagerBridge.js: preloadModel starts the configured local cleanup model in the background (no-op if the provider is not local or the model is not downloaded).
- main.js: calls preloadModel on startup when CLEANUP_PRELOAD_ON_START is set.
- Settings plumbing: two new CleanupSettings fields, synced to the main process through the existing syncStartupPreferences IPC.
- src/components/SettingsPage.tsx: a preload toggle and an unload-when-idle select in the cleanup section, shown for local cleanup models.
- i18n: new keys added to all 10 locales.

## Test plan

- node --test test/helpers/idleTimeout.test.js
- node scripts/check-i18n.js
- Manual: set Unload when idle to Never and confirm the model stays loaded past 5 minutes; enable Preload on startup and confirm the server starts at launch; set a short idle and confirm the model unloads after it.

### DIFF
--- a/main.js
+++ b/main.js
@@ -420,6 +420,18 @@ function initializeDeferredManagers() {
       "clipboard"
     );
   });
+
+  if (process.env.CLEANUP_PRELOAD_ON_START === "true") {
+    const modelManager = require("./src/helpers/modelManagerBridge").default;
+    modelManager.preloadModel().catch((err) => {
+      require("./src/helpers/debugLogger").warn(
+        "cleanup model preload error",
+        { error: err?.message },
+        "model"
+      );
+    });
+  }
+
   clipboardManager.preWarmAccessibility();
   trayManager = new TrayManager();
   globeKeyManager = new GlobeKeyManager();
--- a/src/components/SettingsPage.tsx
+++ b/src/components/SettingsPage.tsx
@@ -418,6 +418,11 @@ function NoteFormattingSettings() {
 
 function AiModelsSection({ useCleanupModel, setUseCleanupModel, toast }: AiModelsSectionProps) {
   const { t } = useTranslation();
+  const cleanupProvider = useSettingsStore((s) => s.cleanupProvider);
+  const cleanupPreloadOnStart = useSettingsStore((s) => s.cleanupPreloadOnStart);
+  const cleanupIdleTimeoutMinutes = useSettingsStore((s) => s.cleanupIdleTimeoutMinutes);
+  const setCleanupPreloadOnStart = useSettingsStore((s) => s.setCleanupPreloadOnStart);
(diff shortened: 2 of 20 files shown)


hastarin:
I've just spent longer than I'd care to admit trying to debug this behaviour. Hopefully the reviewers can get a chance to merge this in amongst the many pull requests.

navarro165:
Still hardcoded in shipped v1.9.0, and there is no user-side workaround, so this is still the fix.

Verified against the 1.9.0 app bundle on macOS:

- `IDLE_TIMEOUT_MS = 5 * 60 * 1000` is still a module constant in `LlamaServerManager`.
- The only environment reads in that class are `INTELLIGENCE_GPU_UUID`, `LLAMA_ARG_FIT` and `LLAMA_CPP_VERSION`; nothing touches the idle window.
- No persisted setting matches idle or timeout, and the exposed IPC surface is only `llamaServerStart` / `llamaServerStop` / `llamaServerStatus`.

Not yet covered in this thread: an external keep-alive cannot work around it. `resetIdleTimer()` has two call sites, `start()` and the `.finally()` on the manager's own request method. The timer is a `setTimeout` living in the Electron main process, so requests sent straight to the llama-server port never reset it and the server is torn down anyway. Today the only escape from the 5-minute unload is to leave `local` mode entirely and run your own llama-server behind `self-hosted`, which gives up the app's model management.

Housekeeping: the branch is currently `CONFLICTING` against `main` and needs a rebase, and no reviews have been submitted yet.

macOS data point, since the related reports are mostly Windows: a warm cleanup on Qwen3.5-4B Q4_K_M returns in 722 ms end to end for a one-sentence transcript. The cold path re-reads a 3.0 GB file to get back to that.

PR #911, opened 2026-06-08
fix(macos): holding Fn/Globe no longer triggers dictation in tap mode

### LINKED ISSUE #865: Can't use fn button for existing system shortcuts with default OpenWhispr shortcut ?
**Describe the bug**
Using the standard shortcut key on macOS:

Image

When I turn my volume up and down OpenWhispr activates: **fn + F12**. 
(note: I didn't tap **fn**, I tap **fn + F12**)

**To Reproduce**
Tap **fn** + any function key

**Expected behavior**
OpenWhispr should not activate tapping **fn + F12**

 - OS: macOS 26.4.1 (25E253)
 - Browser Safari
 - Version 26.4

**Additional context**
I can see the logic in reacting to the **Fn** press, but especially with activation mode set to tap I don't associate **Fn + F12** as also being an Fn press. I don't think that's a crazy assumption? But in either event it's very confusing to find volume controls also activating dictionary mode.

FWIW Wispr Flow defaults to a shortcut of **Fn + Shift**, which gets around this issue. But such a shortcut does not appear possible in OpenWhispr.

renato-rg:
## Summary

- Issue: on macOS with activation mode set to `tap`, pressing and holding the Fn/Globe key (e.g. `Fn+F12` for volume) started dictation.
- Root cause: in `tap` mode, every `"globe-down"` toggled dictation immediately.
- Fix: in `tap` mode, dictation now toggles on `"globe-up"`, and only if the press was held for at most `MIN_HOLD_DURATION_MS`. A longer hold (Fn used as a modifier) is ignored.

Fixes #865.

## Changes

Minimal single-file change in `main.js` (7 lines).

## Tested

Built locally and tested on macOS Tahoe 26.3.1.

### DIFF
--- a/main.js
+++ b/main.js
@@ -1010,7 +1010,7 @@ async function startApp() {
               }
             }, MIN_HOLD_DURATION_MS);
           } else {
-            windowManager.sendToggleDictation();
+            globeKeyDownTime = Date.now();
           }
         } else {
           debugLogger?.debug("[Globe] Ignored — mainWindow not live");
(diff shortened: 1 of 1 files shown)


gabrielste1n:
Closing in favor of #1046, which also moves tap-mode Fn activation to key-up and additionally tracks Fn chords (Fn+Delete, Fn+F12, etc.), covering the #865 case. Thanks for the minimal repro fix.

PR #919, opened 2026-06-08
fix: dictation hotkey silently dying in hyprland on config reload.

### LINKED ISSUE #914: Hyprland global hotkey silently dies after any `hyprctl reload` (runtime `keyword bind` is never re-applied)
**Describe the bug**

On Hyprland, the global dictation hotkey works right after launch, then stops working after a while. Re-recording the hotkey in Settings, or restarting the app, brings it back. The app process and its D-Bus service keep running the whole time — only the Hyprland keybinding is gone.

Root cause: OpenWhispr registers its Hyprland shortcut as a **runtime** keybinding via `hyprctl keyword bind`, which Hyprland discards on any config reload. The app registers it **once at startup** and never re-applies it, so the first `hyprctl reload` after launch permanently kills the hotkey until a manual re-registration. (Full analysis under *Additional context*.)

**To Reproduce**

1. Launch OpenWhispr on Hyprland; confirm the dictation hotkey works.
2. `hyprctl binds` → the `exec, dbus-send ... com.openwhispr.App.Toggle` bind is present.
3. Run `hyprctl reload` (or trigger any config auto-reload, e.g. a theme/wallpaper change that rewrites a `source`d config file).
4. `hyprctl binds` → the OpenWhispr bind is **gone**. The hotkey no longer triggers dictation, although the app and its D-Bus service are still running.

I verified step 3→4 directly: a runtime `hyprctl keyword bind` is removed by `hyprctl reload` on this system. I also observed the live state where, after the app had been running for some time, no OpenWhispr bind was present in `hyprctl binds` while the `com.openwhispr.App` D-Bus service (with the `Toggle()` method) was still registered.

**Expected behavior**

The hotkey should keep working across Hyprland config reloads for as long as the app runs.

- OS: Fedora 43
- Desktop / compositor: Hyprland (Wayland session) — `XDG_SESSION_TYPE=wayland`, `HYPRLAND_INSTANCE_SIGNATURE` set, `XDG_CURRENT_DESKTOP=Hyprland`
- App: OpenWhispr 1.7.2 (Electron), RPM build in `/opt/OpenWhispr`
- GPU: NVIDIA
- Browser: N/A (Electron app)
- Hotkey in use: `Ctrl+Space` (also applies to the default `Control+Super`)
- Note: the app is launched with `--ozone-platform=x11`, but this does **not** change the path taken — hotkey backend selection is based on `XDG_SESSION_TYPE`/`HYPRLAND_INSTANCE_SIGNATURE`, so the Hyprland native path is always chosen here.

**Additional context**

*What actually happens (verified against the packaged 1.7.2 source):*

1. `HotkeyManager.initializeHotkey` selects the Hyprland backend and, after a 1 s delay, calls `registerHyprlandHotkey` **exactly once** (`setTimeout(registerHyprlandHotkey, HOTKEY_REGISTRATION_DELAY_MS)` in `src/helpers/hotkeyManager.js`).
2. `HyprlandShortcutManager.registerKeybinding` (`src/helpers/hyprlandShortcut.js`) runs:
   `hyprctl keyword bind "<MODS>, <key>, exec, dbus-send --session --type=method_call --dest=com.openwhispr.App /com/openwhispr/App com.openwhispr.App.Toggle"`
   Pressing the hotkey makes Hyprland run that `dbus-send`, which calls the app's `Toggle()` method.
3. There is **no listener on Hyprland's `configreloaded` event** and no periodic health-check anywhere in the main process (no reference to `configreloaded` or `.socket2` in the non-`node_modules` code). The only code paths that call `registerKeybinding` again are the **manual** IPC handlers `update-hotkey` and `set-hotkey-listening-mode` — i.e. the Settings UI. That is why re-recording the hotkey "fixes" it.

*Why the hotkey dies:*

`hyprctl keyword ...` changes are runtime-only; Hyprland drops them on `hyprctl reload`. Since `misc:disable_autoreload` defaults to `0`, Hyprland also **auto-reloads** whenever the config (or a `source`d file) changes on disk. After any such reload, the app's runtime bind is gone and is never re-created — matching the "works, then stops; re-register/restart fixes it" symptom exactly.

Both halves of this are directly observed on my machine via the Hyprland event socket (`.socket2.sock`):
- A real in-place content write to a `source`d config file (`source = hyprland/colors.conf` in `hyprland.conf`, regenerated by matugen-based theming) makes Hyprland emit `configreloaded`. A pure `mtime` touch does **not** — Hyprland reacts to content modification, not timestamp changes.
- A runtime `hyprctl keyword bind` is present in `hyprctl binds` before a reload and absent after.

*Suggested fix:*

Re-apply the keybinding after reloads. Options:
- Subscribe to Hyprland's event socket (`$XDG_RUNTIME_DIR/hypr/$HYPRLAND_INSTANCE_SIGNATURE/.socket2.sock`) and re-run `registerKeybinding` on `configreloaded>>`.
- And/or use the XDG Desktop Portal GlobalShortcuts interface where available, which persists across reloads.
- At minimum, document that on Hyprland the bind must be placed in the user's config to survive reloads.

*Working workaround (for other Hyprland users):*

Put the bind in your **own** Hyprland config (it is re-applied on every reload, so it survives). Example for `Ctrl+Space`:

```
bind = Ctrl, Space, exec, dbus-send --session --type=method_call --dest=com.openwhispr.App /com/openwhispr/App com.openwhispr.App.Toggle
```

(Requires OpenWhispr to be running, since the bind talks to its session D-Bus service.)

*What I did NOT directly observe (stated honestly):*

- I confirmed the causal chain by component, but not as a single uninterrupted timeline of the app's *own* bind dying: I observed (a) a content write to the `source`d `colors.conf` makes Hyprland emit `configreloaded`, (b) a runtime `hyprctl keyword bind` disappears across a reload, and (c) after the app had run for a while its bind was absent while its D-Bus service was still up. What I did **not** capture is matugen (or anything else) firing on its own, during idle, without manual action — so the link between "inactivity" specifically and a reload is inferred. The mechanism does not depend on idle: any config reload (theme/wallpaper change, a config edit, a manual `hyprctl reload`, etc.) is sufficient. On this setup hypridle only does `dpms off/on` (no reload), so idle itself is not the cause.
- Line numbers/symbol names are from the packaged 1.7.2 build (`app.asar`); they may differ in the repository source.

*Possibly related (not independently verified):* #869 (Linux/COSMIC, cross-app) — different environment and likely a different root cause, but also a Linux global-hotkey failure.

IdrisGit:
persist dictation hotkey to the hyprland config file to persist it across reloads and fallback to the keyword bind method if config doesn't exist.

fixes #914

### DIFF
--- a/preload.js
+++ b/preload.js
@@ -291,6 +291,7 @@ contextBridge.exposeInMainWorld("electronAPI", {
   setHotkeyListeningMode: (enabled, newHotkey) =>
     ipcRenderer.invoke("set-hotkey-listening-mode", enabled, newHotkey),
   getHotkeyModeInfo: () => ipcRenderer.invoke("get-hotkey-mode-info"),
+  getHyprlandConfigStatus: () => ipcRenderer.invoke("get-hyprland-config-status"),
   startWindowDrag: () => ipcRenderer.invoke("start-window-drag"),
   stopWindowDrag: () => ipcRenderer.invoke("stop-window-drag"),
   setMainWindowInteractivity: (interactive) =>
--- a/src/components/OnboardingFlow.tsx
+++ b/src/components/OnboardingFlow.tsx
@@ -34,6 +34,7 @@ import { formatHotkeyLabel, getDefaultHotkey, isGlobeLikeHotkey } from "../utils
 import { useAuth } from "../hooks/useAuth";
 import { HotkeyInput } from "./ui/HotkeyInput";
 import { useHotkeyRegistration } from "../hooks/useHotkeyRegistration";
+import { useHotkeyModeInfo } from "../hooks/useHotkeyModeInfo";
 import { getValidationMessage } from "../utils/hotkeyValidator";
 import { validateHotkeyForSlot } from "../utils/hotkeyValidation";
 import { getCachedPlatform, getPlatform } from "../utils/platform";
@@ -105,12 +106,8 @@ export default function OnboardingFlow({ onComplete }: OnboardingFlowProps) {
   const [skipAuth, setSkipAuth] = useState(false);
   const [pendingVerificationEmail, setPendingVerificationEmail] = useState<string | null>(null);
   const [isModelDownloaded, setIsModelDownloaded] = useState(false);
-  const [isUsingNativeShortcut, setIsUsingNativeShortcut] = useState(false);
-  const [isUsingHyprland, setIsUsingHyprland] = useState(false);
-  const [hyprlandConfigStatus, setHyprlandConfigStatus] = useState<{
-    canWrite: boolean;
-    path: string;
-  } | null>(null);
+  const { isUsingNativeShortcut, isUsingHyprland, hyprlandConfigStatus, supportsPushToTalk } =
+    useHotkeyModeInfo("onboarding");
   const readableHotkey = formatHotkeyLabel(hotkey);
   const { alertDialog, confirmDialog, showAlertDialog, hideAlertDialog, hideConfirmDialog } =
     useDialogs();
@@ -181,28 +178,10 @@ export default function OnboardingFlow({ onComplete }: OnboardingFlowProps) {
   const showProgress = currentStep > 0;
 
   useEffect(() => {
-    const checkHotkeyMode = async () => {
-      try {
-        const info = await window.electronAPI?.getHotkeyModeInfo();
-        if (info?.isUsingNativeShortcut) {
-          setIsUsingNativeShortcut(true);
-          if (!info.supportsPushToTalk) {
-            setActivationMode("tap");
-          }
-        }
-        if (info?.isUsingHyprland) {
-          setIsUsingHyprland(true);
-          const configStatus = await window.electronAPI?.getHyprlandConfigStatus?.();
-          if (configStatus) {
-            setHyprlandConfigStatus(configStatus);
-          }
-        }
-      } catch (error) {
-        logger.error("Failed to check hotkey mode", { error }, "onboarding");
-      }
-    };
-    checkHotkeyMode();
-  }, [setActivationMode]);
+    if (isUsingNativeShortcut && !supportsPushToTalk) {
+      setActivationMode("tap");
+    }
+  }, [isUsingNativeShortcut, supportsPushToTalk, setActivationMode]);
 
   // Update wizard UI when backend falls back to a different hotkey.
   // Only update local state — don't persist to localStorage so the app
--- a/src/components/OnboardingFlow.tsx
+++ b/src/components/OnboardingFlow.tsx
@@ -100,6 +100,7 @@ export default function OnboardingFlow({ onComplete }: OnboardingFlowProps) {
   const [pendingVerificationEmail, setPendingVerificationEmail] = useState<string | null>(null);
   const [isModelDownloaded, setIsModelDownloaded] = useState(false);
   const [isUsingNativeShortcut, setIsUsingNativeShortcut] = useState(false);
+  const [isUsingHyprland, setIsUsingHyprland] = useState(false);
   const readableHotkey = formatHotkeyLabel(hotkey);
   const { alertDialog, confirmDialog, showAlertDialog, hideAlertDialog, hideConfirmDialog } =
     useDialogs();
@@ -170,6 +171,9 @@ export default function OnboardingFlow({ onComplete }: OnboardingFlowProps) {
           if (!info.supportsPushToTalk) {
             setActivationMode("tap");
           }
+        }
+        if (info?.isUsingHyprland) {
+          setIsUsingHyprland(true);
         }
       } catch (error) {
         logger.error("Failed to check hotkey mode", { error }, "onboarding");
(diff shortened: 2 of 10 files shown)


Grisemine:
Tested this branch on a real Hyprland session and it fixes the issue. Setup: Fedora 43, Hyprland 0.51.1, OpenWhispr 1.7.2, hotkey `Ctrl+Space`, on an illogical-impulse/quickshell config with matugen theming. That theming is the original trigger here: a wallpaper change rewrites a `source`d `colors.conf`, Hyprland auto-reloads, and the runtime bind used to get wiped.

Results:
- On startup the app writes `~/.config/hypr/openwhispr-binds.conf` and adds `source = ./openwhispr-binds.conf` to `hyprland.conf`.
- The bind survives `hyprctl reload`. I checked it both manually and across a real wallpaper/theme change (the exact condition that used to break it), and the hotkey keeps working.
- `unregisterKeybinding` cleans the bind back out of the file.
- The relative `source = ./openwhispr-binds.conf` resolves correctly on Hyprland 0.51.1.

One minor nit: in `_writeBindToConfig`, the line filter keeps comment lines (`startsWith("#")`), so the previous `# OpenWhispr keybinds (managed automatically)` header stays in the body and a fresh header gets prepended on top. The header line ends up duplicated on every write (roughly once per app launch or re-register). It's harmless, but the file slowly accumulates duplicate comment lines; stripping the managed header before rewriting would avoid it.

Thanks for the quick fix!

IdrisGit:
@Grisemine thanks for testing, I have pushed the header duplication fix

xAlcahest (commented):
Nice fix, and good that it got tested on the exact setup from #914. A few things I'd flag before merge, mostly around the config writes.

The persistence calls run inside `registerKeybinding()`'s try block, before the `hyprctl keyword bind`. If `_ensureSourceInMainConfig()` throws (hyprland.conf as a read-only symlink: NixOS home-manager, chezmoi and friends), the catch returns false and the runtime bind never gets registered. On main those users had a hotkey that worked until the first reload; with this PR it never works at all. I'd register the runtime bind first and make the file writes best-effort in their own try/catch that only logs. That also covers the unwritable-config case, which the new alert doesn't catch (it only checks for a missing file).

There's also a stale-bind path after a crash: the bind stays in `openwhispr-binds.conf`, Hyprland loads it from config on the next reload, and when the app restarts `currentBinding` is null, so it skips the unregister and stacks a runtime bind on top of the config one. Hyprland fires every bind on the same combo, so each press would run the dbus-send twice and Toggle would start and immediately stop the recording. An unconditional `hyprctl keyword unbind` before binding would cover it.

The two new translation keys are only in `en/translation.json`; new keys need to land in all 9 locale files.

Smaller things, take or leave: `unregisterKeybinding` cleans the bind line but the `source =` line stays in `hyprland.conf` forever, so if the user uninstalls and deletes the binds file, Hyprland shows the config-error banner on every reload. The `hyprland.conf` write isn't atomic (read + `writeFileSync`), so a crash mid-write can truncate the user's main config; temp file + rename would make it safe. And the alert text hardcodes `~/.config/hypr/hyprland.conf` even when `HYPRLAND_CONFIG` or `XDG_CONFIG_HOME` point elsewhere.

For what it's worth, the other option from #914 (listening for `configreloaded` on Hyprland's socket2 and re-applying the runtime bind) would avoid touching user config files entirely. The file approach works too, as long as the writes are best-effort.

IdrisGit:
@xAlcahest thanks for the review, I will get on it on Sunday and update the PR.

IdrisGit:
@xAlcahest I have pushed some changes based on your review.

1. `hyprctl keyword bind` now runs before persistence writes, and `_writeBindToConfig()` / `_ensureSourceInMainConfig()` are best-effort with logging only. This makes sure the new hotkey is always added atleast to the hyprland runtime keybindings (same as the old implementation), which solves the hyprland.conf as a read-only symlink: NixOS home-manager, chezmoi and friends; basically the unwritable-config case.

2. Registration now runs `hyprctl keyword unbind <combo>` before binding, preventing stacked binds for the same shortcut after config reload/crash.This addresses the stale duplicate case on app crash.

3. The app no longer empties/removes `openwhispr-binds.conf`; it keeps a managed file with guidance comments. This prevents normal reload errors, but if users manually delete the file, the source = ./openwhispr-binds.conf line still remains, which is expected because user manually deleted a conf file and we even log this now in the `openwhispr-binds.conf`. I am intentionally not touching user's hyprland config file, as it requires parsing and removing stuff from user's main config file, I would assume a hyprland user can read the banner alert and fix it.

4. Instead of rewriting `hyprland.conf`, the code now appends the source line using `fs.appendFileSync`. This avoids whole-file truncation risk and is safer for symlink-managed configs.

5. The app now exposes `getHyprlandConfigStatus()` with `{ path, canWrite }`, and UI warns with a general error message that OpenWhispr can't write to the config file and hotkey will not survive any config reload.

6. The UI warning now uses the resolved config path, so it respects HYPRLAND_CONFIG, XDG_CONFIG_HOME, and the default fallback path.

7. I have intentionally not added translation keys in other languages as I don't know them, if you are fine with me using any AI/Online/Google tool for that let me know, I will do it.

can you please do another round of review, I have also enabled edits by maintainers, feel free to make any edits if you want.

gabrielste1n:
Thanks so much for this, @IdrisGit — really appreciate you taking this on and putting together such a clean fix. 🙏 And thank you @Grisemine for the excellent bug report; your diagnosis (runtime `keyword bind` never getting re-applied after `hyprctl reload`) made the root cause obvious and pointed straight at the fix. Thanks too, @xAlcahest, for taking the time to review. 🙌

I pushed a few small follow-ups to the branch:

- **Lint cleanup** — switched to a bare `catch` in `getHyprlandConfigStatus` since the error binding wasn't used.
- **Merged latest `main`** to bring the branch up to date (no conflicts).
- **Small dedupe** — pulled the shared `getHotkeyModeInfo` / `getHyprlandConfigStatus` logic out of `SettingsPage` and `OnboardingFlow` into a single `useHotkeyModeInfo` hook, since both were doing the same thing.

Thanks again, everyone — merging now! 🚀

gabrielste1n (approved):

IdrisGit:
@gabrielste1n Happy to contribute, I have been enjoying using OpenWhispr, will try to contribute more as well.

PR #944, opened 2026-06-13
feat: Add Support for Tinfoil model provider

danielmccannsayles:
## Add support for [Tinfoil](https://tinfoil.sh/), a verifiably private AI company. 

Screenshot 2026-06-12 at 9 28 54 PM

At tinfoil, we run models inside secure enclaves (backed by NVIDIA Confidential Computing), such that no one can see the data, not even us. All our code is [open source](https://docs.tinfoil.sh/verification/attestation-architecture) and publicly committed to a transparency log. The client SDK (in this case, [tinfoil-js](https://github.com/tinfoilsh/tinfoil-js)) fetches the enclave measurements from the transparency log and checks that it matches the runtime attestation.

This PR adds Tinfoil as a first class model provider, along with support for our streaming STT model (`voxtral-mini-4b-realtime`). Bring your own tinfoil inference API key.

_Future: Tinfoil also offers voxtral small (non-realtime) as well as open source LLMs, with ofc. the same privacy guarantees. Wiring these up to openwhispr would be awesome!_

###  Make the realtime endpoint configurable
The realtime streaming transcription provider is hardcoded to `wss://api.openai.com/v1/realtime`.
This makes it configurable

### Support Tinfoil
We offer a streaming STT model (`voxtral-mini-4b-realtime`) behind an OpenAI Realtime-compatible WebSocket. See [here](https://github.com/tinfoilsh/confidential-realtime-models) or our [docs](https://docs.tinfoil.sh/models/audio)

### Sampling format (pre-existing bug)
There's a small bug in the current openWhisper code: the client declares `format: { type: "audio/pcm", rate: 24000 }` in `session.update`. This is probably bc. 24khz is the only PCM rate OpenAI's Realtime API documents. However, the capture pipeline runs at 16khz. This should be changed on the client side to either capture at 24k or upsample to 24k. 

###  QOL
I want previews of the transcription. This was only on for local models, so I had to add the ability to turn this on or off for speciifc cloud models (right now only tinfoil, openAi could be added too, left as an excercise to the reader)

_Tested locally in dev mode, works using tinfoil & an openai key_


danielmccannsayles:
@gabrielste1n Hi Gabe, I work at Tinfoil. I think OpenWhispr is cool, and have started using it locally with our tinfoil models. (At Tinfoil we run open source models in secure enclaves so we can't see any of our clients data) 

It would be awesome if everyone could use private tinfoil models through OpenWhispr, easily! This PR is just the first of many possible steps, adding realtime STT (my personal favorite use case), but we also offer batch STT, LLMs, and more, all private ofc.

Hope this PR is good, tried to keep it as minimal as possible. Lmk!

danielmccannsayles:
@xAlcahest  Hi! Sorry to bother you all again, but I'm really excited about this. Please let me know if there's any way I can help get this PR merged!

gabrielste1n:
@danielmccannsayles sorry only seeing this now! Will work on getting this released in v1.7.4 :) Thanks so much for the contribution!

gabrielste1n (approved):
awesome work thank you sir

PR #952, opened 2026-06-15
Quick duck patch for windows 

sriharshaguthikonda:
## Summary

Adds configurable audio ducking for dictation, so background/system audio can be reduced while recording instead of being fully paused.

This reuses the existing `pauseMediaOnDictation` path and switches behaviour to ducking when configured via environment variables:

```env
OPENWHISPR_MEDIA_ON_DICTATION=duck
OPENWHISPR_AUDIO_DUCKING_VOLUME=25
```

## What changed

- Adds a duck/restore flow that saves the current system volume when dictation starts, lowers it to a configured percentage, then restores the original volume after dictation ends.
- Adds Windows system volume control using Core Audio through PowerShell/C#.
- Adds macOS support through `osascript` volume control.
- Adds Linux support through `pactl`, with `wpctl` fallback.
- Adds defensive logging for Windows volume read/set failures.
- Documents the new environment variables in `.env.example`.

## Testing

Tested manually on Windows with an installed OpenWhispr build:

- Set `OPENWHISPR_MEDIA_ON_DICTATION=duck`
- Set `OPENWHISPR_AUDIO_DUCKING_VOLUME=25`
- Enabled **Settings → General → Sound Effects → Pause media during dictation**
- Started dictation while speaker audio was playing
- Confirmed the app attempted ducking from the recording path
- After improving the Windows Core Audio helper, audio ducking worked as expected locally

## Notes

This is intentionally implemented as an environment-configurable MVP to keep the change small. A future UI pass could replace the current boolean setting with something like:

- Do nothing
- Pause media
- Duck audio

plus a duck-volume slider.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -20,6 +20,12 @@ LANGUAGE=
 # Optional: UI language for app interface and menus (en, es, fr, de, pt, it)
 UI_LANGUAGE=en
 
+# Optional: Duck system audio during dictation instead of pausing media.
+# Requires Settings > General > Sound Effects > Pause media during dictation to be enabled.
+# Volume is a percentage from 0 to 100 and is restored after recording stops.
+OPENWHISPR_MEDIA_ON_DICTATION=
+OPENWHISPR_AUDIO_DUCKING_VOLUME=25
+
 # Anthropic API Configuration (optional - for Claude AI processing)
 # Get your API key from: https://console.anthropic.com/api-keys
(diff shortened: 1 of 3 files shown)


gabrielste1n:
Closing — audio ducking configured only via env vars (plus the unrelated workflow change) isn't a shape we can ship; the settings-based approach in #778 and the meeting echo-cancellation toggle in #1135 cover this space. A ducking option layered onto those with a proper settings UI would be welcome.

PR #954, opened 2026-06-15
fix(mac): restrict Fn-key combinations to hardware-gated keys (#906)

### LINKED ISSUE #906: Bug: Fn/Globe Key Combos behave inconsistently on MacOS
**Describe the bug**
On macOS, the hotkey picker lets me set a Globe/Fn key combination like `Fn+A` or `` Fn+` ``, and it shows up correctly in settings and the widget. But these combos behave inconsistently depending on the base key:

- With a **letter, number, or punctuation key** (e.g. `Fn+A`, `` Fn+` ``, `Fn+-`), the control is effectively reversed: pressing the **base key alone triggers dictation**, while the actual combination `Fn+<base_key>` just **types that base key normally**. So it fires accidentally while typing, and the combo I set does the opposite of what I'd expect.
- With an **arrow / function / navigation key** (e.g. `Fn+Right`, `Fn+F5`), it works as expected: I have to hold Fn for it to trigger.

Same UI, two different behaviors — and the character-key combos are unusable because they fire during normal typing.

**To Reproduce**

1. Open Settings → hotkey input.
2. Hold the Globe/Fn key and press `A` (or `` ` ``, `-`, `=`). It saves as `Fn+A` and displays correctly.
3. Close settings and type normally — press just the letter `a`, without Fn.
4. Dictation starts, even though Fn was never held. Meanwhile pressing `Fn+A` simply types `a`.

(For contrast: set `Fn+Right` instead — that one only triggers when Fn is actually held, as expected.)

**Expected behavior**

A Globe/Fn combo should only trigger dictation when Fn is actually held together with the base key, and should never fire on the base key alone. If a combination can't be reliably Fn-gated, the UI shouldn't allow it to be set.

**Screenshots**

N/A

- OS: macOS
- App: OpenWhispr (Electron desktop app — no browser)
- Version: 1.7.2

- N/A — desktop app

**Additional context**

Root cause: macOS/Electron's global shortcut API has no Fn modifier, so the `Fn+` prefix is stripped and only the base key is registered as a plain global shortcut. For letters/digits/punctuation the OS emits the same key event with or without Fn, so the bare key triggers dictation (and the Fn+key combo falls through to normal typing). For arrow/function/nav keys the OS itself requires Fn, so those happen to work.

I'm happy to help on the PR.

lakshyamehta03:
## Summary
Fixes #906 

Setting a hotkey on macOS using the `Fn` key plus a letter or number (e.g., `Fn+A`) causes a "reversed typing" bug. Because Electron's `globalShortcut` API does not natively support the `Fn` modifier on macOS, the app strips the prefix and registers the base key (`A`) globally. This steals the key from other applications, forcing users to hold `Fn` just to type normally.

This PR introduces a validation-first approach. Instead of requiring high-friction "Input Monitoring" permissions to intercept keys via `CGEventTap`, we block unsupported combinations at the definition level. `Fn` combinations are now strictly restricted to hardware/OS-gated keys (Arrows, Function keys, and Navigation keys), which are handled safely by macOS without conflicting with typing.

## Changes
* `src/utils/hotkeyValidator.ts`: 
  * Added `MACOS_FN_ALLOW_LIST` containing OS-gated keys (Arrows, F1-F12, Home, End, etc.).
  * Updated `validateHotkey` to reject `Fn` combinations where the base key is not in the allow-list, surfacing a clear UI error.
  * Added logic to block `Fn` when combined with other modifiers (e.g., `Command`, `Control`) to prevent unexpected OS shortcut conflicts.

## Verification
- `npm run typecheck` clean.
- `npm run i18n:check` clean.
- `npm run lint` clean (0 errors on changed files).
- Verified standalone `Fn`/`Globe` key continues to trigger dictation correctly.
- Verified valid combinations (e.g., `Fn+Right`) save successfully.
- Verified invalid combinations (e.g., `Fn+A`, `Fn+Command+A`) are blocked with the correct validation error in the UI.
- Verified behavior on Windows/Linux remains unchanged.

### DIFF
--- a/src/utils/hotkeyValidator.ts
+++ b/src/utils/hotkeyValidator.ts
@@ -64,6 +64,27 @@ const SPECIAL_KEYS = new Set(
   ].concat(Array.from({ length: 24 }, (_, i) => `F${i + 1}`))
 );
 
+const MACOS_FN_ALLOW_LIST = new Set([
+  "F1",
+  "F2",
+  "F3",
+  "F4",
+  "F5",
+  "F6",
+  "F7",
+  "F8",
+  "F9",
+  "F10",
+  "F11",
+  "F12",
+  "Home",
+  "End",
+  "PageUp",
+  "PageDown",
+  "Delete",
+  "Insert",
+]);
+
 const MAC_RESERVED_SHORTCUTS = [
   "Command+C",
   "Command+V",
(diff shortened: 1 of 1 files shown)


gabrielste1n:
Heads up — #826 just landed on main, reworking the bare-Fn push-to-talk handling in macos-globe-listener.swift and main.js (Fn now cancels when used as a navigation modifier). This PR is still wanted for the Fn+letter combination case (#906) — could you rebase on current main? Happy to review once it's updated.

PR #964, opened 2026-06-16
feat(history): preserve discarded dictations in transcription history

### LINKED ISSUE #907: Feature Request - Preserve Discarded Dictation in Transcription History
# Feature Request: Preserve Discarded Dictation in Transcription History

## Is your feature request related to a problem? Please describe.

Pressing the Escape key immediately cancels the current dictation and permanently discards the transcription.

This can happen accidentally when Escape is intended for another interface action, such as closing a browser modal or dismissing an overlay. If several minutes of dictation have already been recorded, losing the entire transcription is extremely frustrating and requires starting again from the beginning.

## Describe the solution you'd like

When dictation is cancelled using the Escape key, save the transcription to history with a **Discarded** status rather than permanently deleting it.

Discarded transcriptions should:

* Be hidden from the standard transcription history by default.
* Be accessible through an **Show discarded transcripts** toggle or filter at the top of the history screen.
* Be clearly labelled as discarded.
* Allow the user to recover, copy, or permanently delete the transcription.
* Preserve the transcription text and relevant metadata, such as the date and duration.

This behaviour could be controlled by an opt-in setting, such as **Save discarded transcriptions**, for users who prefer not to retain cancelled dictations.

## Describe alternatives you've considered

Possible alternatives include:

* Displaying a confirmation dialog before discarding longer dictations.
* Providing a temporary Undo action after Escape is pressed.
* Requiring Escape to be pressed twice to cancel an active dictation.
* Allowing the Escape shortcut to be disabled or reassigned.

These options may help prevent accidental cancellation, but preserving discarded transcriptions provides the most reliable recovery mechanism.

## Additional context

The main use case is accidentally pressing Escape while trying to dismiss something in another application or browser window. The dictation may already contain several minutes of content, so silently losing it can result in significant lost work.

Discarded transcriptions should remain visually separate from completed transcriptions so they do not clutter the normal history view.

xAlcahest:
## Summary

Pressing Escape during a dictation discarded the recording for good. An accidental Escape (meant for a browser modal or another window) could wipe out minutes of speech. With the new opt-in setting, a cancelled dictation is kept in history instead, marked as discarded and recoverable.

`cancelRecording()` in `audioManager.js` now builds the audio blob before clearing the chunks and saves it as a `status='discarded'` row (no text, audio on disk). `getTranscriptions` hides those rows by default and a "Show discarded" toggle in `HistoryView` reveals them with a badge; Recover reuses the existing `retry-transcription` IPC to re-transcribe from the saved audio and mark the row completed. That handler used to reset `audio_duration_ms` on retry, so it now keeps the stored duration, which also fixes duration loss when retrying a failed transcription. The toggle stays disabled unless data retention is on and audio retention is above zero.

Fixes #907

## Changes

- `src/helpers/discardedRecording.js`: new pure helper, the gate that decides when a cancelled recording is worth keeping (opt-in + retention on + at least 1s).
- `test/helpers/discardedRecording.test.js`: node:test cases for the gate.
- `src/helpers/audioManager.js`: `cancelRecording` captures the blob before discarding and saves a discarded row via the new `saveDiscardedTranscription`.
- `src/helpers/database.js`: `getTranscriptions` takes `includeDiscarded` and excludes discarded rows by default.
- `src/helpers/ipcHandlers.js`: pass the option through `db-get-transcriptions`; preserve `audio_duration_ms` on `retry-transcription`.
- `preload.js`: forward the options object to `getTranscriptions`.
- `src/types/electron.ts`: add `discarded` to `TranscriptionStatus` and update the `getTranscriptions` signature.
- `src/stores/settingsStore.ts`, `src/hooks/useSettings.ts`: new `saveDiscardedTranscriptions` boolean, default off.
- `src/components/SettingsPage.tsx`: "Save discarded dictations" toggle in the Data Retention section, disabled when retention is off.
- `src/stores/transcriptionStore.ts`: refetch with `includeDiscarded`, drop discarded rows from live adds while the toggle is off.
- `src/components/ControlPanel.tsx`: show-discarded state and refetch wiring.
- `src/components/HistoryView.tsx`: "Show discarded" toggle.
- `src/components/ui/TranscriptionItem.tsx`: discarded row rendering (badge, duration, Recover action).
- `src/locales/*/translation.json`: toggle, badge, and settings strings across all 10 locales.


gabrielste1n:
## Code review — preserve discarded dictations

Nice feature, and the test coverage on `shouldSaveDiscardedRecording` plus the `!isDiscarded` UI gating are solid. A few things worth addressing before merge — two real bugs, the rest are cleanup/consistency.

### Tooling status
- `tsc --noEmit`: clean
- `prettier --check`: all changed files formatted
- `eslint`: no errors/warnings on the changed `.tsx` (the `src/helpers/*.js` files are ignored by eslint config, so they aren't linted at all)
- i18n: all 8 new keys (`controlPanel.history.discarded.*`, `settingsPage.privacy.saveDiscarded*`) are present and resolve in all 10 locales 👍

---

### 🔴 Bugs

**1. Toggle state desyncs from the store global → discarded items silently vanish**
`ControlPanel.tsx:138` / `transcriptionStore.ts:82`

"Show discarded" is tracked in two places that must stay in lockstep: React state `showDiscarded` (ControlPanel) and the module-global `currentIncludeDiscarded` (read by `addTranscription`). The mount effect calls `initializeTranscriptions()` with no args and has deps `[showAlertDialog, t]`.

After the user enables "Show discarded", anything that re-runs that effect — e.g. a language switch changing `t`'s identity — calls `initializeTranscriptions()` again, resetting `currentIncludeDiscarded` back to `false` while `showDiscarded` stays `true`. The toggle still reads "Hide discarded," the discarded rows disappear, and `addTranscription` then drops every newly-discarded item. Only a double-toggle recovers it. Suggest a single source of truth (read the flag back from the store, or pass it through every `initializeTranscriptions` call site).

**2. A discarded row whose audio-save fails is permanently unrecoverable**
`audioManager.js` `saveDiscardedTranscription` / `TranscriptionItem.tsx:184`

The DB row is saved first (`transcription-added` with `has_audio=0`), then audio is saved in a separate IPC call whose failure is only `logger.warn`-ed. The Recover button is gated on `isDiscarded && hasAudio`. If `save-transcription-audio` throws (disk full, permissions, bad path), `has_audio` stays `0` → no Recover button ever renders, and copy/raw/retry are all `!isDiscarded`-gated, so the row is a permanent empty "Discarded recording" with only Delete. The success path self-heals via `transcription-updated`, so only the error path is affected — but the audio the user explicitly opted to keep is silently lost.

---

### 🟠 Semantics

**3. "Recover" overloads the retry handler**
`TranscriptionItem.tsx:259` → `ipcHandlers.js` retry handler

Recover reuses `handleRetry` → `onRetryTranscription`, the failed→completed path: it force-sets `status="completed"`, runs the cleanup model, and emits `t("retrySuccess")` = **"Re-transcribed"**, which is the wrong message for an item that was never transcribed. Coupling also means future retry changes silently change recover. A recover-specific toast key (at minimum) would help.

---

### 🟡 DRY / simplification / altitude

**4. `getTranscriptions` builds two near-identical SQL strings** (`database.js:601`) and threads an `includeDiscarded` boolean through preload → ipc → store → component. The next hide-able status doubles the branches. Prefer one query with a conditional fragment, or a general status-filter param:
```js
const statusFilter = includeDiscarded ? "" : " AND status != 'discarded'";
const sql = `SELECT * FROM transcriptions WHERE deleted_at IS NULL${statusFilter} ORDER BY timestamp DESC LIMIT ?`;
```

**5. `saveDiscardedTranscription` is a near-verbatim copy of `saveFailedTranscription`** (`audioManager.js`). Both do `saveTranscription("", null, {status})` → `result.id` guard → `debouncedPush` → `durationMs = Math.round(s*1000)` → `saveTranscriptionAudio(id, buf, {durationMs, provider:null, model:null})`. Three copies of this save-record-then-save-audio dance now exist; extract a shared private helper so the audio-metadata contract lives in one place.

**6. `discardedDuration` re-implements mm:ss formatting** (`TranscriptionItem.tsx:73`) — the same `Math.floor(t/60)`/`padStart(2,"0")` logic is already inlined in `NoteBottomBar.tsx` and `DictationWidget.tsx`. No shared helper exists today; extracting `formatMmSs(seconds)` would collapse all three.

**7. `toggleShowDiscarded` duplicates the mount effect** (`ControlPanel.tsx:557`) — the `setIsLoading` try/finally + `showAlertDialog({ couldNotLoadTitle, couldNotLoadDescription })` block is verbatim the mount effect's. It also hardcodes `initializeTranscriptions(50, next)` instead of `DEFAULT_LIMIT` / the current limit, so toggling truncates a grown history back to 50. A small `loadTranscriptions(includeDiscarded)` helper would cover both.

---

### 🟡 Consistency

**8. Settings toggle is disabled-but-not-reset** (`SettingsPage.tsx:3541`) — when `dataRetentionEnabled` is turned off or `audioRetentionDays` hits 0, the toggle is disabled but its stored `true` persists, so re-enabling retention later silently resumes capturing discarded audio without a fresh opt-in. The capture-time gate keeps it safe data-wise, so this is a consent/UX point rather than a leak.

**9. CLI/REST bridge hides discarded with no opt-in** (`cliBridge.js:355`) — `GET /v1/transcriptions/list` calls `getTranscriptions(limit)`, inheriting the new default exclusion, with no query param to surface discarded rows. Minor desktop/API asymmetry.

---

**Verdict:** Fix **#1** and **#2** before merge. #3–#9 are quality/consistency improvements that can land here or as follow-ups.

xAlcahest:
Thanks for the thorough review. Fixed:

- `#1` desync: `includeDiscarded` now lives only in the zustand store, read by both `addTranscription` and the UI (`useShowDiscarded`). `initializeTranscriptions` keeps the current value instead of resetting to `false`, so a re-run of the mount effect (e.g. a language switch) no longer drops discarded rows.
- `#2` unrecoverable row: if the audio save fails, the just-created discarded row is deleted instead of left as an empty entry with only Delete.
- `#3` toast: Recover now shows a dedicated "Recovered" message (new `discarded.recovered` key in all 10 locales) instead of "Re-transcribed".
- `#4`: `getTranscriptions` uses a single conditional fragment.
- `#7`: extracted a shared `loadTranscriptions` helper for the mount effect and the toggle, using the current limit instead of a hardcoded 50.

Left as follow-ups: `#5` (shared save helper, touches the failed path), `#6` (`formatMmSs` extraction across components), `#8` (reset toggle on retention off, UX), `#9` (CLI bridge opt-in).

gabrielste1n (approved):

PR #965, opened 2026-06-16
fix: make discardedRecording an ESM module for Vite dev

xAlcahest:
## Summary

The renderer imports `shouldSaveDiscardedRecording` from `src/helpers/discardedRecording.js`, but that helper was written as CommonJS (`module.exports`). Vite does not transpile CommonJS source files in dev, so the named import failed with a `SyntaxError` and the whole window rendered blank under `npm run dev`. Production RPM builds go through Rollup, which handles the interop, so the breakage only showed up in dev. This converts the helper to ESM like every other helper in `src/helpers/`, and switches its test to the dynamic `import()` pattern already used by `dictationRouting.test.js`.

Based on #964 (that PR adds the helper).

## Changes

- src/helpers/discardedRecording.js: export with `export const`/`export function` instead of `module.exports`
- test/helpers/discardedRecording.test.js: load the helper through dynamic `import()` so node:test can run an ESM module


xAlcahest:
Folded into #964: the fix belongs to that feature branch, so I moved the commit there and closed this one.

PR #978, opened 2026-06-18
fix(audio): retry on default mic when the pinned deviceId is stale

### LINKED ISSUE #900: Recording fails with generic "Failed to access microphone" when a saved mic deviceId goes stale (OverconstrainedError unhandled)
## Summary
When a specific microphone is selected (so `selectedMicDeviceId` is saved and `preferBuiltInMic` is false), recording can suddenly start failing with the generic error:

> **Recording Error — Failed to access microphone**

The real cause is that Chromium/Electron device IDs rotate over time. Once the saved `selectedMicDeviceId` no longer matches a current device, `getUserMedia` throws an `OverconstrainedError`, which is not handled, so the user just sees the generic catch-all message with no hint that the *saved device* is the problem.

## Environment
- OpenWhispr v1.7.2
- Windows 11
- Mode: local Whisper (BYOK), but the bug is in the capture path so it applies to all modes

## Steps to reproduce
1. In settings, select a specific (non-default) microphone, so `preferBuiltInMic=false` and `selectedMicDeviceId` is persisted.
2. Use the app normally (works fine).
3. Let the Chromium media device ID rotate (re-salt over time, device replug, profile/storage changes, etc.).
4. Trigger the dictation hotkey.

## Expected
Recording either succeeds by falling back to an available mic, or the error clearly explains the saved device is no longer available.

## Actual
`getUserMedia` throws `OverconstrainedError` and the user sees the generic "Recording Error — Failed to access microphone". Nothing indicates the saved device is stale, so it looks like a permission/hardware/OS problem. (In my case this sent me down a multi-hour false trail blaming antivirus mic protection.)

## Root cause
`src/helpers/audioManager.js` → `getAudioConstraints()` hard-pins the saved device:
```js
if (!preferBuiltIn && selectedDeviceId) {
  return { audio: { deviceId: { exact: selectedDeviceId }, ...noProcessing } };
}
```
`{ exact: ... }` makes a stale ID fatal. In `startRecording()`, the catch block handles `NotAllowedError`, `NotFoundError`, and `NotReadableError`, but **not** `OverconstrainedError`, so it falls through to the generic `Failed to access microphone: ${error.message}` branch.

## Suggested fix
On `OverconstrainedError` (or as a general fallback when an exact-device constraint fails), retry `getUserMedia` without the `exact` deviceId constraint, and clear the stale `selectedMicDeviceId` so the app self-heals to the default mic:
```js
catch (error) {
  if (error.name === "OverconstrainedError" && constraints?.audio?.deviceId?.exact) {
    // saved device is gone — clear it and retry with the default mic
    clearSelectedMicDeviceId();
    return this.startRecording(); // or retry getUserMedia with { audio: noProcessing }
  }
  // ...existing handling
}
```
Adding an explicit `OverconstrainedError` branch with a clear message ("Your previously selected microphone is no longer available — switched to the default mic") would also help.

## Workaround for affected users
Close OpenWhispr, delete the `selectedMicDeviceId` key from `%APPDATA%/open-whispr/Local Storage/leveldb` (or re-pick a mic from the live list in settings), and it records again on the default device.

xAlcahest:
## Summary

Selecting a specific microphone, or relying on the built-in one, pins it with `{ deviceId: { exact } }` in `getAudioConstraints`. Chromium rotates media device IDs over time, so the saved id eventually matches no device and `getUserMedia` throws `OverconstrainedError`. Neither `startRecording` nor `startStreamingRecording` handled that error, so the user just saw the generic "Failed to access microphone" with no hint that the saved device was the problem.

When `getUserMedia` fails with a stale-device `OverconstrainedError`, recording now retries once on the system default mic. The saved preference is left untouched, so a device that comes back (a briefly unplugged USB mic) is picked up again automatically. A new `isStaleDeviceError` helper gates the fallback on the `deviceId` constraint specifically, so an unrelated constraint failure never triggers it.

Fixes #900

## Changes

- src/helpers/staleMicDevice.js: new pure helper `isStaleDeviceError`, true only for an `OverconstrainedError` whose failed constraint is `deviceId` (or unnamed)
- src/helpers/audioManager.js: `getAudioConstraints(forceDefaultMic)` returns default constraints when forced; `startRecording` and `startStreamingRecording` catch a stale-device error and retry once on the default mic, guarded against re-looping and logged; the streaming retry honors a stop requested mid-start
- test/helpers/staleMicDevice.test.js: 12 tests for deviceId/empty/unnamed constraints, non-deviceId constraints, and non-OverconstrainedError inputs

### DIFF
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -11,6 +11,7 @@ import {
   recordLocalSpeechWindow,
 } from "./localSpeechGate";
 import { reacquireIfDead } from "./micTrackHealth";
+import { isStaleDeviceError } from "./staleMicDevice";
 import { shouldSaveDiscardedRecording } from "./discardedRecording";
 import { getSettings, getEffectiveCleanupModel, isCloudCleanupMode } from "../stores/settingsStore";
 import { shouldSkipTranscriptionApiKey } from "./transcriptionAuth";
@@ -292,7 +293,7 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
     return this.sttConfig?.streamingProvider || defaultProvider;
   }
 
-  async getAudioConstraints() {
+  async getAudioConstraints(forceDefaultMic = false) {
     const { preferBuiltInMic: preferBuiltIn, selectedMicDeviceId: selectedDeviceId } =
       getSettings();
 
@@ -306,6 +307,13 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
       autoGainControl: false,
       channelCount: 2,
     };
+
+    // Pinned device was unavailable (Chromium rotates IDs / device unplugged); fall back to the
+    // system default for this capture without discarding the saved preference. See #900.
+    if (forceDefaultMic) {
+      logger.debug("Using default microphone (pinned device unavailable)", {}, "audio");
+      return { audio: noProcessing };
+    }
 
     if (preferBuiltIn) {
       if (this.cachedMicDeviceId) {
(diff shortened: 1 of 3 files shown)


gabrielste1n (approved):
great work

Issue #131, opened 2026-01-21
Allow adding Azure AI Foundry models


Ashish-Bansal:
**Describe the solution you'd like**
Ability to use models deployed in Azure AI Foundry. 

It needs following variables:
```
  azureOpenAIApiInstanceName
  azureOpenAIApiDeploymentName
  azureOpenAIApiVersion
  azureOpenAIApiKey
  azureOpenAIEndpoint
```

gabrielste1n:
@Ashish-Bansal feel free to make a PR! Will pick this up when I can 🙏

bikramjitk:
@gabrielste1n @xAlcahest can you review https://github.com/OpenWhispr/openwhispr/pull/997? Thanks

gabrielste1n:
on it thank you legend @bikramjitk

gabrielste1n:
The core request here shipped in v1.7.4 (#997): you can add Azure AI Foundry / Azure OpenAI models with your own endpoint and deployment vars. Realtime streaming via Azure is tracked separately in #1009. Closing as released — if anything from the original ask is still missing on 1.7.5, comment and we'll reopen.

PR #997, opened 2026-06-24
feat(transcription): add Azure AI Foundry / Azure OpenAI speech-to-text

bikramjitk:
Addresses #131.

## Problem

Connecting an Azure AI Foundry (Azure OpenAI) speech-to-text deployment as a custom transcription provider fails with `404 DeploymentNotFound`, even when the deployment works in the Azure playground and via curl.

The custom STT path could only emit the plain OpenAI shape — `{base}/audio/transcriptions` with `Authorization: Bearer` and no `api-version`. Azure requires the deployment in the URL path (`/openai/deployments/{deployment}/audio/transcriptions`), an `api-version` query string, and the `api-key` auth header (Bearer is reserved for Entra ID tokens). The enterprise *language* model already works because it routes through `@ai-sdk/azure`; the transcription path had no equivalent.

## Changes

- `src/utils/urlUtils.ts` — add `isAzureOpenAIEndpoint()` and `buildAzureTranscriptionUrl()` helpers.
- `src/helpers/audioManager.js` — the STT endpoint resolver detects Azure hosts (`*.cognitiveservices.azure.com`, `*.openai.azure.com`, `*.services.ai.azure.com`) and builds the deployment-style URL; auth switches to the `api-key` header on Azure hosts. `getTranscriptionEndpoint()` now takes the deployment (model) name.
- `api-version` defaults to a transcribe-capable preview (`2025-03-01-preview`) — deliberately not the enterprise LLM `azureApiVersion` default (`2024-10-21`, which does not support `gpt-4o-transcribe`). Users can override it by appending `?api-version=...` to the endpoint URL; a full `.../audio/transcriptions` URL is respected as-is.
- UI — a contextual Azure hint under the custom endpoint fields, i18n keys added to all 10 locales (`check-i18n.js` passes).
- Tests — `test/helpers/azureTranscriptionUrl.test.js`.
- CHANGELOG — Unreleased entry.

## Usage

Custom provider → Endpoint URL = your resource endpoint (e.g. `https://your-resource.cognitiveservices.azure.com/openai/v1`), Model = your exact deployment name (e.g. `gpt-4o-transcribe`).

## Testing notes

URL-builder logic verified against representative Azure values; all locale JSON validated and the i18n consistency check passes. The `node --test` suite (including the new test) requires the repo-pinned Node 24 (native `.ts` type-stripping + ESM). Verified end-to-end against a live Azure `gpt-4o-transcribe` deployment in a local packaged build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Transcription
+
+- **Azure AI Foundry / Azure OpenAI speech-to-text.** The custom transcription provider now recognizes Azure endpoints (`*.cognitiveservices.azure.com`, `*.openai.azure.com`, `*.services.ai.azure.com`) and builds the deployment-style URL Azure requires (`/openai/deployments/{model}/audio/transcriptions?api-version=...`) instead of the plain OpenAI `{base}/audio/transcriptions` shape, which returned `404 DeploymentNotFound`. Auth now uses Azure's `api-key` header on Azure hosts. Enter your resource endpoint as the URL and your exact deployment name in the Model field; `api-version` defaults to a transcribe-capable preview and can be overridden by appending `?api-version=...` to the endpoint.
+
 ## [1.7.3] - 2026-06-23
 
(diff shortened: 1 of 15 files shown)


gabrielste1n (approved):
epic work @bikramjitk thank you!

PR #1002, opened 2026-06-25
fix: searchable, grouped model picker for large cloud provider lists

### LINKED ISSUE #999: [UI Bug] Model selection UI unusable with cloud providers that expose 300+ models (e.g. OpenRouter) — no search, no grouping, no pagination, no selection feedback
**Describe the bug**
When using cloud providers like OpenRouter that expose 300–400 models, the language model selection UI becomes unusable. There is no search/filter functionality, no pagination or virtual scrolling, and no indication of which models are currently selected. Scrolling through hundreds of entries to find a specific model is impractical.

**To Reproduce**
1. Go to Settings → Language Models
2. Add OpenRouter as a cloud provider
3. Observe that 300–400 models populate the selection list
4. Attempt to find a specific model (e.g. anthropic/claude-sonnet-2026-04-01)
5. Scroll through the entire unsorted/flat list
6. See no search bar, no grouping, no pagination
7. Select a model — the UI gives no clear visual feedback on which models are currently enabled

**Expected behavior**
- Search/filter input to quickly narrow down models by name or provider prefix
- Grouping by provider (e.g. Anthropic, OpenAI, Meta) or category
- Virtual scrolling or pagination for performance
- Show the slug or the friendly name of the model, and ellipsize the description.
- Perhaps a "selected models" summary section

**Screenshots**
Image

Image
Image

Desktop (please complete the following information):
- OS: Linux
- Browser: Firefox / Chrome
- Version: latest - 1.7.3
Smartphone
N/A

**Additional context**
OpenRouter exposes models like openai/gpt-4o, anthropic/claude-sonnet-4-20250514, meta-llama/llama-4-scout, etc. A flat list of 350+ entries with no search makes selection nearly impossible. Grouping by provider prefix (the string before /) would be a significant improvement.

gabrielste1n:
Started as a fix for the unusable model-selection UI with 300+ model providers (#999) and grew into the OpenRouter integration + a secret-key refactor to support it cleanly. Now rebased on `main`, so it also carries the Tinfoil provider and folds it into the same refactor.

## What's in this PR

1. **Searchable, grouped, virtualized model picker** — the custom / OpenAI-compatible panel rendered a provider's `/models` (300–400 for OpenRouter) as a flat, unsearchable list. Now: search, grouping by provider prefix with counts, `@tanstack/react-virtual` windowing, keyboard/listbox navigation, a pinned "Selected" row, and ellipsized descriptions. Small lists keep the plain renderer.

2. **Per-row provider icons + stripped labels** — inside a provider group the header already names the provider, so labels drop the `provider/` prefix (`openai/gpt-5.5` → `gpt-5.5`) and show the provider's icon (mapped from our existing assets; the API returns no icons). The pinned "Selected" row keeps the full id so its provider stays identifiable.

3. **OpenRouter as a first-class provider** — its own tab with its own **global, encrypted `openrouterApiKey`** (not the shared per-scope custom key). Resolves to OpenRouter's endpoint across both reasoning paths, the AI-SDK factory, the provider registry, and `getModelProvider`. Uses the **Chat Completions** dialect (OpenRouter doesn't implement the OpenAI Responses API), so Agent Mode works.

4. **Secret-key manifest refactor** — the per-key boilerplate that spanned `environment.js`, `ipcHandlers.js`, `preload.js`, and the settings store now derives from one `src/config/secretKeys.js` manifest. Every method name and IPC channel is preserved (generated, not renamed), so nothing downstream changes. **8 BYOK keys** now flow through it: OpenAI, Anthropic, Gemini, Groq, xAI, Mistral, OpenRouter, and Tinfoil (folded in during the `main` merge). `preload.js` inlines the `{base, get, save}` tuples (a sandboxed preload can't `require()` a local module), guarded by a test that asserts they mirror the manifest.

5. **UTM attribution on partner links** — outbound "Get API key" pages + HuggingFace model links are tagged via a shared `GetApiKeyLink` component + a `withUtm` helper that only touches `http(s)` and never clobbers an existing `utm_source`.

## How to test

Build the branch (`npm run dev`). The highest-value checks are the **non-breaking secret refactor** (all keys) and **OpenRouter/Tinfoil inference**.

### A. Every API key still saves, loads, and persists — do this first

The secret plumbing for all 8 BYOK providers is now generated from one manifest, so the goal is to prove nothing regressed.

1. **Upgrade path:** on your current release, save keys for whatever providers you have (OpenAI / Anthropic / Gemini / Groq / xAI / Mistral). Build this branch → **Settings → AI Models** and confirm each key still loads.
2. **Round-trip each of the 8:** OpenAI, Anthropic, Gemini, Groq, xAI, Mistral, **OpenRouter**, **Tinfoil** — enter a key, fully quit and relaunch → the key persists (encrypted round-trip through the manifest path).
3. **Transcription BYOK** (OpenAI / Groq / xAI / Mistral): keys still save/load.
4. `node --test test/helpers/secretKeys.test.js` → **4 passing** (round-trip + clear for all 8, plus the preload-mirror drift guard).

### B. The app launches and dictates (preload runs in sandboxed windows)

The dictation overlay, notification, and transcription-preview windows run a **sandboxed** preload; this verifies `window.electronAPI` is live there.

1. Launch → the dictation overlay appears and is interactive.
2. Dictate something → transcription + cleanup run and the text pastes.
3. Trigger a meeting notification / transcription preview → they render.

### C. OpenRouter (its own key + inference)

1. **Settings → AI Models → Cloud → OpenRouter** tab.
2. *Without* a key the model list still populates (its `/models` is public) — search `claude`, confirm grouping-by-provider with counts + icons + stripped labels.
3. Click **Get API key**, paste your key, pick a model. Quit + relaunch → OpenRouter tab + model + key all persist.
4. **Cleanup:** dictate and let cleanup run → goes through OpenRouter with no 401.
5. **Agent Mode:** set an OpenRouter model for the agent and run an agent command that uses a tool (e.g. search notes) → it reaches OpenRouter's chat endpoint and tools stay enabled.
6. Switch to the **Custom** tab → your OpenRouter key is **not** there (separate key, no cross-provider leak).

### D. Tinfoil still works (merged from main; key plumbing moved into the manifest)

1. **Cloud → Tinfoil** tab: save a key, quit + relaunch → persists.
2. Tinfoil cleanup / realtime transcription still works (`getTinfoilKey` is now generated from the manifest).
3. Both **OpenRouter and Tinfoil** tabs appear (before Custom) and the app starts with **no duplicate-IPC-channel crash**.

### E. Model picker UX

1. Switch between provider tabs — fixed-catalog tabs preset their default model; the OpenRouter/Custom tabs **clear the previous provider's selection** and show a “select a model” hint, so a stale id from another provider can't silently persist. Mid-refresh, the panel keeps your selection and query.
2. In the search box: arrow keys move through results, **Enter** selects, "no models match" shows even while a model is selected, and the list scrolls back to top on a new query.

### F. Regression

- **Custom** tab: any OpenAI-compatible endpoint + key still works, URL still editable.
- **OpenAI / Anthropic / Gemini / Groq** tabs: keys load, models list, cleanup works.

## Verification done

- `tsc` ✓, eslint ✓, prettier ✓ (whole repo), `i18n:check` ✓ — full suite **175 passing / 0 failing** (incl. the secret round-trip + preload-mirror tests).
- Merge with `main` resolved by unioning OpenRouter + Tinfoil everywhere they collide; Tinfoil folded into the manifest so its old hand-written IPC handlers/bridges are removed (no double-registration).
- Secret channel alignment traced across all 8 BYOK keys (manifest → IPC → preload → env var).

Closes #999.


xAlcahest:
Took a careful pass over this. The feature itself is clean. Most of what I found is edge-case polish, plus one credential-routing thing I'd fix before merge.

## Worth fixing

**The API key gets sent to the wrong provider when you switch between the OpenRouter and Custom tabs.** Both tabs render the same `OpenAICompatiblePanel` and share `customReasoningApiKey` (`ReasoningModelSelector.tsx:554`). `handleCloudProviderChange` swaps the base URL (`:477`/`:483`) but never the key, so after the switch `OpenAICompatiblePanel`'s fetch effect re-runs and sends `Authorization: Bearer <old key>` to the new provider's `/models` (`OpenAICompatiblePanel.tsx:128-134`). OpenRouter → Custom ships your `sk-or-...` to `api.openai.com`; Custom → OpenRouter ships the custom key to `openrouter.ai`. It only 401s, so it's bounded to the wrong vendor's request logs rather than an attacker, but it's a credential going somewhere it shouldn't. Clearing the key alongside the base URL would fix it.

**A non-conformant endpoint can take down the whole control panel.** `OpenAICompatiblePanel.tsx:153-160` casts the `/models` fields with `as string` (a no-op at runtime). A numeric `id` survives the `if (!value)` guard and later hits `value.indexOf("/")` / `label.localeCompare(...)` in `SearchableModelList`'s render; an object `description` renders as a React child. Either throws during render and trips the global `ErrorBoundary`. The cast pattern predates this PR, but since the new list now runs string ops on these values, this is the natural place to coerce (`String(item?.id ?? "")` and friends).

## UX papercuts in the new list

- The "no results" empty state never shows once a model is selected. `selectedOption` is computed without the query (`SearchableModelList.tsx:50`) and the pinned Selected row is always pushed (`:81-93`), so `rows.length` is never 0 and `:132` can't render `noSearchResults`. Type a non-matching query and you just see the selected model, no feedback.
- Scroll position isn't reset when the query changes (`SearchableModelList.tsx:137`). With a few hundred models, scroll down then type and the first match sits off-screen above the viewport.
- Refresh wipes the in-progress search. `loadRemoteModels` runs `setModelOptions([])` mid-fetch (`OpenAICompatiblePanel.tsx:105`), which drops `displayedModels.length` below the threshold (`:332`) and unmounts `SearchableModelList` along with its `query` state.
- Typing OpenRouter's own suggested URL into the Custom tab (the example at `OpenAICompatiblePanel.tsx:262`) silently flips you onto the locked OpenRouter tab, because the mount effect re-derives the tab from the base URL (`ReasoningModelSelector.tsx:411` via `storedToTab`). Clicking back to Custom then resets your URL to OpenAI's (`:483`).
- OpenRouter shows "API Key (Optional)" (`OpenAICompatiblePanel.tsx:271`), but its `/models` is public so the list populates and a model persists without a key, then inference 401s later. The key isn't optional for OpenRouter.

## Smaller stuff

- Prefix icon aliasing applies to any endpoint with >12 models, not just OpenRouter (`SearchableModelList.tsx:27` → `providerIcons.ts:57`), so a self-hosted `google/...` id renders the Gemini logo even though `gemma.svg` exists.
- A few grouping edge cases on arbitrary endpoint ids: an id ending in `/` gives a blank label (`SearchableModelList.tsx:31`); case-variant prefixes (`OpenAI/` vs `openai/`) produce two headers because grouping keeps the raw case while search lowercases (`:68`); duplicate ids produce duplicate React keys (`:103`).
- `useVirtualizer` has no `getItemKey` (`SearchableModelList.tsx:109`), so the measurement cache stays index-keyed and a reused row can briefly render with a stale height after filtering. `getItemKey: (i) => rows[i].key` handles it.
- i18n: the extracted `ModelCard` hardcodes "Cancel"/"Download" (`ModelCardList.tsx:200`/`:213`) and the empty state hardcodes "No models available..." (`:255`), while the same component uses `t()` for the surrounding badges. `common.cancel`/`common.download` already exist.
- Small dedupe: `MODEL_SEARCH_THRESHOLD` (`SearchableModelList.tsx:11`) and `SEARCH_THRESHOLD` (`ui/LanguageSelector.tsx:26`) are the same 12; the `getProviderIcon` + `isMonochromeProvider` pair at `ReasoningModelSelector.tsx:392` and `TranscriptionModelPicker.tsx:737` is exactly what `getRemoteProviderIcon` already returns; `TranscriptionModelPicker.tsx:968` re-implements `GetApiKeyLink`.
- `withUtm` (`externalLinks.ts:18`) returns non-http(s) URLs unchanged and `open-external` (`ipcHandlers.js:2406`) has no scheme allowlist before `shell.openExternal`. Not reachable today since every caller passes hardcoded https, but a scheme guard on the new helper would be cheap defense-in-depth.

xAlcahest:
Re-ran the review after the new commits. Nice turnaround on the first-round items: the dedicated `openrouterApiKey` kills the shared-key leak, and the list picked up the empty-state fix, the scroll reset, `getItemKey`, and the removal of the tab-vs-URL mapping. I also ran the branch end to end (voice agent → OpenRouter → `openai/gpt-4o-mini`) and it works once a model is selected.

The new OpenRouter plumbing has a few spots where the provider isn't wired all the way through:

**`isAvailable()` doesn't know about OpenRouter.** The provider short-circuits cover custom/bedrock/azure/vertex and the final key sweep checks openai/anthropic/gemini/groq/tinfoil/local, never `getOpenrouterKey` (`ReasoningService.ts:899-915`). With `cleanupProvider="openrouter"` and no other key configured it returns false, and `audioManager` gates the whole reasoning pipeline on that and caches it, so every dictation pastes the raw transcript with nothing surfaced. Regression vs the previous revision, where the OpenRouter tab stored `provider="custom"` and passed the custom short-circuit. The same silence shows up one more way: since the tab intentionally doesn't preset a model, a fresh OpenRouter setup with no model picked also falls back to cleanup without a hint. That combination bit me while testing, the voice agent quietly did nothing until I noticed no model was selected.

**Thinking suppression fires on every OpenRouter request, with no way to turn it off.** Vendor-prefixed ids aren't in the local registry, so the `supportsThinking` exemption never applies (`thinkingSuppression.ts:47-50`), and with `disableThinking` defaulting to true in all four scopes every chat request gets `reasoning_effort:"none"` plus `chat_template_kwargs:{enable_thinking:false}`. OpenRouter forwards unknown params upstream, strict backends reject them, and reasoning models can't be exempted because the toggle only renders for `custom` or registry models with `supportsThinking` (`InferenceConfigEditor.tsx:152-156`). In my test `openai/gpt-4o-mini`'s upstream tolerated the params, but that's upstream luck.

**OpenRouter requests inherit the OpenAI request shape.** Vendor-prefixed ids fall into `getOpenAiApiConfig`'s final branch (`ModelRegistry.ts:392`), so every OpenRouter call sends `max_completion_tokens` and never `temperature`, and `getEndpointCandidates(OPENROUTER_BASE)` has no stored preference, so the first call per install probes `/responses` before `/chat/completions` (`openai.ts:60-77`). Pinning OpenRouter to chat completions with `max_tokens` would settle both.

**`updateApiKeys()` silently drops `openrouterApiKey`.** Every other `ApiKeySettings` member is handled there, including `tinfoilApiKey` (`settingsStore.ts:1704`). One-liner.

**Switching a scope's mode away and back wipes an OpenRouter selection.** `isProviderValidForMode` checks the registry's `cloudProviders`, which don't include `openrouter`, so the mode-change path resets provider and model to `""` (`InferenceConfigEditor.tsx:25-33`, `:118-121`).

Smaller things, take or leave:

- Users who had OpenRouter configured through the Custom tab keep their key in the old shared slot; the new tab starts empty with no migration.
- `secretKeys.test.js` mocks `safeStorage` off but not `@napi-rs/keyring` (`secretCrypto.js:53`), so on a machine with a working keyring the round-trip exercises the keyring path rather than the plaintext fallback its comment describes. And nothing runs it yet: no `test` script and no CI job, so the preload drift guard it provides never fires.
- The fallback capitalizer renders the brand as "Openrouter" in the key-missing error (`ReasoningService.ts:129`) and in PromptStudio's Test tab (`PromptStudio.tsx:36` has no openrouter entry).
- `environment.js` imports the manifest but still hand-maintains `SECRET_KEYS` next to it (`:8-10`), one more list to forget when adding a key.
- `docs/network-allowlist.md` doesn't list `openrouter.ai`; egress-allowlisted setups will hit it.
- `REMOTE_PROVIDER_ALIASES` maps the `microsoft` prefix to the Azure icon (`providerIcons.ts:67`), so Microsoft-published models like Phi get Azure branding.
- The listbox hardcodes `id="model-listbox"` / `id="model-opt-…"` (`SearchableModelList.tsx:186`), which collides if two instances ever mount together, and the keyboard cursor can go stale when rows rebuild without a query change (`:123`).

Still as described in my earlier comment: the `as string` casts on the `/models` payload, Refresh wiping the in-progress search, the "API Key (Optional)" header on the OpenRouter tab, the prefix icon aliasing on non-OpenRouter endpoints, the grouping edge cases, and the hardcoded "Cancel"/"Download"/"No models available" strings.

aryadovoy:
Where can I select OpenRouter provier?
image

With custom I get error: `API Error: 401 {"error":{"message":"Missing Authentication header","code":401}}`, app version is 1.7.5.

Issue #581, opened 2026-04-14
Realtime streaming hardcodes wss://api.openai.com — ignores custom transcription endpoint


benjaminharr:
## Bug Description

When the user has configured a **Custom transcription provider** (with a custom Base URL and API key), the **batch transcription** endpoint correctly routes to the custom URL, but the **Realtime streaming** transcription **always connects to `wss://api.openai.com/v1/realtime`** — completely bypassing the custom endpoint.

This causes silent fallback behavior:
- If the user also has `OPENAI_API_KEY` set → streaming silently uses the OpenAI key (no error, but the custom endpoint is bypassed without notice)
- If `OPENAI_API_KEY` is not set → streaming fails with `WebSocket closed before ready (code: 3000)` because OpenAI rejects the custom-provider key

This is the same class of issue as #430 (silent fallback to OpenAI) — only this time for the streaming/realtime path instead of the batch path.

## Steps to Reproduce

1. Set up a self-hosted OpenAI-compatible proxy that supports the Realtime API (e.g. LiteLLM ≥ v1.83 — `wss://your-proxy/v1/realtime` works correctly with proper auth)
2. In OpenWhispr, configure **Custom** transcription provider:
   - Base URL: `https://your-proxy/v1`
   - API key: your custom key
3. Remove `OPENAI_API_KEY` from the env to disable fallback
4. Start a streaming transcription (e.g. dictation or meeting note) with `gpt-4o-mini-transcribe`

**Expected**: WebSocket connects to `wss://your-proxy/v1/realtime?intent=transcription` using the custom key
**Actual**: WebSocket connects to `wss://api.openai.com/v1/realtime?intent=transcription`, fails with "WebSocket closed before ready (code: 3000)" because the custom key is rejected by OpenAI

## Root Cause

`src/helpers/openaiRealtimeStreaming.js` line 47 hardcodes the URL:

```js
const url = "wss://api.openai.com/v1/realtime?intent=transcription";
```

There is no option to override it via `connect()`, even though the surrounding code (`src/config/constants.ts` → `normalizeBaseUrl`, `buildApiUrl`) already supports custom base URLs for batch transcription.

The callers in `src/helpers/ipcHandlers.js` (lines 3168, 3436) instantiate `OpenAIRealtimeStreaming` without passing a base URL — the user's custom transcription endpoint config is never consulted.

## Suggested Fix

1. Accept `baseUrl` in `OpenAIRealtimeStreaming.connect()` options:

   ```js
[...]

PR #1009, opened 2026-06-25
feat(transcription): Azure OpenAI realtime streaming transcription (BYOK)

bikramjitk:
## Summary

Adds a bring-your-own-key **Azure OpenAI** cloud transcription provider that streams dictation over the OpenAI **realtime WebSocket** protocol exposed by Azure OpenAI / AI Foundry resources. Today, Azure/Foundry users fall back to the slower batch upload path; this gives them the same low latency as the existing OpenAI streaming path.

## Why

`OpenAIRealtimeStreaming` was hardcoded to `wss://api.openai.com/v1/realtime`, so dictation streaming only worked with OpenAI directly. Azure exposes the identical realtime protocol at a resource-scoped URL with `api-key` auth, so we can reuse the same client with a generic URL builder.

## What changed

- **`openaiRealtimeStreaming.js`** — generic Azure realtime URL builder that normalizes whatever endpoint form the user pastes and targets the correct surface:
  - `…/openai/v1` (or bare host) → GA `…/openai/v1/realtime`
  - `…/openai` → preview `…/openai/realtime` with a `deployment` param
  - `http(s)`→`ws(s)`, trailing `/realtime` tolerated, `intent=transcription`
  - `api-key` auth header; configurable input sample rate.
  - **The OpenAI (non-Azure) path is unchanged — the Azure branch is gated entirely on `endpoint` being present.**
- **`audioManager.js`** — gate Azure streaming on the cloud provider selection + presence of the three credentials; stream at **24 kHz** (Azure realtime rejects input below 24 kHz) via a rate-aware `AudioContext`; everything else continues at 16 kHz.
- **`ipcHandlers.js` / `environment.js` / `preload.js`** — persist & load the Azure transcribe endpoint, deployment and api-version (`.env`) and API key (`safeStorage`), plus a dedicated dictation connect branch. Kept separate from the existing enterprise Azure **reasoning** keys so the two deployments don't clobber each other.
- **settings store / types / model registry / `TranscriptionModelPicker`** — new "Azure OpenAI" transcription tab with endpoint, deployment, api-version and key fields.

## Scope

This targets **Azure OpenAI transcription deployments** (`gpt-4o-transcribe`, `gpt-4o-mini-transcribe`, `whisper`). Non-OpenAI Foundry speech-to-text (Voice Live API / Azure AI Speech / `mai-transcribe`) uses a different endpoint, auth and message schema and is intentionally out of scope; the deployment field help text states the supported deployments, and a wrong/non-transcribe deployment surfaces a connection error rather than silently degrading.

## Validation

- Assumptions cross-checked against Microsoft Learn (realtime WS URL forms for GA vs preview, `api-key` auth for non-browser clients, `session.type: transcription` schema, ≥24 kHz PCM16).
- Verified end-to-end against a live Azure AI Foundry endpoint: correct WS URL, 24 kHz capture, accurate transcription, low latency.
- `tsc --noEmit` clean, `eslint` clean, `i18n:check` consistent across all 10 locales, renderer build passes.

## Backward compatibility

No change for existing OpenAI / OpenWhispr Cloud / Corti users — all new behavior is gated behind selecting the Azure provider and supplying credentials.

## Related issues

- Related to #581 (realtime streaming hardcodes `wss://api.openai.com` and ignores custom transcription endpoints) — this PR removes that hardcoding for the **Azure** path. The generic custom OpenAI-compatible proxy case in #581 is a separate slice and is not addressed here.
- Complements #131 (Azure AI Foundry models) on the transcription side.

### DIFF
--- a/preload.js
+++ b/preload.js
@@ -398,6 +398,17 @@ contextBridge.exposeInMainWorld("electronAPI", {
   saveAzureDeployment: (value) => ipcRenderer.invoke("save-azure-deployment", value),
   getAzureApiVersion: () => ipcRenderer.invoke("get-azure-api-version"),
   saveAzureApiVersion: (value) => ipcRenderer.invoke("save-azure-api-version", value),
+  getAzureTranscribeKey: () => ipcRenderer.invoke("get-azure-transcribe-key"),
+  saveAzureTranscribeKey: (key) => ipcRenderer.invoke("save-azure-transcribe-key", key),
+  getAzureTranscribeEndpoint: () => ipcRenderer.invoke("get-azure-transcribe-endpoint"),
+  saveAzureTranscribeEndpoint: (value) =>
+    ipcRenderer.invoke("save-azure-transcribe-endpoint", value),
+  getAzureTranscribeDeployment: () => ipcRenderer.invoke("get-azure-transcribe-deployment"),
+  saveAzureTranscribeDeployment: (value) =>
+    ipcRenderer.invoke("save-azure-transcribe-deployment", value),
+  getAzureTranscribeApiVersion: () => ipcRenderer.invoke("get-azure-transcribe-api-version"),
+  saveAzureTranscribeApiVersion: (value) =>
+    ipcRenderer.invoke("save-azure-transcribe-api-version", value),
   getVertexProject: () => ipcRenderer.invoke("get-vertex-project"),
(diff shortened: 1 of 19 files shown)


bikramjitk:
+ @gabrielste1n whenever you get the chance :)

PR #1015, opened 2026-06-26
fix:style: show line breaks in the history view

IdrisGit:
The line breaks of the final LLM output after all the cleaning process, is preserved properly in the db and when copying the text, however on the main window, the history view doesn't render those line break, it's just a case of missing CSS and this PR adds that line break styling.

Here is out it's rendered currently in the latest release:
image

Here is the screenshot of the fix:
image


gabrielste1n (approved):
Verified the fix: history rows now render stored line breaks via whitespace-pre-wrap, matching how the preview overlay renders the same transcript text, and the class updates are behavior-neutral. Lint, typecheck, prettier, and the unit tests all pass.

PR #1020, opened 2026-06-29
fix: preserve rich clipboard formats on restore

### LINKED ISSUE #834: Clipboard restore step destroys the formatted-data of the original clipboard contents
**Describe the bug**

When the "Keep transcription in clipboard" toggle (Settings → General → Clipboard) is **off**, OpenWhispr's clipboard restore step destroys the formatted-data representations of the original clipboard contents. Plain-text clipboard contents survive the round trip; multi-format clipboard contents (e.g. from MS Word, Outlook, Notes, or anywhere else producing HTML+RTF+plain-text+image-preview) come back as a single representation — typically a small PNG preview — with the HTML, RTF, and plain-text formats silently dropped.

In other words, the toggle does the *opposite* of "non-destructive" for any clipboard payload richer than plain text: the original is captured before dictation but is not actually fully restored after the paste step.

**To Reproduce**

1. Settings → General → Clipboard → ensure **"Keep transcription in clipboard"** is OFF.
2. In Microsoft Word for Mac (any modern version — repro'd on Word 15 / "Aptos" default theme), type a short sentence (e.g. `This is text in a Word document`) and Cmd-C it. This puts four representations on the macOS pasteboard:
   - `public.html` (Word's HTML)
   - `public.rtf` (Word's RTF)
   - `public.utf8-plain-text` (plain text fallback)
   - `public.png` / `public.tiff` (Word's auto-generated preview image — Word always includes this)
3. Open https://evercoder.github.io/clipboard-inspector/ in a browser and paste into the page **before dictating** to confirm all 4 representations are present. (Result: `.types: 3 type(s) available` with `text/html`, `text/rtf`, `text/plain`, plus `.items: 3 string item(s)` and the original PNG attached separately.)
4. Without changing the clipboard, focus a target app and trigger an OpenWhispr dictation. Let it auto-paste, and let the clipboard restore step run.
5. Paste into the clipboard-inspector page again.

https://github.com/user-attachments/assets/5da89087-5187-416f-bcb6-0029426e80c6

**Expected behavior**

Step 5 produces the same dump as step 3 — all 4 representations intact. The "Keep transcription in clipboard = off" setting should *exactly* round-trip whatever was on the clipboard before dictation, regardless of the formats it carried.

**Actual behavior**

Step 5 produces only one representation:

- `.types: 1 type available` — `Files`
- `.items: 1 item available` — kind `file`, type `image/png`, name `image.png`, size `3314` bytes

The Word-formatted HTML, the RTF, and the plain text are gone. What's left is just the small PNG preview that Word had attached as a fourth representation in the original clipboard payload. That PNG was an *adjacent* representation in the source clipboard, not a derived re-rendering — OpenWhispr captured it correctly at snapshot time but restored *only* it.

Reproduces every time. Does **not** reproduce with plain-text clipboard contents (single-format clipboards survive the round trip fine).

**Screenshots**

Video: `Post-dictation clipboardData reformatted.mp4` (attached) shows the full repro: the clipboard-inspector before/after, and the dictation in between.

Pasted clipboard-inspector dumps (before and after the dictation cycle) are in the **Additional context** section below.

**Desktop:**
- OS: macOS 26.3.1 (a), Apple Silicon (M4 Pro, 14-core / 20-core GPU, 48 GB RAM)
- App: OpenWhispr 1.7.2
- Source app for the clipboard payload: Microsoft Word for Mac 15 (default "Aptos" theme — i.e. brand-new Word document)
- Paste target / inspector: Chrome on macOS, https://evercoder.github.io/clipboard-inspector/

**Smartphone:**
- N/A (Electron desktop app)

**Additional context**

**Root cause is visible in the bundled code.** Searching `app.asar` for clipboard API calls finds these reads:

```
clipboard.readHTML(
clipboard.readImage(
clipboard.readText(
```

…and these writes:

```
clipboard.writeImage(original.data
clipboard.writeText(original.data
clipboard.writeText(text          // plus various single-arg call sites
```

Two problems jump out:

1. **No `clipboard.readRTF()` / `clipboard.writeRTF()` anywhere.** RTF is captured neither at snapshot nor at restore time, so any payload that carries RTF will lose it on restore even before the second bug below kicks in. (This is by far the most common loss on macOS — every Office/Apple app puts RTF on the clipboard alongside HTML.)
2. **No atomic `clipboard.write({ text, html, rtf, image, ... })` call** — only the single-format setters `writeText` and `writeImage`. Each single-format setter *clears* the rest of the clipboard. So if the restore path looks like (pseudo-code):

   ```js
   const snap = {
     text: clipboard.readText(),
     html: clipboard.readHTML(),
     image: clipboard.readImage(),
   };
   // …paste dictation result…
   if (!snap.image.isEmpty()) clipboard.writeImage(snap.image);
   else if (snap.html) clipboard.writeText(snap.html); // or similar
   else clipboard.writeText(snap.text);
   ```
   …then whenever the source app provided *any* image representation (Word always does, even for trivial text), the restore writes only the image and clobbers everything else. That matches the observed symptom exactly: pure-text clipboards round-trip, anything-with-image clipboards come back as image-only.

**Suggested fix:** use Electron's atomic multi-format write API, which exists for exactly this reason. Roughly:

```js
const snap = {
  text:   clipboard.readText() || undefined,
  html:   clipboard.readHTML() || undefined,
  rtf:    clipboard.readRTF()  || undefined,           // currently missing entirely
  image:  clipboard.readImage().isEmpty() ? undefined : clipboard.readImage(),
  bookmark: clipboard.readBookmark?.()?.url ? clipboard.readBookmark() : undefined,
};
// …paste dictation result…
clipboard.write(snap); // single atomic call preserves ALL captured formats
```

Caveat: Electron's `clipboard.write({...})` does not support every format the macOS pasteboard can carry (e.g. file lists, custom UTIs, Word's `com.microsoft.Word…` private types). A perfect restore for those would need an `NSPasteboard`-level snapshot via a small native helper. But preserving `text + html + rtf + image + bookmark` in the documented Electron API would already eliminate this bug for ~all everyday cases (Word, Outlook, Apple Notes, Mail, Pages, Markdown editors, web pages with formatting).

**Clipboard-inspector dumps**

*Before dictation:* 3 string types available — `text/html` (Word HTML with `xmlns:o`, `xmlns:w`, `<w:WordDocument>`, etc. — full Office namespace bundle), `text/rtf` (Word RTF with Aptos font tables, color tables, latentstyles, themedata, colorschememapping — i.e. a complete Word export), `text/plain` (`This is text in a Word document`). Plus 3 items of corresponding `string` kind. `.files: 0`.

*After dictation:* 1 type available — `Files`, value: empty string. 1 item: kind `file`, type `image/png`, `image.png` (3,314 bytes). `.files: 0` (because the PNG comes through `items` as a file-kind string blob, not as a real `DataTransfer` file). Every text representation is gone.

Happy to attach the full raw clipboard-inspector text dumps from before/after if useful — they're sizable (mostly Word's RTF style table) but I have them.

That said, the report as written is self-contained and the maintainer has everything they need to land the fix.

dpersek:
## Summary
- Preserve rich clipboard representations together when restoring the pre-dictation clipboard.
- Snapshot Electron-supported text, HTML, RTF, and image data and restore multi-format payloads with `clipboard.write(...)`.
- Add focused regression coverage for image + HTML + RTF + plain text restoration.

Fixes #834

## Verification
- `mise x node@24.15.0 -- node --test test/helpers/clipboardRestore.test.js` -> 4 pass, 0 fail
- `mise x node@24.15.0 -- npm run lint` -> passed; emitted an existing module-type warning from `src/eslint.config.js`
- `mise x node@24.15.0 -- npm run typecheck` -> passed
- `git diff --check upstream/main...HEAD` -> clean

## Caveats
- `npm run quality-check` was not used as the PR proof because it reports unrelated pre-existing Prettier warnings in untouched helper files.
- I did not run a full desktop dictation smoke test; this change is covered by the focused clipboard restore regression test.

### DIFF
--- a/src/helpers/clipboard.js
+++ b/src/helpers/clipboard.js
@@ -621,10 +621,7 @@ class ClipboardManager {
       if (html) data.html = html;
     }
 
-    if (
-      typeof clipboard.readRTF === "function" &&
-      (formats.includes("text/rtf") || formats.includes("public.rtf"))
-    ) {
+    if (formats.includes("text/rtf") || formats.includes("public.rtf")) {
       const rtf = clipboard.readRTF();
       if (rtf) data.rtf = rtf;
     }
@@ -651,9 +648,7 @@ class ClipboardManager {
     if (original.type === "formats") {
       clipboard.write(original.data);
     } else if (original.type === "image") {
-      if (!original.data.isEmpty()) clipboard.writeImage(original.data);
-    } else if (original.type === "html") {
-      clipboard.write({ text: original.text, html: original.html });
+      clipboard.writeImage(original.data);
     } else {
       clipboard.writeText(original.data);
     }
--- a/src/helpers/clipboard.js
+++ b/src/helpers/clipboard.js
@@ -611,18 +611,46 @@ class ClipboardManager {
 
   _saveClipboard() {
     const formats = clipboard.availableFormats();
-    if (formats.some((f) => f.startsWith("image/"))) {
-      return { type: "image", data: clipboard.readImage() };
-    } else if (formats.includes("text/html")) {
-      return { type: "html", text: clipboard.readText(), html: clipboard.readHTML() };
-    } else {
-      return { type: "text", data: clipboard.readText() };
+    const data = {};
+
+    const text = clipboard.readText();
+    if (text) data.text = text;
+
+    if (formats.includes("text/html")) {
+      const html = clipboard.readHTML();
+      if (html) data.html = html;
     }
+
+    if (
+      typeof clipboard.readRTF === "function" &&
+      (formats.includes("text/rtf") || formats.includes("public.rtf"))
+    ) {
+      const rtf = clipboard.readRTF();
+      if (rtf) data.rtf = rtf;
+    }
+
+    if (formats.some((f) => f.startsWith("image/"))) {
+      const image = clipboard.readImage();
+      if (image && !image.isEmpty()) data.image = image;
+    }
+
+    const keys = Object.keys(data);
+    if (keys.length === 1 && keys[0] === "image") {
+      return { type: "image", data: data.image };
+    }
+    if (keys.length === 1 && keys[0] === "text") {
+      return { type: "text", data: data.text };
+    }
+    if (keys.length > 0) return { type: "formats", data };
+
+    return { type: "text", data: text };
   }
 
   _restoreClipboard(original) {
     if (!original) return;
-    if (original.type === "image") {
+    if (original.type === "formats") {
+      clipboard.write(original.data);
+    } else if (original.type === "image") {
       if (!original.data.isEmpty()) clipboard.writeImage(original.data);
     } else if (original.type === "html") {
       clipboard.write({ text: original.text, html: original.html });
--- a/test/helpers/clipboardRestore.test.js
+++ b/test/helpers/clipboardRestore.test.js
@@ -5,7 +5,10 @@ const Module = require("node:module");
 const fakeClipboard = {
   text: "",
   html: "",
+  rtf: "",
+  image: null,
   formats: ["text/plain"],
+  writes: [],
   availableFormats() {
     return this.formats;
   },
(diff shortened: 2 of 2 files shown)


gabrielste1n:
thank you sir this is great! appreciate you 🙏 @SyntaxSawdust

gabrielste1n (approved):

Issue #998, opened 2026-06-24
[Linux] "Keep transcription in clipboard" copies nothing when auto-paste is off (renderer navigator.clipboard.writeText blocked while window unfocused)


gleitz:
## Summary

On Linux, when **auto-paste is OFF** and **"Keep transcription in clipboard" is ON**, the transcription never reaches the clipboard — dictation appears to do nothing. The cause is that this code path writes the clipboard from the **renderer** via `navigator.clipboard.writeText()`, which the browser/Electron rejects because the OpenWhispr window is not focused during dictation. The auto-paste-*on* path is unaffected because it writes the clipboard from the **main process** (`write-clipboard` IPC), which is not focus-gated.

## Environment

- OpenWhispr **1.7.3** (Linux .deb)
- Ubuntu 24.04.4 LTS, **X11** session
- Reproduces regardless of transcription backend (seen with cloud + raw fallback)

## Steps to reproduce

1. Settings → Clipboard: **disable** "Automatic pasting", **enable** "Keep transcription in clipboard".
2. Focus any other window (e.g. an editor) and dictate.
3. Check the system clipboard (`xclip -o -selection clipboard`).

**Expected:** clipboard contains the transcription.
**Actual:** clipboard is unchanged; nothing is pasted/copied. The transcription is still saved to history, so it silently "disappears."

## Root cause

In the renderer's dictation-completion handler, the branch is roughly:

```js
const { autoPasteEnabled, keepTranscriptionInClipboard } = settings;
if (autoPasteEnabled) {
  await safePaste(text, { restoreClipboard: !keepTranscriptionInClipboard, ... }); // main-process clipboard write — works
} else {
  keepTranscriptionInClipboard && await navigator.clipboard.writeText(text);        // renderer clipboard write — FAILS when unfocused
}
saveTranscription(...);
```

The Async Clipboard API (`navigator.clipboard.writeText`) requires the document to be focused; during dictation the OpenWhispr window is in the background, so the promise rejects with a `NotAllowedError` ("Document is not focused") and the text is never written. There is no log line on this path, so it fails silently.

## Suggested fix

Use the same main-process bridge the auto-paste path already relies on. The preload already exposes it:

```js
// preload.js
writeClipboard: (text) => ipcRenderer.invoke("write-clipboard", text)
```

So the else-branch becomes:

```js
keepTranscriptionInClipboard && await window.electronAPI.writeClipboard(text);
```

[...]

PR #1022, opened 2026-06-30
fix(clipboard): use main-process write for clipboard-only mode

sochotnicky:
When auto-paste is disabled, the transcription was written via the renderer's navigator.clipboard.writeText. The async Clipboard API requires the document to be focused, but the OpenWhispr window isn't focused during dictation, so in Wayland (e.g. Sway) the write is silently rejected and the clipboard ends up empty.

Route through the existing write-clipboard IPC instead, which uses wl-copy in the main process and works regardless of which window is focused. Harmless on X11/macOS/Windows, where it falls back to Electron's clipboard.writeText.

Note: I only tested on Linux/NixOS on Sway but this works

### DIFF
--- a/src/hooks/useAudioRecording.js
+++ b/src/hooks/useAudioRecording.js
@@ -160,7 +160,12 @@ export const useAudioRecording = (toast, options = {}) => {
               "streaming"
             );
           } else if (keepTranscriptionInClipboard) {
-            await navigator.clipboard.writeText(result.text);
+            // Route through the main process (write-clipboard IPC) rather than the
+            // renderer Clipboard API: navigator.clipboard.writeText requires the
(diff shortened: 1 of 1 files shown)


sochotnicky:
Ah, just noticed https://github.com/OpenWhispr/openwhispr/issues/998 

I believe this is a fix for that (even has the same suggestion)

greatcoat:
Closing — fixed in v1.10.0 via #1979. Clipboard-only delivery and the failed selection-edit fallback both go through the main-process `writeClipboard` bridge now, and history saves independently of the clipboard result.

You identified the root cause first — renderer clipboard writes being focus-gated while the dictation window deliberately never takes focus. Thanks for that.

PR #1024, opened 2026-06-30
Add com.openwhispr.App D-Bus endpoint for global shortcuts on Linux

sochotnicky:
Adds a `com.openwhispr.App` session D-Bus endpoint for global shortcuts on Linux, since Electron's `globalShortcut` can't register hotkeys on many Wayland compositors (e.g. Sway). GNOME already drove shortcuts through this D-Bus interface; this extracts it into one shared, compositor-agnostic service and makes it available everywhere.

- **`dbusToggleService.js`** (new): single owner of `com.openwhispr.App` and its methods — `Toggle`/`ToggleAgent`/`ToggleMeeting`/`ToggleVoiceAgent` (tap) plus `StartDictation`/`StopDictation` (push-to-talk).
- **Universal on Linux:** the endpoint starts on every session (X11 + Wayland). Each desktop layers its own keybinding auto-registration on top and routes through it — GNOME gsettings, Hyprland hyprctl, wlroots manual `dbus-send` binds; KDE/X11 keep their native mechanism with the bus additionally available.
- **`gnomeShortcut.js` / `hyprlandShortcut.js`:** reduced to keybinding registrars (Hyprland's duplicate bus/interface removed).
- Having the bus up does **not** force tap-to-talk — X11 keeps `globalShortcut` + push-to-talk.
- Push-to-talk needs a compositor that can bind key release (Sway `bindsym --release`, Hyprland `bindr`).
- Docs: `CLAUDE.md` §15a.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -88,6 +88,14 @@ OpenWhispr is an Electron-based desktop dictation application that uses whisper.
   - Converts Electron hotkey format to Hyprland bind format (`MODS, key`)
   - Only active on Linux + Wayland + Hyprland (detected via `HYPRLAND_INSTANCE_SIGNATURE`)
   - D-Bus transport: `@homebridge/dbus-native` (pure JavaScript, no native addons)
+- **dbusToggleService.js**: Compositor-agnostic owner of the `com.openwhispr.App`
+  session-bus service and its no-arg methods (`Toggle`, `ToggleAgent`,
+  `ToggleMeeting`, `ToggleVoiceAgent`). Started once per Linux session by
+  `hotkeyManager.ensureDbusEndpoint()`; `gnomeShortcut.js`/`hyprlandShortcut.js`
+  are keybinding registrars that `dbus-send` to it. Single source of truth for the
+  D-Bus interface — no gsettings/compositor logic. D-Bus transport:
+  `@homebridge/dbus-native`. Claims the name with `DO_NOT_QUEUE`; if another owner
+  holds it, logs and returns failure so callers fall through to normal detection.
 - **kdeShortcut.js**: KDE Wayland global shortcut integration
   - Uses D-Bus to communicate with KGlobalAccel for global hotkey registration
   - Registers hotkeys via `setShortcut`/`doRegister` D-Bus calls on the KGlobalAccel interface
@@ -496,7 +504,7 @@ On GNOME Wayland, Electron's `globalShortcut` API doesn't work due to Wayland's 
 
 1. `main.js` enables `GlobalShortcutsPortal` feature flag for Wayland
 2. `hotkeyManager.js` detects GNOME + Wayland and initializes `GnomeShortcutManager`
-3. `gnomeShortcut.js` creates D-Bus service at `com.openwhispr.App`
+3. `gnomeShortcut.js` registers gsettings keybindings that `dbus-send` to the shared `com.openwhispr.App` endpoint (owned by `dbusToggleService.js`)
 4. Shortcuts registered via `gsettings` as custom GNOME keybindings
 5. GNOME triggers `dbus-send` command which calls the D-Bus `Toggle()` method
 
@@ -526,7 +534,7 @@ On Hyprland (wlroots Wayland compositor), Electron's `globalShortcut` API and th
 
 1. `main.js` enables `GlobalShortcutsPortal` feature flag for Wayland (fallback)
 2. `hotkeyManager.js` detects Hyprland + Wayland and initializes `HyprlandShortcutManager`
-3. `hyprlandShortcut.js` creates D-Bus service at `com.openwhispr.App` (same as GNOME)
+3. `hyprlandShortcut.js` registers `hyprctl` binds that `dbus-send` to the shared `com.openwhispr.App` endpoint (owned by `dbusToggleService.js`, same as GNOME)
 4. Shortcuts registered via `hyprctl keyword bind` (runtime keybinding)
 5. Hyprland triggers `dbus-send` command which calls the D-Bus `Toggle()` method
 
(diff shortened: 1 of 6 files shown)


sochotnicky:
I think in theory the whole Hyprland could be changed as well since it seems to support Global shortcuts: https://wiki.hypr.land/Configuring/Basics/Binds/#global-keybinds

But I left that alone

xAlcahest:
Nice refactor. Pulling bus ownership into a single `DBusToggleService` and dropping the duplicate `requestName` that GNOME and Hyprland each did is a real cleanup, and cutting both managers down to plain keybinding registrars reads well.

A few minor things, none blocking:

`dbusToggleService.js` ignores the `requestName` reply:

```js
await this.bus.requestName(DBUS_SERVICE_NAME, 0);
```

In dbus-next the call resolves with the numeric reply code and only rejects on an actual D-Bus error, so a non-primary result (`IN_QUEUE`) still looks like success: it exports the interface, logs "initialized successfully", and returns `true`. The single-instance lock in `main.js` makes a second owner unreachable in a normal install, so this is mostly defensive. The one concrete gap is that `CLAUDE.md` §15a says "If the bus name can't be claimed, it logs and falls through to normal detection", but the code never detects that case. Cheap to make real if you want the doc to hold: pass `NameFlag.DO_NOT_QUEUE` and treat any reply other than `RequestNameReply.PRIMARY_OWNER` / `ALREADY_OWNER` as failure (both are exported from dbus-next).

`ensureDbusEndpoint` returns early when `dbusToggleService` already exists, so a second `initializeHotkey` with a different dictation callback keeps the old one. The callback is stable today so it's harmless, but a `setDictationCallback(callback)` before the early return would close the gap.

`startWindowsPushToTalk` / `handleWindowsPushKeyUp` are now the generic PTT primitives (they only touch `winPushState` and send IPC, no Windows API), so the name reads a bit off on the D-Bus path. Not worth churning here.

Otherwise it holds up: no dangling refs after the `GNOME_NATIVE_SLOTS` rename, the `DBUS_*` constants are still used to build the `dbus-send` commands, and the wiring through `registerSlot` + `setDbusToggleCallbacks` lines up.

sochotnicky:
Eh, in the meantime there seems to have been a change of dbus library to homebridge/dbus-native? I'll get this reworked

xAlcahest:
Rebased this onto current main to clear the conflicts. The one deliberate change beyond conflict resolution: main has since added a `translation` hotkey slot, so I wired `ToggleTranslation` into the shared `DBusToggleService` alongside the other Toggle* methods, otherwise the translation shortcut would have silently fallen back to plain dictation on the D-Bus path. Nothing else changed intentionally. If you have local work on this branch, pull before pushing.

xAlcahest (changes requested):
I went through this one carefully since it touches the startup path and the hotkey mechanism across all of Linux. Having a single owner for `com.openwhispr.App` is the right call, and collapsing the duplicate bus in `gnomeShortcut.js`/`hyprlandShortcut.js` is a clear improvement. Two things I think need fixing before this lands, though.

## Blockers

**1. `await this._requestName()` can hang forever, and it sits in the startup path**

`dbusToggleService.js:71` is the first time hotkey init awaits a D-Bus round trip. In `@homebridge/dbus-native`, pending reply callbacks live in `bus.cookies` (`lib/bus.js:23`) and are only removed when a `methodReturn` with a matching `replySerial` arrives (`lib/bus.js:148-150`). `bus.js` subscribes only to the connection's `message` event (`lib/bus.js:91`), never to `error` or `end`, so nothing ever settles those pending callbacks. A stream error or a failed handshake just does `self.emit('error', err)` (`index.js:79` and `index.js:99`), and the `connection.on("error")` listener added at line 64 only logs.

So with a stale `DBUS_SESSION_BUS_ADDRESS` or a dead socket, exactly the case the comment in the code calls out, the promise never settles. `start()` never returns, so neither do `ensureDbusEndpoint` and `initializeHotkey`, and since that chain is awaited inside `createMainWindow()` the app stalls before the control panel, the tray, and slot registration. The old `initDBusService` used fire-and-forget `requestName(name, 0)` and couldn't hang, so this is a new regression.

Suggestion: a timeout on `_requestName()` (two or three seconds is plenty), plus settling the promise on the connection's `error` and `end`.

**2. `useDbus = true` disables the native evdev listener on wlroots**

`hotkeyManager.js:843` sets `useDbus = true`, which feeds into `isUsingNativeShortcut()` (line 1466). That flag makes `windowManager.js:535` pass `[]` to `linuxKeyManager`, so `linux-key-listener` never spawns. The immediate `return` on the next line also skips `loadSavedHotkeyOrDefault`, leaving the `dictation` slot empty.

The catch is that on Sway/river/dwl that evdev listener is the only hotkey mechanism that works today. `getEffectiveDefaultHotkey()` (line 1048) returns `Control+Super` on non-GNOME Wayland, `isModifierOnlyHotkey` classifies it as modifier-only, and `getNativeListenerKeys` (line 380) routes exactly those to the native listener regardless of compositor. Before this PR those sessions fell through to `loadSavedHotkeyOrDefault` and the listener started. After it, anyone upgrading ends up with no working hotkey at all until they hand-write compositor binds. For the users this PR targets that's a downgrade.

## Should be fixed, not blocking

- `hotkeyManager.js:844`: the early return skips both `this.currentHotkey` and `notifyActiveHotkey()`, which every other native path sets. As a result `getCurrentHotkey()` stays `null`, which makes `effectiveHotkey` null in `set-hotkey-listening-mode` (`ipcHandlers.js:3019`), effectively disables cross-slot conflict detection in `updateHotkey` (`_findSlotConflict` can never see a dictation hotkey), and leaves the renderer without the `dictation-key-active` event that `settingsStore.ts:2330` relies on to correct the startup default.
- `hotkeyManager.js:1363`: `updateHotkey` branches on `useGnome`, `useHyprland` and `useKDE` but not `useDbus`, so on wlroots it falls through to `setupShortcuts` and `globalShortcut.register`. Since `dictation` isn't in `DBUS_TOGGLE_SLOTS`, nothing in that path routes to the shared endpoint either. The user changing their hotkey in Settings gets a globalShortcut registration that can never fire on wlroots, and a success or failure message that has nothing to do with the compositor bind they actually need to edit.
- `hotkeyManager.js:716`: `initializeGnomeShortcuts` now returns `false` when `!this.dbusToggleService`. With `DO_NOT_QUEUE` and the strict reply check, any instance that loses the name drops to `globalShortcut`. On GNOME Wayland that registers nothing usable; on GNOME X11 it's worse, since the losing instance registers real accelerators and can steal the hotkey from the owner. Concrete case: a dev build running alongside the installed build, which have separate single-instance locks but share `com.openwhispr.App`. The previous code ignored the reply and always carried on.
- `hotkeyManager.js:217`: the D-Bus callbacks are wired before the GNOME/KDE/globalShortcut registration is attempted, so a slot whose registration then fails still has its `Toggle*` method live. `unregisterSlot` has no `useDbus` branch and never clears them either, so clearing the agent hotkey in Settings leaves `ToggleAgent` responding for the rest of the session.

## Minor

- `dbusToggleService.js:173`: a mid-session connection drop only reaches the debug log. The service stops responding, nothing reconnects, and the user sees nothing, so every compositor bind they set up silently stops working. Worth surfacing, or at least attempting one reconnect.
- `dbusToggleService.js:88`: `_requestName` rejects with whatever dbus-native hands it, which for D-Bus error replies isn't always an `Error` (it can be a string or an array from the body), so the `err.message` in the catch at line 89 can print `undefined`.
- `main.js:941`: `startDictationCallback` calls `windowManager.startWindowsPushToTalk()` with no key, leaving `winPushState.key` undefined. The key-match guard in `handleWindowsPushKeyUp` (`windowManager.js:431`) only applies when `this.winPushState.key` is truthy, so on a session where both the endpoint and the native listener are active, releasing an unrelated hotkey stops a recording that D-Bus started.
- `Toggle` is a silent no-op when the activation mode is push: `createHotkeyCallback` returns early on Linux in that case (`windowManager.js:244`), and `supportsPushToTalk` is computed as `linuxKeyManager.isAvailable()` on Linux (`ipcHandlers.js:3172`), which is just "the bundled binary exists". So the auto-switch to tap at `SettingsPage.tsx:1212` doesn't fire on wlroots even though `isUsingNativeShortcut` is now true. A user left in push mode who binds the documented `Toggle` gets nothing, and the native listener is disabled in the meantime per blocker 2.
- `shouldAutoUseDbus()` relies on `GnomeShortcutManager.isWayland()`, which only checks `XDG_SESSION_TYPE`. Not this PR's doing, the method predates it, but it limits the reach of the fix: a compositor started from a TTY keeps the logind session type at `tty` unless the launch script sets the variable or uses the logind libseat backend, which is a known problem (hyprwm/Hyprland#9384). Those are exactly the setups most likely to need this feature. `WAYLAND_DISPLAY` would be the more reliable signal.

## Tests

`shouldAutoUseDbus()` is a pure static and would test nicely in the style of the existing `test/helpers/` suites, covering the combinations of `XDG_SESSION_TYPE` and `XDG_CURRENT_DESKTOP`. The `_requestName` timeout would also be testable with a stubbed bus. As it stands the PR adds no tests.

## Things I checked that look fine

The ordering between `setDbusToggleCallbacks` and `initializeHotkey` is correct, window creation is awaited first, and `ensureDbusEndpoint` passes the dictation callback into `start({ dictation: callback })` so `Toggle` is wired. The `RequestName` constants match the spec (PRIMARY_OWNER 1, ALREADY_OWNER 4, DO_NOT_QUEUE 0x4). The `DBUS_*` constants are still defined in `hyprlandShortcut.js` after the removal. No callers of `initDBusService` remain. `start-dictation` and `stop-dictation` are wired unconditionally in the renderer, so D-Bus push-to-talk doesn't depend on the UI activation mode. The double `close()` on the failure path is idempotent.

The direction here is right, especially unifying the bus. It's mainly the two blockers I'd want addressed; the rest is polish.

PR #1026, opened 2026-06-30
fix(linux): stop overlay focus theft on wlroots (Sway)

### LINKED ISSUE #719: Feature: Voice recording icon should not steal focus
When using the `Auto-hide when idle` setting, the creation of the icon window steals focus from the current window. This can cause the auto-paste to fail as the text box is no longer focused. I think it would be a decent idea to force the icon window to be unfocusable (if this is something that can be done).

If you are also experiencing this on Sway/i3, the following config line fixes this:
```
no_focus [instance="open-whispr"]
```

I noticed this issue describing the same problem specifically for Claude Code on macOS - but the problem can be applied in the general case: https://github.com/OpenWhispr/openwhispr/issues/668

Thanks! Keep up the good work on OpenWhispr!

sochotnicky:
On wlroots compositors the dictation/preview/notification overlays steal focus on map. Root cause: the overlays are created focusable:false, which makes Electron map them as override-redirect XWayland windows. wlroots' XWayland focuses override-redirect windows on map unless their _NET_WM_WINDOW_TYPE is in its no-focus set (notification/tooltip/utility/splash/menus) — and the overlays are TOOLBAR/NORMAL, which are not. Mutter/KWin honor focusable:false directly, so GNOME/KDE never had the bug; forcing XWayland (the prior fix) doesn't help here because it's already XWayland.

Map the non-interactive overlays (dictation mic, transcription preview, meeting notification) as type "notification" on wlroots, which is in wlroots' no-focus set — clicks still work, keyboard focus isn't grabbed. GNOME/KDE keep their existing types (no regression). The agent overlay is left focus-capable since it needs keyboard input for chat.

Verified against xprop on Sway: preview was override-redirect + TOOLBAR (focus stolen); no_focus rules and get_tree don't apply to override-redirect windows.

Fixes the focus stealing on Sway for me at least and should fix https://github.com/OpenWhispr/openwhispr/issues/719

### DIFF
--- a/src/helpers/windowConfig.js
+++ b/src/helpers/windowConfig.js
@@ -10,13 +10,29 @@ const isKDEWayland =
   process.env.XDG_SESSION_TYPE === "wayland" &&
   /kde/i.test(process.env.XDG_CURRENT_DESKTOP || "");
 
+// wlroots compositors (Sway, Hyprland, river, …). Our overlays are created with
+// focusable:false, which makes Electron map them as override-redirect XWayland
+// windows. wlroots' XWayland focuses override-redirect windows on map *unless*
+// their _NET_WM_WINDOW_TYPE is in its no-focus set (notification, tooltip,
+// utility, splash, menus). TOOLBAR/NORMAL are NOT in that set, so the overlays
+// steal focus on wlroots — but not on GNOME/KDE, whose WMs honor focusable:false
+// directly. So on wlroots we map the non-interactive overlays as "notification".
+const isWlrootsWayland =
+  process.platform === "linux" &&
+  process.env.XDG_SESSION_TYPE === "wayland" &&
+  (/sway|hyprland|wayfire|river|dwl|labwc/i.test(process.env.XDG_CURRENT_DESKTOP || "") ||
+    !!process.env.SWAYSOCK ||
+    !!process.env.HYPRLAND_INSTANCE_SIGNATURE);
+
 const MAIN_OVERLAY_TYPE =
   process.platform === "darwin"
     ? "panel"
     : process.platform === "linux"
-      ? isGnomeWayland || isKDEWayland
-        ? "normal"
-        : "toolbar"
+      ? isWlrootsWayland
+        ? "notification"
+        : isGnomeWayland || isKDEWayland
+          ? "normal"
+          : "toolbar"
       : "normal";
 
 const FLOATING_OVERLAY_TYPE =
@@ -27,6 +43,13 @@ const FLOATING_OVERLAY_TYPE =
         ? "normal"
         : "toolbar"
       : "normal";
+
+// Type for non-interactive, never-focus overlays (transcription preview, meeting
+// notification). On wlroots, "notification" keeps the override-redirect window
+// from stealing focus (see isWlrootsWayland above); elsewhere it matches the
+// regular floating overlay type. NOT used for the agent overlay, which needs
+// keyboard focus for chat input.
+const NOFOCUS_OVERLAY_TYPE = isWlrootsWayland ? "notification" : FLOATING_OVERLAY_TYPE;
 
 const WINDOW_SIZES = {
   BASE: { width: 96, height: 96 },
@@ -117,7 +140,7 @@ const NOTIFICATION_WINDOW_CONFIG = {
     sandbox: true,
   },
   visibleOnAllWorkspaces: process.platform !== "win32",
-  type: FLOATING_OVERLAY_TYPE,
+  type: NOFOCUS_OVERLAY_TYPE,
 };
 
 const TRANSCRIPTION_PREVIEW_SIZE_LIMITS = {
@@ -148,7 +171,7 @@ const TRANSCRIPTION_PREVIEW_CONFIG = {
     sandbox: true,
   },
   visibleOnAllWorkspaces: process.platform !== "win32",
-  type: FLOATING_OVERLAY_TYPE,
+  type: NOFOCUS_OVERLAY_TYPE,
 };
 
 class WindowPositionUtil {
--- a/src/helpers/windowConfig.js
+++ b/src/helpers/windowConfig.js
@@ -10,25 +10,30 @@ const isKDEWayland =
   process.env.XDG_SESSION_TYPE === "wayland" &&
   /kde/i.test(process.env.XDG_CURRENT_DESKTOP || "");
 
-// wlroots compositors (Sway, Hyprland, river, …). Our overlays are created with
-// focusable:false, which makes Electron map them as override-redirect XWayland
-// windows. wlroots' XWayland focuses override-redirect windows on map *unless*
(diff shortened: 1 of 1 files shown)


xAlcahest:
Really nice to see someone chase this one down, the override-redirect + window-type angle is something I'd probably have missed.

One thing I wasn't sure about, figured I'd just ask. The `SWAYSOCK` / `HYPRLAND_INSTANCE_SIGNATURE` checks are gated behind `XDG_SESSION_TYPE === "wayland"`. From what I've read that var comes from logind at login, so a session started straight from a TTY can come up as `tty` instead of `wayland` (Hyprland ran into it in [#9384](https://github.com/hyprwm/Hyprland/issues/9384), and Sway doesn't seem to set it itself). If that's right, `XDG_CURRENT_DESKTOP` is usually empty in that setup too, so I *think* detection could miss even with `SWAYSOCK` sitting right there. Does that match what you see, or am I overthinking a corner case? If it's real, maybe those two vars could count on their own since they only exist when the compositor's actually up, but you know these setups better than me.

Separate small question: the remapped windows have clickable bits (preview copy/dismiss, meeting start/dismiss), so do `notification` windows still take clicks on the other compositors the regex now matches, or has Sway been the only one you could test on? Not a blocker, just curious.

Either way, good find, thanks for chasing it.

sochotnicky:
@xAlcahest XDG_SESSION_TYPE is set by wlroots itself (https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/backend/session/session.c#L88).

As for XDG_CURRENT_DESKTOP - when I run sway from tty without login manager I still get that set to sway. But for me distro takes care of that (https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/sw/sway/package.nix#L37). In general Sway has that documented (https://github.com/swaywm/sway/wiki#xdg_current_desktop-environment-variable-is-not-being-set) and if people run things from tty they should know what they're doing?

Re clickable bits - I've confirmed they can still be clicked as usual on sway. But you were right to ask! I checked and in hyprland I could not click on "copy" in the notification overlay anymore in my version. This gets into ugly "fix one, break others" territory. Perhaps I rework this to have a list of desktop environs which use NO_FOCUS type and we just add sway in this PR and as people complain you can easily add more?

sochotnicky:
OK, I've narrowed this down to just Sway so hyprland won't get messed up

xAlcahest:
The mechanism checks out. I read the wlroots xwm source and `notification` is genuinely in the no-focus set that `wlr_xwayland_or_surface_wants_focus` filters on, while TOOLBAR and NORMAL are not, so override-redirect overlays really do get focused on map today. Dragging is unaffected since `dragManager` uses `setPosition` rather than `_NET_WM_MOVERESIZE`. My comments are all about the detection predicate and about which windows it ends up retyping.

**The predicate misses the case its own fallback was added for.** `isNoFocusOverlayCompositor` is gated behind `XDG_SESSION_TYPE === "wayland"`, and that variable comes from pam_systemd at login, so a plain `exec sway` from a TTY leaves it at `tty` unless you run sway under `LIBSEAT_BACKEND=logind` or a wrapper that exports it. In that session `shouldForceXWayland()` in `src/helpers/xwayland.js` also returns false, but Electron 41 defaults to `--ozone-platform-hint=auto`, which decides by reading the same variable and falls back to X11. So the app still runs under XWayland, the overlays are still override-redirect TOOLBAR windows, the focus theft still reproduces, and the new code does nothing. The gate for the fix and the gate for XWayland read the same env var, so they miss that session together.

Following from that, the `|| !!process.env.SWAYSOCK` fallback can almost never fire. On a display-manager Sway session `XDG_CURRENT_DESKTOP` already contains "sway" and the first test matches. On a TTY session the `XDG_SESSION_TYPE` gate rejects everything before `SWAYSOCK` is ever read, which is exactly the case the fallback was meant to cover. The only window left is a session where `XDG_SESSION_TYPE=wayland` but `XDG_CURRENT_DESKTOP` has no "sway" in it.

What that fallback does buy is a false positive. `SWAYSOCK` outlives its session when it gets pushed into the systemd user manager, which the sway wiki recommends via `dbus-update-activation-environment ... SWAYSOCK`, and it survives logout with lingering enabled. Stale tmux and screen servers keep it too. A later Hyprland session started from one of those inherits `SWAYSOCK`, `XDG_SESSION_TYPE=wayland`, and lands on exactly the configuration your comment at lines 21-26 documents as broken, where the preview and meeting buttons stop responding. And since the new branch is evaluated before `isGnomeWayland || isKDEWayland` in `MAIN_OVERLAY_TYPE`, the same stale variable on a KDE session silently downgrades it from `normal` back to `notification`. To be clear this needs lingering or a stale multiplexer, so it is hardening rather than a live bug, but it is cheap to close: match `XDG_CURRENT_DESKTOP` when it is non-empty, accept `WAYLAND_DISPLAY` as a second signal the way `clipboard.js:34-37` and `ensureYdotool.js:196-198` already do in this repo, bail out on `HYPRLAND_INSTANCE_SIGNATURE`, and order the wlroots check after the GNOME/KDE one.

**`NOTIFICATION_WINDOW_CONFIG` is shared by two windows.** `windowManager.js:1186` is the meeting prompt, but `:1283` is the update-available prompt, and that one has real action buttons at `UpdateNotificationOverlay.tsx:78` and `:112`. The "non-interactive overlays" wording in the description and in the comment at lines 52-56 undersells what is being retyped, and any compositor added to the allowlist later inherits the same exposure. Worth a line in the comment and, if you have a Sway box handy, a quick check that the update prompt's buttons still work.

**The dictation panel is the most interactive overlay of the three.** `src/App.jsx:283-296` registers a document keydown handler that closes the command menu or the panel on Escape, and on Sway that currently works precisely because the overlay steals focus. Once the window is in the no-focus set that handler stops firing there, matching what already happens on GNOME, KDE, macOS and Windows thanks to `focusable: false`. There is no text input in the panel and the click-outside handler is mouse-driven, so the trade looks fine to me, but the description shouldn't call that overlay non-interactive since it carries the mic button, cancel, drag and the command menu.

One non-blocking note. This file now has three levels of nested platform ternaries and no test coverage, while the repo's own precedent for exactly this kind of branching is a pure resolver with injected env, `dockPolicy.js` alongside `test/helpers/dockPolicy.test.js`. A `resolveOverlayTypes(env, platform)` would make the Sway, Hyprland and KDE matrix assertable instead of verified by hand, and would pin the ordering question above.

Last thing, and it's the one claim in here nobody can check by reading: the allowlist rests on the Sway-works / Hyprland-breaks split. Do you have a note on how that was tested? If it's a Hyprland version or a specific button that broke, dropping it in the comment would help whoever adds the next compositor.

gabrielste1n:
Thanks for tracking down the wlroots/XWayland window-type mechanism and validating the behavior on Sway. This has now been addressed by #1718, which incorporated this approach with hardened session detection and regression coverage, and #719 is closed.\n\nClosing this PR as superseded.

PR #1027, opened 2026-07-01
fix(meeting): notes recording chat selection bug

B3rK-3:
When meeting transcription is running in Notes, each incoming transcript update can disrupt text selection in the embedded chat.                                                                     
                                                                                                                                                                                                      
 Repro:                                                                                                                                                                                               
 1. Open a note.                                                                                                                                                                                      
 2. Start meeting recording/transcription.                                                                                                                                                            
 3. Open that same note’s embedded chat.                                                                                                                                                              
 4. Select text from an assistant response.                                                                                                                                                           
 5. Wait for a transcript partial/final line to arrive.                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                      

Fix:
https://github.com/user-attachments/assets/cd6ac331-4878-4f18-8edc-b4404a1cb858

1.7.3:
https://github.com/user-attachments/assets/d498d0bc-2504-49b5-93bb-37cf54f09c19

                                                                                                                                             
 Expected:                                                                                                                                                                                            
 - Transcript updates should update only the transcript UI.                                                                                                                                           
 - They should not re-render/reconcile the note editor or embedded chat DOM.                                                                                                                          
 - Text selection in chat should remain stable while transcription continues.                                                                                                                         
                                                                                                                                                                                                      
 Root cause:                                                                                                                                                                                          
 - Live transcript state was subscribed too high in the Notes view/editor tree.                                                                                                                       
 - Every transcript update caused a larger React subtree to re-render, including chat/editor DOM.                                                                                                     
 - Browser selection was lost/remapped during reconciliation.                                                                                                                                         
                                                                                                                                                                                                      
 Fix:                                                                                                                                                                                                 
 - Scope live transcript subscriptions to the transcript component only.                                                                                                                              
 - Keep PersonalNotesView and the embedded chat/editor outside the live transcript update path.                                                                                                       
 - Read transcript state imperatively only for persistence/action execution.


gabrielste1n (approved):
Thank you @B3rK-3 — great find and exactly the right fix. Scoping the live transcript subscriptions to a leaf component is the correct pattern here, and the repro videos made this easy to verify.

I rebased onto main (reconciling with #1119, which had touched the same action-transcript code) and layered a few refinements on top:
- Actions button now uses a narrow boolean selector so it enables during recording without re-subscribing to transcript text
- Diarization-complete fallback is gated to the recorded note, so store segments can't leak into a different note
- Deduplicated the speaker list logic into a shared `buildKnownSpeakers` helper and derived the wrapper's props from `MeetingTranscriptChat` itself

Typecheck, lint, and full test suite pass. Merging — thanks again for the contribution!

PR #1030, opened 2026-07-01
fix(media): resume playback when recording stops, not after transcription

### LINKED ISSUE #947: BUG: Auto-pause feature for music - music remains paused until the transcription is finished
The auto-pause feature for music behaves in a way that, when I click to speak and transcribe, the music pauses. However, when I release the button, the music remains paused until the transcription is finished. Since I sometimes speak for several minutes, this takes a very long time for the music to resume, making the feature practically useless. The expected behavior would be for the music to resume immediately after I finish speaking, while the transcription process continues in the background.

xAlcahest:
## Summary

With "pause media on dictation" on, playback paused at record start but only resumed inside `onTranscriptionComplete`, so a long dictation left the music paused for the entire transcription (minutes). The fix moves the resume to the moment recording stops: it fires on the `isRecording` false edge in `onStateChange` (`src/hooks/useAudioRecording.js`), gated by a new `wasRecordingRef` so it runs once at the stop instant, and the late `onTranscriptionComplete` resume is removed. Music comes back on button release while transcription keeps running in the background. That edge is the one point every stop path passes through (normal stop, streaming stop, cancel, empty-recording drop, connection-loss auto-stop), and `mediaPlayer.resumeMedia()` is idempotent, so the remaining defensive resume calls stay safe.

Fixes #947

## Changes

- src/hooks/useAudioRecording.js: resume media on the recording-stopped edge in `onStateChange` via a new `wasRecordingRef`; drop the resume from `onTranscriptionComplete`.
- src/helpers/audioManager.js: update the stale silence-path comment that referred to the removed resume.

### DIFF
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -3096,7 +3096,7 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
         "streaming"
       );
     } else {
-      // Silence: still fire callback so media playback resumes.
+      // Silence: still fire callback to dismiss the preview and show the no-audio toast.
       this.onTranscriptionComplete?.({ success: true, text: "" });
     }
 
--- a/src/hooks/useAudioRecording.js
+++ b/src/hooks/useAudioRecording.js
@@ -18,6 +18,7 @@ export const useAudioRecording = (toast, options = {}) => {
   const audioManagerRef = useRef(null);
   const startLockRef = useRef(false);
   const stopLockRef = useRef(false);
+  const wasRecordingRef = useRef(false);
   const { onToggle } = options;
 
   const performStartRecording = useCallback(async ({ voiceAgentRequested = false } = {}) => {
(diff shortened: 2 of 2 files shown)


gabrielste1n (approved):
Verified the two load-bearing claims against the code:

- Every stop path in audioManager.js fires the `isRecording: false` edge (batch onstop, degenerate-recording drop, silence gate, cancel, cancel-processing, streaming stop/completion, and all three streaming start-failure paths), and the `wasRecordingRef` gate yields exactly one resume per recording.
- `resumeMedia()` is pause-state-tracked on all three platforms (`_didPause` / `_pausedPlayers` / `_pausedWinApps`), so the defensive resumes and the false edges that fire before any pause happened are safe no-ops.

Streaming UX checks out too: the false edge fires at the start of `stopStreamingRecording`, before inline finalization/reasoning, so playback resumes at button release.

One pre-existing (not introduced here) race worth a follow-up issue: a quick tap during streaming startup defers the stop inside `startStreamingRecording`, so `performStartRecording` pauses media after the recording already ended, leaving it stuck paused. Re-checking `getState().isRecording` before pausing would close it.

PR #1032, opened 2026-07-01
fix(transcription): reload local GPU model on wake from sleep

### LINKED ISSUE #766: Feature: Option to leave local transcription model loaded in VRAM while idle

xAlcahest:
## Summary

The literal ask in #766 (keep the model in VRAM while idle) is already how it works: the local whisper-server holds the model resident the whole time it sits idle. The real problem, surfaced later in the thread, is sleep. On Windows the GPU powers down during S3 sleep and the model is evicted from VRAM; the whisper-server process survives but its CUDA context is dead, and since `start()` in `whisperServer.js` short-circuits when the server is already up with the same model, nothing reloads it. So the first dictation after waking pays the full model reload and clips the opening words.

This hooks the existing `powerMonitor` "resume" handler in `main.js`. A few seconds after wake (enough for the NVIDIA driver to reinitialize) it calls a new `whisperManager.onWakeFromSleep()`, which restarts the local server so the model is back in VRAM before the user dictates. It only fires for a running local CUDA server (remote and CPU servers keep their model across sleep, and Parakeet is CPU only), replays the last `start()` options so the reloaded server keeps the VAD/thread signature the next dictation expects, and skips when a transcription or another re-warm is already in flight.

Fixes #766

## Changes

- src/helpers/whisper.js: add `onWakeFromSleep()` to restart a running local CUDA whisper-server, replaying the last start options; add the `shouldRewarmOnWake()` gate; track `_transcribing` and `_rewarmInFlight`; wrap `transcribeViaServer` so it flags in-flight work around the extracted `_runServerTranscription`.
- src/helpers/whisperServer.js: remember the last `start()` options so the wake reload matches the next dictation's VAD/thread signature.
- main.js: on `powerMonitor` "resume", schedule a debounced, delayed `onWakeFromSleep()` and clear the timer on teardown.
- test/helpers/whisperWakeRewarm.test.js: unit tests for the `shouldRewarmOnWake()` gate.

### DIFF
--- a/main.js
+++ b/main.js
@@ -318,6 +318,8 @@ let ipcHandlers = null;
 let cliBridge = null;
 let globeKeyAlertShown = false;
 let authBridgeServer = null;
+const WHISPER_WAKE_REWARM_DELAY_MS = 3000;
+let wakeRewarmTimer = null;
 
 function parseAuthBridgePort() {
   const raw = (process.env.OPENWHISPR_AUTH_BRIDGE_PORT || "").trim();
@@ -932,6 +934,14 @@ async function startApp() {
     if (googleCalendarManager) {
       googleCalendarManager.onWakeFromSleep();
     }
+    // Sleep evicts the local GPU model from VRAM; reload it once the driver settles. See #766.
+    if (wakeRewarmTimer) clearTimeout(wakeRewarmTimer);
+    wakeRewarmTimer = setTimeout(() => {
+      wakeRewarmTimer = null;
+      whisperManager?.onWakeFromSleep().catch((err) => {
+        debugLogger.debug("whisper wake re-warm error (non-fatal)", { error: err.message });
+      });
+    }, WHISPER_WAKE_REWARM_DELAY_MS);
   });
 
   // Non-blocking server pre-warming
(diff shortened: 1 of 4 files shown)


gabrielste1n (approved):
Reviewed and verified locally: the no-op guard in whisperServer.js start() confirms the diagnosis, the CUDA-only gating keeps remote/CPU/macOS untouched, and the concurrent stop/start paths converge safely on startupPromise. Full test suite passes locally (139/139) including the 6 new tests. Reporter confirmed the fix on Windows/CUDA. Thanks for the careful investigation on #766.

PR #1038, opened 2026-07-02
Fix macOS clipboard restore delay

### LINKED ISSUE #1023: [macOS] Dictation pastes twice / previous clipboard not available — 1.7.3 raised the restore delay 450ms → 2000ms
**Describe the bug**

On macOS, with auto-paste on and "Keep transcription in clipboard" **off**, OpenWhispr holds the transcription in the system clipboard for **2 full seconds** after pasting, before restoring the user's original clipboard contents. Any manual paste during that 2-second window pastes the transcription again instead of the user's original clipboard data.

Concretely: I copy a link, dictate (the transcription auto-pastes into my prompt), then immediately ⌘V to paste my link after it — and I get the transcribed text a second time instead of my link. The dictated text ends up in the prompt twice and my copied link is effectively unavailable until the restore fires.

This is a **regression in 1.7.3**. The identical workflow worked in 1.7.2.

**Root cause (confirmed in source)**

`src/helpers/clipboard.js`, the `RESTORE_DELAYS` constant changed between releases:

| | v1.7.2 | v1.7.3 |
|---|---|---|
| `RESTORE_DELAYS.darwin` | `450` | `2000` |
| restore path | unconditional `setTimeout(() => this._restoreClipboard(originalClipboard), RESTORE_DELAYS.darwin)` | `_restoreClipboardAfterDelay()` — waits 2000ms, then **skips restore entirely** if `clipboard.readText() !== expectedText` ("Skipping clipboard restore because clipboard changed") |

The 4.4× longer delay widens the window in which the clipboard still holds the transcription from ~0.45s to 2s — long enough that a normal "dictate, then paste my link" interaction lands inside it. The new `expectedText` guard adds a second failure mode: if anything changes the clipboard during the 2s window, the original is **never** restored at all.

This is purely a local timing issue — it reproduces with a fully local pipeline (local Whisper STT + local LLM cleanup, zero network), so cloud/transcription latency is not involved.

**To Reproduce**

1. Settings → ensure auto-paste is **on** and "Keep transcription in clipboard" is **off**.
2. Copy a URL (or any plain text) to the clipboard — e.g. a link you want to paste alongside dictated text.
3. Focus a text field. Trigger dictation and say a sentence. Let it auto-paste.
4. Immediately (within ~2s) press ⌘V to paste your copied link after the dictated text.
5. **Result:** the dictated transcription is pasted again instead of your link.

**Expected behavior**

After the transcription is pasted, the original clipboard contents should be available again quickly enough that a normal follow-up paste returns the user's data, not the transcription — as in 1.7.2 (≤450ms).

**Actual behavior**

For 2 seconds after dictation, ⌘V yields the transcription. If the clipboard is touched during that window, the original is never restored.

**Suggested fix**

- Lower `RESTORE_DELAYS.darwin` back toward the 1.7.2 value (~450ms), or make the restore adaptive rather than a fixed 2000ms wait; and/or
- Expose the paste/restore delay as a user setting (there is currently no user-facing delay control and no type-out output mode — clipboard paste is the only path on macOS); and/or
- Reconsider the `expectedText` skip-guard so a legitimate restore isn't silently dropped.

**Additional context**

- **Not a duplicate of #834** (rich-format HTML/RTF→PNG loss; that report notes plain text *survives* — here plain text is the thing lost, via timing).
- **PR #1020 does not address this.** It's the open fix for #834 and touches the same code region (`_restoreClipboard`), but it only changes *which formats* are restored — it leaves the 2000ms `RESTORE_DELAYS.darwin` delay and the `expectedText` skip-guard untouched. So this timing regression remains even with #1020 merged.
- **PR #335** (custom `com.openwhispr.transcription` pasteboard type) is a partial mitigation for clipboard-manager users — it lets Maccy/Flycut ignore OpenWhispr's transient writes — but it does not fix the underlying restore-window timing either.
- Related area: #641, #305.

**Desktop**

- OS: macOS 26.5.2, Apple Silicon
- App: OpenWhispr 1.7.3 (arm64)
- Pipeline: local Whisper STT + local LLM cleanup (no cloud) — confirms the issue is local timing, not network latency
- Regression from: 1.7.2 (same workflow worked)

dpersek:
## Summary

- Restore the macOS clipboard after 450ms instead of 2000ms after a successful paste.
- Add coverage for both the fast-paste path and the osascript fallback to keep the shared macOS restore delay at 450ms.

## Why

Fixes #1023.

OpenWhispr 1.7.3 increased the macOS clipboard restore delay to 2000ms. That leaves the dictated text on the clipboard for longer after paste, which can cause follow-up paste behavior to use the dictated text rather than the user's previous clipboard content. This restores the shorter macOS delay while keeping the existing expected-text guard that avoids overwriting a newer user clipboard write.

## Tests

- `mise x node@24 -- node --test test/helpers/clipboardRestore.test.js` passed, 5/5.
- `mise x node@24 -- npm run lint` passed.
- `mise x node@24 -- npm run typecheck` passed.
- `mise x node@24 -- npx prettier --check src/helpers/clipboard.js test/helpers/clipboardRestore.test.js` passed.
- `git diff --check upstream/main` passed.
- `mise x node@24 -- npm run format:check` ran but failed on existing unrelated formatting issues in `helpers/audioManager.js`, `helpers/speakerEmbeddings.js`, and `helpers/textEditMonitor.js`; the two changed files pass Prettier directly.

## Scope / non-goals

- Scope is limited to the macOS clipboard restore delay and focused unit coverage for the macOS paste success paths.
- No changes to Windows or Linux clipboard timing.
- No changes to paste behavior beyond the restore delay value.

## Caveats

- I did not run a manual macOS dictation smoke test.
- PR #1020 also touches `src/helpers/clipboard.js` and `test/helpers/clipboardRestore.test.js` for #834 rich clipboard restoration. This PR fixes a separate #1023 timing regression; if #1020 lands first, I will rebase this branch before requesting review.

### DIFF
--- a/src/helpers/clipboard.js
+++ b/src/helpers/clipboard.js
@@ -54,7 +54,7 @@ const PASTE_DELAYS = {
 };
 
 const RESTORE_DELAYS = {
-  darwin: 2000,
+  darwin: 450,
   win32_nircmd: 500,
   win32_pwsh: 500,
   linux: 800,
--- a/test/helpers/clipboardRestore.test.js
+++ b/test/helpers/clipboardRestore.test.js
@@ -1,6 +1,8 @@
 const test = require("node:test");
 const assert = require("node:assert/strict");
 const Module = require("node:module");
+const { EventEmitter } = require("node:events");
+const childProcess = require("node:child_process");
 
 const fakeClipboard = {
   text: "",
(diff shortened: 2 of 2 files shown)


gabrielste1n:
thank you for the awesome work here @SyntaxSawdust !!

gabrielste1n (approved):

PR #1042, opened 2026-07-03
fix(linux): fall back to --no-sandbox when user namespaces are restricted

### LINKED ISSUE #1033: Linux AppImage 1.7.3 don't work
**Bug**
AppImage release don't start.

**To Reproduce**
Try to execute the AppImage release 1.7.3.

**Expected behavior**
Run the application

**Desktop:**
- OS: Ubuntu 24.04
- Version 24.04.4 LTS (Noble Numbat)

**Additional context**
[4095462:0701/172134.534169:FATAL:sandbox/linux/suid/client/setuid_sandbox_host.cc:166] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /tmp/.mount_OpenWh94iFfF/chrome-sandbox is owned by root and has mode 4755.

xAlcahest:
## Summary

Ubuntu 23.10+ enables `kernel.apparmor_restrict_unprivileged_userns=1` by default, which denies the user namespaces Chromium's sandbox needs to any process without an AppArmor profile. Electron then falls back to the setuid `chrome-sandbox` helper, which inside an AppImage FUSE mount can never be root-owned with mode 4755, so the app aborts at launch (electron/electron#41066). deb/rpm installs are unaffected because `after-install.sh` already sets the SUID bit.

The launcher script generated by `wrapLinuxBinary` in `scripts/afterPack.js` now checks at startup: if `chrome-sandbox` next to the binary is not setuid root and `unshare --user --map-root-user true` fails, it starts with `--no-sandbox` and prints a notice to stderr. This is the same probe electron-builder ships in AppRun since 26.9 (electron-userland/electron-builder#9590), which we don't get on 26.4 with the default AppImage toolset. The uid mapping in the probe matters: on Ubuntu 24.04 namespace creation succeeds but capabilities inside are denied, so a plain `unshare --user` reports the sandbox as usable when it isn't. This also covers tar.gz, while deb/rpm and distros with working namespaces keep the sandbox.

Fixes #1033

## Changes

- scripts/lib/linux-launcher.js: new module that generates the Linux launcher script (template moved out of afterPack.js), adding the sandbox fallback and executable name validation
- scripts/afterPack.js: wrapLinuxBinary now builds the wrapper via the new module
- test/helpers/linuxLauncherSandbox.test.js: tests for the fallback (failing probe, working namespaces, setuid helper present, unshare missing, flag ordering, name validation)

### DIFF
--- a/scripts/afterPack.js
+++ b/scripts/afterPack.js
@@ -6,13 +6,16 @@
 //
 // 1. Strips non-target platform/arch binaries from onnxruntime-node
 //    (saves 150–180 MB per build).
-// 2. Wraps the Linux binary in a shell script that forces XWayland and
-//    reads user flags from ~/.config/open-whispr-flags.conf.
+// 2. Wraps the Linux binary in a shell script that forces XWayland, reads
+//    user flags from ~/.config/open-whispr-flags.conf, and falls back to
+//    --no-sandbox where the Chromium sandbox cannot work (AppImage/tar.gz
+//    on distros that restrict unprivileged user namespaces).
 
 const fs = require("fs");
 const path = require("path");
 const { execFileSync } = require("child_process");
 const { Arch } = require("app-builder-lib");
+const { buildLinuxWrapperScript } = require("./lib/linux-launcher");
 
 // ---------------------------------------------------------------------------
 // macOS resource binary signing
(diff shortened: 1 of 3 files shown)


gabrielste1n (approved):
Verified the launcher behavior in a Linux container across all scenarios (restricted userns fallback, working namespaces, setuid chrome-sandbox, missing unshare, flag ordering) — all correct. deb/rpm installs keep the sandbox via after-install.sh, and the lockfile is on electron-builder 26.8.1 so we don't get the upstream AppRun probe. Thanks for the thorough fix and tests!

PR #1043, opened 2026-07-03
fix(transcription): make self-hosted model configurable

### LINKED ISSUE #1013: Self Hosted - model seems to be hard coded to gpt-4o-mini-transcribe
**Describe the bug**
When using "Self hosted" OpenWhispr expects the model "gpt-4o-mini-transcribe".  There is no option in the settings to specify the model id / name that I am self hosting.

**To Reproduce**
Self host a model.  I used lemonade and then Whisper-Large-v3-Turbo.

In OpenWhispr settings, select self hosted and set your server URL, in my case http://localhost:13305/api/v1/

When transcribing OpenWhispr returns an error from the server: 

`"error":{"code":"model_not_found","message":"Model 'gpt-4o-mini-transcribe' was not found`

I have verified the local server using curl and a sample file:

`curl http://localhost:13305/v1/audio/transcriptions -F file=@sample_speech.wav -F model=Whisper-Large-v3-Turbo`

`{"text":" Good morning, Steve.\n Good morning, Katie.\n Have you tried the latest real-time diarization in\n Microsoft Speech Service,\n which can tell you who said what in real-time?\n Not yet.\n I've been using the batch transcription with diarization\n functionality,\n but it produces diarization result until whole audio get\n processed.\n Is the new feature can diarize in real-time?\n Absolutely.\n That's exciting.\n Let me try it right now.\n"}`

**Expected behavior**
I think we need a field in the settings to specify the model that is being self hosted.

**Desktop:**

- OS: Windows 11

**Additional context**
I am self hosting in order to get gpu acceleration on AMD APU (on my laptop).

xAlcahest:
## Summary
In self-hosted mode the model was never configurable: `getTranscriptionModel()` in `audioManager.js` has no self-hosted branch, so it resolved the model from leftover cloud provider settings and for a fresh setup always fell back to `gpt-4o-mini-transcribe`, which most self-hosted servers reject with `model_not_found`. The self-hosted panel now has a Model field (stored as `remoteTranscriptionModel`) that is sent verbatim when set; left empty, everything behaves exactly as before.

Tracing the same path also showed that `shouldUseStreaming()` never checked self-hosted mode: a self-hosted user with a stored OpenAI key and the default realtime model was silently routed to OpenAI's realtime WebSocket instead of their own server. Self-hosted mode now always goes through batch HTTP to the configured URL.

Fixes #1013

## Changes
- src/helpers/selfHostedTranscription.js: new helper with isSelfHostedTranscription and resolveSelfHostedTranscriptionModel
- src/helpers/audioManager.js: getTranscriptionModel returns the self-hosted model when set, shouldUseStreaming returns false in self-hosted mode
- src/helpers/transcriptionAuth.js: shouldSkipTranscriptionApiKey delegates to the shared predicate
- src/stores/settingsStore.ts: new remoteTranscriptionModel key with setter
- src/hooks/useSettings.ts: expose the new key and setter
- src/components/SelfHostedPanel.tsx: optional model input, rendered only when wired
- src/components/SettingsPage.tsx: pass the model field through the transcription section
- test/helpers/selfHostedTranscription.test.js: node --test coverage for the helper


gabrielste1n (approved):
Verified both fixes against main: the model now flows into the self-hosted request, and self-hosted mode no longer routes to OpenAI's realtime WebSocket. All unit tests pass locally. Thanks for the clean contribution!

PR #1044, opened 2026-07-03
fix(transcription): harden realtime dictation streaming connection

### LINKED ISSUE #988: Initial GPT-4o transcription is intermittently very poor, re-transcribing produces way better results
The initial transcription with GPT-4o or GPT-4o Mini is sometimes extremely poor. This can continue for hours or days, despite good audio quality.

When I open the UI and retranscribe the exact same recording, the result is usually much better. Whisper also often works correctly during these periods.

xAlcahest:
## Summary
Live dictation with GPT-4o and GPT-4o Mini streams audio to OpenAI over a realtime WebSocket in `openaiRealtimeStreaming.js`, while retry from history always re-uploads the saved audio file directly, bypassing that connection entirely. Three defects in the realtime path explain why the first pass is often bad and a retry fixes it: audio recorded while the connection was still being established got dropped instead of buffered, a connection that died silently (network change, sleep/wake) stayed marked as usable and got reused for hours, and the session declared the wrong sample rate to OpenAI (24kHz instead of the actual 16kHz capture rate). This PR buffers audio through the full connection window, adds a ping/pong keep-alive to catch dead connections before they get reused, and fixes the declared sample rate.

Fixes #988

## Changes
- src/helpers/openaiRealtimeStreaming.js: buffers audio from the start of a connection attempt instead of only once the socket exists, adds a keep-alive ping/pong that terminates stale connections
- src/helpers/ipcHandlers.js: assigns the streaming instance before the token fetch so early audio gets buffered, declares the correct 16kHz input rate for OpenAI sessions
- src/helpers/audioManager.js: updates a comment that no longer matched the buffering behavior
- test/helpers/openaiRealtimeStreaming.test.js: new tests for the buffering and keep-alive behavior

### DIFF
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -2639,7 +2639,8 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
       }
 
       // 2. Set up audio pipeline so frames flow the instant WebSocket is ready.
-      //    Frames sent before WebSocket connects are silently dropped by sendAudio().
+      //    Frames sent before the connection is open are buffered (bounded) by
+      //    sendAudio(), not dropped.
       const audioContext = await this.getOrCreateAudioContext();
       this.streamingAudioContext = audioContext;
       this.streamingSource = audioContext.createMediaStreamSource(stream);
--- a/src/helpers/ipcHandlers.js
+++ b/src/helpers/ipcHandlers.js
@@ -5367,29 +5367,40 @@ class IPCHandlers {
 
       const connectInner = async () => {
         const isCloud = options.mode !== "byok";
-        const apiKey = await fetchRealtimeToken(event, {
-          mode: options.mode,
-          provider: options.provider,
-        });
         const streaming = new OpenAIRealtimeStreaming();
         setupDictationCallbacks(streaming, event);
-        if (options.provider === "tinfoil-realtime") {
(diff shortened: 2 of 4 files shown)


gabrielste1n:
Pushed a follow-up commit to this branch. The buffering and keep-alive fixes are solid, but the `inputRate: 16000` change broke the OpenAI BYOK path — the realtime API rejects session rates below 24kHz:

```
invalid_request_error: Invalid 'session.audio.input.format.rate': integer below minimum value. Expected a value >= 24000, but got 16000 instead.
```

Since `session.updated` never arrives, every BYOK connect would hang 15s and fail. The diagnosis was right though: the worklet really does capture 16kHz while both BYOK and cloud sessions are pinned to 24kHz, so the model hears everything at 1.5x speed — likely the core of #988.

The new commit keeps the declared rate at 24kHz and linearly upsamples the 16kHz capture in `sendAudio()` instead. That also fixes preconfigured cloud sessions (configured server-side as pcm16/24kHz, out of reach of a client `session.update`). Verified against the live API: session accepted, exact 1.5x byte ratio on the wire, no error events, clean disconnect. Also reworked the connect()-preserves-buffer test to drive `connect()` through a fake socket factory so it exercises the real code path.

gabrielste1n (approved):

PR #1046, opened 2026-07-04
fix(dication): macOS Fn hotkey triggering during Fn+Delete chords    

B3rK-3:
## Summary                                                                                                                                                                         
                                                                                                                                                                                      
   Fixes a macOS Globe/Fn hotkey conflict where OpenWhispr could trigger dictation or agent actions when the user intended to use Fn as a normal macOS modifier key, such as          
 `Fn+Delete` for Forward Delete.                                                                                                                                                      
                                                                                                                                                                                      
   The previous behavior handled the Fn hotkey on key-down in toggle mode, so pressing `Fn+Delete` could immediately activate OpenWhispr before macOS completed the chord. This made  
 the Delete/Forward Delete workflow feel broken when Fn was configured as an OpenWhispr hotkey.                                                                                       
                                                                                                                                                                                      
   ## Changes                                                                                                                                                                         
                                                                                                                                                                                      
   - Track when Fn/Globe is used together with another key.                                                                                                                           
   - Emit distinct native listener events for:                                                                                                                                        
     - standalone Fn release                                                                                                                                                          
     - Fn used as a modifier/chord                                                                                                                                                    
     - Fn becoming chorded while held                                                                                                                                                 
   - Move toggle-mode Fn/Globe activation from key-down to key-up.                                                                                                                    
   - Skip OpenWhispr activation when Fn was used with another key.                                                                                                                    
   - Keep push-to-talk behavior starting from key-down, but cancel the pending hold if Fn becomes part of a chord.                                                                    
   - Apply the same key-up/chord guard to Fn-based agent and voice-agent hotkeys.                                                                                                     
                                                                                                                                                                                      
   ## User impact                                                                                                                                                                     
                                                                                                                                                                                      
   - `Fn+Delete` on macOS now works normally as Forward Delete.                                                                                                                       
   - Standalone Fn/Globe can still be used as an OpenWhispr hotkey.                                                                                                                   
   - Push-to-talk Fn behavior is preserved.                                                                                                                                           
   - OpenWhispr no longer steals Fn modifier chords in toggle mode.                                                                                                                   
                                                                                                                                                                                      
   ## Verification                                                                                                                                                                    
                                                                                                                                                                                      
   Ran:                                                                                                                                                                               
                                                                                                                                                                                      
   ```bash                                                                                                                                                                            
   node --check main.js                                                                                                                                                               
   node --check src/helpers/globeKeyManager.js                                                                                                                                        
   npm run compile:globe                                                                                                                                                              
   npx eslint main.js                                                                                                                                                                 
   npx eslint --no-ignore src/helpers/globeKeyManager.js                                                                                                                              
 ```                                                                                                                                                                                  
                                                                                                                                                                                      
 Results:                                                                                                                                                                             
                                                                                                                                                                                      
 - Main process syntax check passed.                                                                                                                                                  
 - Globe key manager syntax check passed.                                                                                                                                             
 - macOS Globe listener rebuilt successfully.                                                                                                                                         
 - ESLint passed for main.js.                                                                                                                                                         
 - ESLint passed for src/helpers/globeKeyManager.js; only the existing module-type warning from src/eslint.config.js was printed.

### DIFF
--- a/main.js
+++ b/main.js
@@ -1063,13 +1063,11 @@ async function startApp() {
     const { isGlobeLikeHotkey, isMouseButtonHotkey } = require("./src/helpers/hotkeyManager");
     let globeKeyDownTime = 0;
     let globeKeyIsRecording = false;
-    let globeKeyWasUsedAsModifier = false;
     let globeLastStopTime = 0;
     const MIN_HOLD_DURATION_MS = 150;
     const POST_STOP_COOLDOWN_MS = 300;
 
     globeKeyManager.on("globe-down", async () => {
-      globeKeyWasUsedAsModifier = false;
       const currentHotkey = hotkeyManager.getCurrentHotkey && hotkeyManager.getCurrentHotkey();
       const mainWindowLive = isLiveWindow(windowManager.mainWindow);
       debugLogger?.debug("[Globe] globe-down received", {
@@ -1101,11 +1099,7 @@ async function startApp() {
             globeKeyDownTime = pressTime;
             globeKeyIsRecording = false;
             setTimeout(async () => {
-              if (
-                globeKeyDownTime === pressTime &&
-                !globeKeyIsRecording &&
-                !globeKeyWasUsedAsModifier
-              ) {
+              if (globeKeyDownTime === pressTime && !globeKeyIsRecording) {
                 globeKeyIsRecording = true;
                 debugLogger?.debug("[Globe] Starting dictation (push hold)");
                 windowManager.sendStartDictation();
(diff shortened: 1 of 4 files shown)


gabrielste1n:
#826 just landed on main — it adds an FN_INTERRUPTED keyDown monitor to macos-globe-listener.swift and a globe-interrupted cancel path in main.js, touching the same code this modifies. The chord tracking and key-up activation here are still wanted (they cover the tap-mode #865 case that #826 doesn't) — could you rebase on current main? I'd like to get this in next.

B3rK-3:
Rebased.
Fixed:
- Removed the duplicate FN_CHORDED cancellation path, so an Fn chord cancels rather than accidentally stopping/finalizing dictation.
- Kept the floating OpenWhispr icon visible when Fn is used as a modifier (not sure if that was intended behaviour but the icon would disappear as the #826 added  windowManager.hideDictationPanel();)
- Queued a hold-release received during async audio startup, preventing dictation from staying active after Fn is released, a bug I observed.

Still needs fixing:
- Spamming Fn can trigger overlapping microphone warmups before the recording start lock is acquired. Deduplicate the in-flight warmup in AudioManager and await it before starting Fn-held dictation.

PR #1050, opened 2026-07-04
fix: snippet triggers containing Turkish İ and dotless ı never match

UmutEmreOnder:
## Problem

Snippet trigger matching used `toLowerCase()` for the lookup keys and the regex `/i` flag for matching, and neither handles the Turkish dotted/dotless I pair:

- `"İ".toLowerCase()` (U+0130) produces **two** code units — `"i"` + U+0307 combining dot — so a trigger saved as `İmza` produced a Map key that could never equal any regex match.
- The `/i` flag's simple case folding maps neither `İ` (U+0130) nor `ı` (U+0131): `/i/iu.test("İ")` and `/ı/iu.test("I")` are both `false`.

Result: any snippet whose trigger contains `İ` or `ı` (or is spoken with one) silently never expands.

## Fix

- Fold `İ` → `i` before lowercasing, giving Map keys a canonical form.
- A capital `I` in a trigger is ambiguous (English `i` vs Turkish `ı`), so both readings are registered as keys; an explicit trigger always wins over a generated variant.
- The compiled pattern matches `İ`/`ı` explicitly via `[iİ]` / `[ıI]` character classes.
- The transcript is NFC-normalized so a decomposed `I` + U+0307 recombines into `İ` before matching.
- Lookup falls back to the `I` → `ı` reading, since JS `toLowerCase()` maps `I` to `i` (`"IŞIK".toLowerCase()` is `"işik"`, not `"ışık"`).

Deliberately avoids `toLocaleLowerCase("tr")`, which would fix Turkish triggers but break English ones (`I` → `ı` everywhere).

## Tests

`test/utils/snippets.test.js`: covers `İmza`/`imza`/`İMZA`/`Imza`, `ışık`/`Işık`/`IŞIK`, an all-caps English trigger (`IBAN`), NFD input, word boundaries, and plain-ASCII regression cases. All 10 pass; existing tests untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixes
+
+- **Snippets now match triggers containing Turkish İ and ı.** Trigger matching used `toLowerCase()`, whose lowercase form of İ (U+0130) is "i" plus a combining dot, so a snippet like "İmza" never matched the transcript — and the regex `/i` flag case-folds neither İ nor dotless ı (U+0131). Triggers and matches are now folded to a canonical key, the pattern matches İ/ı explicitly, and the transcript is NFC-normalized, so "imza", "İmza", "İMZA" — and "Işık"/"IŞIK" for an "ışık" trigger — all expand the same snippet.
+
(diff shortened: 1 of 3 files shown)


gabrielste1n (approved):
awesome stuff thank you!

Issue #67, opened 2026-01-03
macOS: Unable to use existing whisper install


heydonovan:
**Describe the bug**

I'm unable to get past the `npm run dev` or `npm start` step:

```
$ npm run dev
...
Failed to install Whisper: Error: Error invoking remote method 'install-whisper': Error: error: externally-managed-environment × This environment is externally managed ╰─> This Python installation is managed by uv and should not be modified. note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages. hint: See PEP 668 for the detailed specification.
```

```
$ npm start
...
Failed to load main window: -6 ERR_FILE_NOT_FOUND file:///Users/heydonovan/projects/open-whispr/src/dist/index.html`
```

I've tried installing `whisper` via uv and brew:

```
$ uv tool install openai-whisper
$ ls -l ~/.local/bin/whisper
lrwxr-xr-x@ 1 heydonovan  staff  64 Jan  3 02:03 /Users/heydonovan/.local/bin/whisper@ -> /Users/heydonovan/.local/share/uv/tools/openai-whisper/bin/whisper
```

```
$ brew install openai-whisper
$ ls -l /opt/homebrew/bin/whisper
lrwxr-xr-x@ 1 heydonovan  admin  47 Jan  3 02:20 /opt/homebrew/bin/whisper@ -> ../Cellar/openai-whisper/20250625_2/bin/whisper
```

Can confirm that $PATH contains them both. I'm still prompted to install it.

Thinking `uv` was too new and that `pyenv` might be the solution, gave that a shot:

```
$ pyenv install 3.14.0
$ OPENWHISPR_PYTHON=~/.pyenv/shims/python3.14 npm run dev
```

**Desktop (please complete the following information):**

```
$ sw_vers
ProductName:		macOS
ProductVersion:		26.2
BuildVersion:		25C56
```

gabrielste1n:
@heydonovan thanks for raising this - will look into it asap.

gabrielste1n:
@heydonovan i think this should be addressed in 1.1.0  but please let me know if the issue still persists.

Ganbin:
I still have the same issue on Mac (14.7.7 (23H723))

gabrielste1n:
[...]

Issue #61, opened 2025-12-11
Not working on Windows.


cygnussystems:
The installation worked fine on Windows 11. But I was unable to generate text. There was an error message that popped up on bottom right, but it was only partially visible and only for a couple of seconds - I was unable to copy the text.  Would be great if you could get this working on Windows. When I clicked on the backtick - it just printed a backtick on the screen.  In future, please allow a double key like Win+Ctrl like WhisprFlow does it. This would be ideal from my perspective.

hankyukim89:
Same. It doesn't work at all. I've tried everything. Everything.

ceebeel:
Same here, with standalone version on Windows 11.
Transcription Mode: Local Processing | Model: Base

<img width="221" height="219" alt="Image" src="https://github.com/user-attachments/assets/d598e297-ee3c-4aee-bce1-385f238d3e93" />

MathieuCGit:
I can confirm the exact same error on my side, windows 10 22H2 (19045.3324). Where are stored the error logs ?

gabrielste1n:
Hey all (@cygnussystems @hankyukim89 @ceebeel @MathieuCGit ) , I just merged #51 into main and released v1.0.14 - would really appreciate if you could try the latest version out to see if it resolves your issues. Thanks so much for raising them.

MathieuCGit:
Hello @gabrielste1n ,
Transcription is now working, but I get an issue with AI assistant. I have reported the issue here https://github.com/HeroTools/open-whispr/issues/68

BTW Thanks for the time, effort and energy you put into this piece of software :)

gabrielste1n:
Yay amazing - glad that is resolved, will address #68 today. Appreciate your support! 💪

Issue #19, opened 2025-08-20
Interface Elements Disappear and Become Inaccessible After Minimization


PabloLION:
**Describe the bug**
Voice recognition interface elements (circular status indicator and settings window) disappear and become permanently inaccessible after minimization operations. The app lacks proper taskbar integration, making it impossible to restore the interface once minimized.

**To Reproduce**
1. Launch OpenWhispr and open the settings window
2. Start voice recognition (circular status indicator appears)
3. Minimize the settings window or double-click on interface elements
4. Try to restore the interface - circular indicator and settings window are gone
5. Check Docker/taskbar - no way to restore the interface

**Expected behavior**
Interface elements should remain accessible through taskbar integration or system tray. Minimized windows should be restorable, and the voice recognition interface should persist or be easily recalled.

**Screenshots**
Not applicable - interface elements completely disappear.

**Desktop (please complete the following information):**
- OS: macOS 15
- Version: OpenWhispr v1.0.3

**Additional context**
- Official v1.0.3 installer doesn't work, requiring self-compilation
- Voice recognition icon disappears after showing in Docker taskbar
- Models download to ~/.cache instead of app directory, complicating uninstallation
- Performance issues: slow recognition even with turbo model, no audio queue
- Recognition quality problems with frequent "thank you" artifacts

PabloLION:
conflict of interest: I made a toy tool on https://github.com/PabloLION/whisper-note . I refactored https://github.com/HaujetZhao/CapsWriter-Offline/pull/191 and maybe I'll make a new adaptive STT tool. Since macOS dictation is not very satisfying.

PabloLION:
BTW I tested all the models

  - base.pt - 138MB (recommended balance)
  - large-v2.pt - 2.9GB (older large model)
  - large-v3-turbo.pt - 1.5GB (fast + high quality)
  - large-v3.pt - 2.9GB (latest large model)
  - medium.en.pt - 1.4GB (English-only medium)
  - medium.pt - 1.4GB (multilingual medium)
  - small.en.pt - 461MB (English-only small)
  - small.pt - 461MB (multilingual small)

they are all pretty slow.

I checked that it really not connecting to internet. I was afraid that the recognition result would be uploaded to somewhere.

PabloLION:
the text was "The recording appears to be silent. Please check that your microphone is working and not muted."

gabrielste1n:
[...]

PR #1051, opened 2026-07-05
chore(deps): clear all critical/high Dependabot alerts — 1 critical + 10 high (SOC 2 Secure Code finding)

greatcoat:
## Why
SOC 2 Type 2 auditor finding **Secure Code** requires zero open Critical/High dependency vulnerabilities org-wide. This repo had **1 critical + 10 high** Dependabot alerts: form-data critical #12 + highs #64/#65, tar ×6 (#14–#19), undici #67, ws #61.

## What
- **ws 8.20.1 → 8.21.0** (direct pin) — supersedes Dependabot PR #985, which can be closed
- **Root-cause removal instead of patching:** the form-data critical + tar@6 highs all lived in an ancient `node-gyp@7` bundled under `usocket` (a Linux-only *optional* dep of `dbus-next` — never installed on mac/win, but present in the lockfile where Dependabot scans). Added override `usocket → node-gyp ^11.4.2`, which drops the deprecated `request` + `form-data@2.3.3` + `tar@6.2.1` chain from the tree entirely.
- Lockfile regenerated (required to flush the stale optional-subtree entries); `undici` resolved to ≥6.27.0 — supersedes Dependabot PR #983, which can also be closed.

No source changes; package.json changes are the ws pin + the overrides block.

## Verification (local, macOS)
- `npm install` clean · `npm run typecheck` ✅ · `npm run lint` ✅
- `npm audit`: **high 0, critical 0** (2 moderates remain — outside the finding's bar)
- Lockfile scan: zero copies of request / form-data<2.5.6 / form-data@4<4.0.6 / tar<7.5.16 / ws@8<8.21 / undici<6.27

## Reviewer notes
- The one behavioral surface worth a look: Linux builds (`dbus-next` → `usocket` native build now uses node-gyp 11 instead of 7). Mac/Win unaffected.
- Closing #983 + #985 after this merges keeps the PR queue clean. Nothing auto-merges.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -65,7 +65,7 @@
         "tw-animate-css": "^1.3.4",
         "unbzip2-stream": "^1.4.3",
         "unzipper": "^0.12.3",
-        "ws": "8.20.1",
+        "ws": "8.21.0",
         "zod": "^4.3.6",
         "zustand": "^5.0.11"
       },
@@ -145,8 +145,6 @@
     },
     "node_modules/@ai-sdk/gateway": {
       "version": "3.0.143",
-      "resolved": "https://registry.npmjs.org/@ai-sdk/gateway/-/gateway-3.0.143.tgz",
-      "integrity": "sha512-RCH60KsUaNiZkI/fBuyau4yvYrVBIEgAcN+Ain94QpL1kVm28GduQzFKfGffAiJU2We0ZrmN4BHkoCZzACK96Q==",
       "license": "Apache-2.0",
       "dependencies": {
         "@ai-sdk/provider": "3.0.13",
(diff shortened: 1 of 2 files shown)


Chadpiha:
**One strong suggestion (fine as a fast-follow, doesn't need to hold up this PR):** the lockfile regeneration stripped `resolved` + `integrity` fields from most entries. Integrity hashes went from 198 on main to **53** on this branch — 1,008 of 1,062 packages (95%) now have neither field. Without integrity hashes, `npm ci` in CI/release builds installs those packages with no tamper verification, which is the same risk class this SOC2 work is meant to close. (Main was already partially stripped at ~82%, so this is a longstanding issue rather than something this PR introduced — but it's a good opportunity to fix it properly.)

This happens when npm rebuilds lockfile entries from an existing `node_modules` tree instead of registry metadata. To fix, in a dedicated lockfile-repair PR:

```bash
rm -rf node_modules package-lock.json
nvm exec 24 npm install
grep -c '"integrity"' package-lock.json   # should be ~1,000+, not ~53
nvm exec 24 npm ci                        # confirm it installs clean
```

Heads up that the full regen will also float all in-range transitive versions (expect a large diff of unrelated patch bumps — that's normal and one-time), so please run the test suite after.

**Also worth considering in that follow-up:** add a `lockfile-lint` step to CI that fails if any lockfile entry is missing an integrity hash or resolves to a non-HTTPS/unexpected registry URL. That would catch this class of issue at PR time and doubles as a SOC2 supply-chain control.

Chadpiha (approved):

PR #1069, opened 2026-07-06
chore(deps): restore lockfile integrity hashes + lockfile-lint CI guard (follow-up to #1051)

greatcoat:
## Why

Follow-up to #1051, driven by @Chadpiha's review comment: https://github.com/OpenWhispr/openwhispr/pull/1051#issuecomment-4893568084

The lockfile on main had `resolved` + `integrity` fields on only **53** entries (~95% missing), meaning `npm ci` in CI/release builds installed most packages with **no tamper verification** — the same supply-chain risk class this SOC 2 work is meant to close.

## What

1. **Regenerated `package-lock.json` from registry metadata** (Node 24 / npm 11.16.0, `rm -rf node_modules package-lock.json && npm install`).
2. **Added `.github/workflows/lockfile-lint.yml`** (Chad's suggested CI guard): runs `lockfile-lint --type npm --validate-https --allowed-hosts npm --validate-integrity` on every PR, failing if any entry loses its integrity hash or resolves to a non-HTTPS / unexpected host. The new lockfile passes it locally.

`package.json` is byte-identical to main — only the lockfile + the new workflow changed.

## Integrity hashes: 53 → 967

- Before (main): **53** entries with `integrity`/`resolved`
- After: **967 of 967** package entries (100%) — zero entries without an integrity hash

## Note on the large diff

As Chad flagged, the full regen floats all in-range transitive versions, so the diff includes a large one-time batch of unrelated in-range bumps. Expected and normal.

## Deviation worth knowing about: better-auth cluster kept at main's locked versions

A fully unpinned regen is currently **impossible** with a plain resolver due to upstream registry state:

- `@better-auth/sso@1.6.19` (exact-pinned in package.json) declares peer `better-call@1.3.6`
- its peer range `@better-auth/core@^1.6.19` greedily resolves to core ≥ 1.6.21, which declares peer `better-call@1.3.7` → **ERESOLVE**
- no `--before` date window exists either (`tinfoil@^1.1.7` published 2026-06-30; the core peer change landed 2026-06-26)

Rather than reach for `--force`/`--legacy-peer-deps` (which would bake a peer-inconsistent tree, or silently drop auto-installed peers like `electron-builder-squirrel-windows` needed by Windows release builds), the better-auth cluster (18 nodes) was seeded at **main's exact locked versions** — `@better-auth/core@1.6.20`, `better-call@1.3.6`, etc. — with integrity backfilled from registry `dist` metadata, and everything else re-resolved fresh. The resulting tree is fully peer-consistent and identical in shape to main for that cluster.

**Long-term fix** (out of scope here, needs a package.json change): bump or un-pin `@better-auth/sso` so it aligns with current `@better-auth/core` peers.

## Verification

- ✅ `npm ci` from the new lockfile: clean (proves self-consistency)
- ✅ `npm run typecheck`: pass
- ✅ `npm run lint`: pass
- ✅ `npm audit`: **high = 0, critical = 0** (2 pre-existing moderates: `xml2js <0.5.0` via `dbus-next`, no fix available upstream — unchanged from main)
- ✅ `npx lockfile-lint --type npm --validate-https --allowed-hosts npm --validate-integrity`: no issues detected

### DIFF
--- a/.github/workflows/lockfile-lint.yml
+++ b/.github/workflows/lockfile-lint.yml
@@ -1,0 +1,12 @@
+name: Lockfile Lint
+
+on:
+  pull_request:
+
+jobs:
+  lockfile-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Lint package-lock.json
+        run: npx --yes lockfile-lint --path package-lock.json --type npm --validate-https --allowed-hosts npm --validate-integrity
--- a/package-lock.json
+++ b/package-lock.json
@@ -97,12 +97,15 @@
       }
     },
     "node_modules/@ai-sdk/amazon-bedrock": {
-      "version": "4.0.96",
+      "version": "4.0.129",
+      "resolved": "https://registry.npmjs.org/@ai-sdk/amazon-bedrock/-/amazon-bedrock-4.0.129.tgz",
+      "integrity": "sha512-z+JAe15rDrsK5tdguGJFFV/CGAlBBTHcjA1FCgdH34F5+Qo9raMRc1fyTeVICoWJ2/Lf9g+A3yuGf6S8UVBdzg==",
       "license": "Apache-2.0",
       "dependencies": {
-        "@ai-sdk/anthropic": "3.0.71",
-        "@ai-sdk/provider": "3.0.8",
-        "@ai-sdk/provider-utils": "4.0.23",
+        "@ai-sdk/anthropic": "3.0.93",
+        "@ai-sdk/openai": "3.0.80",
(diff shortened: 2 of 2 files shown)


gabrielste1n (approved):
Verified independently: package.json untouched; 967/967 lockfile entries have sha512 integrity + HTTPS registry.npmjs.org resolved URLs; the hand-seeded better-auth cluster matches main's versions and its integrity hashes match the npm registry dist metadata exactly. All platform builds passing on npm ci proves the hashes match what the registry serves. Production floats are all in-range (largest: onnxruntime-node 1.24.3 → 1.27.0) — worth a manual smoke test of dictation/diarization/auto-update before the next release since there's no automated test suite.

PR #1070, opened 2026-07-06
feat(sync): re-sync on window focus and a 5-minute interval

Chadpiha:
## Summary

The desktop app only ran a full syncAll() once, when ControlPanel mounts (app launch). During a long-lived session it never re-pulled, so dictionary, snippet, and note changes made on other devices (e.g. mobile) wouldn't appear on desktop until the app was restarted or some unrelated action (transcription/folder delete) happened to trigger a sync.

This adds ambient re-sync triggers that mirror mobile's foreground behavior: on window focus / tab becoming visible, and on a 5-minute interval. A new throttled entry point (requestSyncAll) prevents these triggers from stacking redundant passes, and leaves the existing canSync() gating and syncing mutex fully intact — so sync semantics are unchanged, only its cadence improves.

### DIFF
--- a/src/components/ControlPanel.tsx
+++ b/src/components/ControlPanel.tsx
@@ -384,14 +384,17 @@ export default function ControlPanel({ initialSettingsSection }: ControlPanelPro
         syncService.requestSyncAll("focus");
       }
     };
+    const onOnline = () => syncService.requestSyncAll("online");
     window.addEventListener("focus", onFocus);
     document.addEventListener("visibilitychange", onVisibility);
+    window.addEventListener("online", onOnline);
 
     const interval = setInterval(() => syncService.requestSyncAll("interval"), 5 * 60 * 1000);
 
     return () => {
       window.removeEventListener("focus", onFocus);
       document.removeEventListener("visibilitychange", onVisibility);
+      window.removeEventListener("online", onOnline);
       clearInterval(interval);
     };
   }, []);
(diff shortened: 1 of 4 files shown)


gabrielste1n:
Pushed a follow-up commit addressing review findings:

- **Throttle now measures completed syncs, not started ones.** `lastSyncAllAt` is stamped after a pass finishes successfully, so an in-flight or failed pass no longer suppresses subsequent triggers for 20s.
- **Requests that land mid-flight are coalesced instead of dropped.** `syncAll()` now flags a single follow-up pass when called while a pass is running (same pattern as the existing `dictionaryDirty` re-run), so e.g. a deletion that happens during a long sync still pushes right after.
- **Single entry point.** The four direct `syncAll()` call sites (transcription delete/clear, folder delete, cloud-backup enable) now go through `requestSyncAll("manual")`, which puts the previously-unused `manual` reason to work and makes the throttle bookkeeping global — no more redundant back-to-back full syncs after a manual action.
- **Re-sync on network reconnect** via an `online` listener, alongside focus/visibility/interval.

Typecheck, lint, and Prettier all pass. Known limitation left for a follow-up: sync scheduling still lives in the ControlPanel renderer, so a start-minimized session that never opens the panel has no sync triggers (pre-existing).

gabrielste1n (approved):
Reviewed the sync coalescing and entry-point unification — logic traced end-to-end, typecheck/lint/format pass. LGTM.

PR #1072, opened 2026-07-07
refactor(sync): move ambient sync scheduling into SyncService

gabrielste1n:
## Summary

Follow-up to #1070. That PR added ambient re-sync triggers, but they lived in ControlPanel's mount effect — so a start-minimized session that never opens the control panel had **no sync at all**: dictations stayed local-only and edits from other devices never pulled down until the panel was opened. This moves sync scheduling to where it belongs and makes it window-independent, with no hidden background window needed.

## Changes

- **`SyncService.startAutoSync()`** now owns all ambient triggers: initial pass, window focus/visibility, network reconnect (`online`), and the 5-minute interval. It's bootstrapped from `AppRouter` in both the dictation overlay and the control panel, so tray-only sessions sync too.
- **Cross-window coordination via a Web Lock** (`navigator.locks`): each renderer has its own SyncService instance, but they share `localStorage` and the local DB. `syncAll()` runs under a shared lock with `ifAvailable` — if another window is mid-pass, this window skips, since that pass covers the same data.
- **Shared throttle stamp**: the auto-sync throttle now reads `lastSyncedAt` (written only when a pass completes, shared across windows) instead of a per-renderer in-memory field, so all windows share one 20s throttle window.
- ControlPanel's sync effect is deleted; its manual triggers (`requestSyncAll("manual")`) are unchanged. The `"mount"` reason is renamed `"start"`.

## Behavior

- Start-minimized users now get full background sync (push and pull) without opening the control panel.
- With both windows open, only one syncs at a time; the other's triggers are throttled or skip via the lock — no duplicate passes.
- Agent overlay and transient notification overlays are excluded; `canSync()` still gates everything (signed in + cloud backup + subscribed).

## Verification

- Typecheck: no new errors (5 pre-existing `tinfoil` module errors on `main` are unchanged).
- ESLint and Prettier: clean.

### DIFF
--- a/src/AppRouter.jsx
+++ b/src/AppRouter.jsx
@@ -57,6 +57,14 @@ function MainApp() {
         import("./components/OnboardingFlow.tsx").catch(() => {});
       }
     }
+
+    // Both the dictation overlay and the control panel keep cloud data fresh,
+    // so sync runs even in tray-only sessions where the panel is never opened.
+    if (!isAgentPanel) {
+      import("./services/SyncService.js")
+        .then(({ syncService }) => syncService.startAutoSync())
+        .catch(() => {});
+    }
   }, [isAgentPanel, isControlPanel]);
 
   useEffect(() => {
--- a/src/AppRouter.jsx
+++ b/src/AppRouter.jsx
@@ -58,8 +58,8 @@ function MainApp() {
       }
     }
 
-    // Both the dictation overlay and the control panel keep cloud data fresh,
-    // so sync runs even in tray-only sessions where the panel is never opened.
+    // Sync runs in every non-agent window, so tray-only sessions where the
+    // control panel is never opened still stay fresh.
     if (!isAgentPanel) {
       import("./services/SyncService.js")
         .then(({ syncService }) => syncService.startAutoSync())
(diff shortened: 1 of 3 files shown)


gabrielste1n:
Second-pass review (end-to-end trace of user journeys, cross-window lock semantics, and update path for existing users) surfaced three things, all addressed in the follow-up commit:

1. **First sync after auth changes could wait up to 5 minutes.** The `canSync()` keys (`isSignedIn`, `isSubscribed`, `cloudBackupEnabled`) are written asynchronously during onboarding/re-login, and nothing reacted when they flipped. A `storage` listener on those keys now triggers a pass in the other window the moment sync becomes possible (storage events fire only in windows that didn't write the change — exactly where the trigger is needed). Covers: fresh onboarding, re-login after session expiry, enabling cloud backup.
2. **Manual syncs could be silently skipped.** With `ifAvailable`, a manual sync (transcription delete/clear, folder delete) requested while the other window was mid-pass would skip — and the deletion push would wait for the next ambient trigger. Manual passes now wait for the lock; ambient passes still skip (the in-flight pass covers them).
3. Comment accuracy fixes (throttle is now global across windows, not per-window).

Verified for existing users updating: `lastSyncedAt` is the same ISO format written pre-PR, both windows share the default session/partition and origin in dev and prod (the app already relies on shared localStorage across windows, e.g. `onboardingCompleted` gating in AppRouter), and Web Locks are supported since Chromium 69 (app is on Electron 41). Locks auto-release if a window is destroyed mid-pass, so no leak. Typecheck adds zero errors vs main; ESLint and Prettier clean.

PR #1074, opened 2026-07-07
feat: consume and enforce organization policy

gabrielste1n:
## Consume and enforce organization policy

Desktop side of enterprise policy enforcement. Part of a 3-repo change; depends on the **openwhispr-api** PR (adds `GET /api/workspace-policy` + the `managed` flag on `stt-config`). Safe to merge after the API is deployed — with the API absent, `get-workspace-policy` 404s and the app treats the user as unmanaged (no behavior change).

### What's included
- `src/types/policy.ts` — `OrgPolicy` type mirroring the API/console schema.
- `src/stores/policyStore.ts` — store + `isModeAllowed` / `isProviderAllowed` / `enforceModeOptions` selectors.
- `get-workspace-policy` IPC — authenticated fetch, disk cache at `userData/workspace-policy.json`, **fail-closed** (serves cache on failure; authoritative `managed:false` clears it). Adds `x-openwhispr-version` to cloud requests.
- Fetches policy on sign-in; transcription mode picker locks disallowed modes with a "Managed by your organization" badge; managed banner in the settings modal.

Server-side checks (transcribe, reason, agent, sharing) remain the backstop for cloud paths. Remaining UI filtering (BYOK/LLM/enterprise provider tabs, retention-clamp plumbing, selection auto-migration) builds on these selectors — see the handoff doc.

### Verification
`cd src && npx tsc --noEmit` clean, except 5 pre-existing errors in `services/ai/tinfoil*.ts` (missing `tinfoil` module, unrelated — present on `main`). eslint passes on changed files.

Full context: `~/Documents/Plans/openwhispr-admin-console-REVIEW-HANDOFF.md`.


Chadpiha:
## Policy enforcement: complete field coverage + hardening

The first two commits on this branch added policy consumption and mode-allowlist
enforcement. A code-quality review found that enforcement was effectively visual-only, and
several schema fields were typed but never consumed. This round fixes that and enforces
**every** field the policy schema defines.

### Enforcement made real

- **Click-through bypass fixed.** The mode selector fires `onSelect` even for disabled
  options (the sign-in gate depends on that to launch onboarding), so a managed user could
  click a policy-blocked mode and switch anyway. Policy is now enforced in the select
  handlers — transcription in `SettingsPage`, and all four LLM scopes in
  `InferenceConfigEditor`, which previously ignored the `llm` scope entirely.
- **Stored state reconciled.** `reconcileSettingsWithPolicy` runs whenever a managed policy
  loads and forces disallowed modes, providers, feature flags, and retention values onto
  allowed ones, mirroring a manual change. The pipeline reads modes/providers from the
  settings store, so this closes the runtime gap, not just the UI.
- **Malformed policies fail closed.** The IPC handler validates the structure of every
  field the renderer dereferences. A managed response with a bad policy is served from the
  disk cache instead of being cached and silently unlocking settings; cache reads
  re-validate so caches from older builds can't crash the renderer. Validator extracted to
  `policyValidation.js` with `node --test` coverage.

### Field-by-field enforcement

| Field | Enforcement |
|---|---|
| `transcription.allowedByokProviders` | `TranscriptionModelPicker` filters providers at the single memo the tabs/models/auto-migration derive from; UI-injected `custom` tab and base-URL auto-detect gated separately |
| `llm.allowedByokProviders` | `ReasoningModelSelector` filters provider tabs, guards tab clicks and the local→cloud switch that used to write the remembered provider back unchecked |
| `llm.allowedEnterpriseProviders` | `EnterpriseSection` disables disallowed tabs with the managed badge and guards selection |
| `features.agentEnabled` | Forces `useDictationAgent` off (also kills the voice-agent hotkey via reachability), locks the toggle, hides the sidebar chat tab, blocks the chat view render, auto-hides the agent overlay — which is its own BrowserWindow, so it fetches the policy itself on mount |
| `features.webSearchEnabled` | Gates `web_search` tool registration; part of the tool-registry cache key so a policy flip can't serve a stale registry |
| `sharing.externalLinkSharing` | `domain_only` hides "anyone with the link"; `disabled` also hides the domain option — enforced in the menu, in `applyVisibility` (the one server write path), and in the silent auto-apply of the remembered default (clamped to `invited`) |
| `dataRetention.localHistoryMode` | `always_on`/`always_off` force and lock the history toggle |
| `dataRetention.cloudBackupAllowed` | Forces `cloudBackupEnabled` off via the localStorage-writing setter that `SyncService.canSync()` reads; toggle locked |
| `dataRetention.audioRetentionMaxDays` | Clamps the stored value, disables out-of-range options in the retention select |
| `minAppVersion` | Persistent update-required banner in the control panel (all views) when the app is older than the org minimum; new dotted-version comparator in `src/utils/version.ts` |

### Cleanups

- One in-flight `fetchPolicy` shared across all `useAuth` consumers (was up to 7 fetches
  at startup); stale stt-config bootstrap comment and unread `managed` type flag removed
- `isProviderValidForMode` moved into `ModelRegistry`; `x-openwhispr-version` header
  spread extracted to a helper
- All new UI strings (`common.managedByOrg`, `settingsModal.managedByOrg`,
  `controlPanel.updateRequiredByOrg.*`) added to all 10 locale catalogs — no inline `t()`
  defaults

### Verification

typecheck ✅ · eslint (all changed files) ✅ · `i18n:check` ✅ · `node --test` 190 pass / 0 fail
(9 new validator tests)

### Known limitations

- Enforcement is renderer-level by design; the CLI bridge and raw IPC don't consult the
  policy. Authoritative enforcement for cloud features belongs in the API, which also
  needs the matching schema fields (`openwhispr-api/lib/policy-schema.ts`).
- Policy refreshes once per session (sign-in/startup); admin changes take effect after an
  app restart.
- Pre-existing inconsistency, flagged not fixed: regular transcription audio is still
  saved when audio retention is "Disabled" (0) — only discarded-recording saves respect
  the 0 value today.

Chadpiha (approved):

PR #1086, opened 2026-07-07
fix: link default folders to existing cloud folders instead of registering duplicates

Chadpiha:
## Bug

Both desktop and mobile seed the default "Personal" and "Meetings" folders locally,
each with its own randomly generated `client_folder_id`. Folder sync matches purely
on that id, so the two platforms registered the *same* conceptual folders as
*different* folders.

Concretely, when mobile synced first and desktop signed in afterwards:

- **Push**: desktop sent its defaults with unknown client ids; the server fell
  through to a plain INSERT and hit its `UNIQUE (user_id, name)` constraint, which
  500'd the entire batch — including any legitimately new folders. The defaults
  stayed `pending` and re-failed on every sync.
- **Pull**: the incoming cloud defaults matched nothing locally, so the upsert tried
  to INSERT a second "Personal"/"Meetings" and crashed on the local unique folder
  name, aborting the pull before the sync cursor was saved. Folder sync was wedged
  in both directions, and notes filed in the defaults synced unfiled.

Mobile already solves this with a name-based adoption step on first sync
(`initialBackfill`); desktop had no equivalent.

## Fix

- **`SyncService.adoptDefaultFolders()`** (new): runs before the first folder push.
  If any default folder is still unlinked, fetch the cloud folder list, match
  defaults by name (case-insensitive), and adopt the cloud's `client_folder_id` +
  cloud id instead of pushing our own. Mirrors mobile's reconciliation.
- **`pushPendingFolders`**: map batch-create responses to local folders by position
  (the server preserves input order — same contract mobile relies on) instead of by
  `client_folder_id`, and adopt the returned identity when the server hands back an
  existing folder (see companion API MR).
- **`pullFolders`**: when a cloud default arrives with an unknown client id, match a
  local default by name and adopt its identity instead of INSERTing a duplicate row
  that violates the local unique name.
- Plumbing: new `adoptFolderIdentity` DB method + IPC handler + preload binding +
  `electronAPI` typing.

## Companion MR

Requires/pairs with the `openwhispr-api` `fix/notes-sync` MR (server-side adoption
on default-name collision). Desktop's pre-push adoption alone fixes the reported
bug; the positional mapping makes the server-side fallback effective too.

### DIFF
--- a/preload.js
+++ b/preload.js
@@ -825,6 +825,8 @@ contextBridge.exposeInMainWorld("electronAPI", {
   upsertFolderFromCloud: (cloudFolder) =>
     ipcRenderer.invoke("db-upsert-folder-from-cloud", cloudFolder),
   markFolderSynced: (id, cloudId) => ipcRenderer.invoke("db-mark-folder-synced", id, cloudId),
+  adoptFolderIdentity: (id, clientFolderId, cloudId, updatedAt) =>
+    ipcRenderer.invoke("db-adopt-folder-identity", id, clientFolderId, cloudId, updatedAt),
   getFolderIdMap: () => ipcRenderer.invoke("db-get-folder-id-map"),
   getPendingFolderDeletes: () => ipcRenderer.invoke("db-get-pending-folder-deletes"),
   hardDeleteFolder: (id) => ipcRenderer.invoke("db-hard-delete-folder", id),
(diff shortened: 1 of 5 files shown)


gabrielste1n (approved):
Approve. Faithfully mirrors mobile's folder reconciliation (name-match default adoption + positional batch-create mapping). Note→folder linkage keys on cloud_id so adopting client_folder_id is safe; idempotent via sync_status flip; types and conventions check out. Ensure the companion openwhispr-api fix/notes-sync MR lands alongside.

PR #1091, opened 2026-07-08
nix: update to 1.7.4

gabrielste1n:
Automated bump of the Nix package to v1.7.4. (Branch was pushed by the update-nix workflow; PR created manually because the workflow's PR step was blocked before the Actions PR-creation permission was enabled. Catches up from 1.7.2 — the 1.7.3 bump hit the same permission block.)

### DIFF
--- a/nix/package.nix
+++ b/nix/package.nix
@@ -6,11 +6,11 @@
 
 let
   pname = "openwhispr";
-  version = "1.7.2";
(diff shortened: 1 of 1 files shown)


Juanal07:
hey @gabrielste1n can you automate this to always have the last version? nix users are in old one :)

PR #1093, opened 2026-07-08
fix(meeting): require transcript match before dropping held-back mic segments

gabrielste1n:
## Problem

During meetings, genuine local speech is silently deleted from the transcript. Field debug logs (Windows, 2026-07-02) show utterances like *"How do you make money off of this one"* and *"Can you hear me?"* being discarded with:

```
Dropping flagged-bleed mic segment after holdback { averageCorrelation: "0.742", ... }
```

Root cause: `flushPendingMicFinals` dropped a held-back mic segment **unconditionally on audio-only echo evidence** (`hasBleedEvidence`) — even though the text-dedupe check it had just run (`shouldSkipDuplicateMicSegment`) confirmed **no system transcript matched** the segment. The audio heuristic's own gate is correlation ≥ 0.7, and every falsely dropped utterance in the field log scored 0.728–0.812 during double-talk — the heuristic cannot distinguish "local speech while remote audio plays" from "echo" in exactly the situation where it decides to drop. Losing real speech is silent and unrecoverable; a leaked echo duplicate is visible and removable.

## What this does

**1. Transcript confirmation is now the only condition that can drop a held-back mic segment at flush time.** The holdback window exists to give the system channel time to produce matching text; if it never does, the segment is released instead of destroyed. The decision logic is extracted into a pure helper (`src/helpers/meetingMicHoldback.js`, `partitionPendingMicFinals`) following the `transcriptText.js` pattern, so the policy is unit-testable. Deferral, force-flush at meeting stop, timer rescheduling, and queue ordering are unchanged (verified by review — see below).

**2. The local-mode holdback now outlasts one transcription cycle.** Local mode transcribes in 5s chunks; a remote utterance straddling a chunk boundary gets its system-side tail transcribed one cycle later. The old 4.5s holdback expired *before* that confirming transcript could arrive. The holdback is now derived as `LOCAL_MEETING_CHUNK_INTERVAL_MS + 1000` (6s), so echo is confirmed and dropped while still buffered in the common case.

**3. The racing retract now works for late confirmations.** `removeRacingMicEntriesFor` — which retracts an already-committed mic segment when a matching system transcript arrives — gated on capture timestamps with a hard 4s window. For held-back segments that gate is unreachable by construction: a segment released at `capture + holdback` can only meet its confirming transcript *more* than the holdback away from its capture timestamp. Retract eligibility now also races the segment's **commit time** (`committedAt`, stamped in `storeMeetingDiarizationSegment`), with bleed-flagged segments getting the full `DUPLICATE_TRANSCRIPT_WINDOW_MS` (6s) window; the early-scan `break` is corrected to the max window. This covers late confirmations in both local and streaming modes; the renderer already handles `retract` events end-to-end (`meetingRecordingStore.ts`).

**4. Stop-time transcripts are joined in spoken order.** Held-back segments are appended at release time, after system segments spoken later; the stop-time transcript string now sorts by timestamp via `buildOrderedTranscriptText` (transitive comparator), used by both the local and streaming stop paths.

## Echo containment after this change

A released segment is still covered by text-gated nets, each verified end-to-end in code:

1. **While buffered**: `removePendingMicFinalsFor` drops it the moment a matching system transcript arrives (now covers a full local transcription cycle).
2. **At flush**: `shouldSkipDuplicateMicSegment` re-checks against all accumulated system transcripts (strict matcher for bleed-flagged, relaxed for double-talk).
3. **After commit**: the racing retract removes it from `meetingDiarizationSegments` and the renderer's live transcript — now genuinely reachable in both modes via the commit-time race (point 3 above); a second clean-room verification pass proved the capture-timestamp-only version was mathematically dead in local mode.

High-confidence capture-time suppression (`suppress: true`, reason `render_bleed` — no double-talk, strong gates) is intentionally **unchanged**: it exists for pure echo where the system channel may produce no text at all.

Known limitation (pre-existing, documented in the helper's doc comment): the post-diarization merge dedupe (`dedupeMicAgainstSystem`) filters only the diarization payload; the renderer's note merge preserves already-committed segments, so that net cannot remove text from the saved note. The live nets above are the effective containment; making the note merge respect the deduped payload is a candidate follow-up.

## Scope / risk

- Pure JS in the main process; identical behavior on macOS, Windows, and Linux, and in both cloud-streaming and local (whisper/Parakeet) meeting modes.
- No IPC, schema, settings, or renderer changes. No new strings.
- Live-view latency for risky **local-mode** mic segments increases 4.5s → 6s (still within one transcription cycle of the system channel's own inherent lag).
- Worst-case regression: when system-channel ASR returns empty for remote speech (echo has no text to match, seen in field logs as "Parakeet returned empty text for non-silent audio"), an echoed line can appear attributed to the mic instead of being silently dropped — the deliberate trade, since the previous behavior deleted real user speech in the same situation.

## Testing

- New `test/helpers/meetingMicHoldback.test.js` (13 tests): the field-log regression (bleed-flagged, correlation 0.742, no transcript match → must be released), the inverse (transcript match → still dropped), the meeting-stop force-flush × duplicate interaction, the `releaseAt === now` timer boundary, deferred entries never evaluated early, and the retract-window arithmetic — including a regression test proving a held-back segment's late confirmation matches on commit time where capture time categorically cannot.
- Full suite: 194 pass / 0 fail (11 skipped, platform-specific). `tsc` ✅ ESLint ✅ Prettier ✅.
- Two adversarial review passes with independent verification of every finding. Pass 1 (4 lenses over the core change) confirmed the flush refactor preserves old semantics except the intended policy change, and surfaced the local-mode holdback/cycle arithmetic. Pass 2 (clean-room trace of the final state) proved the first retract widening was still unreachable in local mode on capture timestamps — fixed via the commit-time race — and confirmed streaming provider timestamps (Deepgram/AssemblyAI/OpenAI/Corti) are epoch-ms comparable for the sorted stop join. All confirmed findings are addressed; refuted findings are documented in the review artifacts.

## Verification on real hardware

The definitive check is the field scenario from the debug log: Windows laptop, speakers + mic array (no headphones), local Parakeet meeting mode, both parties talking over each other. Expect: local speech during double-talk appears in the transcript (possibly after the ~6s holdback); echoed remote speech either never appears or is retracted within seconds.

### DIFF
--- a/src/helpers/ipcHandlers.js
+++ b/src/helpers/ipcHandlers.js
@@ -21,6 +21,7 @@ const AudioStorageManager = require("./audioStorage");
 const TINFOIL_REALTIME_MODEL = "voxtral-mini-4b-realtime";
 const liveSpeakerIdentifier = require("./liveSpeakerIdentifier");
 const MeetingEchoLeakDetector = require("./meetingEchoLeakDetector");
+const { partitionPendingMicFinals } = require("./meetingMicHoldback");
 const { applySmartSpacing } = require("./smartSpacing");
 const {
   transcriptsOverlap,
@@ -3963,7 +3964,13 @@ class IPCHandlers {
     const DUPLICATE_TRANSCRIPT_WINDOW_MS = 6000;
     const DUPLICATE_TRANSCRIPT_MERGE_LIMIT = 3;
     const STREAMING_RISKY_MIC_SEGMENT_HOLDBACK_MS = 3000;
-    const LOCAL_RISKY_MIC_SEGMENT_HOLDBACK_MS = 4500;
+    const LOCAL_MEETING_CHUNK_INTERVAL_MS = 5000;
+    // Must outlast one local transcription cycle: when a remote utterance
+    // straddles a chunk boundary, its system-side tail is only transcribed in
+    // the next cycle, and the buffered mic echo has to still be held back for
+    // that transcript to confirm it as a duplicate.
+    const LOCAL_RISKY_MIC_SEGMENT_HOLDBACK_MS = LOCAL_MEETING_CHUNK_INTERVAL_MS + 1000;
+    const RACING_MIC_RETRACT_WINDOW_MS = 4000;
 
     const buildNearbyTranscriptCandidates = (
       targetSource,
@@ -4035,9 +4042,19 @@ class IPCHandlers {
       for (let i = meetingDiarizationSegments.length - 1; i >= 0; i -= 1) {
         const candidate = meetingDiarizationSegments[i];
         if (candidate.source !== "mic" || candidate.timestamp == null) continue;
-        if (systemTimestamp != null && Math.abs(candidate.timestamp - systemTimestamp) > 4000) {
-          if (candidate.timestamp < systemTimestamp) break;
-          continue;
+        if (systemTimestamp != null) {
+          // Bleed-flagged segments get the full duplicate window: their
+          // confirming system transcript can land after the holdback released
+          // them (local mode transcribes in cycles), and the retract is the
+          // last live chance to remove the echo.
+          const retractWindowMs =
+            candidate.hasBleedEvidence || candidate.likelyRenderBleed
+              ? DUPLICATE_TRANSCRIPT_WINDOW_MS
+              : RACING_MIC_RETRACT_WINDOW_MS;
+          if (Math.abs(candidate.timestamp - systemTimestamp) > retractWindowMs) {
+            if (candidate.timestamp < systemTimestamp - DUPLICATE_TRANSCRIPT_WINDOW_MS) break;
+            continue;
+          }
         }
         const hasMicDuplicateRisk =
           candidate.likelyRenderBleed ||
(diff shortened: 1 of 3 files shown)


LauraGPT:
This policy change looks directionally right for local/private STT providers too: audio-only echo evidence is useful for holdback, but it should not be the final authority that deletes text when the text matcher has already found no matching system transcript.

A few checks I would keep around this boundary:

- Keep the drop condition text-confirmed: `hasBleedEvidence` can delay/retract, but a flush-time delete should require a matching system transcript.
- Preserve the negative case from the field logs: double-talk with correlation around `0.73-0.81`, no system text match, releases the mic segment.
- Keep one provider-failure case explicit: system-side ASR returns empty text or times out while the mic segment has text. That should prefer releasing the mic segment plus later retract/dedupe over silent loss.
- If OpenWhispr later adds more OpenAI-compatible local STT backends, including FunASR/SenseVoice-style final-only endpoints, this same text-confirmed rule is the safer cross-provider contract because timestamp and echo-correlation quality vary by backend.

So from an ASR integration perspective, the tradeoff in this PR is sensible: a visible duplicate is recoverable; silently losing genuine speech is not.

PR #1101, opened 2026-07-08
fix(linux): replace "dbus-next" with "@homebridge/dbus-native" to fix Wayland D-Bus crashes

### LINKED ISSUE #1098: 1.7.4 regression: tray icon missing & custom hotkey resets to default (Ctrl+Super) on every restart [Linux/Fedora/RPM]
## Environment

- **OS:** Fedora (Linux), x86_64
- **Install method:** official RPM (`open-whispr-1.7.4-1.x86_64`, installed to `/opt/OpenWhispr`)
- **Version:** 1.7.4 (upgraded from 1.7.3)

## What broke after upgrading to 1.7.4

Two regressions appeared immediately after the 1.7.4 update and persist on every launch. No other settings of mine were affected — only these two.

### 1. System tray icon no longer appears
OpenWhispr used to show a tray icon (alongside volume/Wi-Fi system icons) while running. Since 1.7.4 the icon is gone; the app runs but there is no tray entry to interact with.

### 2. Custom hotkey resets to default (Ctrl+Super / Ctrl+Meta) on every restart
I set the dictation hotkey to **Ctrl+Shift+Space**. After the 1.7.4 update, the binding resets back to the default **Ctrl+Super** (which surfaces as **Ctrl+Meta** on Linux) every time I quit and reopen the app. I rebind it each session, but it never persists across restarts.

## Confirmed as a 1.7.4 regression by downgrade

I downgraded to **1.7.3** (kept the same RPM install path and config dir):

```
sudo dnf remove -y open-whispr
sudo dnf install -y OpenWhispr-1.7.3-linux-x86_64.rpm
```

On 1.7.4 → bugs reproduce. On 1.7.3 → **both** bugs disappear (tray icon shows, custom hotkey persists across restarts). So the regression was introduced in 1.7.4, not in my setup. Reinstalling 1.7.4 from scratch (after wiping `~/.config/open-whispr`) does **not** fix it either.

## Likely related change

The hotkey behavior points at #1001 (`fix(hotkeys): make modifier-only hotkeys work simultaneously on Windows/Linux`), which reworked the native key listener into a multiplexer and re-runs `reconcileNativeKeyListeners()` at startup plus every hotkey/activation-mode change site. The default dictation key is exactly `Control+Super` (the value my custom binding keeps reverting to), which suggests the persisted custom binding is either not being read, not being recognized as a regular-key combo (it should go through `globalShortcut`, not the native listener), or is being overwritten by the reconcile step on startup. The tray-icon disappearance may share the same startup/init path that #1001 touches (native listener / windowManager reconciliation), though I have not confirmed a direct link.

## Ask

1. Can a maintainer confirm whether the startup `reconcileNativeKeyListeners()` path in #1001 overwrites a persisted regular-key (non-modifier-only) hotkey with the default?
2. Is the tray-icon initialization on Linux known to be affected by the same change?

Thanks!

### LINKED ISSUE #1100: D-Bus service crashes on startup, breaking hotkeys and tray icon on Wayland (Hyprland, GNOME) on the latest 1.7.4 release
**Describe the bug**
On Wayland (Hyprland and GNOME), the shipped AppImage for v1.7.4 fails to register global hotkeys and does not show a tray icon. v1.7.3 works correctly.

**To Reproduce**

1. Launch the v1.7.4 AppImage on any Wayland desktop (Hyprland, GNOME).
2. Check ~/.config/hypr/openwhispr-binds.conf; it exists with only header comments, no bind line.
3. Try the default global hotkey — nothing happens.
4. Tray icon is absent.

- OS: Linux (Arch Linux, Hyprland)

**Additional context**
The issue seems to be due to an error in "dbus-next" dependency

IdrisGit:
> [!IMPORTANT]
> This PR needs regeneration of `package-lock.json` file.
> I have intentionally not commited lockfile, as it's a critical piece for dependency management,
> and I am assuming either @gabrielste1n or any other maintainer should generate and commit that file.
> If you want me to do it, let me know I will commit the file or else I have **allow edits from maintainers** on as well.

## Problem

On Wayland (Hyprland, GNOME), the v1.7.4 AppImage fails to register global hotkeys and does not show a tray icon. v1.7.3 works correctly.

Both symptoms share the same root cause: the D-Bus library (`dbus-next`) depends on `usocket`, a native C++ addon (`uwrap.node`) that crashes on Electron 41 / Node 24. The crash kills the process before it reaches either the hotkey binding step or the tray creation step.

### Root cause chain

Two cascading failures in `dbus-next` → `usocket`:

1. **`util.isError` removed in Node 24** (`DEP0048`). `usocket@0.3.0` calls `util.isError()` during the D-Bus auth handshake. This API was EOL'd in Node 23 and doesn't exist in Node 24.   
`TypeError: util.isError is not a function`

2. **Native `uwrap` addon crashes on D-Bus method calls.** Even after polyfilling `util.isError`, the C++ addon crashes (SIGSEGV) when sending marshalled D-Bus binary messages. The auth handshake (plain text) works, but `requestName()` marshals binary data.
`fatal native crash that kills the Electron process.`

### How it regressed

- https://github.com/OpenWhispr/openwhispr/commit/397da9996ec470f2a741f9ae38539df71c9113e4 (Apr 9): `.nvmrc` upgraded Node 22 to Node 24.
- [`0ce117ab`](https://github.com/OpenWhispr/openwhispr/commit/699b1dabc1d435aae9d80d0f3b463591f510a59d) (Jul 7): Lockfile regenerated, `usocket`'s bundled `node-gyp` changed from `7.1.2` (working) to `11.5.0` (broken), producing a native binary that crashes on Node 24's V8.
- Both combined break the D-Bus transport in the 1.7.4 release

### My initial plan was to polyfill but didn't proceed with it because
- **Polyfill `util.isError`**: Fixes crash no. 1, but crash no. 2 (native SIGSEGV on method calls) persists and kills the process.
- **Pin `usocket`'s `node-gyp` to `7.1.2`**: Produces a working binary today, but `usocket` is unmaintained (last published 2019, uses the deprecated `nan` C++ layer). Any future Electron/Node bump will break it again.

## Solution

Replace `dbus-next` (native `usocket` addon) with [@homebridge/dbus-native](https://www.npmjs.com/package/@homebridge/dbus-native), a pure JavaScript D-Bus implementation from the [Homebridge project](https://github.com/homebridge) (maintained by a team of maintainers and ~20K weekly downloads).

No native addons. No `node-gyp` compilation. No future V8 ABI risk.

## How to test

1. `rm -rf node_modules && npm install`
2. On Hyprland Wayland: `OPENWHISPR_LOG_LEVEL=debug npm run dev`
3. Verify logs show:
  - HyprlandShortcut D-Bus service initialized successfully
  - HotkeyManager Hyprland D-Bus init result: true
  - HyprlandShortcut Keybinding "..." registered successfully
4. Verify `hyprctl binds | grep openwhispr` shows the binding
5. Press the dictation hotkey --> overlay should appear
6. Verify tray icon is visible

fixes #1100 fixes #1098


IdrisGit:
A part of logs when running the latest AppImage in debug mode

```
[INFO] Start minimized { enabled: false }
[154979:0708/220905.959360:ERROR:ui/gfx/x/atom_cache.cc:235] Add _NET_WM_WINDOW_TYPE_TOOLBAR to kAtomsToCache
[DEBUG] [HotkeyManager] Hyprland detection {
  "isLinux": true,
  "isWayland": true,
  "isHyprland": true,
  "XDG_SESSION_TYPE": "wayland",
  "HYPRLAND_INSTANCE_SIGNATURE": "present",
  "XDG_CURRENT_DESKTOP": "Hyprland"
}
Uncaught Exception: TypeError: util.isError is not a function
    at USocket._write (/tmp/.mount_OpenWh9V4khx/resources/app.asar/node_modules/usocket/index.js:64:11)
    at writeOrBuffer (node:internal/streams/writable:570:12)
    at _write (node:internal/streams/writable:499:10)
    at Writable.write (node:internal/streams/writable:508:10)
    at USocket.write (/tmp/.mount_OpenWh9V4khx/resources/app.asar/node_modules/usocket/index.js:197:32)
    at auth (/tmp/.mount_OpenWh9V4khx/resources/app.asar/node_modules/dbus-next/lib/handshake.js:67:10)
    at USocket.<anonymous> (/tmp/.mount_OpenWh9V4khx/resources/app.asar/node_modules/dbus-next/lib/connection.js:138:34)
    at Object.onceWrapper (node:events:630:28)
    at USocket.emit (node:events:509:28)
    at USocket._wrapEvent (/tmp/.mount_OpenWh9V4khx/resources/app.asar/node_modules/usocket/index.js:111:8)
Error stack: TypeError: util.isError is not a function
    at USocket._write (/tmp/.mount_OpenWh9V4khx/resources/app.asar/node_modules/usocket/index.js:64:11)
    at writeOrBuffer (node:internal/streams/writable:570:12)
    at _write (node:internal/streams/writable:499:10)
    at Writable.write (node:internal/streams/writable:508:10)
    at USocket.write (/tmp/.mount_OpenWh9V4khx/resources/app.asar/node_modules/usocket/index.js:197:32)
    at auth (/tmp/.mount_OpenWh9V4khx/resources/app.asar/node_modules/dbus-next/lib/handshake.js:67:10)
    at USocket.<anonymous> (/tmp/.mount_OpenWh9V4khx/resources/app.asar/node_modules/dbus-next/lib/connection.js:138:34)
    at Object.onceWrapper (node:events:630:28)
    at USocket.emit (node:events:509:28)
    at USocket._wrapEvent (/tmp/.mount_OpenWh9V4khx/resources/app.asar/node_modules/usocket/index.js:111:8)
[DEBUG] [AutoLearn] Setting changed { enabled: true }
[INFO][meeting] Updating detection preferences { processDetection: true }
[INFO][meeting] Process detector started {
  platform: 'linux',
  appsMonitored: [ 'Zoom', 'Microsoft Teams' ],
  intervalMs: 30000
}
```

You can see the problematic `Uncaught Exception: TypeError: util.isError is not a function`, which surfaces due to cascading failures in the dbus-next and it's transient dependencies.

UmutEmreOnder:
Nice approach. Pure-JS `@homebridge/dbus-native` is the right long-term call.

~~One thing to catch before this lands: `kdeShortcut.js` still does `require("dbus-next")`, so KDE would lose global hotkeys once dbus-next is removed. Worth porting it too.~~

Edit: you've already migrated `kdeShortcut.js` too, I missed that. Scratch the above.

I also opened #1102 as a smaller patch-based alternative (patches `dbus-next` to use Node's `net` instead of the `usocket` addon). Fewer changes, works with KDE in my testing. Could serve as a quick hotfix if this one needs more testing time.

IdrisGit:
Hey @xAlcahest the CI will not pass because I have intentionally not committed the regenerated lock file.

Let me know if you want me to commit it or the maintainers should do it. As it's an important piece for dependency management.

IdrisGit:
I have pushed the lock file to test CI and it's working fine.

gabrielste1n (approved):
Verified every converted call site in gnomeShortcut.js, hyprlandShortcut.js, and kdeShortcut.js against @homebridge/dbus-native 0.7.7 source and exercised the exportInterface/invoke/signal paths against a live message bus; also confirmed the KDE conflict-check unwrapping fix and the connection error listeners behave correctly on a dead session bus. Lint, typecheck, unit tests, lockfile-lint, and all platform CI builds are green on the current head, and the lockfile diff is a clean dbus-next/usocket removal.

PR #1102, opened 2026-07-08
fix(linux): patch dbus-next to use net instead of usocket (Wayland tray + hotkeys)

### LINKED ISSUE #1098: 1.7.4 regression: tray icon missing & custom hotkey resets to default (Ctrl+Super) on every restart [Linux/Fedora/RPM]

### LINKED ISSUE #1100: D-Bus service crashes on startup, breaking hotkeys and tray icon on Wayland (Hyprland, GNOME) on the latest 1.7.4 release

UmutEmreOnder:
## Problem

On Wayland, v1.7.4 shows no tray icon and global hotkeys don't register (KDE, GNOME, Hyprland); v1.7.3 is fine. Root cause is the Electron 41 / Node 24 runtime bump: `dbus-next` opens its D-Bus connection through `usocket`, a native addon (`uwrap.node`) that fails on the new runtime in two stages:

1. `usocket@0.3.0` calls `util.isError()` during the D-Bus auth handshake. That API was removed in Node 23 (`DEP0048`), so it throws `TypeError: util.isError is not a function`.
2. Even past that, the native addon crashes the process with a native `SIGTRAP`/`SIGSEGV` once it does real socket I/O.

Either failure aborts startup before global shortcuts register or the tray is created. (#1098, #1100)

## Fix

Patch `dbus-next` (via `patch-package`, wired into `postinstall`) so the `unix` branch of `connection.js` opens the D-Bus socket with Node's built-in `net.createConnection` instead of instantiating `usocket.USocket`. This applies to both the `path` and `abstract` socket forms, so `usocket` is never `require`d on the connection path.

`net` speaks the same path/abstract Unix sockets. The only capability dropped is unix-fd (`SCM_RIGHTS`) passing, which none of the shortcut services (KGlobalAccel / GNOME / Hyprland) use. The native addon stays out of the connection path entirely, fixing both failure modes with **no app-code changes**. All three desktops keep using `dbus-next` as-is.

## Why this, as a hotfix

@IdrisGit's #1101 fixes the same root cause the right way for the long term, replacing `dbus-next` with the pure-JS [`@homebridge/dbus-native`](https://www.npmjs.com/package/@homebridge/dbus-native) and removing the native-addon class of bug for good. This PR is offered as a smaller, lower-risk stopgap you might want to ship first while #1101 gets its review and cross-desktop testing:

- **Minimal surface:** one patch file plus the existing `patch-package` wiring, no app-code and no dependency changes. `dbus-next` keeps working as-is on KDE, GNOME, and Hyprland, just over `net` instead of the `usocket` addon.
- **Easy to reason about and revert:** the only behavior change is the socket transport.
- **Drops cleanly** when #1101 (or any pure-JS `dbus` client) lands. Just delete `patches/dbus-next+0.10.2.patch`.

## Testing

Verified on KDE Plasma (Wayland, Fedora 44):

- With the patch, `dbus-next` connects to `org.kde.KGlobalAccel` over `net`, global shortcuts register, and the tray icon appears (StatusNotifierItem registered on the session bus). No `util.isError` error, no native crash; the app runs to full init.
- Without it, startup fails at the D-Bus handshake with `util.isError is not a function`, plus a native `SIGTRAP` if that's polyfilled.

GNOME / Hyprland use the same `dbus-next` transport, so the same `net` path applies; independent confirmation there is welcome.

Closes #1098
Closes #1100

Related: #1101 (long-term pure-JS replacement)


gabrielste1n:
Closing since #1101 fixed this by replacing dbus-next with @homebridge/dbus-native entirely, so the dependency this patches no longer exists in the tree. Your root-cause analysis (usocket breaking on Node APIs removed in Node 24) was spot on — the merged fix was built on the same conclusion. Thanks!

Issue #990, opened 2026-06-23
Disable thinking output breaks reasoning models on Groq: chat_template_kwargs + reasoning_effort:'none' rejected (400 -> ~7s retry, raw transcript pasted)


jay-tau:
## Summary

With **"Disable thinking output"** enabled and a **reasoning model on a cloud OpenAI‑compatible provider** (reproduced with **Groq `openai/gpt-oss-120b`**), the cleanup request is rejected by the provider, retried with exponential backoff, and after ~7 s OpenWhispr gives up and pastes the **raw, un‑cleaned transcription**.

The end result the user sees: dictation that should take ~1 s takes ~8 s **and** isn't actually cleaned up.

## Root cause

`applyThinkingSuppression()` (in `src/services/ai/thinkingSuppression.ts`) sends two parameters that strict cloud APIs reject:

1. **`chat_template_kwargs: { enable_thinking: false }`** is added for every non‑Gemini provider. Cloud APIs (Groq, Cerebras, …) reject unknown fields.
2. **`reasoning_effort: "none"`** is sent for every non‑Ollama provider. Groq/Cerebras only accept `low | medium | high`.

Either one alone produces an HTTP 400, so the cleanup call can never succeed on Groq while thinking‑suppression is active.

### Verified against the live Groq API (`openai/gpt-oss-120b`)

```
POST https://api.groq.com/openai/v1/chat/completions
{ "model":"openai/gpt-oss-120b", "reasoning_effort":"none", "chat_template_kwargs":{"enable_thinking":false} }
  -> 400  "property 'chat_template_kwargs' is unsupported"

{ ... "reasoning_effort":"none" }                 -> 400  "`reasoning_effort` must be one of `low`, `medium`, or `high`"
{ ... "reasoning_effort":"low" }                  -> 200  OK
{ ... "reasoning_effort":"medium" }               -> 200  OK
```

## Real-world trace (debug log)

Single dictation, `cleanupModel: openai/gpt-oss-120b`, `cleanupProvider: groq`, `disableThinking: true`:

```
08:16:03.110  Recording stopped
08:16:03.585  Transcription received                 (STT ok, 474 ms)
08:16:03.591  GROQ_REQUEST           (cleanup attempt 1)
08:16:03.639  GROQ_API_ERROR_DETAIL  (400)
08:16:04.694  GROQ_API_ERROR_DETAIL  (400, +~1.0 s backoff)
[...]

PR #1103, opened 2026-07-08
fix(reasoning): disable thinking via reasoning object on self-hosted endpoints

### LINKED ISSUE #1021: Self-hosted BYOK with Ollama OpenAI-compatible API sends think:false instead of reasoning_effort:none
**Describe the bug**
When using OpenWhispr `1.7.3` with Self-hosted / BYOK connected to Ollama through its OpenAI-compatible API (`/v1/chat/completions`), the `Disable thinking output` option does not actually disable Qwen thinking generation.

OpenWhispr sends:

```json
{
  "think": false,
  "chat_template_kwargs": {
    "enable_thinking": false
  }
}
```

But for Ollama's OpenAI-compatible endpoint, the request should include:

```json
{
  "reasoning_effort": "none"
}
```

or:

```json
{
  "reasoning": {
    "effort": "none"
  }
}
```

As a result, Qwen still generates hidden thinking tokens. OpenWhispr appears to strip the thinking output from the final visible result, but Ollama still spends time decoding those tokens.

**To Reproduce**

1. Go to OpenWhispr settings.
2. Configure Self-hosted / BYOK to use Ollama's OpenAI-compatible API, for example `http://localhost:11434/v1`.
3. Select a Qwen reasoning model, e.g. `qwen3.5:9b`.
4. Enable `Disable thinking output`.
5. Run a short transcription cleanup task, for example: `Try to explain the code`.
6. Capture the request body sent to Ollama or check Ollama server logs.
7. Observe that the request contains `think:false` and `chat_template_kwargs.enable_thinking:false`, but does not contain `reasoning_effort:"none"`.
8. Observe high decoded/eval token count in Ollama logs despite the final visible output being short.

Example Ollama log:

```text
POST "/v1/chat/completions"
...
task.n_tokens = 504
...
eval time = 20481.76 ms / 1619 tokens
total time = 20600.35 ms / 2123 tokens
```

**Expected behavior**
When `Disable thinking output` is enabled and the Self-hosted / BYOK endpoint is using Ollama's OpenAI-compatible API, OpenWhispr should send OpenAI-compatible reasoning controls such as:

```json
{
  "reasoning_effort": "none"
}
```

This should prevent Qwen from generating hidden thinking tokens, not only hide or strip them from the final output.

**Screenshots**
Not applicable.

* OS: Windows 11
* Browser: Not applicable, desktop app
* Version: OpenWhispr `1.7.3`

**Additional context**
I tested Ollama's `/v1/chat/completions` endpoint directly. With `reasoning_effort:"none"`, Qwen output token count is much lower. With thinking enabled, the output volume is about 10x larger.

This suggests that Ollama and the Qwen model can disable thinking correctly, but OpenWhispr's Self-hosted / BYOK path is sending the Ollama-native `think:false` field instead of the OpenAI-compatible `reasoning_effort:"none"` field for `/v1/chat/completions`.

The captured OpenWhispr request body includes:

```json
{
  "model": "qwen3.5:9b",
  "temperature": 0.3,
  "max_tokens": 4096,
  "think": false,
  "chat_template_kwargs": {
    "enable_thinking": false
  }
}
```

xAlcahest:
## Summary

Self-hosted / BYOK always posts to an OpenAI-compatible `/v1/chat/completions` endpoint, but `suppressThinking` in `src/services/ai/thinkingSuppression.ts` still spoke Ollama's native dialect and sent `think: false`, a field Ollama's `/v1` compat layer never parses (it is not in `ChatCompletionRequest` in `openai/openai.go`, and unknown keys are dropped). "Disable thinking output" therefore did nothing against Ollama and Qwen models kept generating hidden thinking tokens on every dictation.

The fix sends `reasoning: {effort: "none"}` instead, which Ollama maps to `think=false` since v0.12.4, while llama.cpp, vLLM and LM Studio silently drop the unknown object and keep relying on `chat_template_kwargs` as before. The object form is deliberate: a flat `reasoning_effort: "none"` is a declared pydantic Literal in vLLM that returns 422 on releases before ~v0.22. For Ollama v0.11.1 to v0.12.3, which rejects the object with a 400 for non-gpt-oss models, both request paths drop the field and retry once, so those users keep exactly the behavior they have today. Verified against Ollama 0.31.1 with Qwen3.5 9B: cleanup goes from ~570 hidden thinking tokens per request to 8 output tokens in 386 ms.

Fixes #1021

## Changes

- `src/services/ai/thinkingSuppression.ts`: the `lan` provider now sends `reasoning: {effort: "none"}` instead of `think: false`; removed `usesOllamaDialect` and the legacy `remoteReasoningType` check, which nothing reads anymore; `chat_template_kwargs` is still sent for llama.cpp and vLLM backends.
- `src/services/ReasoningService.ts`: both the cleanup and streaming request paths retry once without the `reasoning` field on HTTP 400/422 (old Ollama, strict proxies) and cancel the discarded response body; the streaming path now resolves the `lan` dialect instead of passing the raw scope provider, which never matched `"lan"`.


gabrielste1n:
Note: #1156 (merged) added response-side stripping of <think> blocks on the non-streaming chat-completions path, complementing this PR's request-side suppression fix — both are needed (this one stops Ollama generating hidden reasoning tokens; the merged one cleans output when suppression isn't honored). ReasoningService.ts changed on main in #1154/#1156, so a rebase is likely needed before this can merge.

gabrielste1n (approved):
Verified the core claim in code: LAN suppression previously sent Ollama-native think:false, which the OpenAI-compat layer drops, so the toggle was a no-op (#1021); reasoning:{effort:"none"} is honored, and strict backends that reject the object are covered by the 400/422 drop-and-retry. Traced every suppressThinking path — Groq/Gemini/OpenRouter/Corti/OpenAI/custom are unchanged, so no regression of #990/#841. Brought up to date with main (think-tag stripping and LAN key resolution preserved alongside the retry), deduplicated the two retry blocks into one shared helper, full suite/typecheck/lint/format pass.

PR #1104, opened 2026-07-08
fix(transcription): scale local transcription timeouts with audio duration

### LINKED ISSUE #976: Server timeout after exactly 5 minutes during audio transcription
**Description:**

When uploading an audio file and starting transcription, the server times out after exactly 5 minutes, even though the transcription process has not yet completed.

**Steps to reproduce:**

1. Upload an audio file
2. Start transcription
3. Wait — after exactly 5 minutes, the request times out

**Expected behavior:**

The transcription completes successfully, regardless of duration.

**Actual behavior:**

Server returns a timeout error after exactly 5 minutes (indicating a hardcoded timeout limit).

 - OS: Windows 11
 - App
 - Version 1.7.2

xAlcahest:
## Summary

Local transcription dies after exactly 5 minutes on long recordings: the local single-request paths hardcode a 300000 ms request timeout (`whisperServer.js` POST /inference, `parakeetWsServer.js` transcribe, llama post-processing, diarization), and the wall is on inference time, so a long file on a slow CPU with a big model gets killed mid-transcription. The new `src/helpers/transcriptionTimeout.js` scales the timeout with the estimated audio duration (max(300000, seconds \* 10000), with a 24-hour ceiling to stay under the 32-bit setTimeout limit) and uses a flat 60-minute cap where duration is not cheaply available, so short dictations keep exactly today's behavior and a genuinely hung server still fails eventually.

Fixes OpenWhispr/openwhispr#976

## Changes

* src/helpers/transcriptionTimeout.js: new shared module with the floor, per-second, flat-cap and ceiling constants plus computeTranscriptionTimeoutMs()
* src/helpers/whisperServer.js: /inference request timeout scales with the converted WAV buffer duration
* src/helpers/parakeetWsServer.js: per-call timeout computed from the sample buffer instead of the flat 5-minute constant
* src/helpers/llamaServer.js: post-processing timeout uses the 60-minute flat cap since no audio duration exists there
* src/helpers/diarization.js: per-run timeout derived from the WAV file size, falling back to the flat cap
* test/helpers/transcriptionTimeout.test.js: floor, scaling, ceiling and invalid-input coverage

### DIFF
--- a/src/helpers/diarization.js
+++ b/src/helpers/diarization.js
@@ -16,6 +16,7 @@ const {
   transcriptsLooselyOverlap,
   buildMergedCandidates,
 } = require("./transcriptText");
+const { computeTranscriptionTimeoutMs } = require("./transcriptionTimeout");
 
 const DIARIZATION_TIMEOUT_MS = 3600000; // 60 minutes
 const POST_MERGE_CONTEXT_WINDOW_MS = 6000;
@@ -318,6 +319,15 @@ class DiarizationManager {
       wavPath,
     });
 
+    // 16 kHz mono s16le WAV is 32000 bytes/s; scale with length but never below the 60-minute floor.
+    let timeoutMs = DIARIZATION_TIMEOUT_MS;
+    try {
+      timeoutMs = Math.max(
+        DIARIZATION_TIMEOUT_MS,
+        computeTranscriptionTimeoutMs(fs.statSync(wavPath).size / 32000)
+      );
+    } catch {}
+
     return new Promise((resolve) => {
       let stdout = "";
       let stderr = "";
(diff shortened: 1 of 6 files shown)


gabrielste1n (approved):
Thanks @xAlcahest — really nice catch, and the diagnosis was spot on. Local transcription sends the whole file in a single request (chunking only exists on the cloud path), so a long recording genuinely did hit one hardcoded 5-minute wall regardless of how much progress the model had made.

The floor that keeps short dictations on exactly today's budget, the 24-hour ceiling to stay under the 32-bit `setTimeout` limit, and the invalid-input coverage in the tests were all exactly right.

I pushed one follow-up commit on top before merging:

- **Reverted the `llamaServer.js` change.** That request has no audio behind it — it's the reasoning/cleanup pass, and `max_tokens` already bounds generation — so it wasn't part of #976. Going 5 → 60 minutes there only meant a wedged local LLM took 12× longer to surface an error.
- **Shared the `32000` bytes/sec constant.** `whisperServer` and `diarization` were each hardcoding it; both derive from `convertToWav(..., { sampleRate: 16000, channels: 1 })`, which is parameterized, so it's now a single `PCM16_MONO_16K_BYTES_PER_SECOND` export.
- **Collapsed the two 5-minute floors in `parakeetWsServer`.** The online path already scaled by duration before this PR, so `300000` ended up defined twice in one file. Its 2× budget is deliberately unchanged — streaming decodes as the audio drains, and `ONLINE_FINISH_IDLE_TIMEOUT_MS` is the real stall guard there.

Full suite is green (664 tests). Merging once the platform builds finish — thanks again for the fix!

PR #1106, opened 2026-07-08
Update dependency acknowledgment

ofek:
https://ui.shadcn.com/docs/changelog/2026-07-base-ui-default


gabrielste1n:
Closing — upstream shadcn did switch to Base UI, but our vendored components are still the Radix versions (package.json still carries the @radix-ui deps and the components import them), so the current acknowledgment is accurate for this repo. Happy to take this if/when we migrate. Thanks!

PR #1109, opened 2026-07-09
fix(autolearn): ignore redundant same-value auto-learn syncs

### LINKED ISSUE #1080: fix(startup): AutoLearn 'Setting changed { enabled: true }' fires twice — possible double init
## Problem

Startup log v1.7.3 shows `[AutoLearn] Setting changed { enabled: true }` emitted **twice** in sequence:

```
[DEBUG] [AutoLearn] Setting changed { enabled: true }   ← first emission
... (HotkeyManager, meeting detection, qdrant startup) ...
[DEBUG] [AutoLearn] Setting changed { enabled: true }   ← second emission
```

The second emission occurs after the auth renderer syncs (`Auth state sync { isSignedIn: true }`), suggesting AutoLearn is initialized once at app start and again when the renderer process loads settings.

## Expected

AutoLearn initialized once. Duplicate init may cause double event listener registration — check for memory leak or double-transcription side effects.

## Env

- OpenWhispr: 1.7.3
- OS: Windows 11 Enterprise 26200
- Electron: 41.2.0

Thanks for the great work on OpenWhispr!

serhiileniv:
## Summary

Fixes #1080.

On startup the debug log shows `[AutoLearn] Setting changed { enabled: true }` emitted **twice**. The cause is the dual-window architecture: every renderer window syncs its auto-learn preference to the main process on mount (`useSettings.ts` → `setAutoLearnEnabled` → `auto-learn-changed` IPC), so the main overlay and the control panel each send the same value at startup. The `auto-learn-changed` handler ran its full body for every message, re-applying identical state and logging a duplicate "Setting changed" each time.

## Change

- Add a small pure resolver `applyAutoLearnSetting(current, incoming)` in `src/helpers/autoLearnSetting.js` that coerces the incoming value to boolean and reports whether it actually changed.
- Make the `auto-learn-changed` handler a no-op when the value is unchanged, so repeated mount-time syncs no longer do redundant work or emit duplicate logs.

Behaviour on a real toggle (enable ⇄ disable) is unchanged — including the existing debounce-timer/pending-data cleanup on disable.

## Testing

- `node --test test/helpers/autoLearnSetting.test.js` — 6 cases covering enable/disable transitions, idempotent same-value syncs, boolean coercion, and a regression for the #1080 dual-window double-sync. All pass.
- `prettier --check` clean on all changed files.

I don't have a full Electron runtime set up locally, so I verified the extracted logic via the unit test and by tracing the handler; I did not exercise the packaged app end-to-end.

---
🤖 Drafted with the assistance of Claude Code (reviewed by me before submitting).

### DIFF
--- a/src/helpers/autoLearnSetting.js
+++ b/src/helpers/autoLearnSetting.js
@@ -1,17 +1,6 @@
-/**
- * Pure resolver for the "auto-learn-changed" IPC sync.
- *
- * Every renderer window syncs its auto-learn preference to the main process on
- * mount (see `useSettings.ts`). With the dual-window architecture (main overlay
- * + control panel) the main process therefore receives the same value more than
- * once at startup. Returning `changed: false` for a repeated value lets the
- * handler skip redundant work and the duplicate "[AutoLearn] Setting changed"
- * log reported in #1080.
- *
- * @param {boolean} current - the main process's current auto-learn state
- * @param {*} incoming - the raw value received over IPC (coerced to boolean)
- * @returns {{ changed: boolean, enabled: boolean }}
- */
+// Pure resolver for the "auto-learn-changed" IPC sync. Coerces the raw IPC
+// value to boolean and reports whether it differs from the current state, so
+// the handler can ignore repeated same-value syncs (#1080).
 function applyAutoLearnSetting(current, incoming) {
   const enabled = !!incoming;
   return { changed: enabled !== !!current, enabled };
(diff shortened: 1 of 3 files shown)


gabrielste1n (approved):
Thank you for the clean fix and the clear root-cause analysis! Verified the dual-window mount-sync diagnosis against `useSettings.ts`, confirmed the early-return can't skip any needed disable cleanup, and ran the unit tests locally — all green. I pushed a small tidy-up (trimmed comments, folded the duplicate regression test into the idempotency case). Much appreciated! 🙏

PR #1111, opened 2026-07-09
feat: add Corti reasoning provider and route healthcare cleanup to it

xAlcahest:
## Summary
Corti ships as a BYOK medical transcription provider but was missing from the LLM side: `buildReasoningProviders()` picks registry entries up automatically, while everything that renders tabs or makes requests keeps its own hardcoded provider list (`PROVIDER_REGISTRY` in `inferenceProviders/index.ts`, the two streaming allowlists in `ReasoningService.ts`, `getAIModel()` in `providers.ts`, `CLOUD_PROVIDER_IDS` in `ReasoningModelSelector.tsx`), so a registry-only entry would be invisible in the UI and throw "Unsupported reasoning provider" at call time. This adds Corti's OpenAI-compatible Models gateway (`https://ai.eu.corti.app/v1`, EU-hosted only for now) as a working reasoning provider: the four corti-s1 models in the registry, with `corti-s1-instant` first since it skips the reasoning trace and makes the better cleanup default, a `corti.ts` inference provider, and an entry in each of those lists. Auth takes the Corti Console API key as a dedicated secret (`CORTI_API_KEY`), sent verbatim as the bearer per the Models quickstart. An earlier revision derived a base64 packing from the transcription client credentials; that encoding appeared in the gateway's OpenAPI description when this PR was first written and has since been removed from Corti's docs, so it is gone here too.

The other commits close the loop for healthcare onboarding. `startWithCorti()` only called `setCloudTranscriptionForAllScopes()`, so a user accepting Corti got Corti speech-to-text while every LLM feature kept sending text to whatever provider was configured, OpenAI by default. For PHI that is a compliance hole. The new `setCloudReasoningForAllScopes()` in `settingsStore.ts`, modeled on the transcription fan-out, routes provider, model and mode to all four LLM scopes (dictationCleanup, noteFormatting, dictationAgent, chatIntelligence), and the accept path calls it with the first Corti reasoning model from the registry when the selected data region is EU and the Corti API key is present; outside the EU the hint under the region selector disappears and LLM features keep the current provider, since the Models gateway is EU-hosted and rejects US-project credentials. The gateway documents streaming and function/tool calling, so the agent scopes are safe to route. The fan-out logic lives in a pure helper (`src/helpers/reasoningRouting.js`) with node --test coverage; declining Corti and the non-healthcare flow are untouched.

## Changes
- `src/models/modelRegistryData.json`: corti entry in cloudProviders with corti-s1-instant, corti-s1, corti-s1-mini-instant and corti-s1-mini, marked max_tokens plus temperature so requests don't fall through to the gpt-5 defaults
- `src/services/ai/inferenceProviders/corti.ts`: new provider hitting the gateway's /chat/completions, same shape as groq.ts
- `src/services/ai/inferenceProviders/index.ts`: corti registered in PROVIDER_REGISTRY
- `src/services/ReasoningService.ts`: corti entry in the keyGetters map, corti in both streaming allowlists and in the endpoint switch
- `src/services/ai/providers.ts`: corti case in getAIModel pointing createOpenAI at the gateway
- `src/config/constants.ts`: CORTI_MODELS_BASE endpoint
- `src/helpers/environment.js`, `src/helpers/ipcHandlers.js`, `preload.js`, `src/types/electron.ts`: CORTI_API_KEY secret plumbing (secure store, IPC pair, renderer exposure, types)
- `src/components/ReasoningModelSelector.tsx`: corti tab with an API key input and the console link
- `src/components/chat/useChatStreaming.ts`: corti added to the cloud provider allowlist
- `src/stores/settingsStore.ts`: setCloudReasoningForAllScopes fans LLM routing out to all four scopes; cortiApiKey held like the other secrets, and the corti setters invalidate the corti key cache so a changed credential drops the stale bearer
- `src/helpers/reasoningRouting.js`: pure helpers deriveReasoningMode, buildReasoningScopePatches and buildCortiOnboardingPayloads with the EU gate
- `src/components/onboarding/FinishStep.tsx`: startWithCorti routes every LLM scope to Corti alongside transcription when the data region is EU and the API key is set; the pitch collects the API key and shows the hint only for EU
- `test/helpers/reasoningRouting.test.js`: eleven tests covering the fan-out, the onboarding payloads, the EU gate and the missing-registry edges
- `src/locales/*/translation.json`: description keys for the four models plus the pitch hint and API key label strings in all ten locales


gabrielste1n (commented):
Reviewed with a focus on code quality and reuse. The wiring is genuinely clean — the registry entry matches the tinfoil/openai field conventions, `corti.ts` mirrors `groq.ts` exactly, all four hardcoded provider lists are updated, the icon/tab name flow automatically from the registry, all ten locales are covered, and `test/helpers/reasoningRouting.test.js` passes 10/10 locally. Narrowing `setCortiClientId/Secret` to `invalidateApiKeyCaches("corti")` is a nice improvement too.

Two things need resolving before merge, though — details inline:

1. **The bearer format is speculative and diverges from the auth code we already have.** Corti's public Models quickstart says to use the Console API key as the bearer ("OAuth 2.0 access tokens will be supported soon"); nothing public documents the `base:client_credentials:<id>:<secret>` base64 encoding, and `src/helpers/cortiAuth.js` already implements the documented OAuth flow per environment/tenant. See the comment on `buildCortiBearer`.
2. **US-region users get every LLM feature broken by onboarding.** The fan-out is unconditional while the Models gateway is EU-hardcoded and the PR itself says US credentials are rejected there. See the comment in `FinishStep.tsx`.

Smaller notes inline as well (unreachable registry guard + its tests, the arity test, and a small dedup opportunity in `getApiKey`).

src/services/ReasoningService.ts
```diff
@@ -18,6 +18,19 @@ import { getConfiguredOpenAIBase } from "./ai/openaiBase";
 import { applyThinkingSuppression } from "./ai/thinkingSuppression";
 import { clearTinfoilClientCache } from "./ai/tinfoilClient";
 
+// Corti Models gateway bearer: base64 of the credential template documented in
+// its OpenAPI spec, with only the client id/secret substituted.
+function buildCortiBearer(clientId: string, clientSecret: string): string {
```
gabrielste1n: This encoding doesn't appear anywhere in Corti's public docs. The [Models quickstart](https://docs.corti.ai/models/quickstart.md) says *"OAuth 2.0 access tokens will be supported soon. For now, use your Corti API key"* — i.e. a value the user copies from the Corti Console, not something the client derives from client id/secret. Two concrete risks:

- The literal `base` segment looks like a hardcoded tenant/realm name. The transcription side treats tenant as user-configurable (`cortiTenant`, validated in `src/helpers/cortiAuth.js`), so even if the encoding is right for default tenants, non-`base` tenants may get well-formed bearers that 401.
- The PR description says this "deserves a sanity check against a live key" — that check needs to happen **before** merge. If the format is wrong, every LLM call fails for every Corti user, and onboarding routes all four LLM scopes here.

There's also a reuse concern: `src/helpers/cortiAuth.js` already implements the documented OAuth flow (per-environment/tenant token endpoint, 5-minute expiry, cached with a refresh margin). This adds a second, divergent Corti auth mechanism that ignores `cortiEnvironment` and `cortiTenant` entirely and caches a never-expiring credential. If the base64 trick isn't officially supported, the cleaner options are an IPC that mints tokens via the existing helper (as transcription does), or collecting the actual Console API key as its own setting.

src/services/ReasoningService.ts
```diff
@@ -64,8 +77,25 @@ class ReasoningService extends BaseReasoningService {
   }
 
   private async getApiKey(
-    provider: "openai" | "anthropic" | "gemini" | "groq" | "tinfoil" | "custom"
+    provider: "openai" | "anthropic" | "gemini" | "groq" | "tinfoil" | "corti" | "custom"
   ): Promise<string> {
+    if (provider === "corti") {
```
gabrielste1n: Minor: this branch re-implements the generic flow below (cache check, IPC fetch, error log, cache set). A `corti: async () => buildCortiBearer(...)` entry in the existing `keyGetters` map would reuse the caching, logging, and the standard "Corti API key not configured" error. Not blocking — the `custom` branch sets precedent for special-casing.

src/components/onboarding/FinishStep.tsx
```diff
@@ -45,12 +47,15 @@ export default function FinishStep({
...
-    });
+    // Route all four LLM scopes to Corti too so PHI never leaves Corti; skip
+    // reasoning routing if the Corti reasoning entry is absent (Part 1 not shipped).
+    const reasoningProvider = modelRegistry.getCloudProviders().find((p) => p.id === "corti");
+    const { transcription, reasoning } = buildCortiOnboardingPayloads(
+      cortiProvider,
+      reasoningProvider
+    );
+    setCloudTranscriptionForAllScopes(transcription);
+    if (reasoning) setCloudReasoningForAllScopes(reasoning);
```
gabrielste1n: This fan-out is unconditional, but `CORTI_MODELS_BASE` is hardcoded to the EU gateway and the PR description says US project credentials are rejected there. A US healthcare user who accepts Corti gets cleanup, note formatting, dictation agent, and chat all failing at call time. Worse, the region selector directly above promises *"Choose where your audio is processed — US or EU"* while LLM text would silently go to the EU — for a flow whose stated purpose is PHI compliance, that's the opposite of what the UI promises.

`buildCortiOnboardingPayloads` already has a `reasoning: null` path — gate it on `cortiEnvironment === "eu"` (and show `llmHint` only in that case).

src/components/onboarding/FinishStep.tsx
```diff
@@ -45,12 +47,15 @@ export default function FinishStep({
...
 
   const startWithCorti = () => {
-    setCloudTranscriptionForAllScopes({
-      useLocalWhisper: false,
-      cloudTranscriptionMode: "byok",
-      cloudTranscriptionProvider: "corti",
-      cloudTranscriptionModel: cortiProvider?.models[0]?.id,
-    });
+    // Route all four LLM scopes to Corti too so PHI never leaves Corti; skip
+    // reasoning routing if the Corti reasoning entry is absent (Part 1 not shipped).
```
gabrielste1n: Two notes on this guard:

- "(Part 1 not shipped)" is review-facing phrasing that will confuse readers after merge — this PR ships both parts together, so reword to the actual condition.
- Since this same PR adds the corti entry to the bundled `modelRegistryData.json`, the missing-provider case can never trigger post-merge — it's dead code, and three of the ten tests cover it. If you adopt the environment gating suggested below, the null path becomes real and this resolves itself; otherwise consider dropping the guard.

test/helpers/reasoningRouting.test.js
```diff
@@ -0,0 +1,116 @@
...
+test("corti reasoning provider with empty models yields no reasoning payload", async () => {
+  const { buildCortiOnboardingPayloads } = await load();
+  const { reasoning } = buildCortiOnboardingPayloads(
+    { id: "corti", models: [{ id: "corti-transcribe" }] },
+    { id: "corti", models: [] }
+  );
+  assert.equal(reasoning, null);
+});
+
+test("buildCortiOnboardingPayloads takes no environment argument", async () => {
```
gabrielste1n: Asserting `fn.length === 2` is brittle (any default parameter breaks it) and verifies no behavior — and per the `FinishStep` comment, an environment argument is arguably exactly what this function needs. Suggest dropping this test.

src/services/ReasoningService.ts
xAlcahest: The base64 recipe was in the gateway's OpenAPI securityScheme when this was written ("The token is a Base64-encoded client credential string in the format `base:client_credentials:<client_id>:<client_secret>`") — Corti has since removed it from the page, which settles it. Switched to the Console API key as its own secret (`CORTI_API_KEY`), sent verbatim as the bearer; `buildCortiBearer` is gone and tenant/environment no longer matter for LLM auth.

src/services/ReasoningService.ts
xAlcahest: Done — corti is now a plain keyGetters entry on the generic flow.

src/components/onboarding/FinishStep.tsx
xAlcahest: Restored: the reasoning payload is gated on `cortiEnvironment === "eu"` and the hint only renders there. The fan-out also requires the API key to be present, so accepted users are never routed to an unconfigured provider.

src/components/onboarding/FinishStep.tsx
xAlcahest: Reworded; with the EU gate back the null path is real again and its tests cover live behavior.

test/helpers/reasoningRouting.test.js
xAlcahest: Dropped.

PR #1112, opened 2026-07-09
feat(onboarding): route cleanup to Corti when healthcare users opt in

xAlcahest:
## Summary
`startWithCorti()` in `FinishStep.tsx` only called `setCloudTranscriptionForAllScopes()`, so a healthcare user accepting Corti during onboarding got Corti speech-to-text while dictation cleanup and note formatting kept sending text to whatever LLM provider was configured, OpenAI by default. For PHI that is a compliance hole. This adds `setCleanupForAllScopes()` to `settingsStore.ts`, modeled on the transcription fan-out: it routes provider, model, cloud mode and mode to both the dictationCleanup and noteFormatting scopes, and the Corti accept path now calls it with the first Corti reasoning model from the registry. The fan-out logic lives in a pure helper (`src/helpers/cleanupRouting.js`) with node --test coverage; declining Corti and the non-healthcare flow are untouched.

The dictationAgent and chatIntelligence scopes are deliberately left alone: tool calling on the corti-s1 gateway is unverified, so routing the agents there could break them. Worth a follow-up decision.

Depends on #1111.

## Changes
- `src/stores/settingsStore.ts`: setCleanupForAllScopes fans cleanup routing out to dictationCleanup and noteFormatting, mode included so the settings tabs follow
- `src/helpers/cleanupRouting.js`: pure helpers deriveCleanupMode, buildCleanupScopePatches and buildCortiOnboardingPayloads
- `src/components/onboarding/FinishStep.tsx`: startWithCorti routes cleanup to Corti alongside transcription, skipped when the registry has no Corti reasoning provider
- `test/helpers/cleanupRouting.test.js`: nine tests covering the fan-out, the onboarding payloads and the missing-registry edges


xAlcahest:
Folded into #1111 so the provider and the onboarding routing land as one reviewable unit.

PR #1115, opened 2026-07-09
Fix(tinfoil): Make the model list dynamic

danielmccannsayles:
# Summary 

Tinfoil is unique in that our models are more ephemeral than most of the providers (unlike openai, ant, etc.). It's closer to custom/self-hosted models, which fetch from /models, but it's also different from them in that we know the shape of what we should get, eg `max_tokens` and per-model `reasoning`.

This PR changes tinfoil handlers to refresh the list of active models every hour.  It saves these in localStorage, so they are persistent. After fetching the active list, we set this as the current registry for tinfoil. If this fails, we keep the registry the same. In the unfortunate case that a user was using this model, the model is switched to a different tinfoil model and a toast popup alerts them. 

The tinfoil model sync runs in the background, and is only triggered if someone is using tinfoil, so it doesn't fetch for users who don't use TInfoil. The model registry comes with a set of Tinfoil models that will probably be accurate for a while, so users don't have to wait to see what's available

# Problem
Tinfoil's model list was hardcoded in `modelRegistryData.json` and had already drifted (it still listed `deepseek-v4-pro`, which the API no longer serves). Now it's fetched from `https://inference.tinfoil.sh/v1/models`, filtered to `type: "chat"` models that expose `/v1/chat/completions`.

# Other things
Curated translated descriptions for the existing models are kept as the api only returns english descriptions.

The tinfoil model list lives in localStorage on the renderer instead of on the main process because it's needed immediately (The main process actually does the fetch). This follows the pattern in settingsStore for user preferences.

Screenshot 2026-07-09 at 4 08 28 PM

### DIFF
--- a/preload.js
+++ b/preload.js
@@ -381,6 +381,7 @@ contextBridge.exposeInMainWorld("electronAPI", {
   proxyCortiTranscription: (data) => ipcRenderer.invoke("proxy-corti-transcription", data),
   getTinfoilKey: () => ipcRenderer.invoke("get-tinfoil-key"),
   saveTinfoilKey: (key) => ipcRenderer.invoke("save-tinfoil-key", key),
+  getTinfoilChatModels: () => ipcRenderer.invoke("get-tinfoil-chat-models"),
 
   // Custom endpoint API keys
   getCustomTranscriptionKey: () => ipcRenderer.invoke("get-custom-transcription-key"),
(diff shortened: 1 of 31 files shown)


gabrielste1n (approved):
Reviewed end-to-end: fixed provider routing for non-cleanup scopes, threaded the resolved note-formatting provider through note actions and titles, surfaced the dictation panel for the model-retired toast, and verified against the live Tinfoil catalog. Typecheck, lint, and full test suite pass.

gabrielste1n:
thanks legend @danielmccannsayles !

PR #1120, opened 2026-07-10
feat(tinfoil): add batch STT

danielmccannsayles:
# Summary

Previously the only tinfoil STT model configured was our realtime streaming. With tinfoil selected it would either fail to batch (when doing eg audio upload) or it would silently fall back to somewhere else. Both of these were bad. 

This PR makes it so that Tinfoil now offers a 'voxtral' model, that works as realtime streaming or batch when needed. 

New UI:
Screenshot 2026-07-09 at 11 35 41 PM

## Tinfoil special handling in processWithOpenAIAPI

This Pr adds some janky guards to the function `processWithOpenAIAPI`. There is a reason for this. 

Right now failed endpoint resolution defaults to openai. This PR adds two guards to it, just to be sure. 
Also, streaming fallback defaults to openwhispr cloud. This PR makes it so that byok refuses to fall back to openwhispr cloud. If this fallback is intended it should probably be opt-in and toggleable.

_A better fix for the first problem is to move all the proxied providers (mistral, xai, etc.) above the endpoint resolution. Left for a future PR since it's hard to test this area & seems out of scope_

## Streaming/ Batching ontology, and confusion there-in

Right now there's no difference between streaming & batching models under STT. This is a problem if you want to, for instance, require streaming on dictations(I really like seeing the stream come in so I know it's working😄). 

What I do for tinfoil in this Pr is just offer one 'voxtral' model, which can behave as streaming for dictation, or as batch for audio upload. Probably there's a cleaner ontology/way of surfacing this to users that could be done.


gabrielste1n (approved):
Reviewed end-to-end and addressed the outstanding review feedback (self-hosted routing guard, error codes preserved across the proxy IPC boundary, 429/5xx classification, custom dictionary forwarding, signed-out cloud fallback fix, and i18n cleanup). Local quality gate green: format, typecheck, 218 tests (0 fail), i18n consistent. Verified the batch model (voxtral-small-24b) serves /v1/audio/transcriptions on Tinfoil's live catalog.

PR #1133, opened 2026-07-11
Toggle the control panel from the tray icon

sochotnicky:
Clicking the tray icon always (re)opened the control panel; clicking again did nothing. Now it toggles — hide if it's visible, show and focus otherwise.

This mainly helps on Linux with multiple workspaces: a window parked on another workspace still reports as visible, so the first click hides it and the next brings it to the current workspace (worst case two clicks) instead of it staying out of sight. macOS is unchanged — its tray uses the context menu, not a click.


gabrielste1n (approved):
Thanks @sochotnicky — nice quality-of-life fix, especially for multi-workspace Linux setups. While reviewing we extended it slightly: the tray context-menu item now toggles too, with a dynamic Open/Hide Control Panel label across all 10 locales. Appreciate the contribution!

PR #1134, opened 2026-07-11
Sync zoom across all windows and scale the popups

sochotnicky:
Two issues with zoom:

- Ctrl +/-/0 only zoomed the focused window, so the notification, transcription-preview and agent popups stayed at 100%.
- Those windows are fixed-size, so once they did zoom the content clipped.

This routes the View-menu zoom items through a small `ZoomManager` that keeps one shared zoom level, applies it to every window (and persists it across restarts), and scales the fixed-size popups by the zoom factor so their frames grow to fit. Also adds the Ctrl/Cmd+`=` accelerator.

Translations are auto-generated (it's a common text so should be OK, and probably better than no text at all)

### DIFF
--- a/src/helpers/menuManager.js
+++ b/src/helpers/menuManager.js
@@ -2,6 +2,36 @@ const { Menu } = require("electron");
 const { i18nMain } = require("./i18nMain");
 
 class MenuManager {
+  // Zoom items route through onZoom(window, direction) so a single app-managed
+  // zoom level stays in sync across every window (see ZoomManager). Plain
+  // `role` items would only zoom the focused webContents and emit no event.
+  static zoomMenuItems(onZoom) {
+    return [
+      {
+        label: i18nMain.t("menu.actualSize"),
+        accelerator: "CommandOrControl+0",
+        click: (_item, window) => onZoom?.(window, "reset"),
+      },
+      {
+        label: i18nMain.t("menu.zoomIn"),
+        accelerator: "CommandOrControl+Plus",
+        click: (_item, window) => onZoom?.(window, "in"),
+      },
+      {
+        // Common habit: Ctrl/Cmd + "=" (the unshifted "+" key).
+        label: i18nMain.t("menu.zoomIn"),
+        accelerator: "CommandOrControl+=",
+        visible: false,
+        click: (_item, window) => onZoom?.(window, "in"),
+      },
+      {
+        label: i18nMain.t("menu.zoomOut"),
+        accelerator: "CommandOrControl+-",
+        click: (_item, window) => onZoom?.(window, "out"),
+      },
+    ];
+  }
+
   static setupMainMenu(onOpenSettings) {
     if (process.platform === "darwin") {
       const template = [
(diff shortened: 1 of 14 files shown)


xAlcahest (commented):
Nice approach overall: one shared level, applied per window, with the fixed-size popups scaled by the factor. That's the right shape for this, and the i18n side is complete (all 10 locales have the new `menu.*` keys).

Two things I'd want changed before this goes in.

**Zoom range is tighter than intended.** `MIN/MAX_ZOOM_LEVEL = ±3` with `factor = 1.2^level` gives 58%–173%, but the comment right above says `~50%..300%`. The `{ role: "zoomIn" }` items this replaces used Electron's built-in role, whose documented default limits are 300% and 50% of original size, so anyone who had the control panel at 250% now hits a hard stop at 173% with Ctrl+`+` doing nothing past it. Since low-vision users are the main audience for this, I'd raise `MAX_ZOOM_LEVEL` to 6 — `1.2^6` is 298.6%, right at Electron's own ceiling — and leave `MIN_ZOOM_LEVEL` at -3, which is 58% and already inside the 50% floor. `-6` would be 33.5%, below that floor, so Chromium would clamp it back and reintroduce the desync described in the next point.

**Ctrl+wheel past the cap desyncs the origin window.** `setLevel` skips `origin` on the assumption it's already at the requested level, which only holds when no clamping happened. If the shared level is already at the cap and you keep wheeling on the control panel, Chromium takes that webContents to level 4, `setLevel(4, cp)` clamps back to 3, `changed` comes out false so there's no persist and no `_onChange`, and the loop skips the origin. The control panel sits at 200% while every popup stays at 173%, and it doesn't heal until you happen to use a menu item. Applying to the origin too when `clamped !== level` covers it.

Smaller stuff, none of it blocking:

- `applyZoomCommand` returns early when `browserWindow` is null. On macOS the View menu is the app menu and stays installed after the control panel closes, so Cmd+0 with only popups on screen does nothing. Falling back to `getLevel()` as the base would make it work regardless of focus.
- The scaled clamp limits in `windowManager.js` come out fractional (`96 * 1.0954 = 105.16`), so the "bounds unchanged" fast path never matches at non-100% zoom and every ResizeObserver tick issues a real `setBounds`. Rounding the scaled limits fixes it. Worth noting Electron doesn't throw on fractional bounds, it truncates, so this is wasted work rather than a crash.
- The agent animation's height floor stays at an unscaled 200 while the target is scaled, so at zoom-out the window can't reach its target.
- The `isVisible()` guard skips notification windows during their pre-show gap, which can leave an old-factor frame around new-zoom content. Same clipping the PR fixes elsewhere.
- The scaled main overlay size isn't clamped to the work area, so at max zoom the expanded panel runs off the bottom on a 768px-high display.
- `writeFileSync` fires on every wheel notch alongside four `setBounds` calls, all on the main process during a continuous gesture. Debouncing the persist would be enough.

`ZoomManager` is pure enough to unit test the way `dockPolicy.js` is in `test/helpers/dockPolicy.test.js`, and the clamp and origin-skip cases above are exactly what tests would pin down.

PR #1135, opened 2026-07-11
Add an echo-cancellation toggle for meeting recording

### LINKED ISSUE #1025: AEC helper causing missed microphone recording in meetings when used with headphones
**Describe the bug**
I had a hard time where my microphone was working for note taking but in a meeting it was being ignored and not recorder. I found out this was related to my using BT headphones. Rebuilding the appimage without the aec helper made the problem go away

*might* be related to me having 2 mics (one on the headphones - unused, another external mic actually used).

**To Reproduce**

1. Connect bluetooth headphones + 
2. Start meeting recording with some audio in the background and talk

**Expected behavior**
My own voice is captured during the meeting as well

**Screenshots**
None

- OS: Linux, NixOS

**Additional context**
I have my own fix where I added a toggle for AEC helper: https://github.com/sochotnicky/openwhispr/tree/meeting-aec-toggle

I can open a PR with that if there's interest

sochotnicky:
Adds an opt-in echo-cancellation toggle to meeting settings (default off), for cases where system-audio playback leaks into the mic track during a recording. Second commit adds the translation keys across all locales (generated, hopefully not too bad)

Fixes #1025


xAlcahest (approved):
This closes #1025. Before this change `startMeetingAec` ran whenever system audio was supported and the helper binary was available, which is exactly the failure mode from the issue. Now it is gated on the flag from the start options, and the gate holds at all three call sites (warm reuse, local mode, realtime).

What I verified in the code:

- Default is genuinely off for existing users: `readBoolean("meetingAecEnabled", false)` leaves an absent key false, there is no migration, and on the main side `options.aecEnabled === true` treats a missing flag as off, so an older renderer or any caller that does not pass the flag cannot accidentally enable it.
- The toggle reaches the sidecar: the setting flows into the `meeting-transcription-start` options, `meetingAecRequested` is set at the top of the handler before any `startMeetingAec` call, and with AEC off the mic chunks take the pre-existing echo-leak path unchanged.
- All 10 locales carry `settingsPage.transcription.aec.toggle.{title,description}` with identical structure and no placeholders, and the translations read fine in the languages I can judge.
- Dictation is untouched, `meetingAecManager` stays meeting-only.

Two notes, neither a request: the setting is read at meeting start, so a mid-meeting flip applies from the next recording (the inline comment already covers this), and it is worth adding `Fixes #1025` to the PR description so the issue closes on merge.

Default off is the right call here. The commit message gets the asymmetry exactly right: silent total mic loss is a worse failure than residual echo. Mirroring the `meetingSileroEnabled` pattern kept this easy to review, thanks for a clean contribution.

PR #1143, opened 2026-07-11
feat: reprocess saved transcripts with AI cleanup

adamholter:
## What changed

- adds a History action that reruns AI cleanup against the exact saved raw transcript
- updates the existing row only after valid reasoning output and a successful database write
- preserves raw text byte-for-byte and keeps failures non-destructive
- surfaces unavailable-provider and persistence errors without optimistic UI drift
- adds all current locale strings and focused provider/database/UI coverage

## Why

Retry currently covers transcription failures, but users cannot deliberately rerun cleanup after changing models, prompts, or settings. This adds that narrow workflow without retranscribing audio or creating duplicate history rows.

## Validation

- 263 passing tests, 0 failures, 5 macOS-expected Linux skips
- focused provider failure, empty output, missing raw text, and database failure coverage
- formatting, TypeScript, i18n, and renderer build pass
- headless fail-then-success flow passed with exact write-payload assertions and zero console errors


gabrielste1n:
So sorry I'm only seeing this now. I believe we actually have this in the interface now. This was a great idea clearly.

adamholter:
No problem, glad to have it here!

PR #1144, opened 2026-07-11
feat: add dedicated translation dictation mode

adamholter:
## What changed

- adds a first-class translation dictation route that is distinct from cleanup and voice-agent behavior
- supports up to five saved target languages, fast active-target switching, and a dedicated configurable shortcut
- uses an instruction-resistant translation prompt and works even when ordinary cleanup is disabled
- preserves source and translated transcript versions and labels the target in History
- fails closed when translation is unavailable or returns empty output, so untranslated source text is never silently pasted
- supports globalShortcut plus the existing macOS/Linux native-listener, mouse, GNOME, and preload/IPC plumbing
- updates all current locales

## Why

Translation is possible today only as an ad hoc agent request. A dedicated mode makes source detection, target selection, shortcut behavior, history provenance, and failure handling predictable.

## Validation

- 270 passing tests, 0 failures, 5 macOS-expected Linux skips
- formatting, TypeScript, i18n, and renderer build pass
- focused target normalization, routing precedence, empty-output rejection, history persistence, and native-listener conflict coverage
- headless end-to-end flow passed with cleanup disabled, success and provider-failure paths, target switching, shortcut persistence, history provenance, and zero console errors
- streaming audit verifies failed translation cannot fall through to source text and batch fallback output is not translated twice

No visible Electron harness is used by this PR's current validation path.


xAlcahest (commented):
Hey @adamholter, since Gabe went with #1189 for translation mode, I wanted to leave a proper note here rather than let this close silently.

The short version of why: #1189 had already built translation as a separate LLM scope (own provider and model, same pattern as the app's other scopes), with the prompt running through the localized, user-editable pipeline, plus the cleanup-then-translate chain and translation-aware history retry. Picking one base was mostly about which architecture the rest would build on.

Your PR directly improved the result though: the multi-target quick switcher, the injection-resistant prompt wording and the visible failure feedback all made it into #1189, adapted from here. Thanks for building this, and hope to see you around the repo.

Issue #1087, opened 2026-07-07
Empty response in chat.


jtoronto:
**Describe the bug**
LLM responses show up as empty bubbles. 

**To Reproduce**

1. Go to any chat window. 
2. Type in a prompt/request in the text box and hit enter. 
3. A "thinking..." chat bubble appears, and then eventually it changes to an empty bubble. 

**Expected behavior**
Expect to see the LLM response. 

**Screenshots**

<img width="948" height="1163" alt="Image" src="https://github.com/user-attachments/assets/53d199d6-bd08-471b-a7d8-b8979173292d" />

**Desktop (please complete the following information):**

- OS: Kubuntu 26.04 LTS
- Firefox v152
- Openwhisper 1.7.3

This issue is also occurring on my Macbook Pro running Tahoe 26.4.1

**Additional context**
Using Mimo v2.5 free from Opencode-Zen. Same model is configured for dictation cleanup and note formatting and works fine. I've tried other models and it doesn't seem to make a difference. 

Debug log output:

```
[2026-07-07T14:15:28.595Z] [DEBUG] Debug logging enabled {
  "logFile": "/home/joey-dev/.config/open-whispr/logs/debug-2026-07-07T14-15-28-594Z.log"
}
[2026-07-07T14:15:28.595Z] [INFO] System Info {
  "platform": "linux",
  "nodeVersion": "v24.14.0",
  "electronVersion": "41.2.0",
  "appPath": "/opt/OpenWhispr/resources/app.asar",
  "userDataPath": "/home/joey-dev/.config/open-whispr",
  "resourcesPath": "/opt/OpenWhispr/resources"
}
[2026-07-07T14:15:35.870Z] [DEBUG][meeting] Process list refreshed {
  "count": 563
}
[2026-07-07T14:15:45.783Z] [DEBUG] qdrant stdout {
  "data": "2026-07-07T14:15:45.783405Z  INFO actix_web::middleware::logger: 127.0.0.1 \"POST /collections/notes/points/search HTTP/1.1\" 200 229 \"-\" \"qdrant-js/1.17.0\" 0.015034"
}
[2026-07-07T14:16:05.868Z] [DEBUG][meeting] Process list refreshed {
  "count": 565
}
[2026-07-07T14:16:06.543Z] [INFO][meeting] Updating detection preferences {
  "audioDetection": true
}
[2026-07-07T14:16:08.564Z] [DEBUG][ipc] model-get-all called
[...]

PR #1146, opened 2026-07-12
fix: isolate Chat routing and preserve self-hosted auth

cgkades:
## Summary

- Resolve Chat routing exclusively from Chat-owned settings so Dictation Cleanup self-hosted configuration cannot override BYOK, built-in provider, local, or enterprise Chat.
- Preserve explicit Chat self-hosted routing, including its optional API key, in both the direct Chat Completions stream and the AI SDK/tool-enabled stream.
- Centralize the Chat route decision and add focused regression coverage for Custom, Gemini, local, enterprise, self-hosted precedence, blank URLs, and authenticated/unauthenticated self-hosted endpoints.
- Leave the existing onboarding/Corti reasoning-mode helper unchanged.

This addresses the BYOK Custom and authenticated self-hosted Chat cases documented in #1087. Model discovery could succeed because it used the configured key, while Chat either inherited the unrelated Dictation Cleanup LAN endpoint or discarded the self-hosted Chat key during inference.

## Tests

- `node --test test/helpers/chatRouting.test.js test/helpers/reasoningRouting.test.js`
- `npm run format:check`
- `npm run typecheck`
- `npm run build:renderer`
- `git diff --check`
- `npm test` — 28 passed; 7 environment-dependent tests could not run because the local Electron/native binaries are unavailable. The same 7 tests fail on the unchanged base for this local installation.


cgkades:
A smaller fix would be possible here. The minimal version would remove the fallback to `this.isLanCleanupMode()` from both Chat streaming paths and determine self-hosted Chat routing only from `config.lanUrl`:

```ts
const lanUrl = config.lanUrl?.trim();
const isLanChat = !!lanUrl;
```

Authenticated self-hosted Chat would then load:

```ts
apiKey = config.customApiKey?.trim() || "";
```

And the local OpenAI-compatible AI SDK transport would use:

```ts
createOpenAI({ apiKey: apiKey || "no-key", baseURL }).chat(model);
```

That would fix the immediate issue with a relatively small diff.

I chose to extract `resolveChatRoute()` because the two streaming implementations previously duplicated their provider classification and endpoint-selection logic. Keeping that logic in one pure helper prevents the direct Chat Completions path and the AI SDK/tool-enabled path from drifting apart. It also allows the important routing combinations, Custom, built-in providers, local, enterprise, authenticated self-hosted, and unauthenticated self-hosted, to be tested without having to mock Electron, the settings store, streaming fetches, and the AI SDK.

This is intended as a middle ground between a minimal inline fix and a larger redesign where the UI passes a fully explicit route object to the service. If the maintainers would prefer the smallest possible patch, I’m happy to simplify this to the direct inline changes while preserving the authentication fix.

gabrielste1n:
Heads up: part of this PR just landed on main via #1154 — the chat agent's self-hosted API key is now forwarded in both processTextStreaming and processTextStreamingAI (resolved the same way as the non-streaming lan provider), and providers.ts's local factory honors a provided key.

The other half of this PR — isolating chat routing from Dictation Cleanup's self-hosted settings (dropping the isLanCleanupMode consultation) — is still unaddressed on main and remains valuable. Suggest rebasing onto main and narrowing the PR to the route-isolation change plus its regression tests; expect textual conflicts in ReasoningService.ts and providers.ts from the merged key fix.

gabrielste1n (approved):
Verified the core premise: processTextStreaming/processTextStreamingAI are called only from useChatStreaming, so chat routing must not consult dictation-cleanup settings — resolveChatRoute fixes a real cross-scope leak (cleanup self-hosted config hijacking BYOK/provider/local chat). Brought up to date with main twice (self-hosted key forwarding, then #1103's lan suppression + reasoning-field retry — composed the two so suppression keys off the chat route). The 7 regression tests cover the precedence matrix. Full suite (323 tests), typecheck, lint, format pass.

PR #1153, opened 2026-07-12
fix(audio): fall back when the preferred microphone remains muted

### LINKED ISSUE #1152: Windows: built-in microphone preference selects muted Realtek device instead of working default headset
## Environment

- OpenWhispr 1.7.5
- Windows
- Windows default input: Headset Microphone (Plantronics Blackwire 5220 Series)
- Built-in input: Microphone Array (Realtek(R) Audio)

## Reproduction

1. Enable both the Plantronics and Realtek microphones.
2. Set Plantronics as the Windows default input.
3. Leave OpenWhispr's Prefer Built-in Microphone setting enabled.
4. Start dictation.

## Actual behavior

OpenWhispr explicitly selects the Realtek built-in microphone instead of the Windows-default Plantronics headset.

Chromium returns a live but muted Realtek track:

```text
Recording started with microphone {
  label: 'Microphone Array (Realtek(R) Audio)',
  deviceId: '<redacted>',
  sampleRate: 48000,
  channelCount: 2,
  muted: true,
  readyState: 'live'
}
```

The recovery path then selects the same Realtek device again. Recorded samples remain zero and dictation ends with “No audio detected.”

```text
Speech gate skipped transcription {
  reason: 'silence',
  peakRms: '0.0000',
  peakAmplitude: '0.0000'
}
```

## Workaround

Disabling Realtek in Windows Recording Devices causes OpenWhispr to use the Windows-default Plantronics microphone successfully.

The problem survives reinstalling OpenWhispr and recreating the full application profile.

## Expected behavior

OpenWhispr should preserve the existing built-in preference when that microphone is healthy.

If the preferred microphone remains muted after a short readiness period, OpenWhispr should:

1. stop the unusable stream;
2. retry once using the unpinned system-default microphone;
3. verify that the fallback is usable;
4. avoid repeatedly reopening the rejected built-in during the same stable session;
5. surface a clear error if the fallback is also unusable.

## Additional context

This issue is about recovery from an unusable preferred device. It does not propose removing or changing the existing Prefer Built-in Microphone policy.

nishankpt:
## Problem

With Prefer Built-in Microphone enabled, OpenWhispr pins the built-in microphone using an exact device ID.

On the affected Windows system, Chromium returned a live but persistently muted Realtek track. The existing readiness check waited 600 ms, cleared the cached device ID, and requested fresh constraints.

Fresh constraint generation immediately rediscovered and pinned the same Realtek device. The retry stream was not health-checked, and a failed retry could return the original silent stream.

Fixes #1152.

## Fix

- Retain a healthy preferred microphone without adding latency.
- Give an initially muted track 600 ms to emit `unmute`.
- Stop a missing, ended, or persistently muted preferred stream.
- Retry exactly once with unpinned system-default constraints.
- Health-check the fallback before recording.
- Stop and surface an unusable fallback instead of continuing with silence.
- Avoid reopening a rejected automatically selected built-in on every subsequent recording in the stable session.
- Reconsider devices after a `devicechange` event.
- Share the acquisition path between regular and streaming dictation.
- Preserve permission errors and generic acquisition errors for explicit microphone selections.
- Retain the existing default fallback for stale explicit device IDs.

## Tests

Using Node 24.18:

- Focused microphone/device tests: 21 passed, 0 failed
- Full suite: 275 total, 267 passed, 8 skipped, 0 failed
- ESLint: passed
- Prettier: passed
- Type checking: passed
- i18n validation: passed
- Renderer production build: passed
- `git diff --check`: passed

On Windows:

- Focused tests on the hardware-tested core commit: 19 passed, 0 failed
- ESLint: passed
- The full Windows suite had one unrelated existing archive-extraction failure, which reproduced when run independently and is unchanged by this branch.

## Windows validation

Tested using the native Windows development build with both Realtek and Plantronics enabled.

With Plantronics configured as the Windows default:

- OpenWhispr detected that the preferred Realtek stream remained unavailable.
- It logged fallback to the system-default microphone.
- The fallback acquired Plantronics as a live, unmuted track.
- Four repeated recordings remained on Plantronics and contained audible voice.

With Plantronics disconnected:

- The system-default fallback resolved to Realtek.
- Realtek was live, unmuted, and produced audible recordings.
- Two repeated recordings also succeeded.

After reconnecting Plantronics:

- A device-change event caused OpenWhispr to reconsider Realtek.
- Realtek had recovered and was retained, matching the existing built-in preference.
- Physically muting Plantronics throughout a recording confirmed that the Realtek label matched the actual capture source.

After restarting the app, healthy Realtek continued recording successfully.

Audio capture and playback were validated. End-to-end local transcription was not completed because the development Whisper service failed after recording. The microphone capture had already completed successfully and the saved recordings contained audible speech.

## Scope

This change covers microphone acquisition for regular and streaming dictation startup.

It does not change the built-in microphone policy, device classification, microphone settings UI, meeting recording, or recovery from a device disappearing during an active recording.

Windows validation used the development build rather than a packaged installer.

### DIFF
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -441,7 +441,7 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
       },
       logger,
       {
-        fallbackOnAcquisitionError: Boolean(pinnedDeviceId),
+        fallbackOnAcquisitionError: isBuiltInPreference,
         onFallbackSuccess: ({ reason }) => {
           // A health failure or stale ID is stable enough to avoid retrying the same auto-picked
           // built-in on every recording. Generic acquisition failures may be transient, and an
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -1,7 +1,7 @@
 import ReasoningService from "../services/ReasoningService";
 import { API_ENDPOINTS, buildApiUrl, normalizeBaseUrl } from "../config/constants";
 import logger from "../utils/logger";
-import { isBuiltInMicrophone } from "../utils/audioDeviceUtils";
+import { findBuiltInMicrophone } from "../utils/audioDeviceUtils";
 import {
   isSecureEndpoint,
   isAzureOpenAIEndpoint,
@@ -15,7 +15,6 @@ import {
   recordLocalSpeechWindow,
 } from "./localSpeechGate";
 import { reacquireIfDead } from "./micTrackHealth";
-import { isStaleDeviceError } from "./staleMicDevice";
 import { shouldSaveDiscardedRecording } from "./discardedRecording";
 import {
   getSettings,
@@ -208,10 +207,14 @@ class AudioManager {
     };
     window.addEventListener("api-key-changed", this._onApiKeyChanged);
 
+    this.cachedMicDeviceId = null;
+    this.unavailableBuiltInMicDeviceId = null;
+
     // Invalidate the pinned mic device when the OS adds/removes/suspends inputs.
     // Otherwise wake-after-idle keeps requesting a stale deviceId that yields silence.
     this._onDeviceChange = () => {
       this.cachedMicDeviceId = null;
+      this.unavailableBuiltInMicDeviceId = null;
       this.micDriverWarmedUp = false;
     };
     navigator.mediaDevices?.addEventListener?.("devicechange", this._onDeviceChange);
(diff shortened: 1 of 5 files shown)


gabrielste1n:
Thanks for the detailed diagnosis — this exact failure mode (a live but persistently muted pinned mic being re-discovered on retry) was fixed in [v1.7.6](https://github.com/OpenWhispr/openwhispr/releases/tag/v1.7.6) via #1204: when the retry is still silent the recorder hops to the OS default input, remembers the rejected device for the session so it isn't re-pinned, and fails fast with a "Microphone Muted" error if no input delivers audio. Closing as superseded — please give 1.7.6 a try and open an issue if your Realtek case still misbehaves.

Issue #970, opened 2026-06-17
Self-hosted transcription mode is ignored by retry/upload transcription paths


braden-dev:
**Describe the bug**
Self-hosted transcription is not honored consistently across OpenWhispr transcription paths. Live dictation can use the configured self-hosted endpoint, but retry/upload or other transcription flows still appear to use legacy BYOK/cloud provider settings.

This can cause OpenWhispr to send a request to OpenAI/Groq/custom cloud settings even when `transcriptionMode` is set to `self-hosted`.

**To Reproduce**
1. Configure transcription mode as `Self-hosted`.
2. Set the self-hosted transcription URL to `http://localhost:5001`.
3. Leave previous BYOK/cloud transcription settings configured, for example:
   - `cloudTranscriptionMode = byok`
   - `cloudTranscriptionProvider = groq` or `custom`
   - `cloudTranscriptionBaseUrl = https://api.groq.com/openai/v1` or blank/custom
4. Use a transcription path such as retrying a saved transcription or standard audio transcription.
5. See an API-key/cloud-provider error such as:

```text
custom API Error: 401 {
  "error": {
    "message": "You didn't provide an API key...",
    "type": "invalid_request_error"
  }
}
```

**Expected behavior**
When `transcriptionMode = self-hosted` and `remoteTranscriptionUrl = http://localhost:5001`, all transcription flows should use the self-hosted endpoint:

```text
http://localhost:5001/audio/transcriptions
```

Self-hosted transcription should not require a cloud API key and should not fall back to OpenAI/Groq/custom cloud provider settings.

**Screenshots**
<img width="1395" height="1025" alt="Image" src="https://github.com/user-attachments/assets/fcfe8326-ed13-4ff6-a764-f88fc0b4e624" />

**Desktop (please complete the following information):**
 - OS: Linux (Fedora 44)
 - Browser: N/A, OpenWhispr desktop/Electron app
 - Version: OpenWhispr desktop app, version unknown

**Smartphone (please complete the following information):**
 - Device: N/A
 - OS: N/A
 - Browser: N/A
 - Version: N/A

**Additional context**
Current relevant settings:

```text
transcriptionMode = self-hosted
remoteTranscriptionUrl = http://localhost:5001

cloudTranscriptionBaseUrl = https://api.openai.com/v1
cloudTranscriptionMode = byok
cloudTranscriptionModel = gpt-4o-mini-transcribe
cloudTranscriptionProvider = openai
```

[...]

PR #1154, opened 2026-07-13
fix(transcription): honor self-hosted settings on retry

hsusul:
## Summary

- Honor self-hosted transcription settings when retrying audio from History.
- Give explicit self-hosted mode precedence over stale cloud-provider state.
- Forward the configured self-hosted model without looking up or forwarding unrelated cloud credentials.
- Fail closed when self-hosted mode has a missing, malformed, or unsupported URL.
- Preserve existing routing behavior for non-self-hosted modes.

## Behavior

Valid self-hosted URLs are normalized to an OpenAI-compatible
`/audio/transcriptions` endpoint. Base URLs, `/v1` URLs, existing full
transcription paths, whitespace, and trailing slashes are supported.

Local and private-network HTTP endpoints remain supported. Public endpoints
must use HTTPS.

When self-hosted mode has no usable URL, retry returns a clear configuration
error without calling another transcription provider or retrieving a cloud API
key.

## Scope

This PR changes only the History/Retranscribe path.

The separate Upload UI/settings path from #970 remains out of scope.

## Testing

Passed:

- `node --test test/helpers/retryTranscriptionRouting.test.js`
- `node --test test/helpers/selfHostedTranscription.test.js test/helpers/transcriptionAuth.test.js`
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `npm run i18n:check`
- `git diff --check`
- `node --check src/helpers/ipcHandlers.js`

The full `npm test` suite executed with 203 passing, 11 skipped, and
29 failing tests. All 29 failures require the Electron binary, which is
unavailable because dependencies were installed using
`npm ci --ignore-scripts` to avoid Electron postinstall and native rebuild
steps. No failures occurred in the focused retry-routing or related
self-hosted authentication tests.

Refs #970


gabrielste1n (approved):
Verified end-to-end: retry routing fix confirmed against the IPC handler chain, URL validation matches the live-dictation path, and the .ts-from-main-process import was tested against Electron 41's bundled Node. Pushed follow-ups extending the same fail-closed philosophy: chat agent API key now reaches self-hosted endpoints in both streaming paths, and getTranscriptionEndpoint fails closed instead of silently falling back to the cloud default when a self-hosted URL is invalid. Full suite (304 tests), typecheck, lint, and format all pass.

PR #1156, opened 2026-07-13
fix(notes): route self-hosted note formatting to its endpoint

sochotnicky:
Generate Notes with Note Formatting set to self-hosted silently called api.openai.com. runBackgroundAction only forwarded provider/baseUrl/key for cloud and providers/custom modes; self-hosted fell through to provider=undefined, so ReasoningService.processText resolved the provider via getModelProvider(model) and defaulted to OpenAI. The self-hosted remoteUrl and API key were never passed.

Extract buildNoteFormattingOverrides() which maps self-hosted to lanUrl + customApiKey (routing processText to the "lan" provider), and widen generateNoteTitle's config type to forward lanUrl for title generation too. Add unit tests covering cloud, providers/custom, local, and self-hosted.


sochotnicky:
Note that note names still don't get updated and I'm not yet clear why. I'm converting to draft until that's resolved

Edit: Found the cause - it would not regenerate note titles that were not default "Untitled Note" (meetings produce "New Note"). I suppose another option would be fix/change that to be consistent

gabrielste1n (approved):
Reviewed end-to-end: routing fix verified against processText/lan provider, cloud and providers modes are behavior-identical, tests and typecheck pass. Added a fail-fast guard so a self-hosted config with no server URL errors instead of falling through to a cloud provider, and trimmed comments.

PR #1162, opened 2026-07-13
feat(account): add profile editing (name, email, password)

xAlcahest:
## Summary

Users can't edit anything about their account from the app: the Account tab renders name and email as read-only text, and the only actions are sign out and delete account. This adds profile management on top of the Better Auth client already wired in `src/lib/auth.ts`: inline display name editing, a change-email dialog (the server sends a confirmation link before the new address takes effect), and a change-password dialog with current-password check and a default-on "sign out other devices" option. The password block is only rendered when `listAccounts()` reports a credential account, so OAuth-only users don't get a control that can't work for them. All three routes (`update-user`, `change-email`, `change-password`) are already live on the auth server, so this is desktop-only.

## Changes

- `src/components/settings/ProfileSection.tsx`: new Account tab section with the name, email and password rows plus the two dialogs, using the same SettingsPanel/SettingsRow layout as the rest of the tab.
- `src/lib/auth.ts`: `updateDisplayName`, `requestEmailChange`, `changePassword` and `hasCredentialAccount` wrappers, with Better Auth error codes preserved and mapped to localized messages.
- `src/hooks/useAuth.ts`: expose `refetch` from `useSession()` so the UI refreshes after profile changes.
- `src/components/SettingsPage.tsx`: mount `ProfileSection` in the account section.
- `src/locales/*/translation.json`: new `settingsPage.account.profile.*` keys in all 10 languages.
- `src/components/ui/label.tsx`: keep `text-neutral-700` in light mode and add `dark:text-foreground` so form labels stay readable in dark mode.
- `src/utils/validation.ts`: shared `EMAIL_REGEX`, now imported by both `ProfileSection` and `ShareNoteDialog` instead of two local copies.


gabrielste1n:
## Code review

**Verdict: approve.** Checked out the branch, traced every changed file against its callers, and ran `typecheck`, `eslint`, and `i18n:check` — all pass cleanly.

### Verified

- `ProfileSection` mounts inside the `isLoaded && isSignedIn && user` branch of SettingsPage, so `user.name`/`user.email` are always safe to read.
- The new `refetch` exposed from `useAuth.ts` is safe even when the auth client is absent — the `useStaticSession` fallback already returns a `refetch` stub.
- The `label.tsx` change (`dark:text-foreground`) is app-wide, so I audited every `<Label>` usage: the only custom className in the codebase is `text-xs font-medium` with no color, so nothing gets unintentionally overridden. This fixes a real dark-mode readability bug (`text-neutral-700` on dark backgrounds).
- The `EMAIL_REGEX` extraction to `validation.ts` is byte-identical to the one removed from `ShareNoteDialog` — clean DRY move, no behavior change.
- All 10 locale files have identical key structures under `settingsPage.account.profile` (hashed the sorted key paths — one unique hash), and `i18n:check` agrees.
- The `useDialogSession` token pattern correctly guards the close-mid-request race: closing bumps the session token, so a late-resolving `changePassword`/`changeEmail` response can't write state into a reopened dialog.
- Password fields reset on both close and open, so credentials never persist across dialog sessions.
- `hasCredentialAccount()` fails open (shows the password row on error) without caching the failure, and the cache is cleared in `signOut()` so a different account can't inherit it. If an OAuth-only user hits the fail-open path, `CREDENTIAL_ACCOUNT_NOT_FOUND` maps to a clear localized message.
- The hardcoded `callbackURL: "https://openwhispr.com"` matches the existing pattern in the same file (`DESKTOP_OAUTH_CALLBACK_URL`, password-reset `redirectTo`).

### Minor nits (none blocking)

1. **Password row flash for OAuth-only users** — `hasPassword` defaults to `true`, so OAuth-only users briefly see the "Change password" row before `listAccounts()` resolves and it disappears (small layout shift). Defaulting to `null` and rendering the row only once known would avoid it.
2. **Redundant trim** — `ProfileSection` passes an already-trimmed name and `updateDisplayName` trims again. Harmless.
3. **Email-sent copy is vague about the destination** — Better Auth sends the change-approval link to the *current* address for verified users; "We sent a confirmation link" doesn't say where. Saying "to your current email" would prevent users from watching the wrong inbox. Worth a follow-up copy tweak.

xAlcahest:
All three addressed: `hasPassword` now starts as `null` so the password row only renders once `listAccounts()` resolves, dropped the redundant trim in `updateDisplayName`, and the sent-confirmation copy now says the link goes to your current address, in all 10 locales.

Chadpiha (approved):
## Update: email change removed

Dropped the "Change email" flow from this branch. Profile management now covers
**display name** and **password** only. Email is still shown read-only in the
account header — it's just no longer editable.

### What went away

| Area | Change |
|---|---|
| `src/lib/auth.ts` | Removed `requestEmailChange()` (the `authClient.changeEmail` wrapper) |
| `ProfileSection.tsx` | Removed `ChangeEmailDialog` (~135 lines), the Email settings row, and the `email` prop |
| `SettingsPage.tsx` | Dropped `email={user.email}` from the `<ProfileSection>` call |
| 10 locale files | Removed the 12-key `settingsPage.account.profile.email` block |

`updateDisplayName`, `changePassword`, and `hasCredentialAccount` are untouched.

### Cleanups in the same pass

- **Inlined a now-single-consumer abstraction.** `useDialogSession` and
  `DialogBaseProps` existed to be shared by the email and password dialogs. With
  the email dialog gone, both collapsed into `ChangePasswordDialog` as a local
  `sessionRef` + stable `reset` callback. The stale-response guard is unchanged
  in behavior — it still prevents a response that resolves after a close/reopen
  from painting into the new session. Same for `INLINE_ERROR_CLASS`, inlined at
  its one remaining use site.
- **Closed a double-submit window on Save name.** `savingName` flipped back to
  `false` before the refetched `name` prop landed, so the button briefly
  re-enabled while still dirty and a fast second click fired a redundant PATCH.
  Now the dirty check compares against a `savedName` baseline that advances on
  success.
- **`maxLength={80}` → `MAX_NAME_LENGTH`**, alongside the existing
  `MIN_PASSWORD_LENGTH`.
- **`ReferralDashboard`** now imports the shared `EMAIL_REGEX` from
  `src/utils/validation.ts` instead of redeclaring `RE_EMAIL`. Identical pattern,
  no behavior change.
- Fixed a pre-existing JSON formatting defect in all 10 locales where
  `passwordTooLong` and `noPassword` shared a line.

### Verification

`typecheck`, `lint`, `prettier --check`, and `i18n:check` all pass. Locale parity
confirmed for `settingsPage.account.profile` across all 10 languages — 0 missing,
0 orphaned. No leftover references (`grep -rn "profile.email" src/` is clean).

PR #1172, opened 2026-07-14
fix(mic): persist microphone selection across Chromium device ID change

### LINKED ISSUE #1171: The selected microphone is not saved
If you select a microphone under Settings → Microphone → Device, this setting is lost after you restart the computer.

Chadpiha:
## Summary
The saved microphone preference stored only Chromium's deviceId, which rotates over time (and changes on unplug/replug). When the ID went stale, recordings silently fell back to the default mic even though the user's chosen device was still connected (see #900 for the original stale-ID symptom).

This PR persists the device label alongside the ID and reconciles the saved preference against currently available devices at every acquisition point. The saved ID is used while valid; when it goes stale, the selection is remapped by label — but only when exactly one connected input matches, so we never guess between identical devices. If reconciliation can't resolve a device, behavior degrades to exactly what ships today (stale-ID → OverconstrainedError → existing default-mic retry). Existing users are migrated transparently: the label is backfilled the first time their saved ID resolves.

The reconcile-and-persist flow shared by dictation and meeting capture lives in `src/helpers/micSelectionRecovery.js`; the pure resolution logic (unit-tested) lives in `src/helpers/micDeviceSelection.js`. The caller-less `setSelectedMicDeviceId` setter was removed in favor of the label-aware `setSelectedMicDevice`.

Fixes #1171

## Verification
Full test suite passes: 272 tests, 0 failures (Node 24).
`tsc --noEmit`, ESLint, and `prettier --check` are clean across the repo.

## Known limitations
Two identical-model microphones (duplicate labels) intentionally do not remap — the user must re-pick. This is by design to avoid guessing wrong.
The integration paths (caching, persistence side effects) are verified by review only; a manual smoke test is recommended: select a non-default mic → restart → record → confirm the same mic is used.

### DIFF
--- a/src/components/SettingsPage.tsx
+++ b/src/components/SettingsPage.tsx
@@ -707,8 +707,9 @@ export default function SettingsPage({
     setActivationMode,
     preferBuiltInMic,
     selectedMicDeviceId,
+    selectedMicDeviceLabel,
     setPreferBuiltInMic,
-    setSelectedMicDeviceId,
+    setSelectedMicDevice,
     setUseLocalWhisper,
     setUiLanguage,
     setWhisperModel,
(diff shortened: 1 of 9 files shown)


gabrielste1n (approved):
Reviewed end-to-end: every acquisition point that pins an exact deviceId now reconciles against live devices, cross-window settings sync covers the new label key, and all failure paths degrade to the pre-existing default-mic retry. Pushed a cleanup commit extracting the duplicated reconcile block into a shared helper and removing the caller-less setter. Tests (272), typecheck, lint, and formatting all pass.

PR #1178, opened 2026-07-14
fix(notes): support Unicode note search

### LINKED ISSUE #1177: Note search returns no results for non-Latin queries
## Description

Note search removes non-ASCII letters while constructing its SQLite FTS5
query. Searches in Cyrillic, CJK, Arabic, and other scripts can become empty
before SQLite is queried. Accented Latin terms are also modified.

This affects Command Search, local agent note search, and CLI note search
because they use `DatabaseManager.searchNotes()`.

## Reproduction

1. Create a note containing `Привет мир`.
2. Open Command Search.
3. Search for `Привет`.
4. Observe that the note is not returned.

The same preprocessing issue affects queries such as `中文`, `مرحبا`, and
`café`.

## Expected behavior

Unicode letters and numbers should remain searchable, while blank or
punctuation-only input should return no results.

Existing ASCII prefix-search behavior should remain unchanged.

## Actual behavior

The current sanitizer uses an ASCII-only JavaScript word-character class:

```js
query.replace(/[^\w\s]/g, " ")
```

As a result:

* `Привет` becomes empty
* `中文` becomes empty
* `مرحبا` becomes empty
* `café` becomes `caf`

SQLite FTS5 successfully matches the original terms when queried directly, so
the loss occurs during application-side query construction.

## Proposed scope

* Add a Unicode-aware FTS5 query builder.
* Reuse it from `DatabaseManager.searchNotes()`.
* Preserve current prefix matching, deleted-note filtering, ordering, and
  limits.
* Add regression tests for Cyrillic, CJK, Arabic, accented Latin, ASCII, and
  punctuation-only input.

No UI redesign, tokenizer migration, semantic-search changes, or cloud-search
changes are intended.

hsusul:
## Summary

* Preserve Unicode letters, combining marks, and numbers when constructing
  SQLite FTS5 note-search queries.
* Quote every token and apply prefix matching independently without exposing
  raw FTS5 operators.
* Return no results for blank or punctuation-only input without executing an
  invalid or match-all expression.
* Preserve existing note filtering, ranking, limits, IPC, and UI behavior.

## Root cause

`DatabaseManager.searchNotes()` sanitized queries with `/[^\w\s]/g`.
JavaScript `\w` is ASCII-only, so Cyrillic, CJK, Arabic, and other scripts
were removed before SQLite was queried. Accented Latin terms were modified.

## Behavior

Each searchable token is normalized, safely quoted, and independently
prefix-matched using SQLite FTS5 syntax.

Representative verified searches include:

* `Прив` matching `Привет мир`
* `中文` matching `中文测试`
* `東京` matching `東京駅`
* `مرح` matching `مرحبا بالعالم`
* `caf rés` matching `café résumé`
* `hel wor` matching `hello world`

Blank or punctuation-only input returns an empty result set without executing
an invalid or match-all `MATCH` expression.

## Scope

This PR changes only local SQLite note-search query construction and its
regression tests.

It does not change:

* Command Search UI
* IPC or preload APIs
* semantic search or Qdrant
* database schema or FTS tokenizer configuration
* note ranking, ordering, filtering, or limits
* cloud synchronization

## Testing

Passed:

* `node --test test/helpers/noteSearch.test.js`

  * 5 passed, 0 failed

  * 259 passed, 11 skipped, 0 failed
* `npm run typecheck`
* `npm run lint`
* `npm run format:check`
* `npm run i18n:check`
* `git diff --check`
* `node --check src/helpers/database.js`
* `node --check src/helpers/noteSearch.js`

The regression tests use Node 24's in-memory SQLite FTS5 support and require no
Electron startup, native rebuild, network access, model download, or API
credentials.

Fixes #1177

### DIFF
--- a/src/helpers/database.js
+++ b/src/helpers/database.js
@@ -3,6 +3,7 @@ const path = require("path");
 const fs = require("fs");
 const { randomUUID } = require("crypto");
 const debugLogger = require("./debugLogger");
+const { buildNoteSearchQuery } = require("./noteSearch");
 const { app } = require("electron");
 
 // Server-enforced trigger cap (openwhispr-api); enforced here so one oversized
@@ -2097,11 +2098,8 @@ class DatabaseManager {
   searchNotes(query, limit = 50) {
     try {
       if (!this.db) throw new Error("Database not initialized");
-      const term = query
-        .trim()
-        .replace(/[^\w\s]/g, " ")
-        .trim();
-      if (!term) return [];
+      const ftsQuery = buildNoteSearchQuery(query);
+      if (!ftsQuery) return [];
       return this.db
         .prepare(
           `
@@ -2113,7 +2111,7 @@ class DatabaseManager {
         LIMIT ?
       `
         )
-        .all(term + "*", limit);
+        .all(ftsQuery, limit);
     } catch (error) {
       debugLogger.error("Error searching notes", { error: error.message }, "database");
       throw error;
--- a/src/helpers/noteSearch.js
+++ b/src/helpers/noteSearch.js
@@ -1,0 +1,14 @@
+function buildNoteSearchQuery(input) {
+  if (typeof input !== "string") return "";
+
+  const tokens = input
+    .normalize("NFC")
+    .match(/[\p{L}\p{N}_][\p{L}\p{M}\p{N}_]*/gu)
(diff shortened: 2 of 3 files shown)


gabrielste1n (approved):
great work thank you @hsusul !!!

PR #1184, opened 2026-07-15
refactor(transcription): centralized, non-destructive STT provider routing (fixes Parasail wipe + OpenAI leaks)

gabrielste1n:
## What & why

Speech-to-text provider routing had three linked problems, all rooted in how provider/model state was managed:

1. **Custom provider config was wiped on switch** — the "Parasail" bug. Configuring a custom cloud endpoint, trying another top-level provider (e.g. Tinfoil), and switching back destroyed the custom URL/model, because every cloud provider shared one `cloudTranscriptionModel`/`cloudTranscriptionBaseUrl` slot.
2. **Self-hosted and custom modes could leak audio + keys to `api.openai.com`** — several code paths fell open to the OpenAI default when a URL was empty/invalid, or let a leftover BYOK provider (mistral/xai/corti) hijack self-hosted dictation.
3. **Surfaces re-derived routing independently and diverged** — retry had no Corti branch (Corti audio → OpenAI) and defaulted a blank custom URL to OpenAI.

This PR centralizes the routing decision and makes provider config non-destructive.

## Changes (commit by commit)

- **Leak-stoppers** — guard the mistral/xai/corti dispatch on `!isSelfHostedTranscription`; fail closed when self-hosted has no URL and when a custom URL is empty/invalid (no more OpenAI default); add the missing **Corti** branch to retry; route dictation-agent *local* mode through the model lookup instead of the unroutable family-id provider.
- **`resolveTranscriptionRoute`** (new, `src/helpers/transcriptionRoute.js`) — one pure, fail-closed resolver mapping a settings snapshot to a validated route (endpoint, model, auth scheme + secure-key ref, size cap, language) for every batch mode: self-hosted, local, OpenWhispr cloud, openai/groq, custom incl. Azure, and the proxied tinfoil/mistral/xai/corti providers with their quirks. Routes never carry secrets. **12 matrix tests.**
- **Per-provider config** (`transcriptionProviderConfig.js` + store) — each provider remembers its own `{model, baseUrl}` per scope (base/meeting/upload); switching restores the incoming provider's values instead of overwriting. Seeded from the existing flat settings by a guarded one-shot migration; keys stay in the secure-key channel. Pickers delegate the switch to the store. **3 tests incl. the Parasail scenario.**
- **Dictation + retry cutover** — `getTranscriptionEndpoint` and the retry HTTP branch delegate to the resolver (single source of truth, shared fail-closed validation, Azure parity on retry). Transcription source label now reports the real provider instead of a hardcoded `"openai"`.

## Verification

- `node --test test/helpers/*.test.js` — **433 pass, 0 fail** (incl. 15 new).
- `npm run typecheck`, `npm run lint`, `prettier --check` — clean.
- New unit tests cover the full (mode × provider × url-state) matrix, the fail-closed cases, and the Parasail switch-and-return scenario.

## ⚠️ Needs a packaged/dev-app smoke test before merge

Unit tests cover routing logic, but two things can only be verified by running the app:
- The dictation + retry hot paths (main-process execution of the resolved route).
- The dynamic `import()` of the ESM resolver (which imports `.ts` files) from the CJS main process — same precedent as the merged `retryTranscriptionRouting.js`, but worth confirming in a packaged build.

**Suggested smoke test:** configure custom STT (Parasail) → transcribe → switch to Tinfoil → switch back → confirm Parasail config intact and working; retry a Corti transcription and confirm it hits Corti; confirm a bad self-hosted URL shows an error rather than silently using OpenAI.

## Explicitly out of scope (tracked follow-ups)

- **Meeting-surface parity** — meeting transcription is realtime-websocket-only, so self-hosted/custom batch can't serve it; removing the dead picker options + passing language + handling mid-meeting socket close is a separate, realtime-specific change needing its own runtime testing.
- **Per-scope custom API keys** and the **LLM-side resolver** (chat/cleanup/note-formatting) — the sibling architecture, planned as later PRs.
- **Full transcription-error i18n** — pre-existing; errors carry a stable `messageKey` for when the display path is localized.

Plan: `~/Documents/Plans/stt-resolver-refactor-pr-abcd.md`.


gabrielste1n:
Ran an adversarial review over the full diff. It caught one shipping-blocker and one regression, both now fixed in `8dd2b0c5`:

- **CRITICAL (fixed):** `getTranscriptionEndpoint` was called before the mistral/xai/corti proxy branches and now throws for non-HTTP transports — it broke live dictation for those three providers. Endpoint resolution moved below the proxy dispatch so only direct-HTTP providers reach it.
- **MEDIUM (fixed):** retry required an API key for custom endpoints; custom may be keyless (LAN server). Now only the known cloud providers (openai/groq) require a key on retry, matching the live-dictation path.
- **LOW (won't-fix, documented):** a self-hosted URL pointed at an `*.openai.azure.com` host no longer builds the Azure deployment URL. This never actually worked — self-hosted sends no auth header and Azure requires the `api-key` header, so it always 401'd. Azure should be configured as a `custom` provider (which does handle it). Not adding non-functional code for an unsupported config.

The review also positively confirmed the non-destructive switch (no loop/clobber), the migration (idempotent, no data loss), the full-path URL dedup, and the ESM/.ts dynamic-import safety. Still needs the packaged-app smoke test noted above before merge.

PierreF-DP:
Confirming this PR matches a bug I hit on **v1.7.2 (Linux, Ubuntu 24.04.4)** — adding a concrete data point and a workaround for anyone landing here.

**Symptom:** with Transcription mode = **Self-Hosted**, every dictation fails. It's recorded as `error_code = API_KEY_MISSING`, `provider = null`, `model = null`, and the request never leaves the machine — the self-hosted server is never contacted. My server is a plain OpenAI-compatible Whisper endpoint (`POST /v1/audio/transcriptions`, no auth).

**Root cause I traced (lines up with this PR's "fell open to the OpenAI default" / "leftover BYOK provider hijacks self-hosted"):**
- The transcription path resolves the provider as `cloudTranscriptionProvider || "openai"`, then calls `getAPIKey()` unconditionally — even in self-hosted mode.
- `getAPIKey()` throws `API_KEY_MISSING` for the `openai` / `mistral` / `groq` branches; only the `custom` branch returns `null` without throwing.
- Selecting **Self-Hosted** in the UI never sets `cloudTranscriptionProvider = "custom"`, and the Self-Hosted panel exposes **only a URL field** (no provider/key field). So the provider stays at its `"openai"` default → `getAPIKey()` looks for an absent OpenAI key → throws.

**Why it's a hard dead-end for users:** because the Self-Hosted UI has no provider/key control, there's no in-app way to recover — the only fix is editing the app's Local Storage directly.

**Workaround** (until this ships) — force the provider to `custom` so `getAPIKey()` stops requiring an OpenAI key. Two settings need to change, and since the Self-Hosted UI exposes no field for them, you have to write them to the app's Local Storage directly:
- `cloudTranscriptionProvider = "custom"`
- `customTranscriptionApiKey = "sk-local"` (dummy, non-empty; the self-hosted server ignores it)

Easiest way (no need to quit the app): open DevTools on the OpenWhispr window with **Ctrl+Shift+I** (F12 isn't wired in the build I tested), and in the Console run:
```js
localStorage.setItem('cloudTranscriptionProvider', 'custom');
localStorage.setItem('customTranscriptionApiKey', 'sk-local');
```
Then reload (Ctrl+R) or restart. (If the console blocks the paste with a self-XSS warning, type `allow pasting` + Enter first, or just type the two lines by hand.) These are plain `localStorage` values, so `setItem` is enough — no need to touch the LevelDB files by hand.

After that, dictation works immediately and is correctly routed to the self-hosted `/v1/audio/transcriptions` endpoint; `transcriptionMode` stays `self-hosted` and the URL is unchanged.

Could you confirm this PR makes Self-Hosted mode force the no-key routing path (i.e. never fall back to the `openai` provider / never require a key) so a fresh Self-Hosted setup works without touching storage? Happy to test a build. Thanks for the work on this!

gabrielste1n:
Closing rather than rebasing: main has since fixed the custom-URL wipe (#1459) with a different ownership model and woven managed-policy enforcement into the exact regions this PR rewrites, so conflict-resolving would risk stripping policy guards. The still-open issues this PR targeted (custom-URL fail-open to OpenAI for unmanaged users, mistral/xai/corti hijacking self-hosted dictation, missing Corti retry branch, model wiped on provider switch, divergent routing paths) plus the resolver design will be re-landed fresh on top of current main. The resolver + matrix tests carry over largely intact.

PR #1192, opened 2026-07-15
fix(whisper): fall back to CPU when the CUDA server crashes mid-request

### LINKED ISSUE #1139: Graceful fallback when CUDA initialization fails on unsupported GPU architectures
## Feature request: Graceful fallback when CUDA initialization fails on unsupported GPU architectures
###Environment

OpenWhispr version: 1.7.5
OS: Windows 10
GPU: NVIDIA Quadro M2000M
Model: Parakeet TDT 0.6B v3 (multilingual, ~680MB)

### Context
On Windows, selecting GPU-accelerated local Whisper spawns `whisper-server-win32-x64-cuda.exe`. On GPUs below the compute-capability floor this CUDA build was compiled against, CUDA context creation fails during initialization.

**Confirmed crash signature** (Windows Application Event Log):
```
Faulting application name: whisper-server-win32-x64-cuda.exe
Faulting module name: ucrtbase.dll
Exception code: 0xc0000409
```

This is a fail-fast abort (`STATUS_STACK_BUFFER_OVERRUN` / unhandled CRT abort), occurring during CUDA init on a GPU below the supported compute-capability floor — confirmed on an NVIDIA Quadro M2000M (Maxwell architecture, compute capability 5.0).

### Impact
- The process dies before binding its port.
- The app surfaces this only as an opaque `ECONNRESET` on the client side (`whisper-server request failed: read ECONNRESET`), with no indication of the actual cause.
- No Application Event Log entry beyond the raw crash dump — no app-level log, since the process never gets far enough to log anything itself.
- This makes the actual cause very difficult for an end user to diagnose (in this case it took cross-referencing Task Manager process presence/absence, Windows Event Viewer, and manual binary testing to identify).

### Requested behavior
Wrap CUDA execution-provider/context initialization in the whisper-server launch path in a try/catch (or equivalent init-probe) at startup. On failure, fall back automatically to one of:
1. The CPU-only `whisper-server-win32-x64.exe` binary, **or**
2. A DirectML execution provider path, if feasible for the Whisper/whisper.cpp engine — precedent exists within the codebase, since the bundled sherpa-onnx (Parakeet) runtime already appears to successfully use a GPU-accelerated path (likely DirectML) on this same hardware without issue.

### Why this matters
Compute-capability mismatches on older-but-still-functional NVIDIA GPUs (common in older laptops/workstations, e.g. Quadro M-series) currently produce a hard crash with a misleading network-layer error rather than a degraded-but-working experience. A caught failure + automatic fallback + a user-facing notice (e.g., "GPU acceleration unavailable on this device, using CPU instead") would turn a fatal, confusing bug into a transparent, non-blocking degradation.

xAlcahest:
## Summary
On GPUs below the CUDA build's compute capability floor (like the Quadro M2000M in the report), ggml's CUDA init does not validate compute capability, so the CUDA whisper-server loads the model, binds its port and passes the health check, then hard-aborts (0xC0000409) at the first kernel launch during the first inference request. The startup fallback in `_doStart` never sees this, and the user only gets "whisper-server request failed: read ECONNRESET". Now `transcribe()` detects a connection failure from a local CUDA server whose process died, emits the existing `cuda-fallback` event, restarts on the CPU binary and retries the request once, so the dictation still completes. A generation counter and a stop guard keep concurrent transcriptions, model switches and intentional stops from triggering spurious restarts, and the startup path is untouched to stay out of #1185's way. The `cuda-fallback-notification` IPC chain existed end to end but nothing in the renderer consumed it, so the fallback was invisible; it now shows a toast.

Fixes #1139

## Changes
- `src/helpers/whisperServer.js`: request-time CUDA to CPU fallback with a single retry, plus a generation counter and stop guard so concurrent callers, model switches and intentional stops never trigger spurious restarts
- `src/App.jsx`: subscribe to `onCudaFallbackNotification` and show a toast when the fallback happens
- `src/locales/*/translation.json`: new `app.toasts.cudaFallback` key in all 10 languages
- `test/helpers/whisperCudaRequestFallback.test.js`: table tests for the two exported predicates and loopback integration tests covering crash recovery, concurrent callers, a peer's in-flight restart, restart failure, intentional stops and the no-stall paths


gabrielste1n:
Rebased onto main (now includes #1185's Vulkan support and #1173) and pushed a follow-up commit on top — thanks for the excellent groundwork, the generation counter and the loopback tests made extending this straightforward. What changed:

- **Unified the stop flags**: your `_stopRequested` and #1185's `_abortStartup` had identical lifecycles, so they're now one flag consulted by both `waitForReady` and the request-retry guards.
- **Vulkan parity**: the fallback predicate takes `useGpu` (CUDA or Vulkan), the restart clears both GPU flags, and a Vulkan crash emits `gpu-fallback`. The toast now listens to both channels, and the new locale key is renamed `cudaFallback` → `gpuFallback`.
- **Toast ordering**: the event now fires only after the CPU server is up, so "Transcription is using the CPU instead" can't precede a failed restart (the restart-failure test now asserts no notification).
- **Peer-restart hole**: if a concurrent caller restarted the crashed server with the GPU flags still set, the replaced-path retry hit another doomed GPU server and the second connection error propagated — failing the dictation this PR set out to save. That retry now gets the same one-shot crash check and lands on CPU. New test covers it, plus one for the Vulkan path.

Full suite (469), lint, typecheck, prettier, and i18n checks all pass locally.

gabrielste1n (approved):
Traced end to end (request failure → crash detection → CPU restart → retry → toast), adversarially reviewed the concurrency paths, and verified locally. With the rebase and follow-up commit this is solid.

PR #1194, opened 2026-07-15
fix(export): preserve merged transcript start times

### LINKED ISSUE #1193: SRT export starts merged same-speaker cues at the wrong time
## Describe the bug

When adjacent transcript segments from the same speaker occur less than two
seconds apart, OpenWhispr merges them into one exported cue. The resulting SRT
cue starts at the last merged segment's timestamp instead of the first
segment's timestamp.

This leaves the beginning of the merged speech without a corresponding
subtitle cue.

## To reproduce

Given transcript segments beginning at:

* Speaker 1 at `0.250s`
* Speaker 1 at `1.750s`
* Speaker 2 at `4.000s`

Export the transcript as SRT.

## Actual behavior

The first cue begins at `1.750s`:

```text
00:00:01,750 --> 00:00:04,000
Speaker 1: ...
```

## Expected behavior

The merged cue should retain the first segment's start time:

```text
00:00:00,250 --> 00:00:04,000
Speaker 1: ...
```

## Verified root cause

`mergeSegments()` uses the same timestamp field for both:

* the exported cue's original start time
* the most recently processed timestamp used by the rolling two-second merge
  window

When another segment is merged, the cue's original timestamp is replaced.

## Proposed scope

* Track the latest processed segment timestamp separately for merge-gap
  calculation.
* Preserve the first segment's timestamp on the exported cue.
* Add focused regression tests for merged cues, rolling-window chaining, and
  the two-second boundary.

## Non-goals

* Changing the export UI or IPC API
* Adding timestamp support to cloud transcription
* Changing speaker assignment or diarization
* Changing SRT end-time inference

hsusul:
## Summary

* Preserve the first segment's timestamp when consecutive same-speaker
  transcript segments are merged for export.
* Keep the existing rolling two-second merge window based on the most recently
  processed segment.
* Add focused regression coverage for merged SRT cues and merge boundaries.

## Root cause

`mergeSegments()` used `last.timestamp` both as the exported cue start and as
the latest timestamp for merge-gap calculation. Updating it after every merge
shifted the cue start forward and omitted the beginning of the speech from the
subtitle timeline.

## Behavior

A merged cue containing segments beginning at `0.250s` and `1.750s` now starts
at `0.250s`.

Merge eligibility continues to use the most recently processed segment, so the
existing rolling-window behavior is preserved.

Segments at the exact two-second boundary remain separate, preserving the
existing comparison behavior.

## Scope

This PR changes only transcript segment merge bookkeeping and its regression
tests.

It does not change:

* export UI or IPC APIs
* speaker labels or diarization
* cloud transcription
* SRT end-time inference

## Testing

Passed:

* `node --test test/helpers/transcriptFormatter.test.js`

  * 3 passed, 0 failed

  * 422 tests, 411 passed, 11 skipped, 0 failed
* `npm run typecheck`
* `npm run lint`
* `npm run format:check`
* `npm run i18n:check`
* `git diff --check`
* `node --check src/helpers/transcriptFormatter.js`
* `node --check test/helpers/transcriptFormatter.test.js`

The focused tests require no Electron startup, native compilation, network
access, model download, or API credentials.

Fixes #1193


hsusul:
The focused SRT tests pass, and the other required checks completed
successfully. I traced the failing Tests run to a separate, load-sensitive URL
download cleanup race: an aborted `fs.createWriteStream()` can finish opening
after cleanup has already attempted to unlink the destination.

I reproduced that independently with a deterministic delayed-open test and
kept the correction in a separate focused contribution so this PR remains
limited to SRT timestamp behavior.

Could a maintainer rerun the Tests workflow for this PR?

gabrielste1n (approved):
Reviewed end to end. The original fix correctly preserves merged cue start times, and the follow-up commit anchors end-time bookkeeping (final SRT cue end, JSON duration_seconds) to the last merged segment so those don't regress. Full suite, typecheck, lint, format, and i18n checks all pass locally.

PR #1198, opened 2026-07-15
refactor(transcription): registry for proxied STT providers

gabrielste1n:
Collapses the four copy-pasted proxied-provider dictation branches (tinfoil / mistral / xai / corti) in `processWithOpenAIAPI` into a `PROXY_TRANSCRIPTION_PROVIDERS` registry, following the existing `STREAMING_PROVIDERS` pattern: per-provider payload quirks are data, the transcribe flow exists once, and adding a proxied provider becomes a one-entry change instead of a 35-line copy-paste (the pattern that produced the retry-missing-Corti leak).

**Behavior notes** (both strict improvements):
- mistral/xai/corti gain the structured `{error, code, messageKey}` proxy-error handling that was previously tinfoil-only.
- A missing preload function now throws "X transcription is unavailable in this window" for all four providers (previously mistral/xai/corti fell through and hit a more confusing resolver error).

Net **-57 lines**. 433 tests pass, typecheck/lint/prettier clean.

**Stacked on #1184** (it rewrites the same region — basing on main would conflict and drop #1184's self-hosted guards). GitHub retargets this to `main` automatically when #1184 merges; smoke-test both together.

### DIFF
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -190,6 +190,54 @@ const STREAMING_PROVIDERS = {
   },
 };
 
+// Batch providers that must transcribe via a main-process proxy (CORS,
+// non-Bearer auth, OAuth, or attested transport) instead of a renderer fetch.
+const PROXY_TRANSCRIPTION_PROVIDERS = {
+  tinfoil: {
+    displayName: "Tinfoil",
+    ipc: () => window.electronAPI?.proxyTinfoilTranscription,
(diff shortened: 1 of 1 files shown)


gabrielste1n:
Closing: stacked on #1184, which has diverged too far from main (345 commits, policy-enforcement layer landed in the same regions). The PROXY_TRANSCRIPTION_PROVIDERS registry idea is still valid — the four copy-pasted proxy branches remain on main — and will be re-cut directly against main as part of the STT routing stabilization plan.

PR #1199, opened 2026-07-15
fix(downloads): wait for stream close before cleanup

### LINKED ISSUE #1197: Cancelling URL audio downloads can leave partial files
## Bug

Cancelling a URL audio download can leave an incomplete `ow-url-*` destination
behind after the operation has already reported cancellation.

`fs.createWriteStream()` opens its destination asynchronously. The existing
failure path destroys the stream and immediately attempts to unlink the
destination. If the stream has not finished opening, the unlink can observe
`ENOENT`. The pending open may then create the file after cleanup has
completed.

The problem is load-sensitive with an ordinary filesystem but can be
reproduced deterministically by delaying the write stream's open callback.

## Minimal reproduction

1. Start a direct URL audio download using a write stream whose filesystem
   open callback is delayed.
2. Cancel the download before the delayed open completes.
3. Allow the current cleanup path to attempt destination deletion.
4. Release the delayed open callback.
5. Observe that cancellation can settle before the stream closes and that the
   destination may be created after cleanup.

## Expected behavior

Cancellation should not settle until:

1. the destination stream reaches a terminal closed state; and
2. the incomplete destination has been removed.

## Actual behavior

Cleanup calls `destroy()` and immediately attempts deletion. `destroy()` does
not mean that the asynchronous file-open and close lifecycle has completed
when the call returns.

## Verified root cause

The destination unlink is ordered before confirmed write-stream termination.
A pending `open` can therefore occur after the deletion attempt.

## Proposed scope

* Begin waiting for terminal write-stream completion before requesting
  destruction.
* Wait for the stream to close.
* Remove the incomplete destination afterward.
* Ignore `ENOENT` while preserving visibility into unexpected filesystem
  failures.
* Preserve the original cancellation or download error.
* Add deterministic regression tests for delayed open and adjacent lifecycle
  cases.

## Non-goals

* Changing download retry policy
* Changing progress reporting
* Changing URL validation or download sources
* Refactoring unrelated downloader code
* Changing successful-download behavior
* Changing model or runtime binary downloads
* Adding dependencies

hsusul:
## Summary

* Wait for the destination write stream to close before deleting an incomplete
  URL audio download.
* Prevent cancellation from settling while a pending asynchronous open can
  still recreate the destination.
* Preserve the original download or cancellation error.
* Ignore an already-missing destination while retaining visibility into
  unexpected unlink failures.

## Root cause

`fs.createWriteStream()` opens its destination asynchronously. The previous
cleanup path called `destroy()` and immediately attempted to unlink the
destination.

When cancellation occurred before `open`, unlink could observe `ENOENT` and
return. The pending open could then create the destination after cleanup and
after the download promise rejected.

## Corrected lifecycle

Cleanup now:

1. starts waiting for terminal stream completion;
2. requests destination-stream destruction;
3. waits for the stream to close;
4. removes the incomplete destination;
5. rejects with the original error.

This prevents retry callers from proceeding while the previous destination
stream may still open or write.

## Behavior

The deterministic delayed-open regression proves that:

* cancellation does not settle before the destination stream closes;
* the incomplete destination is removed after closure;
* late source errors do not cause duplicate settlement;
* the original `DOWNLOAD_CANCELLED` code is preserved.

## Scope

This PR changes only incomplete-destination cleanup in the shared URL audio
download stream helper.

It does not change:

* retry policy
* progress reporting
* URL validation
* download sources
* successful-download handling
* model or runtime binary downloads

## Testing

Passed:

* `ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/urlAudioDownloader.test.js`

  * 83 passed, 0 failed, 0 skipped

  * 437 total, 426 passed, 11 skipped, 0 failed
* `npm run typecheck`
* `npm run lint`
* `npm run format:check`
* `npm run i18n:check`
* `git diff --check`
* `node --check src/helpers/urlAudioDownloader.js`
* `node --check test/helpers/urlAudioDownloader.test.js`

A supporting real-filesystem stress run completed 500 isolated iterations with
zero failures and no observed leaked files. This stress result supplements the
deterministic regression test rather than replacing it.

Fixes #1197

### DIFF
--- a/src/helpers/urlAudioDownloader.js
+++ b/src/helpers/urlAudioDownloader.js
@@ -8,6 +8,7 @@ const crypto = require("crypto");
 const fs = require("fs");
 const os = require("os");
 const path = require("path");
+const { finished } = require("node:stream/promises");
 const debugLogger = require("./debugLogger");
 const { getSafeTempDir } = require("./safeTempDir");
 const { getFFmpegPath } = require("./ffmpegUtils");
@@ -1064,11 +1065,27 @@ function streamToFile(
       bail(Object.assign(new Error("Download stalled"), { code: "DOWNLOAD_FAILED" }));
     });
 
-    const cleanupFile = () => {
+    const cleanupFile = async () => {
+      const closed = finished(fileStream, { cleanup: true });
       fileStream.destroy();
       try {
-        fs.unlinkSync(tempPath);
+        await closed;
       } catch {}
+      try {
+        fs.unlinkSync(tempPath);
+      } catch (error) {
+        if (error.code !== "ENOENT") throw error;
+      }
+    };
+
+    const rejectAfterCleanup = (err) => {
+      cleanupFile().then(
+        () => reject(err),
+        (cleanupError) => {
+          if (err.cause === undefined) err.cause = cleanupError;
+          reject(err);
+        }
+      );
     };
 
     const bail = (err) => {
@@ -1076,8 +1093,7 @@ function streamToFile(
       settled = true;
       stall.clear();
       if (abortSignal) abortSignal.removeEventListener("abort", onAbort);
-      cleanupFile();
-      reject(err);
+      rejectAfterCleanup(err);
     };
 
     // React to cancel even while stalled (no data events arriving).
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Superseded by #1212, which shipped in [v1.7.6](https://github.com/OpenWhispr/openwhispr/releases/tag/v1.7.6) with the same approach — cleanup now waits for the write stream's close event before unlinking, so a pending async open can no longer recreate the partial file after removal, and the original error is preserved. Thanks for the diagnosis here — closing this one as superseded.

PR #1202, opened 2026-07-16
feat(macos): add a Show Dock Icon setting and stop the icon reappearing (#428)

### LINKED ISSUE #428: Option to permanently hide Dock icon (MacOS)
The Dock icon reappears whenever a new transcription starts. Even if I press the “x” to hide it, it comes back automatically during the next transcription. This makes it impossible to keep the app running only in the menu bar.

A setting that lets users toggle whether the Dock icon should be shown at all. For example, a “Show Dock Icon” option in Settings. If disabled, the app should remain only in the menu bar and never appear in the Dock.

Manually hiding the icon with the “x” and changing `LSUIElement`, but the icon still reappears once a new transcription starts.

**Additional context**
I personally prefer using the app as a menu-bar-only utility and would like the Dock icon to stay hidden permanently.

dreasan:
Closes #428.

## Problem

Two things behind this issue:

1. **The reported symptom is a bug, not a missing feature.** `hide-window` called
   `app.dock.show()`, left behind by 52077bb2 when it swapped `mainWindow.minimize()` for
   `hideDictationPanel()`. Right for a minimized window, wrong for a hidden one. With
   auto-hide-when-idle on, `src/App.jsx` fires `hideWindow()` after every dictation, so the
   icon kept coming back, exactly as the issue describes. Shipped since v1.0.6, and only
   reproduces with auto-hide on.

2. **No way to opt out.** Dock calls were spread over 5 sites in 4 files, with nowhere for a
   setting to gate them.

## Design

- `dockManager.js` is the single owner of the Dock icon, so the setting can veto any caller.
- `dockPolicy.js` holds the decision logic: pure, no Electron import, unit-tested with zero
  dependencies.
- Defaults **on**, so existing installs are unaffected. Off keeps the app menu-bar-only
  (`accessory` policy).
- Settings → General → Startup, macOS-gated, keys in all 10 locales.

## Why not derive from the window's `show`/`hide` events

On macOS those are **occlusion** events: Electron emits them only from
`windowDidChangeOcclusionState`, and `NotifyWindowShow`/`NotifyWindowHide` have zero call
sites in `native_window_mac.mm`. So they also fire when the panel is covered, minimized, or
on another Space, making the icon flicker as you switch windows. Verified on macOS: covering
the control panel fires `hide` while `isVisible()` stays `true`.

Linux and Windows emit them straight from `Show()`/`Hide()` (`native_window_views.cc`), so
this mistake tests green off macOS.

## Known gaps (pre-existing)

- **Launch flash.** `LSUIElement: false` creates the icon before any JS runs, so with the
  setting off it flashes briefly at startup. Removing that means flipping the default for
  everyone.
- **Electron swallows `dock.hide()` within 1s of a `dock.show()`** (`DockHide`,
  `browser_mac.mm`), so closing the control panel immediately after opening it leaves the
  icon up until the next hide. Working around the throttle risks the duplicate-icon bug it
  guards against. Commented at the call site.

## Testing

- `npm test`: 467 pass, 0 fail, 8 skipped (includes 6 new `dockPolicy` tests)
- `npm run typecheck` clean; eslint and prettier clean on touched files
- Verified against real Electron on macOS: `dockManager` + `dockPolicy` drive the Dock icon
  correctly across all setting/panel-state transitions (8/8)
- No functional `app.dock` or `setActivationPolicy` call sites outside `dockManager.js`
- With the control panel closed and auto-hide-when-idle on, the Dock icon stays hidden across
  dictations. That is the #428 repro.
- With the setting off, the app runs menu-bar-only and the control panel keeps working normally and `Command+,` all behave as usual, and Cmd+Q still quits.


gabrielste1n (approved):
Thank you @dreasan — this is a genuinely excellent PR. The root-cause analysis (the leftover `app.dock.show()` from the minimize-era code) was spot on, and centralizing all five scattered dock call sites into a single owner with a pure, tested policy is exactly the right structure. The write-up on why occlusion events can't be trusted as the visibility source saved us from a whole class of flicker bugs.

One change I made on top (merged main in too): I dropped the "Show Dock Icon" setting and kept just the follow-the-panel behavior. Reasoning: the icon appearing while the control panel is open and disappearing when it closes to the tray is the standard macOS pattern for menu-bar utilities, and it alone resolves every report in #428/#134 — nobody asked to suppress the icon *while the panel is open*. Our settings page is already crowded, so we're trying hard not to add a row unless users actually ask for it. If menu-bar purists want a "never show" mode later, your dockManager/dockPolicy structure makes it a five-line addition — which is a credit to the design.

Nice side effect of deriving the launch policy from panel visibility: start-minimized launches no longer flash a Dock icon at all.

dreasan:
@gabrielste1n Thanks, the reasoning makes sense and deriving it only from panel visibility is the right approach. I doubt many users would actually ask for a hide-dock setting. I admit I'm probably one of those purists, but we're few and far between.

PR #1204, opened 2026-07-16
fix(audio): fall back to the default mic when the preferred device stays muted

### LINKED ISSUE #1152: Windows: built-in microphone preference selects muted Realtek device instead of working default headset

xAlcahest:
## Summary

On Windows with Prefer Built-in Microphone enabled, Chromium can hand back a live but permanently muted track for the built-in mic (the OS-level mute state is mirrored into `track.muted` and never flips back). The recovery in `reacquireIfDead` (`src/helpers/micTrackHealth.js`) then made it worse: its retry re-resolved the same built-in device by label and returned the new stream without checking it, so the whole dictation recorded zeros and ended in "No audio detected".

The retry now stays on the same device first (so the idle-wake recovery from #922 keeps working), but its track gets verified: if it is still muted, the device is remembered as rejected for the session (cleared on devicechange) and the mic is reopened with no deviceId constraint so the OS default takes over. If even that input delivers no audio, recording aborts immediately with a clear "Microphone Muted" error instead of another silent failure.

Fixes #1152

## Changes

- src/helpers/micTrackHealth.js: `reacquireIfDead` takes an optional fallback tier that verifies the same-device retry, hops to unpinned default constraints when it stays dead, verifies the fallback too, and signals when no input delivers audio (including when either getUserMedia call throws).
- src/helpers/audioManager.js: tracks `rejectedMicDeviceId` for the session, skips pinning that device in `getAudioConstraints`, and aborts with a "Microphone Muted" error in both `startRecording` and `startStreamingRecording` when the fallback is unusable as well.
- test/helpers/micTrackHealth.test.js: covers the new paths (healthy retry keeps the device, dead retry falls back, failed or dead fallback surfaces the unusable state, muted persistence with mock timers).


gabrielste1n (approved):
Reviewed the fallback logic end-to-end (all six retry/fallback path combinations, stream ownership, and session rejection tracking) — solid fix for #1152 that preserves the #922 idle-wake recovery. Merged latest main and extracted the duplicated mic-acquisition block into acquireHealthyMicStream; tests, lint, typecheck, and format all pass.

PR #1205, opened 2026-07-16
fix(meeting): reliable macOS meeting notifications, unified Google Calendar CTA, and note-title fix

samkawsarani:
## Summary

Improves the meeting-notification experience — primarily making the "meeting in
progress / meeting starting" prompt reliable on **macOS** and making **Google
Calendar** reminders consistent with in-app detection — and fixes a note-title
data-loss bug that the new calendar "Start Recording" flow surfaced.

## Why

- **macOS: the "Start Recording" button was effectively unclickable.** The
  meeting-notification overlay window was created click-through on darwin only
  (`setIgnoreMouseEvents(true, { forward: true })`), so the CTA only registered a
  click if the hover-driven interactivity toggle re-enabled input at just the
  right moment. The app's own update-notification overlay never does this and
  works fine.
- **Google Calendar reminders were inconsistent with meeting detection.** They
  showed a native OS `Notification` whose only action was a whole-body click that
  broadcast an event nothing in the renderer consumed — i.e. no working "Start
  Recording" action, and a different look/behaviour from the in-app detection card.
- **macOS audio detection could silently stop working.** If the native
  `macos-mic-listener` binary is unavailable, detection fell back to a brittle
  `ioreg "IOAudioEngineState" = 1` check and logged nothing.

### Fixes
- **Reliable macOS "Start Recording" CTA** — the meeting-notification overlay is
  now interactive by default (removed the darwin-only click-through), matching the
  update-notification overlay; the hover interactivity path no longer re-enables
  click-through. (`windowManager.js`, `MeetingNotificationOverlay.tsx`)
- **Google Calendar reminders now use the same in-app card with a real CTA** —
  routed through a new `MeetingDetectionEngine.showCalendarReminder()` and the
  shared `handleNotificationResponse("start")` path (creates/opens the meeting
  note and links the calendar event + participants), replacing the dead
  native-`Notification` path. (`meetingDetectionEngine.js`,
  `googleCalendarManager.js`, `main.js`)
- **Hardened macOS audio-detection fallback** — the polling fallback also detects
  active audio I/O clients and logs a warning when it falls back from the native
  listener, so detection no longer fails silently. (`audioActivityDetector.js`)
- **Meeting note title no longer wiped on load** — opening a meeting note (e.g.
  via the calendar CTA) could blank its title: the content autosave and
  note-switch flush were persisting a transiently-empty title during the editor
  remount. Automatic saves can no longer blank a non-empty title, while deliberate
  title edits/clears still work. (`PersonalNotesView.tsx`, `NoteEditor.tsx`)

### Enhancements
- **"Join & Start Recording"** — when a calendar event has a meeting link (Google
  Meet `hangout_link`, or a Zoom/other link inside `conference_data`), the CTA
  opens the link to join the call in addition to starting recording, and the label
  switches accordingly. Events without a link keep "Start Recording".
  (`meetingDetectionEngine.js`, `MeetingNotificationOverlay.tsx`)
- **Skip reminders for solo calendar events** — no reminder for events with no
  other attendees (personal/focus blocks); 1:1s and group meetings still remind.
  (`meetingDetectionEngine.js`)
- **Meeting-mode layout setting** — Settings → Appearance → "Open meeting mode
  in:" (Full width / Side panel, default Full width). It controls whether meeting
  mode snaps to the right-side panel for **all** entry points (hotkey, notification
  CTA, clicking an upcoming meeting), not just the hotkey.
  (`settingsStore.ts`, `useSettings.ts`, `SettingsPage.tsx`, `ControlPanel.tsx`,
  `src/locales/*`)

## Testing

- `npm run lint`, Prettier, `tsc --noEmit`, and `node scripts/check-i18n.js` pass;
  the non-Electron meeting unit tests pass.
- Manual (macOS): mic activity → "Meeting Detected" overlay appears and "Start
  Recording" is clickable on the first click; a Google Calendar event → the same
  in-app card with a working CTA (and "Join & Start Recording" when the event has
  a Meet/Zoom link); solo calendar blocks don't notify; opening a meeting note
  keeps its title; the Appearance layout setting switches meeting mode between
  side-panel and full-width for every entry point.

## Notes

- New i18n keys are added across all locale files. The meeting-notification
  overlay CTA strings remain hardcoded English, matching the existing strings in
  that overlay.
- This branch bundles the macOS notification fix, the calendar-reminder
  consistency work, a couple of enhancements, and the note-title fix. Happy to
  split it into smaller, focused PRs if you'd prefer.


gabrielste1n:
Heads-up: a large part of this shipped in [v1.7.6](https://github.com/OpenWhispr/openwhispr/releases/tag/v1.7.6) via #1182 — meeting prompts (including Google Calendar reminders) now render in the always-on-top in-app overlay instead of macOS notifications, which fixes the unclickable Start Recording action and unifies the calendar CTA ("Join & transcribe" when the event has a link). Rather than closing this outright: the **note-title data-loss fix** looks like it may still be valuable on its own. Would you rebase onto main and slim this down to that fix (plus anything else 1.7.6 didn't cover)? Happy to review it quickly.

samkawsarani:
i will close this PR for now and port the "enhancements" i introduced to a new PR to not conflate things

PR #1206, opened 2026-07-16
feat(meeting): auto-end a meeting recording after silence, with an interruptible countdown

samkawsarani:
## Summary

Automatically finalizes a meeting recording after a period of silence, so you
don't have to remember to stop it (e.g. when everyone has left the call). Before
it stops, an always-on-top "Meeting ended?" prompt gives you a short countdown to
cancel. Opt-out via a setting (default on).

## Why

When you record a meeting, you currently have to manually stop it. If you walk
away when the call ends, transcription keeps running until you notice. This adds
a safe, interruptible auto-stop.

## Behaviour

- While a meeting is recording, speech activity from **both** the mic and system
  audio is tracked (via the live transcription partials/segments — not raw levels,
  which would trip on ambient noise).
- After **60s with no speech from anyone**, an always-on-top **"Meeting ended?"**
  card appears with a **5-second countdown**:
  - **"Keep recording"** cancels the auto-end and resets the silence window.
  - If the countdown elapses (you've walked away), the recording is **stopped and
    finalized** through the normal meeting-stop path.
  - If **anyone starts speaking again** while the prompt is up, it's dismissed
    automatically and recording continues.
- Gated by a new opt-out setting **Settings → Notifications → "Auto-end meeting
  recording"** (default **on**).

## Implementation

- **Silence monitor** lives in `MeetingRecordingMount` (always mounted during
  recording): it tracks a "last speech activity" timestamp from mic/system
  transcription updates and, after the threshold, asks main to show the prompt.
- **Prompt is its own always-on-top overlay window** (`MeetingEndPromptOverlay`,
  routed via `?meeting-end-prompt=true`), modeled on the existing update-
  notification overlay so it's never click-through and the button is reliable on
  macOS. `windowManager` gains `showMeetingEndPrompt` /
  `resolveMeetingEndPrompt` / `dismissMeetingEndPrompt`, wired over new IPC
  channels in `ipcHandlers.js` + `preload.js`.
- A main-process **safety timeout** force-ends if the overlay renderer never
  responds, so a crashed/blocked overlay can't leave a recording running forever.
- New `autoEndMeetingRecording` boolean in the settings store + `useSettings`
  hook + a Settings toggle, and new i18n keys across all locale files.

## Scope note

Only a sustained **silence** signal drives auto-end. Two other candidate "meeting
ended" signals were deliberately **not** used: the scheduled calendar end time
(meetings run long / aren't always on a calendar) and a meeting app quitting or
the mic muting (a muted user or a backgrounded app quitting would cut a live
recording short). The setting lets users turn the behaviour off entirely.

## Testing

- `npm run lint`, Prettier, `tsc --noEmit`, and `node scripts/check-i18n.js` pass;
  the non-Electron meeting unit tests pass.
- Manual: start a meeting recording, stay silent ~60s → the "Meeting ended?" card
  appears counting down; "Keep recording" cancels; letting it reach 0 stops and
  finalizes the recording; speaking during the countdown auto-dismisses it;
  turning the setting off suppresses the prompt.

## Notes

- The prompt's strings are added to all locale files. The 60s silence threshold
  and 5s countdown are fixed values (kept simple; could be made configurable later).


samkawsarani:
@gabrielste1n i rebased this pr after 1.7.6

PR #1209, opened 2026-07-17
fix(gpu): keep GPU acceleration download state after the settings picker remounts

### LINKED ISSUE #1136: Problem in GPU acceleration download UI
**Describe the bug**
When I was downloading the accelerator I accidentally go to other beige so I can update some other settings and when I return to it showed me the button again and it is not applicable anymore when I checked on My Windows resources i can see that download is still going on and there is a network consumption so i believe it is just a network problem

**To Reproduce**
Download then go to settings then speech to text then select Local and click download GPU Acceleration After that go to any other page When you return the button will be there and it will not be clickable.
After that you can go to resources you will see that Open whisper it's consuming network even though the eye don't show the download is running.

**Screenshots**

Image

Image

- OS: Windows 11 Pro, 25H2, 26200.8246

stantheman0128:
## Summary

The GPU acceleration download (Settings > Speech to Text > Local) runs in the main process, so it keeps going when the user navigates to another settings page. The picker's state does not survive that navigation. On return, the progress bar was gone and the "Enable GPU acceleration" button was back even though the download was still running. Clicking the button again hit the manager's "Download already in progress" guard and silently failed, and when the download eventually finished nothing told the remounted picker, so the UI never flipped to the active state. This is what #1136 reports: the button is back and does nothing while Task Manager still shows network traffic.

To reproduce on main: Settings > Speech to Text > Local > start the GPU acceleration download > open any other settings page > come back while the download is running.

The fix:

1. `GpuBinaryManager` extends `EventEmitter` and emits `download-settled` with `{ success, error? }` from the `finally` block of `download()`, so it fires exactly once per started download, whether it installs, fails, or is cancelled. The duplicate-download guard does not emit, since the original download is still in flight.
2. `ipcHandlers.js` forwards the event to all windows as `cuda-download-complete` and `vulkan-whisper-download-complete`, exposed in the preload as `onCudaDownloadComplete` / `onVulkanWhisperDownloadComplete`.
3. On mount, `TranscriptionModelPicker` restores the downloading flag from `get-cuda-whisper-status` / `get-vulkan-whisper-status` (both handlers already reported it, the renderer just never read it) and subscribes to the completion event. A remounted picker shows the progress bar again (progress events already reach the same webContents), keeps the cancel link working, and settles to "GPU acceleration active" once the install finishes.

No new polling and no renderer-side persistence, just the existing status call plus one new event.

Fixes #1136

## Changes

- `src/helpers/gpuBinaryManager.js`: emit `download-settled` once a started download settles
- `src/helpers/ipcHandlers.js`: broadcast the completion to all windows
- `preload.js`, `src/types/electron.ts`: renderer listeners for both backends
- `src/components/TranscriptionModelPicker.tsx`: rehydrate the downloading state on mount and settle it from the completion event
- `test/helpers/gpuBinaryManager.test.js`: regression tests for the settled event (success, failure, and the duplicate-download guard staying silent)

## Verification

The new tests fail without the manager change and pass with it:

```
$ git checkout origin/main -- src/helpers/gpuBinaryManager.js
$ node --test test/helpers/gpuBinaryManager.test.js
not ok 12 - settled event: emitted with success after install and with the error after a failure
not ok 13 - settled event: the duplicate-download guard does not emit
# pass 11
# fail 3

$ git checkout HEAD -- src/helpers/gpuBinaryManager.js
$ node --test test/helpers/gpuBinaryManager.test.js
# tests 14
# pass 13
# fail 1
```

The one remaining failure ("binary is executable") is a pre-existing artifact of running this file on a Windows checkout: the test pins `process.platform` to linux and the chmod execute bit does not survive on NTFS. It fails identically on origin/main; CI runs the suite on ubuntu-latest.

```
$ npm run typecheck
> cd src && tsc --noEmit
(clean)

$ npm run lint
> eslint . && cd src && eslint .
(clean)
```

Full suite on this Windows machine: 434 pass, 24 fail at HEAD, and the failing set is identical by name to a fresh run at origin/main (all pre-existing Windows-host environment failures), so the branch introduces no new failures.

This fix was developed with AI assistance (Claude Code) and verified locally as shown above.

### DIFF
--- a/preload.js
+++ b/preload.js
@@ -269,6 +269,10 @@ contextBridge.exposeInMainWorld("electronAPI", {
     "cuda-download-progress",
     (callback) => (_event, data) => callback(data)
   ),
+  onCudaDownloadComplete: registerListener(
+    "cuda-download-complete",
+    (callback) => (_event, data) => callback(data)
+  ),
   onCudaFallbackNotification: registerListener(
     "cuda-fallback-notification",
     (callback) => () => callback()
@@ -281,6 +285,10 @@ contextBridge.exposeInMainWorld("electronAPI", {
   deleteVulkanWhisperBinary: () => ipcRenderer.invoke("delete-vulkan-whisper-binary"),
   onVulkanWhisperDownloadProgress: registerListener(
     "vulkan-whisper-download-progress",
+    (callback) => (_event, data) => callback(data)
+  ),
+  onVulkanWhisperDownloadComplete: registerListener(
+    "vulkan-whisper-download-complete",
     (callback) => (_event, data) => callback(data)
   ),
   onGpuFallbackNotification: registerListener(
--- a/src/components/TranscriptionModelPicker.tsx
+++ b/src/components/TranscriptionModelPicker.tsx
@@ -543,12 +543,16 @@ export default function TranscriptionModelPicker({
         if (cuda?.gpuInfo.hasNvidiaGpu) {
           setGpuBackend("cuda");
           setGpuDownloaded(cuda.downloaded);
+          // The download runs in the main process, so it survives this
+          // component unmounting; pick it back up after a remount.
+          setGpuDownloading(cuda.downloading === true);
           return;
         }
         const vulkan = await window.electronAPI?.getVulkanWhisperStatus?.();
         if (vulkan?.vulkan.available) {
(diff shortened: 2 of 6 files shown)


stantheman0128:

PR #1211, opened 2026-07-17
fix(reasoning): stop Groq dictation cleanup from failing silently

### LINKED ISSUE #1183: Cleanup via Groq: thinking suppression sends invalid `reasoning_effort: "none"`, and 4xx errors are silently retried with backoff — every dictation delayed ~5-8s, raw text pasted
## Environment
- OpenWhispr 1.7.5, Windows 11
- ASR: local Parakeet v2
- Cleanup: Groq BYOK — reproduced with both `openai/gpt-oss-120b` and `qwen/qwen3-32b`
- Cleanup "disable thinking" toggle: ON (the default)

## Symptom
With dictation cleanup enabled through Groq, every dictation takes ~5+ seconds
between end-of-speech and paste. The pasted text is the raw transcript
(fillers intact), i.e. cleanup is failing every time — but no error is ever
shown.

**Confirmed workaround:** turning OFF the cleanup "disable thinking" toggle
makes responses near-instant and cleanup works, with both models.

## Root cause (two bugs compounding)

**Bug 1 — suppression sends a param Groq rejects.**
`cleanupDisableThinking` defaults to `true` (`settingsStore.ts` ~line 1205).
For thinking-capable models, `applyThinkingSuppression`
(`src/services/ai/thinkingSuppression.ts`) then sets:

```js
requestBody.reasoning_effort = "none";
requestBody.chat_template_kwargs = { enable_thinking: false };
```

This *overwrites* the valid `reasoning_effort: "low"` that
`ReasoningService.ts` (~line 190) deliberately sets for gpt-oss cleanup.
Groq's gpt-oss endpoints accept only `low | medium | high`, so the request
400s. (With the toggle off, the built-in `"low"` survives and everything
works — which is how I confirmed this.)

**Bug 2 — the 400 is retried as if it were a network error.**
In `ReasoningService.ts`, a non-OK response is thrown as a bare Error:

```js
throw new Error(errorMessage);
```

But `createApiRetryStrategy` (`src/utils/retry.ts` ~line 43) decides
retryability via:

```js
if (!error.response) return true; // Network error
```

A bare `Error` never has `.response`, so **every** API failure — including
hard 400s and 429s — takes the "network error, retry" branch. With
`RETRY_CONFIG` (3 retries, 1s initial delay, 2x backoff), that's
1s + 2s + 4s ≈ 7s of silent dead air per dictation before the final throw
and the raw-transcript fallback. The 5xx status check in the strategy is
dead code, since thrown errors never carry a status.

## Suggested fixes
1. Attach the status to the thrown error (e.g. `err.status = res.status` /
   `err.response = { status: res.status }`) so `createApiRetryStrategy`'s
   status check actually runs; don't retry 4xx.
2. In `suppressThinking`, don't send `reasoning_effort: "none"` to Groq —
   for gpt-oss use `"low"` (or simply don't override the value
   `ReasoningService` already set).
3. Surface cleanup failure to the user (toast / preview badge) instead of
   silently pasting the raw transcript — this bug was invisible for that
   reason.

## Related
- Adjacent: the Groq model registry in `modelRegistryData.json` still lists
  `qwen/qwen3-32b`, `llama-3.3-70b-versatile`, `llama-4-scout`, and
  `moonshotai/kimi-k2-instruct-0905`, all of which Groq deprecated
  March–June 2026. When those shut down, users on them will hit Bug 2's
  silent retry loop with no explanation.
- Possibly the same underlying issue as #1089 ("Dictation cleanup slow when
  using self hosted LLM gateway") and related to #1103 / #783
  (thinking-suppression dialect handling).

xAlcahest:
## Summary

Dictation cleanup through Groq fails on every request while the default "disable thinking" toggle is on: `suppressThinking` sends `reasoning_effort: "none"` plus `chat_template_kwargs`, and Groq 400s on both, since gpt-oss only accepts low/medium/high, qwen3 only none/default, and `chat_template_kwargs` is not supported at all. The 400 then gets retried as if it were a network error, because `createApiRetryStrategy` retries any error without a `.response` and the thrown errors never carried a status, so every dictation sits through a 1+2+4s backoff before silently pasting the raw transcript.

The per-provider dialect switch now lives in `thinkingSuppressionDialects.ts` with a Groq branch that sends the right value per model family and never `chat_template_kwargs`. API errors carry their HTTP status, and the retry strategy retries only network faults, 429 and 5xx. When cleanup still fails, a toast says the dictation was pasted raw instead of leaving the user guessing.

Fixes #1183 #990 

## Changes

- `src/services/ai/thinkingSuppressionDialects.ts`: new pure module with the per-provider dialect switch, now model-aware; on Groq, qwen gets `reasoning_effort: "none"`, gpt-oss gets `"low"`, unknown families get nothing, and `chat_template_kwargs` is never sent
- `src/services/ai/thinkingSuppression.ts`: gating unchanged, threads the model through to the dialect switch
- `src/utils/retry.ts`: `createApiRetryStrategy` retries when no HTTP status exists (network fault) and on 429/5xx; 4xx fail fast
- `src/services/ReasoningService.ts`, `src/services/ai/inferenceProviders/openai.ts`, `src/services/ai/inferenceProviders/gemini.ts`: attach the HTTP status to errors thrown on non-OK responses (tinfoil already gets it from its SDK)
- `src/helpers/audioManager.js`: the three cleanup-fallback catch sites (local, cloud, streaming) record the failure, gated to the cleanup route so agent failures stay out of it
- `src/stores/cleanupFailureStore.ts`, `src/components/CleanupFailureToastListener.tsx`, `src/main.jsx`: cleanup-failure toast wired like the existing Tinfoil listener, with `app.toasts.cleanupFailed` keys in all 10 locales
- `src/tsconfig.json`: enable `allowImportingTsExtensions` so `retry.ts` can use an explicit `.ts` import, which Node's test runner needs to resolve it
- `test/helpers/thinkingSuppressionDialects.test.js`, `test/helpers/apiRetryStrategy.test.js`, `test/helpers/cleanupFailureStore.test.js`: unit tests for the dialect table, the retry strategy, and the failure store


gabrielste1n (approved):
Reviewed end to end: Groq dialect table matches each model family's accepted reasoning_effort values, retry strategy now fails fast on 4xx across all four consumers, and the cleanup-failure toast is wired per the existing listener pattern. Verified locally: typecheck, lint, 480 tests, renderer build.

Improvise2500:
hey @gabrielste1n , does this mean I can turn the switch back on? Simply turning it off seemed to fully fix the issue, and I've been using openwhisper, and the groq cleanup has been working with the switch toggled off.

PR #1212, opened 2026-07-17
fix(url-download): remove partial file only after the write stream closes

### LINKED ISSUE #1197: Cancelling URL audio downloads can leave partial files

xAlcahest:
## Summary
Cancelling a URL audio download could leave an orphan `ow-url-*` file in the temp dir. `fs.createWriteStream()` opens its destination asynchronously, and the cleanup path in `streamToFile` (`src/helpers/urlAudioDownloader.js`) called `destroy()` and unlinked the file right away: if cancellation landed before the open completed, the unlink hit ENOENT and the pending open then created the file after cleanup had already run. Cancellation also settled before the stream was actually closed.

Cleanup now waits for the stream's `close` event (Node defers destruction until a pending open completes, and `close` only fires after the fd is closed), then unlinks the file, ignoring ENOENT and logging any other unlink error, and only then rejects with the original error. A regression test reproduces the race deterministically by gating the stream's open through the `fs` option override of `createWriteStream`; it fails on the previous code.

Fixes #1197

## Changes
- src/helpers/urlAudioDownloader.js: `cleanupFile` in `streamToFile` is now async — destroys the stream, waits for `close`, then unlinks (ENOENT ignored, other errors logged); `bail` and the `finish` stat-failure path reject only after cleanup completes.
- test/helpers/urlAudioDownloader.test.js: regression test that holds the write stream's open past cancellation and asserts the rejection happens only after `close`, with no leftover temp file.

### DIFF
--- a/src/helpers/urlAudioDownloader.js
+++ b/src/helpers/urlAudioDownloader.js
@@ -1064,20 +1064,37 @@ function streamToFile(
       bail(Object.assign(new Error("Download stalled"), { code: "DOWNLOAD_FAILED" }));
     });
 
-    const cleanupFile = () => {
-      fileStream.destroy();
-      try {
-        fs.unlinkSync(tempPath);
-      } catch {}
-    };
+    // destroy() returns before the stream's async open finishes, so unlinking
+    // right away can race a pending open into recreating the file. Wait for 'close'.
+    const cleanupFile = () =>
+      new Promise((done) => {
+        const unlink = () =>
+          fs.unlink(tempPath, (e) => {
+            try {
+              if (e && e.code !== "ENOENT") {
+                debugLogger.warn("Failed to remove partial download", {
+                  tempPath,
+                  error: e.message,
+                });
+              }
+            } finally {
+              done();
+            }
+          });
+        if (fileStream.closed) {
+          unlink();
+          return;
+        }
+        fileStream.once("close", unlink);
+        fileStream.destroy();
+      });
 
     const bail = (err) => {
       if (settled) return;
       settled = true;
       stall.clear();
       if (abortSignal) abortSignal.removeEventListener("abort", onAbort);
-      cleanupFile();
-      reject(err);
+      cleanupFile().then(() => reject(err));
     };
 
     // React to cancel even while stalled (no data events arriving).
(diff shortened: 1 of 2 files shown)


gabrielste1n (approved):
Reviewed end-to-end: verified the regression test fails on main and passes with the fix, full suite green locally. Cleanup correctly waits for the stream's close event before unlinking, covering both direct and proxy download paths.

PR #1213, opened 2026-07-17
fix(reasoning): stop Mistral cleanup requests failing with 422

### LINKED ISSUE #844: Mistral "Language Model" Requests Fail with error 422
**Describe the bug**

Configuring Mistral as the cleanup ("reasoning") provider via a custom OpenAI-compatible provider pointed at `https://api.mistral.ai/v1`, causes every cleanup request to fail. The API key is valid and Mistral returns the model list correctly, but cleanup requests come back as HTTP 422 because OpenWhispr's request body includes two fields Mistral's API does not accept:

- `max_completion_tokens` (Mistral uses `max_tokens`; `max_completion_tokens` is the newer OpenAI naming)
- `chat_template_kwargs: { enable_thinking: false }` (not part of Mistral's API as far as I know)

Mistral's validator rejects unknown fields outright rather than ignoring them, so the request never succeeds. In the UI this surfaces only as `Test failed: [object Object]` in Prompt Studio and as `[WARN][notes][renderer] Reasoning failed { "error": "[object Object]" }` in the debug log, because the renderer's error handler stringifies the error object without extracting the response body. The underlying 422 is only visible via DevTools.

**To Reproduce**

1. Go to Settings, AI Processing, Cleanup scope.
2. Set provider to Mistral (or add a custom OpenAI-compatible provider with base URL `https://api.mistral.ai/v1`).
3. Paste a valid Mistral API key and select any Mistral model (e.g. `mistral-small-latest`).
4. Trigger a cleanup, either by dictating with cleanup enabled or by hitting Test in Prompt Studio.
5. Cleanup silently fails; raw transcript is pasted with no cleanup applied. DevTools Network tab shows HTTP 422 from `api.mistral.ai/v1/chat/completions` with the following response body:

```json
{
    "object": "error",
    "message": {
        "detail": [
            {
                "type": "extra_forbidden",
                "loc": ["body", "max_completion_tokens"],
                "msg": "Extra inputs are not permitted",
                "input": 4096
            },
            {
                "type": "extra_forbidden",
                "loc": ["body", "chat_template_kwargs"],
                "msg": "Extra inputs are not permitted",
                "input": {"enable_thinking": false}
            }
        ]
    },
    "type": "invalid_request_error",
    "param": null,
    "code": null,
    "raw_status_code": 422
}
```

**Expected behavior**

Cleanup requests sent to Mistral's API should succeed. The Mistral adapter should send `max_tokens` instead of `max_completion_tokens`, and should not include `chat_template_kwargs` in requests to Mistral (or any other provider that doesn't support it). Separately, the renderer's error handler should log the actual response body or error message instead of `[object Object]`, since the current logging makes this class of bug very hard to diagnose without DevTools.

**Desktop:**
- OS: Kubuntu 24.04 (X11, KDE)
- OpenWhispr version: 1.7.2
- Electron: 41.2.0
- Node: 24.14.0

**Additional context**

Verified directly that the API key, billing, and network path are working: `curl https://api.mistral.ai/v1/models` with the same key returns the full model list. The 422 is purely about the request body shape.

The same bug appears to manifest in two places that share the cleanup provider code path:
- Real dictation cleanup: transcript is pasted raw, debug log shows `Reasoning failed { "error": "[object Object]" }`.
- Prompt Studio test: shows `Test failed: [object Object]` with no further detail.

May be related to issue #812, where a user reported that adding Mistral via the custom provider section was "to no avail." This 422 could be why.

**Screenshots**

Image

xAlcahest:
## Summary

Depends on #1211: builds on the thinkingSuppressionDialects module introduced there. Merge #1211 first.

Adding Mistral as a custom OpenAI-compatible provider made every cleanup request fail with a 422, because Mistral validates strictly and rejects unknown fields. The body carried `max_completion_tokens` (unknown model ids fall into the `getOpenAiApiConfig` catch-all, and Mistral only takes `max_tokens`) and, with thinking disabled, `chat_template_kwargs` from the generic dialect branch. The 422 then surfaced as `[object Object]`, since Mistral returns `message` as an object with a Pydantic `detail` array and the error extraction assumed a string.

`suppressThinking` now has a mistral dialect picked by endpoint host: `reasoning_effort: "none"` (Mistral's own thinking switch, verified against current docs), never `chat_template_kwargs`, and nothing at all for legacy magistral models, which reason natively. The same host detection makes openai.ts send `max_tokens` and keep temperature. Error bodies go through a small extractor that reads the detail entries, so failures now name the rejected fields. The agent/chat streaming path in ReasoningService.ts builds its own request body and had the same two problems; it now goes through the same host detection and error extractor, and known dialect hosts skip the /responses probe.

## Changes

- `src/services/ai/thinkingSuppressionDialects.ts`: `detectEndpointDialect` recognises mistral.ai hosts and returns the request shape (`max_tokens`, temperature supported); new mistral branch in `suppressThinking`
- `src/services/ai/thinkingSuppression.ts`: optional `baseUrl` param, the detected endpoint dialect wins over the generic provider key
- `src/services/ai/inferenceProviders/openai.ts`: uses the dialect's token param and temperature flag, passes the resolved base URL through, goes straight to chat completions on dialect hosts, and extracts error messages via the new helper
- `src/services/ReasoningService.ts`: the streaming path and `callChatCompletionsApi` pass the endpoint to the dialect detection and use the shared error extractor
- `src/services/ai/apiErrorMessage.ts`: pure helper that turns OpenAI-style and FastAPI-style error bodies into a readable string instead of `[object Object]`
- `test/helpers/thinkingSuppressionDialects.test.js`, `test/helpers/apiErrorMessage.test.js`: cover the mistral dialect, host detection edge cases and error extraction shapes

Fixes #844


gabrielste1n (approved):
Verified the root cause (unknown model ids falling into the max_completion_tokens catch-all plus the generic chat_template_kwargs branch) and confirmed reasoning_effort is Mistral's documented thinking switch. Host detection is robust against lookalike domains, the error extractor handles every shape in the wild, and the dialect now covers the cleanup, agent streaming, and AI SDK paths consistently. Typecheck, lint, format, and the full test suite are clean.

PR #1214, opened 2026-07-17
fix(macos): resolve ffmpeg/ffprobe on Nix-managed setups (#912)

### LINKED ISSUE #912: [BUG] ffmpeg not detected on macOS managed via Nix-Darwin
**Describe the bug**
Transcribing audio files recorded via hotkey doesn't work as the app is not detecting ffmpeg and its binaries. 

**To Reproduce**
1. Open a note
2. Click the recording hotkey
3. Finish the recording
4. Error 
```
custom API Error: 500 {"error":{"message":"ffprobe not found (usually installed with ffmpeg)","type":"server_error","param":null,"code":null}}
```

**Expected behavior**
Transcription is processed

 - OS: macOS M3 Pro
 - Browser Chrome
 - ffmpeg installed via nix into `/run/current-system/sw/bin/ffmpeg` and the like for `ffprobe` 

**Additional context**
The issue happens both with local models (whisper turbo) and self hosted models via an OpenAI compatible API.
Issue is that 
1.   the code only checks these hardcoded paths:
    - macOS: /opt/homebrew/bin/ffmpeg, /usr/local/bin/ffmpeg
    - Linux: /usr/bin/ffmpeg, /usr/local/bin/ffmpeg
2. the code does fall back to scanning process.env.PATH, which would work — but only if that path is in the app's environment. Electron apps on macOS are launched by launchd (the GUI session), not the shell. That means they never source ~/.zshrc or ~/.zshenv, so process.env.PATH inside the app is a minimal system PATH containing only the standard directories, not the Nix profile.
3.  main.js:314-333 (setupProductionPath()) only prepends /usr/local/bin, /opt/homebrew/bin, /usr/bin, /bin, /usr/sbin,  /sbin to PATH on macOS — no Nix paths.

**Suggested fix**
1. Add the mentioned path to the hardcoded paths (not nice)
2. Enable the app to figure out where ffmpeg is installed dynamically

stantheman0128:
# fix(macos): resolve ffmpeg/ffprobe on Nix-managed setups (#912)

## Summary

Fixes #912. On macOS managed via nix-darwin, transcription fails with
`ffprobe not found (usually installed with ffmpeg)`. This makes ffmpeg/ffprobe
resolution robust instead of relying on a fixed list of Homebrew/system paths.

## Root cause

The reporter diagnosed this precisely, and the code confirms it:

1. ffmpeg detection checked only a hardcoded list (macOS: `/opt/homebrew/bin`,
   `/usr/local/bin`) and then scanned `process.env.PATH`.
2. Electron apps on macOS are launched by launchd, not a shell, so they never
   source `~/.zshrc`/`~/.zshenv`. `process.env.PATH` is a minimal system PATH
   with no Nix profile, so the PATH scan finds nothing either.
3. `setupProductionPath()` in `main.js` only prepended the usual
   Homebrew/system dirs, never the Nix location where ffmpeg actually lives
   (`/run/current-system/sw/bin`).

Because `process.env.PATH` stays minimal, the yt-dlp sidecar that URL audio
download shells out to also cannot find `ffprobe`, which is the binary the
error message names (the bundled `ffmpeg-static` ships ffmpeg only, no ffprobe).

## Changes

- New `src/helpers/systemBinaryPath.js`: a shared resolver that probes, in
  priority order, the inherited PATH, the real login-shell PATH (obtained by
  spawning `$SHELL -lic` once, cached, best-effort), and known extra install
  locations including Homebrew and the common Nix paths
  (`/run/current-system/sw/bin`, `~/.nix-profile/bin`,
  `/nix/var/nix/profiles/default/bin`). Filesystem and shell access are
  injectable so the logic is unit-testable.
- The three existing ffmpeg resolvers (`ffmpegUtils.js`, `whisper.js`,
  `whisperServer.js`) now delegate their system fallback to
  `findSystemBinary("ffmpeg")` instead of their own hardcoded lists. Bundled
  `ffmpeg-static` resolution is unchanged and still takes priority.
- `setupProductionPath()` in `main.js` rebuilds `process.env.PATH` from the
  login-shell PATH plus those extra locations, so child processes
  (yt-dlp then ffprobe, whisper.cpp) inherit a PATH they can actually resolve
  tools from. Still a no-op on Windows and in development.
- `debugLogger.js` ffmpeg diagnostics now also probe the Nix paths, so debug
  logs attached to future reports show whether a Nix ffmpeg is present.

The Windows `C:\ffmpeg\bin` fallback and the executable-bit check on Unix are
preserved by the resolver.

## Verification / Evidence

Added `test/helpers/systemBinaryPath.test.js`, run with the repo's harness
(`node --test`). It reproduces the bug at the logic level: with a launchd-style
minimal PATH, the old hardcoded darwin list and PATH scan miss a Nix ffmpeg,
while the new resolver finds ffmpeg (and ffprobe) via the extra dirs and the
login-shell PATH.

```
node --test test/helpers/systemBinaryPath.test.js

ok 1 - regression: the old hardcoded darwin list misses a Nix ffmpeg
ok 2 - findSystemBinary finds a Nix ffmpeg via the extra bin dirs
ok 3 - findSystemBinary finds ffprobe via the login-shell PATH
ok 4 - findSystemBinary skips a match that is not executable
ok 5 - findSystemBinary appends .exe and honors the win32 fallback dir
ok 6 - extraBinDirs lists the common Nix locations on darwin
ok 7 - systemBinDirs de-duplicates while preserving priority order
ok 8 - buildAugmentedPath injects Nix dirs a launchd PATH is missing
# tests 8
# pass 8
# fail 0
```

`node --check` passes on every touched file.

### What I could not verify

I do not have access to a macOS or Nix machine, so I could not live-reproduce
the original failure or confirm the fix end to end on a real nix-darwin install.
Verification is at the logic level via the unit test above. The remaining
resolver files import `electron`, which is not installed in my environment, so I
did not run the full `npm test`/lint/build suite; the new resolver and its test
have no electron dependency and run clean. A maintainer on macOS/Nix should
confirm a real transcription and URL download once merged.

## AI assistance disclosure

This change was prepared with AI assistance (Claude) and reviewed before
submission.

### DIFF
--- a/main.js
+++ b/main.js
@@ -341,24 +341,19 @@ const AUTH_BRIDGE_HOST = "127.0.0.1";
 const AUTH_BRIDGE_PORT = parseAuthBridgePort();
 const AUTH_BRIDGE_PATH = "/oauth/callback";
 
-// Set up PATH for production builds to find system tools (whisper.cpp, ffmpeg)
+// Set up PATH for production builds to find system tools (whisper.cpp, ffmpeg,
+// and the ffprobe that yt-dlp shells out to). GUI apps on macOS/Linux are
+// launched by launchd/systemd with a minimal PATH that never sources the user's
+// shell profile, so tools installed by Homebrew or Nix are invisible. Rebuild
+// PATH from the real login-shell PATH plus known extra locations so child
+// processes inherit somewhere they can actually find these binaries.
 function setupProductionPath() {
-  if (process.platform === "darwin" && process.env.NODE_ENV !== "development") {
-    const commonPaths = [
-      "/usr/local/bin",
-      "/opt/homebrew/bin",
-      "/usr/bin",
-      "/bin",
-      "/usr/sbin",
-      "/sbin",
-    ];
+  if (process.platform === "win32" || process.env.NODE_ENV === "development") return;
 
-    const currentPath = process.env.PATH || "";
-    const pathsToAdd = commonPaths.filter((p) => !currentPath.includes(p));
-
-    if (pathsToAdd.length > 0) {
-      process.env.PATH = `${currentPath}:${pathsToAdd.join(":")}`;
-    }
+  const { buildAugmentedPath } = require("./src/helpers/systemBinaryPath");
+  const augmented = buildAugmentedPath();
+  if (augmented) {
+    process.env.PATH = augmented;
   }
 }
 
--- a/src/helpers/debugLogger.js
+++ b/src/helpers/debugLogger.js
@@ -1,4 +1,5 @@
 const fs = require("fs");
+const os = require("os");
 const path = require("path");
 const { app } = require("electron");
 
@@ -323,6 +324,10 @@ class DebugLogger {
         "/usr/local/bin/ffmpeg",
         "/opt/homebrew/bin/ffmpeg",
         "/usr/bin/ffmpeg",
+        // Nix / nix-darwin install locations, which a launchd PATH never sees.
+        "/run/current-system/sw/bin/ffmpeg",
+        path.join(os.homedir(), ".nix-profile", "bin", "ffmpeg"),
+        "/nix/var/nix/profiles/default/bin/ffmpeg",
       ].filter(Boolean);
     }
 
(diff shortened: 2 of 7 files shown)


stantheman0128:

PR #1216, opened 2026-07-17
fix(logging): keep main-process logs off stdout in packaged builds

### LINKED ISSUE #1058: Electron main-process logs written to parent terminal stdout, pollutes CLI sessions
## Problem
Launching OpenWhispr from a shell leaks INFO-level main-process logs to parent stdout, polluting CLI sessions.

## Env
- Version: 1.7.3
- OS: Windows 11

## Steps
1. Launch `OpenWhispr.exe` from PowerShell/Windows Terminal (e.g. `Start-Process` from a shell)
2. Observe parent terminal

## Expected
GUI app runs detached; no output to parent terminal.

## Actual
INFO-level logs (Pipeline timing, Recording started, etc.) print to parent stdout.

## Root cause
Electron on Windows inherits parent console handles unless the BrowserWindow is launched with explicit detach OR pushed through `explorer.exe`.

## Workaround
`Start-Process explorer.exe "C:/Program Files/OpenWhispr/OpenWhispr.exe"`

## Suggestion
In `main.js`, detect if a TTY is attached and either `freopen` stdout to NUL or spawn as a fully detached process.

Thanks!

### LINKED ISSUE #1075: fix(cli-bridge): startup DEBUG/INFO logs injected into cloud-STT dictation input via :8200 bridge
## Problem

When OpenWhispr v1.7.3 launches on Windows 11 with a cloud STT provider selected (no local model downloaded), the CLI bridge on port 8200 is active. The full startup log — `[DEBUG]`, `[INFO]`, `[WARN]` lines from Electron main process — appears verbatim as dictation text in whatever app has focus (e.g. Claude Code, VS Code).

## Repro

1. Install OpenWhispr v1.7.3 on Windows 11, set provider to any cloud model.
2. Do NOT download a local model (Models: None downloaded).
3. Open a text editor / terminal. Trigger dictation.
4. The following lines appear as transcribed text (sample):

```
[INFO][cli-bridge] CLI bridge started { port: 8200 }
[DEBUG] [HotkeyManager] Setting up hotkey: "Control+Super" for slot "dictation"
[INFO][meeting] Meeting detection engine started { processDetection: true, audioDetection: true }
[INFO] qdrant started successfully { port: 6333 }
```

## Expected

Startup logs are written to the log file only (`AppData\Roaming\open-whispr\logs\debug-*.log`). Nothing reaches the dictation input stream.

## Root cause hypothesis

CLI bridge (`port: 8200`) reads from stdout during startup before filtering by log level. When cloud STT is active and no local server is started, the bridge may forward the raw startup stdout to the active dictation channel.

## Env

- OpenWhispr: 1.7.3
- OS: Windows 11 Enterprise 26200
- Provider: cloud (BYOK), no local model downloaded
- Electron: 41.2.0, Node: v24.14.0

Thanks for the great work on OpenWhispr!

stantheman0128:
Closes #1075
Closes #1058

## What was broken

With a cloud STT provider selected and no local model downloaded, main-process
startup log lines were showing up in the user's active window and reading as
dictation text, for example:

```
[INFO][cli-bridge] CLI bridge started { port: 8200 }
[DEBUG] [HotkeyManager] Setting up hotkey: "Control+Super" for slot "dictation"
[INFO][meeting] Meeting detection engine started { ... }
[INFO] qdrant started successfully { port: 6333 }
```

## Verified root cause

The reporter's hypothesis was that the port-8200 CLI bridge reads stdout and
forwards it to the dictation channel. That mechanism does not exist. At this
HEAD `src/helpers/cliBridge.js` is a loopback REST server whose routes are
notes, folders, and transcriptions CRUD; it never reads stdout and never
touches the paste/type path. The dictation "type into the focused app" path
(`ipcHandlers.js` `paste-text` to `clipboardManager.pasteText`) only ever
receives the renderer's transcription result text, never a log line.

The real vector is the logger. `debugLogger.write()` mirrored every INFO and
higher line to `console.log` / `console.warn` / `console.error` unconditionally,
in addition to the optional log file. A packaged Electron app normally has no
console attached, so this is invisible when the app is double-clicked. But when
the packaged app is launched from, or spawned by, a terminal (a shell, VS Code,
Claude Code), the main process inherits that console's stdio, so those lines
print into whatever window has focus. The exact formatting the reporter saw is
produced by `debugLogger.write()` (`[LEVEL][scope][source] message` plus the
meta object), which confirms the source.

This is the same defect described in #1058 ("main-process INFO logs written to
parent stdout on Windows, pollutes CLI sessions"), viewed from two angles.

## The fix

Gate the logger's console output on packaged state, in `src/helpers/debugLogger.js`:

- Dev builds keep console output (developers expect it).
- Packaged builds write to the log file only, so nothing reaches stdout/stderr.
- An explicit `--log-level` argument or `OPENWHISPR_LOG_LEVEL` / `LOG_LEVEL`
  env var opts a packaged build back in, so the app stays debuggable from a
  terminal on demand.

File logging is unchanged. The gate lives entirely in the logger, so it also
covers the identical stdout pollution in #1058.

## Sibling issue #1058

Recommendation: this PR resolves #1058 as well. Both issues are the same root
cause, the unconditional console output in `debugLogger`. Every routine
startup line the two reports mention flows through `debugLogger.info` /
`debugLogger.debug`, and the gate suppresses all of it in packaged builds.

A handful of direct `console.error` / `console.log` calls remain in the main
process (uncaught-exception and unhandled-rejection handlers in `main.js`,
failure branches in `dragManager.js` and `modelManagerBridge.js`). These are
error-path or user-interaction events, not routine startup logging, and most
target stderr. They do not reproduce either reported symptom, so I left them in
place to preserve crash diagnostics. Fully detaching stdio (the freopen / spawn
detached approach floated in #1058) is not needed to fix either issue and would
sacrifice those diagnostics, so I do not recommend it.

## Verification

Windows 11. Local Node is 22 while the repo pins 24, and dependencies were
installed with `--ignore-scripts`, so the Electron binary, native
`better-sqlite3`, and the TypeScript test loader are absent in this sandbox.
That does not affect the changed file, which is plain JS.

New focused unit test (`test/helpers/debugLoggerConsoleGate.test.js`) asserts
the gate directly: no console output in a packaged build while the file stream
still receives the line, console output preserved in dev, and the explicit
opt-in re-enabling console in a packaged build.

```
$ node --test test/helpers/debugLoggerConsoleGate.test.js
# tests 4
# pass 4
# fail 0
```

```
$ npm run lint
> eslint . && cd src && eslint .
(exit 0)

$ npm run typecheck
> cd src && tsc --noEmit
(exit 0)
```

Full helper suite for context:

```
$ npm test   (node --test "test/helpers/*.test.js")
# tests 396
# pass 307
# fail 74
# skipped 15
```

The 74 failures are all environmental in this sandbox, not from this change:
`Unknown file extension ".ts"` (tests importing `.ts` sources with no TS
loader), `Electron failed to install correctly` (installed with
`--ignore-scripts`), and `better-sqlite3` / `no such module: fts5` (native
module not built). None touch `debugLogger` or console behavior, and the four
new tests pass inside this same run. The gate function is wrapped in try/catch
and always returns a boolean, so it cannot make a previously passing test throw.

## AI-assisted disclosure

This change was investigated and implemented with the help of Claude (Claude
Code). A human reviewed the diff and is submitting the PR.

### DIFF
--- a/src/helpers/debugLogger.js
+++ b/src/helpers/debugLogger.js
@@ -31,6 +31,12 @@ const readArgLogLevel = () => {
   return null;
 };
 
+// True when the user explicitly asked for a log level (CLI arg or env). We treat
+// that as opt-in to console output even in packaged builds, so the app stays
+// debuggable from a terminal on demand.
+const hasExplicitLogLevel = () =>
+  readArgLogLevel() != null || Boolean(process.env.OPENWHISPR_LOG_LEVEL || process.env.LOG_LEVEL);
+
 class DebugLogger {
   constructor() {
     this.logLevel = this.resolveLogLevel();
@@ -40,6 +46,7 @@ class DebugLogger {
     this.logStream = null;
     this.fileLoggingEnabled = false;
     this.fileLoggingPending = this.debugMode; // Track if we need to initialize file logging later
+    this.consoleLoggingEnabled = this.resolveConsoleLogging();
 
     // IMPORTANT: Do NOT call initializeFileLogging() here!
     // It uses app.getPath() which is unsafe before app.whenReady().
@@ -110,8 +117,29 @@ class DebugLogger {
     return "info";
   }
 
+  // Decide whether logs may go to the console (stdout/stderr). Packaged builds
+  // must not: when the app is launched from a terminal, or spawned by one (a
+  // shell, VS Code, Claude Code), the main process inherits that console's stdio
+  // and every log line pollutes it. Users read that noise as dictation text
+  // typed into the focused window (#1075) or as CLI-session spam (#1058). File
+  // logging is unaffected. Dev builds keep console output, and an explicit
+  // --log-level or OPENWHISPR_LOG_LEVEL opts a packaged build back in for
+  // on-demand debugging.
+  resolveConsoleLogging() {
+    if (hasExplicitLogLevel()) return true;
+    try {
+      if (app && typeof app.isPackaged === "boolean") {
+        return !app.isPackaged;
+      }
+    } catch {}
+    // app state unavailable (unit tests, very early boot): mirror the dev signals
+    // main.js uses so development console output is never accidentally silenced.
+    return process.env.NODE_ENV === "development" || Boolean(process.defaultApp);
+  }
+
   refreshLogLevel() {
     const nextLevel = this.resolveLogLevel();
+    this.consoleLoggingEnabled = this.resolveConsoleLogging();
     if (nextLevel === this.logLevel) return;
 
     this.logLevel = nextLevel;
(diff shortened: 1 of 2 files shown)


stantheman0128:

PR #1220, opened 2026-07-17
docs: add custom ASR shim proxy example

### LINKED ISSUE #972: [Docs] Add example: shim proxy for non-OpenAI-compatible ASR backends
The self-hosted transcription mode is a great escape hatch, but it quietly assumes the backend speaks the OpenAI /audio/transcriptions protocol (multipart/form-data → {"text": "..."}). Not every ASR API does.

A real case that triggered this

I ran into this while trying to connect a vendor ASR API (StepFun StepAudio 2.5) that uses a different wire format: JSON body with base64-encoded PCM audio, SSE streaming response. The pattern applies equally to any ASR that deviates from the OpenAI spec — several Chinese cloud providers (Alibaba, Baidu, Tencent) and enterprise speech platforms use similar non-standard formats.

Listen on localhost:8765 at POST /audio/transcriptions
Receive OpenWhispr's multipart request, convert audio to PCM via ffmpeg
Forward to the vendor API, collect SSE events, return {"text": "..."}
Point OpenWhispr's self-hosted URL at http://localhost:8765
Full working shim: https://gist.github.com/ErogosZhou/4eb2c4bab1059b404fb652df7bfe24ac

Why document this pattern?

The shim approach generalizes to any ASR vendor with a non-standard API
Users hitting this wall currently have no guidance; the 501/400 errors aren't obvious
No changes to OpenWhispr itself needed — just a docs page or a examples/ directory entry
Happy to open a PR if this direction makes sense.

stantheman0128:
## Summary

Self-hosted transcription assumes the OpenAI `/audio/transcriptions` wire format (multipart in, `{"text": "..."}` out). Vendors that use a different shape (StepFun StepAudio 2.5, and several Chinese cloud / enterprise ASR APIs) fail with opaque 400/501 errors and no guidance in-repo.

This adds `examples/custom-asr-shim/`: a localhost proxy that speaks OpenAI on the OpenWhispr side and your vendor on the other. `shim_template.py` is the fill-in-one-function skeleton; `stepaudio_shim.py` is a working StepAudio 2.5 port of @ErogosZhou's gist. README covers the request/response contract, HTTP-vs-HTTPS private-host rules, and Self-Hosted UI setup. Root README links the example for discoverability.

Closes #972

Prior closed attempt: #1010 (same idea; UI path was still "Cloud -> Custom". This PR retargets Settings -> Transcription -> Self-Hosted / Server URL, which matches current main.)

## AI assistance

Prepared with AI assistance (Cursor/Grok). Reviewed against OpenWhispr's Self-Hosted transcription path and `isSecureEndpoint` private-host rules before opening.

## Changes

- `examples/README.md`: index for the examples directory
- `examples/custom-asr-shim/README.md`: wire protocol, HTTPS rules, quick start
- `examples/custom-asr-shim/shim_template.py`: vendor-neutral shim (stdlib multipart parser, ffmpeg, 25 MB guard)
- `examples/custom-asr-shim/stepaudio_shim.py`: StepAudio 2.5 (PCM base64 + SSE)
- `examples/custom-asr-shim/test_shim.py`: unittest coverage, no network/ffmpeg
- `README.md`: link under Documentation

## Evidence

```
python test_shim.py
Ran 14 tests in 4.404s
OK
```

Python 3.13.3, Windows 11. Stdlib only (no pip). ffmpeg not required for the unit tests (transcode and vendor call stubbed).

## Test plan

- [x] `python examples/custom-asr-shim/test_shim.py` (14/14)
- [ ] Optional live: set `STEP_API_KEY`, run `stepaudio_shim.py`, point Self-Hosted Server URL at `http://localhost:8765`, dictate once

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,10 @@ Visit **[docs.openwhispr.com](https://docs.openwhispr.com)** for:
 - [MCP server setup](https://docs.openwhispr.com/integrations/mcp)
 - [Troubleshooting](https://docs.openwhispr.com/troubleshooting)
 
+Repo examples:
+
+- [Custom ASR shim](examples/custom-asr-shim/) for Self-Hosted transcription against non-OpenAI-compatible ASR APIs
+
 ## Tech stack
 
 React 19, TypeScript, Tailwind CSS v4, Electron 41, better-sqlite3, whisper.cpp, sherpa-onnx, shadcn/ui
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,5 @@
+# Examples
+
+Community examples and recipes for extending OpenWhispr without changing its source.
+
+- [custom-asr-shim/](custom-asr-shim/) — a small local proxy that lets OpenWhispr's Self-Hosted transcription talk to ASR backends that do not speak the OpenAI `/audio/transcriptions` protocol (StepFun StepAudio, and any other vendor you adapt it to).
--- a/examples/custom-asr-shim/.gitignore
+++ b/examples/custom-asr-shim/.gitignore
@@ -1,0 +1,2 @@
+__pycache__/
+*.pyc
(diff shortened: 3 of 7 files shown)


xAlcahest (approved):
Thanks for picking this up and finishing the retarget. This is the code from #1010 brought up to date with where main actually is, and the update is right: I re-checked the wire contract against current main. Self-Hosted resolves `{Server URL}/audio/transcriptions` (a URL that already ends in `/audio/transcriptions` is kept as-is), no Authorization header goes out on this path, and plain http is accepted for loopback and private hosts. Handling `/v1/audio/transcriptions` too covers people who type a `/v1` base URL. Ran `test_shim.py` locally, 14/14.

One nuance, not blocking and fine to leave as-is: if the Model field in the Self-Hosted panel is left empty, the app falls back to its cloud default model name and with that default it also appends `stream=true` to the form. Nothing breaks (the shim ignores the extra field and OpenWhispr picks the response parser by content-type), so at most a `model=gpt-4o-mini-transcribe` shows up in your shim logs.

LGTM.

PR #1224, opened 2026-07-17
fix(windows): media-key fallback when GSMTC sees no sessions (#993)

### LINKED ISSUE #993: Pause-on-dictation doesn't pause players that aren't registered with Windows SMTC (e.g. Plexamp)
Running 1.7.2 on Windows 11. I have "pause media on dictation" turned on. It works for some players, but it does nothing for Plexamp. The music just keeps going while I dictate.

The player itself is fine. My keyboard's play/pause key pauses Plexamp no problem. So the key works, OpenWhispr just isn't reaching it.

I poked at src/helpers/mediaPlayer.js to see why.

_pauseWindows goes through GSMTC. It lists the media sessions and only pauses the ones reporting PlaybackStatus -eq 4. Plexamp never shows up in that list. I ran the same GlobalSystemMediaTransportControlsSessionManager query by hand with Plexamp actively playing and got back zero sessions. Plexamp grabs the media key through its own global shortcut. It doesn't publish an SMTC session at all, so GSMTC can't see it.

There's already a fallback that would handle this. _sendWindowsMediaKey sends 0xB3, which is exactly the key that pauses Plexamp. The catch is it only runs when the GSMTC script throws or returns GSMTC_FAIL. When GSMTC runs clean and just finds nothing playing, _pauseWindows returns false and the fallback never fires. So any player Windows can't see gets silently skipped.

What I'd suggest: when the session scan comes back empty, fall back to the media key. Gate it behind an opt-in setting, off by default, something like "pause players Windows can't see." It has to be opt-in because the media key is a blind toggle. If nothing is actually playing it could start something instead. Resume already covers this path through _didPause, so it's mostly a matter of routing the empty-session case into _pauseWindowsFallback() and adding the toggle.

Glad to put up a PR if that sounds reasonable.

stantheman0128:
## Summary

Fixes #993: with Pause media on, Windows pause-on-dictation only paused apps that publish a GSMTC/SMTC session. Players like Plexamp never appear in that list, so dictation left them playing even though the hardware play/pause key works.

Root cause: `_pauseWindows` already has a media-key fallback for GSMTC failures, but empty session lists were treated as a no-op. This PR adds an opt-in `pauseMediaKeyFallback` setting (default off) so empty GSMTC can use the same media-key path without forcing blind toggles on everyone.

## Changes

- Extract `resolveWindowsPausePath` for the GSMTC vs media-key decision
- Opt-in Settings toggle on win32 when pause-on-dictation is enabled
- Pass the option through preload / IPC / `useAudioRecording`
- Keep prior behavior when fallback is off (empty GSMTC stays a no-op)
- Unit tests for the path matrix

## AI assistance

This PR was prepared with AI assistance (Cursor / Grok). I reviewed the diff, ran the tests below, and take responsibility for the change.

## Evidence

```text
$ node --test test/helpers/mediaPlayerPausePath.test.js
✔ GSMTC success with paused apps uses gsmtc path
✔ GSMTC_FAIL always falls back to media key
✔ PowerShell failure always falls back to media key
✔ empty GSMTC sessions are a noop without mediaKeyFallback (#993)
✔ empty GSMTC sessions fall back when mediaKeyFallback is on (#993)
✔ whitespace-only GSMTC output is treated as empty sessions
ℹ tests 6
ℹ pass 6
ℹ fail 0
```

## Test plan

- [ ] Windows: Pause media on, fallback off, empty GSMTC app (or mocked empty) -> media stays playing
- [ ] Windows: enable media-key fallback, start dictation with Plexamp (or similar non-SMTC player) -> playback pauses
- [ ] Windows: Spotify/SMTC app still pauses via GSMTC path
- [ ] Non-Windows: new setting not shown


stantheman0128:

PR #1226, opened 2026-07-17
fix(i18n): Simplified vs Traditional Chinese script for zh STT

### LINKED ISSUE #975: Select default between simplified and traditional chinese
If speech in English and chinese happens frequently, then having to choose one language in settings and switching constantly is tedious so I leave it on auto detect. For whatever reason, (when chinese is being spoken) it always transcribes into (the less used/popular) traditional chinese with seemingly no way to set it to simplified chinese (without the tedious switching in settings everytime I switch between English and Chinese speaking). It is impossible to tell (without explicitly saying) whether a user would want it to transcribe into simplified or traditional chinese so a setting to choose between simplified or traditional would be useful. Further, putting traditional and simplified chinese in the "language you speak" setting doesn't really make sense in the first place since neither of them are spoken languages. Reading a sentence sounds the same regardless of whether it was written in simplified or traditional chinese. Also, even if I set the setting to simplified chinese, it still transcribes it in traditional chinese. (I am using openAI whisper since when I use the Nvidia models, it doesn't seem to recognise chinese at all)

stantheman0128:
## Summary
- Whisper STT only accepts language `zh`, so choosing Chinese (Simplified) / Chinese (Traditional) or leaving Auto-detect could not pin the written script. Transcripts often came back in Traditional even when Simplified was selected.
- After STT (batch, cloud, and streaming), convert CJK text with OpenCC `cn`/`twp` when the target script is known: `zh-CN` always Simplified, `zh-TW` always Traditional, and Auto uses a new Settings preference.
- Bias Whisper prompts with a short Simplified/Traditional sample, and map Auto + script preference to `zh-CN`/`zh-TW` cleanup language instructions.

Closes #975

## AI assistance
This PR was prepared with AI assistance (Cursor/Grok). Human author: Stan Shih (@stantheman0128). Please review carefully.

## Verification / Evidence
`node --test test/helpers/chineseScript.test.js`

`
tests 9
pass 9
fail 0
`

`npm run typecheck` (pass)
`npm run lint` (pass)
`node scripts/check-i18n.js` (Locale keys and placeholders are consistent.)

Manual note (zh-TW niche): conversion uses OpenCC Taiwan phrase tables (e.g. 軟體 <-> 软件). Spoken Chinese is script-agnostic; this setting only controls the written form after recognition.


Chadpiha (approved):
Pushed some changes to this branch — summary here so nothing is a surprise.

### Fixed

**Failed Chinese-target translations corrupted the fallback text.**
`translationRequested` stayed true when translation was unreachable, skipped, or
threw, but the script target still resolved to the translation target — so the
untranslated source ran through OpenCC. A failed ja → zh-CN run turned
会議の資料を確認してください into 会议の数据を确认してください.
`executeTranslationChain` now returns `translated`, and conversion uses the
target language only when the translate step actually produced text. Verified
across all four failure modes; three regression tests added.

**The Whisper prompt bias broke dictionary-echo detection.**
`matchesDictionaryPrompt` requires 70% of the prompt's words to appear in the
text. Chinese has no spaces, so the bias normalizes to a single extra token —
enough to drop an exact echo below the threshold on short dictionaries, meaning
prompt echoes get pasted as real transcripts instead of raising "No audio
detected":

| dictionary | before this PR | with bias | after fix |
|---|---|---|---|
| `OpenWhispr` | detected | **missed** | detected |
| `OpenWhispr, Parakeet` | detected | **missed** | detected |
| `OpenWhispr, Parakeet, Whisper` | detected | detected | detected |

The check now tests the dictionary-only prompt alongside the full sent prompt.

**History retry now honors the preference.** Retry's cleanup prompt was already
inheriting the language instruction (long-standing behavior, not introduced
here) with nothing deterministically enforcing it, so a retry could still come
back Traditional with Simplified selected. Conversion now runs before
`updateTranscriptionText` persists.

### Scope

Conversion is centralized in `audioManager` plus retry — two `applyChineseScript`
call sites. Meeting transcription (`meetingRecordingStore.ts`) and uploaded audio
(`UploadAudioView.tsx`) strip zh-CN/zh-TW to `zh` and store the response
unconverted, so they honor no part of the setting. Rather than widen this PR I
narrowed the setting's copy to say "dictation" across all 10 locales, and noted
the widening path in the `chineseScript.js` header. Happy to file a follow-up
issue for those two surfaces.

### Cleanups

- `ChineseScriptPreference` extracted to `src/types/electron.ts` (the union was
  written out in four places; matches the existing `LocalTranscriptionProvider`
  convention)
- streaming path guards on `usedBatchFallback` instead of relying on OpenCC
  round-trip idempotency
- dropped an unreachable default parameter on `getCleanupLanguage`

Issue #668, opened 2026-04-25
Auto-paste not working specifically in Claude (desktop and browser) on macOS


adberf:
Auto-paste works correctly in every application I have tested except Claude. This affects both the Claude desktop app and claude.ai in Brave browser on macOS.
Environment:

OpenWhispr version: 1.6.10
OS: macOS Tahoe 26.2, MacBook Air M3 2024
Affected: Claude desktop app and claude.ai in Brave browser
Works correctly in: every other app tested

Expected behavior:
Transcribed text automatically pastes into Claude's input field after dictation, the same way it does in all other applications.

Current behavior:
Transcription completes successfully. Text is available in clipboard. Auto-paste does not trigger. Manual Cmd+V works as a workaround.

Root cause hypothesis:
Claude's input field uses a non-standard rich text editor component that behaves differently from regular text inputs. OpenWhispr's paste mechanism does not handle this edge case. Wispr Flow auto-pastes into Claude correctly on the exact same machine with the exact same setup, confirming this is not a Claude restriction — it is specific to OpenWhispr's paste implementation and is solvable.

Request:
Please investigate how Wispr Flow handles pasting into Claude's input field and apply the same approach in OpenWhispr.

adberf:
Just noticed: When dictation is triggered while Claude is the active app, a blue recording indicator appears in the top right corner of the screen. This appears to steal focus from Claude's input field, causing auto-paste to fail because the target input is no longer active when transcription completes.

P.S. Tried different shortcuts & it did not solve the issue.

gabrielste1n:
@adberf will investigate thanks for reporting

PR #1227, opened 2026-07-17
fix(windows): restore pre-recording input focus before dictation paste (#859)

### LINKED ISSUE #859: Prevent Input focus from being repositioned.

stantheman0128:
# fix(windows): restore the focused window before paste (#859)

## Why

Fixes #859. On Windows, starting a dictation from a focused input field and then
letting it transcribe often drops the result in the wrong place. The moment
processing begins, focus leaves the original field and the transcript lands
somewhere else (the reporter saw it hit Brave's settings button), so they have
to re-click the field or paste by hand. SuperWhisper keeps focus, so this is a
clear gap.

## Root cause

The Windows paste helper sends its keystroke to whatever window is foreground at
paste time. `resources/windows-fast-paste.c` calls `GetForegroundWindow()` right
before `SendInput`, with no memory of the field the user recorded from. Nothing
captures the pre-recording target and nothing restores it.

macOS and Linux already handle this. On macOS `TextEditMonitor.captureTargetPid()`
grabs the frontmost app's PID at record start and `activateTargetPid()`
re-activates it before the paste (added in #668, for exactly this Brave and
Claude Desktop focus loss). On Linux `pasteLinux()` re-detects the target window
and passes it to the paste tool with `--window`. Windows had no equivalent, so
if anything takes the foreground during the record to transcribe to paste window
(a subprocess console flash, or Chromium moving focus into its own toolbar), the
paste follows the wrong window.

The dictation overlay itself is `focusable: false` and shown with
`showInactive`, so it is not the thief. The problem is the missing capture and
restore, not the overlay stealing focus.

## The fix

Mirror the macOS capture and restore design on Windows, reusing the native
helper that already ships in the paste path (no new dependency):

- `resources/windows-fast-paste.c` gains two modes:
  - `--print-foreground` reports the current foreground window handle, so the
    caller can stash it at record start.
  - `--restore-window <hwnd>` re-activates that window before pasting, using the
    standard `AttachThreadInput` foreground-lock workaround (the same technique
    nircmd's `win activate` uses), then sends the paste keystroke to it.
- `src/helpers/textEditMonitor.js`: `captureTargetPid()` now also captures the
  foreground HWND on Windows, at the same record-start seams that already
  capture the macOS PID (they run before the overlay shows). It invokes the
  binary with both `--detect-only` and `--print-foreground` so that an old
  binary that predates these flags takes its existing non-pasting detect
  early-return instead of injecting a paste (see the old-binary safety note
  below).
- `src/helpers/ipcHandlers.js`: the `paste-text` handler passes the captured
  handle through to the clipboard manager on Windows.
- `src/helpers/clipboard.js`: `pasteWithFastPaste` forwards it to the binary as
  `--restore-window <hwnd>`.

Whole feature is gated on the fast-paste binary. If it is missing, capture
returns nothing and paste behaves exactly as before.

### Old-binary safety

`windows-fast-paste.exe` is downloaded as a prebuilt binary from a separate
GitHub release and cached, so new JS commonly runs against an OLD exe built
before this change. Both new invocations are safe on that old binary, and
neither triggers a premature paste:

- Restore (paste time): the old `main()` has no `--restore-window` handler, so
  it ignores the flag and falls through to its normal paste. That is exactly the
  pre-#859 behavior (paste into the current foreground), which is correct for
  the paste stage.
- Capture (record start): the old `main()` recognizes only `--detect-only` in
  its arg loop. Passing `--detect-only --print-foreground` sets `detectOnly`
  true; the old binary then reaches `if (detectOnly) { print class info;
  return 0; }` and returns BEFORE `SendPasteNormal`/`SendPasteTerminal`, so no
  Ctrl+V is injected. Its output has no `FOREGROUND <hwnd>` line, so capture
  yields null, no restore happens, and behavior falls back to exactly pre-#859.

  Passing `--print-foreground` alone would be unsafe on the old binary: the flag
  is unknown, `detectOnly` stays false, and control falls through to the paste,
  injecting the clipboard at record start. The added `--detect-only` is what
  prevents that. A current binary checks `--print-foreground` before
  `--detect-only`, so it still reports the handle as `FOREGROUND <hwnd>`.

## Verification

This was verified at the code level. Interactive GUI focus behavior (recording
from a real input field and watching where the transcript lands) was not
exercised. A headless agent cannot drive that flow, so a maintainer should
confirm the end to end behavior on a real Windows desktop.

Control flow, before and after:

- Before: record starts, overlay shows inactive, transcription runs, paste
  spawns `windows-fast-paste.exe` which pastes into `GetForegroundWindow()` at
  that instant. If focus drifted during transcription, the transcript lands in
  the wrong window.
- After: record starts and the foreground HWND is captured first. At paste time
  the binary re-activates that captured window, waits briefly for the activation
  to land, then pastes into it.

What was run:

- Targeted unit tests for the clipboard module pass, including cases asserting
  the captured handle is threaded to the binary as `--restore-window <hwnd>`,
  that no window argument is sent when nothing was captured, and that the
  capture path invokes the binary with the old-binary-safe
  `--detect-only --print-foreground` form (yielding null when no `FOREGROUND`
  line is printed):
  `node --test test/helpers/clipboardRestore.test.js` -> 10 passed, 0 failed.
- `node --check` on the three changed JS files: clean.

Not run in this environment, please run in CI or locally:

- `npm run lint` / `npm run format` and `npm run build`: this worktree does not
  have `node_modules` installed (no `electron`), so the Electron-dependent
  suites and the linter could not run here. The clipboard test runs because it
  mocks `electron`.
- `test/helpers/textEditMonitorPrecedingChar.test.js` requires `electron`
  (via `debugLogger`) and cannot run without an install; the change to
  `textEditMonitor.js` only adds a Windows branch and leaves the existing
  macOS-only `activateTargetPid` path untouched.

## Note on the prebuilt binary

`windows-fast-paste.exe` is downloaded as a prebuilt binary during
`prebuild:win` (with a local compile fallback). The new `--print-foreground`
and `--restore-window` modes take effect once the binary is rebuilt from this
updated source (the `build-windows-fast-paste.yml` workflow does this on
release). Until then, as detailed in the old-binary safety section above, both
new invocations are safe against the still-cached old exe: neither injects a
premature paste, and the app falls back to exactly pre-#859 behavior.

## Scope check

The gap was confirmed against the actual code at this branch's base
(HEAD `fa6dd8cf`): `windows-fast-paste.c` pastes into `GetForegroundWindow()`
with no restore, and `pasteWithFastPaste` spawned the binary with no arguments.
The capture and restore added here are additive and gated on the captured handle
being present.

---

This change was written with AI assistance (Claude) and reviewed before
submission.

### DIFF
--- a/resources/windows-fast-paste.c
+++ b/resources/windows-fast-paste.c
@@ -10,6 +10,14 @@
  *   1. Window class name (fast, works for native terminals)
  *   2. Executable name (fallback, catches Electron-based terminals like Termius)
  *
+ * Focus restore (issue #859): the caller captures the target window handle at
+ * record start with --print-foreground, then passes it back as
+ * --restore-window <hwnd> at paste time. We re-activate that window before
+ * sending the keystroke so dictation lands in the field the user was in, even
+ * when something (a subprocess console flash, Chromium moving focus to its
+ * toolbar) stole the foreground during transcription. This mirrors the macOS
+ * PID capture/activate path and the Linux --window path.
+ *
  * Compile with: cl /O2 windows-fast-paste.c /Fe:windows-fast-paste.exe user32.lib
  * Or with MinGW: gcc -O2 windows-fast-paste.c -o windows-fast-paste.exe -luser32
  */
@@ -17,6 +25,8 @@
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 #include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
 #include <string.h>
 
 static const char* TERMINAL_CLASSES[] = {
@@ -144,6 +154,45 @@ static int SendPasteNormal(void) {
     return (sent == 4) ? 0 : 1;
 }
 
+/* Bring a captured target window back to the foreground before pasting.
+   SetForegroundWindow alone is blocked by Windows' foreground lock when the
+   calling process isn't the current foreground owner, so we attach this
+   thread's input queue to both the current foreground thread and the target
+   thread first — the standard workaround, and what nircmd's "win activate"
+   does internally. Returns TRUE if the target ended up foreground. */
+static BOOL RestoreForegroundWindow(HWND target) {
+    if (!target || !IsWindow(target)) return FALSE;
+
+    if (GetForegroundWindow() == target) return TRUE;
+
+    if (IsIconic(target)) {
+        ShowWindow(target, SW_RESTORE);
+    }
+
+    HWND fg = GetForegroundWindow();
+    DWORD thisThread = GetCurrentThreadId();
+    DWORD targetThread = GetWindowThreadProcessId(target, NULL);
+    DWORD fgThread = fg ? GetWindowThreadProcessId(fg, NULL) : 0;
+
+    BOOL attachedTarget = FALSE;
+    BOOL attachedFg = FALSE;
+    if (targetThread && targetThread != thisThread) {
+        attachedTarget = AttachThreadInput(thisThread, targetThread, TRUE);
+    }
+    if (fgThread && fgThread != thisThread && fgThread != targetThread) {
+        attachedFg = AttachThreadInput(thisThread, fgThread, TRUE);
+    }
+
+    BringWindowToTop(target);
+    BOOL ok = SetForegroundWindow(target);
+    SetFocus(target);
+
+    if (attachedFg) AttachThreadInput(thisThread, fgThread, FALSE);
+    if (attachedTarget) AttachThreadInput(thisThread, targetThread, FALSE);
+
+    return ok && GetForegroundWindow() == target;
+}
+
 static int SendPasteTerminal(void) {
     INPUT inputs[6];
     ZeroMemory(inputs, sizeof(inputs));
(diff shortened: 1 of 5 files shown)


gabrielste1n:
Thanks for the work on this. We’re closing this PR because the implementation captures only the top-level foreground HWND. In the Brave case reported in #859, focus can move between elements inside the same top-level window, so GetForegroundWindow remains unchanged and RestoreForegroundWindow returns early without restoring the original input.\n\nThe branch also duplicates the Windows target-capture path now handled by SelectionManager on main and no longer applies cleanly to the current helper and capability infrastructure.\n\nWe’ll investigate the current Windows focus path and pursue a replacement from current main that distinguishes cross-window activation changes from same-window control or DOM focus changes. Thanks again for putting this together.

stantheman0128:
Thanks for writing up the close.

Agreed on the limit: a top-level HWND restore cannot follow caret moves inside the same Brave window, which is the #859 case. I would not have stretched this PR into UI Automation.

#1725 already took `RestoreForegroundWindow` and attached it to the HWND main captures at record start. That is the better seam.

PR #1228, opened 2026-07-17
fix(windows): resolve llama-server with where.exe without shell

stantheman0128:
## Summary
- On Windows, `getSystemBinaryPath` used `where` with `shell: true`, which routes through cmd.exe and is fragile for PATH lookup.
- Switch to `where.exe` with `shell: false` and `windowsHide: true`.
- Split stdout on `/\r?\n/` so CRLF from where.exe does not leave a trailing CR on the first path.

## AI assistance
This PR was prepared with AI assistance (Cursor/Grok). Human author: Stan Shih (@stantheman0128). Please review carefully.

## Verification / Evidence
Live Windows 11 smoke (where.exe + shell: false):

```
code 0 first "C:\\Program Files\\nodejs\\node.exe" hasCR true
```

`npm run typecheck` (pass)
`npm run lint` (pass)


gabrielste1n (approved):
Verified the fix — where.exe with shell:false eliminates the cmd.exe console flash during llama-server PATH resolution, and the CRLF-aware split fixes a latent bug where the resolved path kept a trailing \r. Merged latest main and trimmed a redundant comment plus the default shell option. Full suite, lint, and prettier are green. Thanks! 🙏

PR #1240, opened 2026-07-18
fix(streaming): reject premature AssemblyAI closes

### LINKED ISSUE #1239: AssemblyAI streaming startup hangs when the WebSocket closes before Begin
## Description

AssemblyAI streaming startup can remain pending indefinitely when the WebSocket opens but closes before the provider signals readiness with its `Begin` message.

Both the warmup and cold-connect paths clear their startup timeout during the `close` event. However, neither path currently rejects its pending promise. Callers awaiting `warmup()` or `connect()` therefore never receive success or failure.

## Steps to reproduce

1. Create an AssemblyAI streaming instance.
2. Point `buildWebSocketUrl()` to a WebSocket server that accepts the connection.
3. Have the server close with code `1008` before sending a `Begin` message.
4. Await either `warmup({ token })` or `connect({ token })`.

Using a bounded race, both promises remain pending on the affected implementation.

## Expected behavior

Both operations should reject promptly with an error containing the WebSocket close code, allowing recording startup to report the failure or invoke its fallback path.

## Actual behavior

The close handler clears the startup timeout and cleans up the socket without settling the pending promise.

## Impact

A provider-side policy rejection, gateway failure, or another early transport close can leave AssemblyAI recording startup stuck instead of returning an actionable error.

## Environment

- OpenWhispr `main` at the current upstream base
- AssemblyAI streaming provider
- Reproduced with Node.js 24 and a loopback WebSocket server
- The affected transport logic is platform-neutral

hsusul:
## Summary

Reject AssemblyAI warmup and cold-connect promises when their WebSocket closes before the provider signals readiness with `Begin`.

Fixes #1239

## Root cause

The WebSocket `close` handlers cleaned up the socket and cleared the startup timeout without settling the promise awaited by recording startup.

As a result, both `warmup()` and the cold `connect()` path could remain pending indefinitely when the socket opened successfully but closed before readiness.

## Changes

- guard warmup against multiple settlement paths
- reject warmup when the socket closes before readiness
- reject the pending cold-connect promise when the socket closes before readiness
- include the WebSocket close code in the resulting errors
- preserve established-session close handling and automatic rewarming behavior
- add deterministic loopback WebSocket regression tests for:
  - premature warmup close
  - premature cold-connect close
  - successful warmup `Begin`
  - successful cold-connect `Begin`

## Validation

- `ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/assemblyAiStreaming.test.js`
- `ELECTRON_OVERRIDE_DIST_PATH=/tmp npm test`
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `npm run i18n:check`
- `npm run build:renderer`
- `node --check src/helpers/assemblyAiStreaming.js`
- `node --check test/helpers/assemblyAiStreaming.test.js`
- `git diff --check`

All applicable checks passed after rebasing onto the latest upstream `main`.

## Testing notes

The failure and normal readiness paths are covered with a local loopback WebSocket server.

A live credential-backed AssemblyAI session and packaged Electron build were not run because they require provider credentials and the full native/download setup.


gabrielste1n (approved):
Approving — thanks @hsusul, this is a good one. A promise that never settles is about the worst failure mode there is, because nothing surfaces it.

Verified the bug is real by reconstructing the pre-fix file and running your own test bodies against it: both `warmup()` and cold `connect()` stay permanently pending when the socket closes before `Begin` (`actual: 'pending'`). The mechanism is exactly as you describe — the old close handlers cleared the timeout and called `cleanup()`, destroying the last remaining settle path. The blast radius is a little worse than the PR states: `assemblyai-streaming-start` sets `streamingStartInProgress = true` and resets it in a `finally`, so a hung connect wedges that flag for the life of the process and every later streaming start returns "Operation in progress".

Two things I checked closely:

- **The auto-rewarm chain is intact.** `wasReady === true` implies `settled === true`, so the new `!settled` early-return can never swallow a rewarm — confirmed by probe, a ready socket that drops still re-warms to a new session id.
- **No double-settle.** All four settlement sites in the cold path (Begin, error, close, and the pre-existing timeout) are mutually exclusive via the `pendingReject` sentinel.

Nice touch that this brings `assemblyAiStreaming.js` in line with `deepgramStreaming.js` and `cortiStreaming.js`, which already used the `settled` flag and sentinel-nulling — it was the outlier. And your tests are real regression tests: they hang on the old code and assert on the close code rather than merely on rejection.

One optional follow-up, not worth holding this: the cold-path *timeout* handler rejects without nulling `pendingReject`, so it's the one settlement site that leaves the sentinel set. Harmless today — the later close-handler call lands on an already-settled promise and is a no-op — but nulling it there would make all four paths symmetric.

Verified against current main: 724 tests / 0 failures, typecheck, lint and prettier clean.

gabrielste1n:
Follow-up code-quality pass before merge (`a7ec8c0`): replaced `capturePromptOutcome` with `assert.rejects` inside try/finally. It hand-rolled a promise race that duplicates `assert.rejects`, neither call site distinguished its `resolved` from its `pending` state, and the 250ms window was a flake risk — on a loaded runner a socket close slower than that would report `pending` and fail the test spuriously. The try/finally also guarantees `cleanupAll()` runs when an assertion fails, which the old form skipped, and makes all four tests in the file structurally identical. Net −19 lines.

The production diff I left completely alone — it adds no comments, and the `pendingReject` block already mirrors the `error` handler above it exactly. Re-verified: 724 tests / 0 failures, typecheck, lint and prettier clean. Merging. Thanks!

PR #1242, opened 2026-07-19
feat(notch): macOS Dynamic Island style dictation popup

tanujdargan:
Adds an optional Dynamic Island-style popup for dictation on macOS.

When it's enabled and you start a dictation, black wings slide out from the physical notch — an elapsed timer on the left and a mic button on the right. Click the mic to stop, right-click it to open the Control Panel. There's also an expanded mode that shows a live teleprompter-style transcript right below the notch, using streaming partials on the Nemotron models or the 1.5s chunked previews everywhere else.

A few things I put care into:

- **It never gets in the way of dictation.** The whole thing is best-effort, if anything fails (no notch, geometry looks wrong, window creation fails), you just get the classic floating pill like before.
- **Notch detection is per display.** There's no public API for the notch, so it's detected from the menu bar inset and the width is estimated from the display's logical resolution. External monitors and non-notch Macs fall back to the pill automatically.
- **It plays nice with the rest of the system.** The window sits at screen-saver level so it draws above third-party notch apps, and it's content-protected so it never shows up in screen shares or recordings.

Both settings (the popup toggle and the expanded transcript, with a hint when your model doesn't stream) only show up on macOS, strings are translated in all 10 locales, and the display math has `node --test` coverage. Design-wise it's OLED black with sub-300ms transform/opacity animations and reduced-motion fallbacks.

Off by default, nothing changes unless you turn it on in Settings → General.


tanujdargan:
Closing to shed the force-push noise from iterating on this branch — superseded by a fresh PR with a single clean commit.

PR #1251, opened 2026-07-19
feat(calendar): Microsoft Calendar (Graph) integration

gabrielste1n:
Adds Outlook / Microsoft 365 as the third calendar provider, on **all platforms**, via Microsoft Graph v1.0 (delegated `Calendars.Read`, common tenant, public-client PKCE loopback OAuth — no client secret).

> **Stacked on #1237** — based on `feat/apple-calendar-eventkit`; only the Microsoft commit is in this diff. Retargets to `main` once #1237 merges.

## What's here

**Shared REST-provider machinery (pure refactor, lands first)**
- `oauthLoopbackFlow.js` — PKCE loopback flow extracted from the Google OAuth helper (ephemeral 127.0.0.1 server, state validation, desktop-callback redirects, 120s timeout)
- `calendarSyncInterval.js` — sync interval runner with exponential backoff (2min → ×2 → 30min cap) and 30s focus-sync throttle
- Google OAuth + manager migrated onto both with no behavior change

**Microsoft provider**
- `microsoftCalendarOAuth.js` — common-tenant auth with `prompt=select_account`, rotated refresh-token persistence (Microsoft rotates on every refresh), email resolution via id_token claims with `GET /me` fallback
- `microsoftCalendarManager.js` — `calendarView/delta` incremental sync with `@odata.nextLink` paging and 410 full-resync fallback. Graph deltaLinks permanently encode their original date window, so the manager syncs a 14-day window and discards the token after 7 days (`sync_token_expires_at`), keeping coverage above the app's 7-day lookahead
- Timestamps normalized to `...Z` at map time so SQLite `datetime()` ordering/dedupe hold; unaccepted invitations (`showAs=tentative`) stay `confirmed` for parity with Google

**Data layer**
- `microsoft_calendar_tokens` / `microsoft_calendars` tables + 1:1 mirrors of the Google DB methods
- Dedupe generalized: Apple yields to any REST provider (`MAX(provider != 'apple')`); REST rows are never collapsed
- `removeEventsFromDeselectedCalendars(provider)` parameterized via a whitelisted provider→table map

**Renderer**
- `mcal-*` IPC channels, preload bridges, typed API surface
- Integrations UI: header block + account sub-rows extracted into local `ProviderRow` / `CalendarAccountRows` components (shared by Google, Microsoft, Apple); Microsoft row shows on all platforms
- `mcalAccounts` / `mcalPrimaryOnly` store keys, third provider in `useUpcomingEvents`, Outlook-style calendar icon, strings in all 10 locales

**Fix**
- `gcal-disconnect` dropped the account email end-to-end, so unlinking one Google account cleared all of them; the email now flows through preload → IPC → `disconnect(email)`

## Verification
- `npm run lint`, `cd src && tsc --noEmit`, `node --test` (546 pass), `npm run i18n:check`, prettier — all green
- Dedupe SQL smoke-tested against better-sqlite3 with mixed Google/Microsoft/Apple fixtures in both timestamp formats

## Prerequisites before release
- Azure/Entra app registration; `MICROSOFT_CALENDAR_CLIENT_ID` in env (plumbing mirrors `GOOGLE_CALENDAR_CLIENT_ID`)
- Confirm `openwhispr.com/auth/desktop-callback` forwards `mcal_*` query params

## Known limitations (v1)
- Google↔Microsoft duplicates (same meeting on both addresses) are not deduped — only Apple yields
- No per-calendar selection UI (parity with the other providers — primary-only toggle)
- Disconnect deletes local tokens only; Microsoft has no public revocation endpoint for this flow


Chadpiha (commented):
Two additions on top of the original Microsoft Calendar work: a sync-up with the base branch, and a round of review fixes.

### d204b163 — Merge updated `feat/apple-calendar-eventkit`

The base branch picked up the EventKit hardening work (shared helper extraction, dev TCC fix) after this branch was cut, so I merged it in and adapted the Microsoft manager to the patterns it introduced:

- Uses the shared `windowBroadcast` helper instead of a private `broadcastToWindows` copy
- Calls the reminder scheduler with a provider-scoped key (`reset("microsoft")`) so disconnecting Microsoft can't re-fire reminders already delivered for Google/Apple events
- Constructor slimmed to `(databaseManager, reminderScheduler)`, matching the Apple manager — the `windowManager` dependency was unused
- Dropped the unreferenced `MicrosoftCalendar`/`AppleCalendar` interfaces from `types/calendar.ts`

### 185562ac — Review fixes

**Correctness**
- `cleanup-app` now stops the Apple calendar helper alongside the Google/Microsoft sync runners before the DB is closed and deleted — previously the helper kept pushing snapshots into a closed database
- Both OAuth `_httpsPost` helpers (Google + Microsoft) get the 10s `AbortSignal.timeout` used elsewhere; a stalled token refresh could previously hang a background sync indefinitely
- Full (non-incremental) re-syncs now prune stale events via `removeStaleCalendarEvents`: when the sync/delta token is invalid (410, TTL expiry, app offline for a week), deletions never arrive as removed items, so cancelled meetings lingered and could still fire reminders. Rows referenced by meeting notes are retained, mirroring the Apple snapshot behavior. Applied to both Google and Microsoft
- The AI agent's calendar tool now registers when **any** calendar provider is connected, not just Google — it reads the shared provider-deduped `calendar_events` table, so Microsoft-only and Apple-only users were wrongly excluded. This also wires up the previously-unread `mcalConnected` store field, and the tool-registry cache key now tracks the combined flag

**Cleanup**
- `FOCUS_SYNC_THROTTLE_MS` deduplicated — exported from `calendarSyncInterval.js`, imported by the Apple manager
- CLAUDE.md updated to the multi-provider reality: documents the Microsoft/Apple managers, `CalendarReminderScheduler`, `CalendarSyncInterval`, and `oauthLoopbackFlow`, and fixes the meeting-detection section that still named `GoogleCalendarManager` as the only calendar source

**Tests** (9 new)
- `calendarSyncInterval.test.js`: backoff math, failure rescheduling (mock timers), focus-sync throttle
- `calendarDatabase.test.js`: stale-event pruning — keeps fresh/note-linked/other-provider rows, drops the rest
- `microsoftCalendarManager.test.js`: Graph timestamp normalization and the full `_mapEvent` mapping (response-status translation, self-attendee detection, join-URL fallback from body text)

Full suite: 575 tests, 0 failures. Lint clean.

Chadpiha:
**Review fixes — 65a2a476**

All six review items addressed.

**1. Refresh-token rotation race** (`microsoftCalendarOAuth.js`)
Microsoft rotates the refresh token on every refresh, so a focus sync racing the
interval sync could persist competing rotations and invalidate the stored token.
`getValidAccessToken` now shares a single in-flight refresh per account via a
`_refreshInFlight` map; refresh + persist moved into `_refreshAndSaveTokens`.

**2. Provider-scoped event deletes** (`database.js`)
`removeMicrosoftAccount` deleted `calendar_events` rows by `calendar_id` alone.
Now scoped to `provider = 'microsoft'`. Applied the same scoping to
`removeGoogleAccount`, which had the identical unscoped delete — it matters now
that the events table is shared across providers.

**3. Silent error swallow** (`ipcHandlers.js`)
`mcal-get-connection-status` returned its fallback without logging, unlike the
`acal`/`gcal` handlers. It now logs first.

**4. Dead default parameter** (`database.js`)
Dropped the unused `accountEmail = null` default on `saveMicrosoftCalendars` —
the only caller always passes it.

**5. Connect failures surfaced** (`IntegrationsView.tsx` + all 10 locales)
Google and Microsoft OAuth failures were silent; the spinner just stopped. Both
now raise an AlertDialog driven by a shared `oauthErrorKey`, suppressed for
`access_denied` (the user cancelling at the consent screen). Added
`connectFailed` / `connectFailedDescription` to both provider blocks in every
locale.

**6. Fail fast on missing client ID** (`microsoftCalendarOAuth.js`)
`startOAuthFlow()` now throws immediately when `MICROSOFT_CALENDAR_CLIENT_ID` is
unset, instead of opening the browser on a `client_id=undefined` URL and hanging
for the full 120s loopback timeout.

**Still outstanding before release**
The Entra app registration is configured — loopback redirect URIs
(`http://localhost`, `http://127.0.0.1`) under the Mobile & desktop applications
platform, supported accounts set to any Entra tenant + personal Microsoft
accounts. But `MICROSOFT_CALENDAR_CLIENT_ID` is not injected in CI yet: it needs
the repo secret plus an `echo ... >> .env` line alongside the Google ones in
`release.yml` (3 blocks) and `build-and-notarize.yml`. Until that lands,
packaged builds will hit the new fail-fast path rather than connecting.

**Verification**
`npm test` 1007 pass / 0 fail (120 better-sqlite3 suites skip locally, run in
CI), `npm run typecheck` clean, `npm run i18n:check` consistent across all 10
locales.

**Testing note for reviewers**
Entra-only admin accounts (`@*.onmicrosoft.com` with no Exchange licence) have no
mailbox — Graph returns no calendars, and Outlook Web itself fails with
`CannotResolveExternalDirectoryOrganizationIdException`. Test with a personal
Microsoft account or a licensed M365 account.

Chadpiha (approved):

PR #1252, opened 2026-07-20
fix(downloads): wait for write stream close on cancellation

hsusul:
## Summary

* defer model-download cancellation settlement until the active write stream closes
* guard competing abort, request, timer, and pipeline failures behind a single settlement path
* add a deterministic regression test for cancellation while the temporary file is still opening

## Problem

`fs.createWriteStream()` opens its destination asynchronously.

When a shared model download was cancelled, `downloadFile()` destroyed the active write stream and rejected immediately. The caller could therefore attempt to remove the temporary file before the stream had finished opening.

If the delayed open completed afterward, it could recreate the `.tmp` file after cancellation cleanup had already run. This could leave stale temporary artifacts and interfere with later download attempts.

## Fix

Failure settlement now:

1. prevents competing failure paths from settling the download more than once;
2. clears the active abort handler;
3. destroys the active write stream;
4. waits for the stream’s `close` event;
5. rejects the download only after the temporary file is safe for outer cleanup.

Existing retry, resume, redirect, error-shape, and temporary-file ownership behavior remains unchanged.

## Testing

Added a deterministic regression test that delays the underlying file open and verifies that:

* cancelling does not reject while the stream open is still pending;
* the stream closes before the download promise rejects;
* no temporary file remains after cancellation cleanup.

Validation performed:

* `node --test test/helpers/downloadUtils.test.js`
* `npm test` — 643 passed, 15 skipped, 0 failed
* `npm run typecheck`
* `npm run lint`
* `npm run i18n:check`
* `npm run build:renderer`
* targeted Prettier checks for both changed files
* `node --check src/helpers/downloadUtils.js`
* `node --check test/helpers/downloadUtils.test.js`
* `git diff --check`

`npm run format:check` continues to report the pre-existing formatting differences in:

* `src/locales/zh-CN/translation.json`
* `src/locales/zh-TW/translation.json`

Neither locale file is modified by this pull request.


gabrielste1n (approved):
Approving — thank you @hsusul. Subtle async-lifetime bug, and the fix is structured well.

The core insight is right: `fs.createWriteStream()` opens asynchronously, so destroying the stream and rejecting in the same tick let the caller reach its `unlink` before a pending open could recreate the temp file. Making `cleanup()` return a promise that resolves on `close`, and funnelling every exit through a single `fail()`, is the correct shape — it collapses what were six independent reject sites into one settlement path.

I traced the details and they hold up:

- `cleanup()` is called only from `fail()`, so there's no site that ignores the returned promise.
- `file.closed` short-circuits the already-closed case, and `destroy()` on an fs stream always emits `close`, so the promise can't dangle.
- The `settled` guard in the pipeline callback correctly makes the post-abort "premature close" error a no-op instead of a double-settle.
- The success path nulls `activeFile` after `pipeline` has already finished the stream, so a later `cleanup()` is a no-op.
- Changing the `response.on("data")` abort branch from bare `cleanup()` to `fail()` is a genuine improvement — that path previously cleaned up without ever settling.

Verified against current main: 721 tests / 0 failures, typecheck, lint and prettier clean.

gabrielste1n:
Follow-up code-quality pass before merge (`bcbcc08`): removed `activeFile = null` from the pipeline success branch. Since `cleanup()` is now reached only through `fail()`, and `fail()` returns early once `settled` is true — which the line above sets — that assignment can no longer be observed.

Everything else I left as-is. The `fail()` helper collapses six near-identical `cleanup(); reject(...)` pairs into one call site, so the diff is a net reduction in noise, and the one comment you added is a genuine non-obvious WHY that should stay. 721 tests / 0 failures, typecheck, lint and prettier clean. Merging — thanks!

PR #1256, opened 2026-07-20
Control panel navigates to external links instead of opening them in the browser

dakotahp:
## Summary

Clicking an external link in the release notes opens the link in Electron.

- In production builds, `DevServerManager.getAppUrl(true)` returns `null` (production loads via `loadFile()`, not `loadURL()`).
- The control panel's `will-navigate` guard in `src/helpers/windowManager.js` called `appUrl.startsWith("http")` on that value without a null check, throwing before `event.preventDefault()` ran.
- Net effect: the external-link guard silently no-ops in every packaged build. Clicking any external link inside the control panel (e.g. the changelog link in Settings -> System -> Software Updates) replaces the entire window's content with that page instead of opening it in the OS browser -- with no way to navigate back.
- The bug only reproduces in packaged builds; in dev mode `getAppUrl(true)` returns the dev-server URL string, so `.startsWith` never throws and the issue is invisible under `npm run dev`.
- `main.js` already null-checks `getAppUrl()` before use elsewhere (see `applySessionTokenAndRefresh`), so this fix follows the same existing pattern.

## Screenshots

Visual context on where the issue is:
Control_Panel_and_npm_run_dev_and_fix__control_panel_navigates_to_external_links_instead_of_opening_them_in_the_browser_by_dakotahp_·_Pull_Request__1_·_dakotahp_openwhispr

Before a fix:
Control_Panel_and_npm_run_dev_and_fix__control_panel_navigates_to_external_links_instead_of_opening_them_in_the_browser_by_dakotahp_·_Pull_Request__1_·_dakotahp_openwhispr

## Fix
- Null-guard `appUrl` before calling `.startsWith` on it, and drop the now-unreachable `file://${appUrl}` reconstruction (dead code -- `getAppUrl` never returns a bare path; it's either an `http(s)` URL or `null`). The pre-existing `url.startsWith("file://")` check already covers all production navigation.
- Extracted the decision logic into `src/helpers/navigationPolicy.js` (pure `isInternalNavigation(url, appUrl)`), mirroring the existing `dockManager.js`/`dockPolicy.js` split in this repo. `windowManager.js` keeps the Electron side effects; the policy module is unit-tested directly.
- Added `test/helpers/navigationPolicy.test.js`, including a regression test for the exact null-`appUrl` production case that caused the bug.

## Test plan
- [x] `node --test test/helpers/navigationPolicy.test.js` -- 4/4 passing
- [x] `npm test` -- no new failures (pre-existing sandbox-unrelated failures only)
- [ ] `npm run pack` a build, open Settings -> System, trigger an update-available state (or point release notes HTML at any external link), click a link, and confirm it opens in the default OS browser instead of navigating the control panel window away.

### DIFF
--- a/src/helpers/navigationPolicy.js
+++ b/src/helpers/navigationPolicy.js
@@ -1,17 +1,3 @@
-// Control panel navigation policy.
-//
-// The control panel is the only window that renders arbitrary HTML from
-// outside the app (e.g. update release notes via dangerouslySetInnerHTML),
-// so its will-navigate handler must tell the app's own content apart from
-// anything else and hand the "anything else" off to the OS browser instead
-// of navigating the window itself.
-
-// Whether a navigation target is the app's own content and should be left
-// alone (return true) rather than redirected to the OS browser.
-//
-// appUrl is the control panel's own URL, and is only non-null in development
-// (loadURL() against the dev server). Production loads via loadFile(), so
-// appUrl is null there and only the file:// branch matters.
 export function isInternalNavigation(url, appUrl) {
   return Boolean(
     (appUrl && url.startsWith(appUrl)) ||
--- a/src/helpers/navigationPolicy.js
+++ b/src/helpers/navigationPolicy.js
@@ -1,0 +1,21 @@
+// Control panel navigation policy.
+//
+// The control panel is the only window that renders arbitrary HTML from
+// outside the app (e.g. update release notes via dangerouslySetInnerHTML),
+// so its will-navigate handler must tell the app's own content apart from
+// anything else and hand the "anything else" off to the OS browser instead
+// of navigating the window itself.
+
+// Whether a navigation target is the app's own content and should be left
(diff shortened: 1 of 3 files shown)


greatcoat:
Thank you for this, and apologies that it sat open — the fix landed separately in #1776, before v1.9.2, and nobody closed your PR at the time.

Your diagnosis was exactly right: `DevServerManager.getAppUrl(true)` returns null in packaged builds, so the `will-navigate` handler threw before reaching `preventDefault()`, silently no-opping the external-link guard in every packaged build. Main now carries the same null-safety you proposed, at `src/helpers/windowManager.js:1247-1251`:

```js
const appUrl =
  DevServerManager.getAppUrl(true) ??
  pathToFileURL(DevServerManager.getAppFilePath(true).path).href;
```

and then routes through `isAllowedAppNavigation()`. Closing as already resolved on main.

PR #1259, opened 2026-07-20
feat: meeting audio saving + whisper large-v3 re-transcription

futuregerald:
## Summary

- **Save meeting audio**: After a call ends, mic and system audio are encoded to separate Opus files (~32 kbps mono) and stored in `userData/audio/`, gated on the data retention setting. Paths recorded in two new `notes` columns.
- **High-quality re-transcription**: "Re-transcribe (high quality)" button runs a post-call whisper.cpp large-v3 pass + diarization re-run, replacing the live transcript with a higher-accuracy result.
- **Auto-start fix**: URL check now distinguishes "timeout/error" from "no meeting found" — device signal is trusted when the check can't run (Automation pending/denied).

### Context
OpenWhispr captures meeting audio as separate mic + system PCM streams during recording. The system channel is written to a temp file for post-call diarization and then deleted. The mic stream was never persisted. The live transcript uses Parakeet TDT (0.6B) for speed, but this trades accuracy for low latency.

### The problem
1. **No audio retention**: Meeting audio was discarded after transcription — no way to replay, re-process, or improve the transcript later.
2. **Live transcript quality ceiling**: Parakeet TDT is fast but less accurate than whisper.cpp large-v3 for final transcripts. No post-call improvement path existed.
3. **Auto-start silent failure**: When macOS Automation permission was pending or all browsers timed out, `checkForActiveMeetingUrl` returned `{ matched: false }` identically to "no meeting found" — auto-start silently did nothing.
4. **Dead UI**: MCP integration card was a cloud-only feature, non-functional in the local fork.

### The fix
1. **Audio saving**: Added a parallel mic PCM write stream alongside the existing system diarization stream. At stop, the system PCM is `copyFileSync`'d before diarization cleanup (which deletes the original), then both tracks are encoded to Opus via FFmpeg and stored. The renderer passes `saveAudio: dataRetentionEnabled` through the stop IPC since the main process has no access to the settings store.
2. **Re-transcription**: New `retranscribe-meeting-note` IPC handler reads the saved Opus, feeds it to whisper-server large-v3, re-runs diarization on the system track, and overwrites `notes.transcript`. UI button shows in the note editor when saved audio exists.
3. **URL checker**: `checkForActiveMeetingUrl` now returns `{ matched: false, unavailable: true }` when all browsers fail. `_handleCallActive` trusts the device signal in this case instead of silently returning.
4. **Gain diagnostic**: RMS level logged every 100 system chunks (tag: `meeting-gain`) to validate capture levels for future loudnorm decisions.

### Pre-existing fixes
- Removed dead `McpIntegrationCard` import/usage from `IntegrationsView.tsx` (cloud feature, file kept for upstream merge compatibility).

## Test plan

- [ ] Manual: Start a meeting recording, speak 30+ seconds, stop. Verify `getNoteAudioPaths(noteId)` returns Opus file paths and both files are playable (`afplay`)
- [ ] Manual: Toggle data retention OFF, repeat — verify no audio files created and temp PCM cleaned up
- [ ] Manual: Click "Re-transcribe (high quality)" on a meeting note with saved audio — verify transcript updates
- [ ] Manual: Revoke Automation permission, enable auto-start, join a call — verify auto-start fires with "trusting device signal" in debug log
- [ ] Manual: Open Photo Booth with Automation granted — verify auto-start does NOT fire
- [ ] TypeScript typecheck passes (`cd src && npx tsc --noEmit`)
- [ ] All 10 translation JSON files valid

## Rollback plan

Revert this PR. The `ALTER TABLE notes ADD COLUMN` migrations are additive and tolerate re-application (`duplicate column` guard). No data loss on revert — new columns simply become unused. Any Opus files already saved in `userData/audio/` remain on disk (harmless).

---

**18 tasks across 6 feature areas** (full plan: `docs/plans/2026-07-17-meeting-audio-and-retranscription.md`)

Generated with [Claude Code](https://claude.com/claude-code)


futuregerald:
Opened against wrong repo — intended for futuregerald/openwhispr fork.

PR #1264, opened 2026-07-21
feat(dictation): edit highlighted text in place with voice agent commands

Chadpiha:
## What & why
Voice agent commands ("Hey <agent>, …" or the voice-agent hotkey) previously always typed the model's output at the cursor. This MR makes them selection-aware: if text is highlighted in the target app, the transcript and the selected text are sent to the configured agent model as a structured edit instruction, and the selection is replaced in place with the result. With no selection, behavior is unchanged (types at the cursor).

The design fails closed at every step, because the failure mode is overwriting text the user never chose:

- Selections are captured into single-use sessions (5-minute TTL) and re-verified byte-for-byte immediately before pasting; any mismatch, target-app change, or ambiguous read aborts with a toast instead of pasting.
- Selection reads are non-destructive where the OS allows: macOS reads AXSelectedText via a native helper (AppleScript fallback), Linux Wayland prefers AT-SPI. Windows and Linux X11 fall back to a sentinel-guarded synthetic Ctrl+C with clipboard save/restore.
- Model output is only trusted when complete: a per-request completion marker plus finish_reason truncation checks across all providers; truncated or empty output aborts the edit.
- Editors that copy the whole line on empty-selection Ctrl+C (VS Code, JetBrains, Scintilla apps, Visual Studio) are guarded: a single line with a trailing terminator from those apps is treated as "no selection".

This MR also fixes paste targeting: the target app and selection are now captured on every toggle press and recording start (previously gated on a renderer-desync-prone flag, which could paste into a stale target), and fixes the paste-text IPC handler returning a non-cloneable Promise across the IPC boundary.


Treidge:
Incredibly useful feature, can't wait to see it in action. 👍 This bridges the gap between the voice input and the text that already exists. Also, looks like it would be possible to do fast iterations on the text piece with voice instructions via the Voice Agent.

gabrielste1n (approved):

PR #1267, opened 2026-07-21
fix(models): persist local download progress in main process and allow parallel downloads

### LINKED ISSUE #1266: Model download progress bar disappears when Settings panel is closed
**Describe the bug**
When downloading a local model, closing the Settings panel or navigating to another section makes the progress bar disappear. The download itself continues in the background and the model eventually lands on disk, but the UI gives no indication of progress or completion when you return to Settings.

**To Reproduce**

1. Open Settings  --> AI/Language Models
2. Start downloading any local model
3. Close the Settings panel while the download is in progress
4. Reopen Settings --> the download is still running, but no progress bar is shown

**Expected behavior**
The progress bar should still be visible when returning to Settings, reflecting the in-flight download, and a confirmation should appear when the download completes.

**Screenshots**

https://github.com/user-attachments/assets/913bf140-b50e-4bf4-abd2-2d73731878c9

- OS: Linux
- Version: 1.7.6

IdrisGit:
> [!IMPORTANT]
> While this work was in progress, #1052 landed with a fix for restoring local model download progress after picker remounts.
> This PR follows the same goal, but uses a main process download-status registry so the renderer can restore all active downloads rather than a single active model.

## What changed
- Track active Whisper, Parakeet, and local LLM download status in the main process.
- Send lifecycle events to the control panel instead of only to the renderer that initiated the download.
- Hydrate the renderer from the main-process active-download snapshot when Settings mounts.
- Render progress, cancellation, and terminal state per model.

### LLM downloads
- Restore concurrent downloads for distinct local LLM models.
- Reject duplicate requests for the same model.
- Keep Whisper and Parakeet single-flight because their native download managers use one active process.

### Kept from #1052
- Race-safe model download snapshots.
- Safe cancellation cleanup after the underlying writer exits.
- Existing native download guards, extraction handling, and disk-space checks.

## Why this extension
#1052 modeled download state as one active model. The main-process registry is keyed by model type and model ID, so it also handles multiple active LLM downloads and per-model cancellation without relying on the renderer to remain mounted.

## Screen recording

https://github.com/user-attachments/assets/300682f0-1d4e-4c93-91dd-eb7925babaef

fixes #1266


IdrisGit:
Quick clarification: #1052 already added single-download recovery and the new background download tray. This PR builds on that work—it restores all active downloads from the main process, supports parallel downloads for different local LLMs, and shows each one in the new Settings/tray UI with its own progress and cancel state. Whisper and Parakeet remain single-download.

Chadpiha (approved):
Thanks @IdrisGit !

Reviewed the download lifecycle and onboarding flows. The follow-up fixes address duplicate-request ownership, installed-model selection, concurrent progress, and independent retries, with regression coverage for each.

PR #1272, opened 2026-07-21
fix(history): group UTC timestamps by local date

### LINKED ISSUE #1271: History date groups interpret UTC timestamps as local time
**Describe the bug**

History date-group headers interpret zone-less SQLite timestamps as local wall time, while the transcription row itself interprets the same timestamp as UTC. In non-UTC time zones near a day boundary, the row can therefore appear under the wrong date header.

**To Reproduce**

On current `main` (`7c06935a`), run this deterministic reproduction with Node 24:

```bash
TZ=America/Los_Angeles node --input-type=module -e '
import { formatDateGroup } from "./src/utils/dateFormatting.ts";
const NativeDate = Date;
globalThis.Date = class extends NativeDate {
  constructor(...args) {
    super(...(args.length ? args : ["2024-06-15T01:00:00Z"]));
  }
  static now() {
    return NativeDate.parse("2024-06-15T01:00:00Z");
  }
};
console.log(formatDateGroup("2024-06-15 00:30:00", (key) => ({
  "controlPanel.history.dateGroups.today": "Today",
  "controlPanel.history.dateGroups.yesterday": "Yesterday",
})[key] || key));
'
```

At the mocked time, both the current time and the transcription timestamp fall on June 14 in Los Angeles. The command prints `Jun 15, 2024`.

**Actual behavior**

`HistoryView` passes the SQLite timestamp string directly to `formatDateGroup()`, which calls `new Date(date)` and treats the zone-less value as local time.

`TranscriptionItem` separately appends `Z` before formatting the row time, so the date header and row time use different interpretations of the same value.

**Expected behavior**

Zone-less SQLite timestamps should be interpreted as UTC consistently. The example above should be grouped under `Today` (June 14 local time).

Database timestamps that already contain an explicit `Z` or numeric offset should keep that zone rather than receiving another `Z` suffix.

**Desktop**

- OS: macOS (the pure date-formatting reproduction is platform-independent)
- Node.js: 24.9.0
- OpenWhispr revision: `7c06935a`

**User impact**

Users outside UTC can see recent transcriptions grouped under the wrong Today, Yesterday, or calendar-date header near midnight UTC.

**Root-cause hypothesis**

The shared database-date normalizer already treats zone-less timestamps as UTC, but `formatDateGroup()` bypasses it for string inputs.

**Proposed focused scope**

- Normalize string inputs to history date grouping through the existing database-date helper.
- Preserve explicit timezone suffixes.
- Add deterministic non-UTC regression coverage around a midnight boundary.
- Leave upcoming calendar-event grouping, display layout, database schema, IPC, and timestamp storage unchanged.

hsusul:
## Summary

- normalize history database timestamp strings before local calendar-day grouping
- preserve timestamps that already carry a `Z` or numeric timezone offset
- add deterministic regression coverage for a non-UTC midnight boundary

## Problem

SQLite stores OpenWhispr timestamps in UTC without a zone suffix. The history row display already appends `Z` before converting those values to local time, but the date-group header passed the same string directly to `new Date()`.

Outside UTC, a transcription near midnight UTC could therefore display a local time on one day while appearing under the next day's header.

## Root cause

`formatDateGroup()` bypassed the existing database-date normalizer for string inputs. That made JavaScript interpret zone-less SQLite values as local wall time.

The normalizer also appended `Z` to timestamps that already used a numeric offset, producing an invalid date for synced ISO timestamps such as `2024-01-15T10:30:00+02:00`.

## Solution

History date grouping now routes string inputs through `normalizeDbDate()` so zone-less database values are interpreted as UTC consistently with the row display.

`normalizeDbDate()` now appends `Z` only when no explicit `Z` or numeric offset is present.

## Behavior before and after

At June 14, 2024 6:00 PM in Los Angeles, the SQLite timestamp `2024-06-15 00:30:00` represents June 14 at 5:30 PM locally.

- Before: the row was grouped under `Jun 15, 2024`.
- After: the row is grouped under `Today`.

## Scope and non-goals

This PR changes only database timestamp normalization and history date grouping.

It does not change:

- upcoming calendar-event grouping
- timestamp storage or database schema
- history layout or translation strings
- IPC or public return shapes
- dependencies or `package-lock.json`

## Tests

- `node --test test/helpers/dateGroupFormatting.test.js` — 8 passed, 0 failed
- `TZ=UTC node --test test/helpers/dateGroupFormatting.test.js` — 8 passed, 0 failed
- `TZ=Asia/Tokyo node --test test/helpers/dateGroupFormatting.test.js` — 8 passed, 0 failed
- `npm test` — 678 total, 663 passed, 15 skipped, 0 failed

The skipped full-suite cases are existing platform/native-dependent tests.

## Validation

- `npm run lint` — passed
- `npx prettier --check src/utils/dateFormatting.ts test/helpers/dateGroupFormatting.test.js` — passed
- `node --check src/utils/dateFormatting.ts` — passed
- `node --check test/helpers/dateGroupFormatting.test.js` — passed
- `git diff upstream/main --check` — passed

`npm run format:check` continues to report the pre-existing formatting differences in `src/locales/zh-CN/translation.json` and `src/locales/zh-TW/translation.json`. The same result was reproduced in a clean `upstream/main` worktree; neither locale file is modified here.

## Compatibility and risk

The public function signatures and `Date` input behavior remain unchanged. `formatUpcomingDateGroup()` is untouched. The regression test controls the timezone explicitly and restores it after the assertion.

## Documentation

No documentation change is needed because this restores consistent interpretation of existing database timestamps without changing a user-facing option or API.

Fixes #1271.

PR #353 changed only sticky history-header layout. PR #1250 added date-formatting coverage but did not normalize the history grouping path.


gabrielste1n (approved):
Approving — thanks @hsusul. Clean, minimal, and correctly scoped.

Routing `formatDateGroup` through `normalizeDbDate` is clearly right: the row display already appended `Z` before converting, so the header was the one place interpreting a zone-less SQLite value as local wall time. Having the row and its own group header disagree about the day is exactly the kind of bug that looks like a data problem and isn't.

The `normalizeDbDate` half is the better catch of the two — appending `Z` to a value that already carried a numeric offset produced `...+02:00Z`, an outright Invalid Date, for any synced ISO timestamp. That's a latent crash-shaped bug in every caller, and there are eight of them across CommandSearch, ConversationList/Item, NoteEditor, EmbeddedChat and NoteListItem, so this fix reaches well beyond the history view.

I specifically checked your non-goal and it's justified: `formatUpcomingDateGroup` looks like it has the same defect, but its input is `calendar_events.start_time`, which is written from `toISOString()` and from the Google API — always zone-carrying. So `new Date()` handles it correctly there and leaving it alone is right, not an oversight.

The second `formatDateGroup` caller (`AddNotesToFolderDialog`, on `note.updated_at`) benefits from the same fix. Deterministic test with a pinned non-UTC boundary rather than a wall-clock-dependent one is the right call.

PR #1274, opened 2026-07-21
fix(export): include meeting participants

### LINKED ISSUE #1273: Transcript exports omit meeting participants
## Bug description

TXT and Markdown transcript exports omit the participant metadata stored on calendar-linked notes.

## Affected component

Transcript export formatting in `src/helpers/transcriptFormatter.js`.

OpenWhispr stores note participants with the current attendee shape (`displayName`, `email`, `responseStatus`, and `self`), but `extractMetadata()` reads `p.name`. As a result, every current attendee maps to `undefined`, so the formatter silently leaves out the Participants header.

## Steps to reproduce

1. Create or open a note linked to a calendar event with named attendees.
2. Confirm the note contains participants (they are visible in the note UI).
3. Export the transcript as TXT or Markdown.
4. Open the exported file.

A deterministic formatter-level reproduction is to pass a note with:

```js
participants: JSON.stringify([
  { displayName: "Ada Lovelace", email: "ada@example.com" },
  { displayName: "Grace Hopper", email: "grace@example.com" },
])
```

to `formatTxt()` or `formatMd()`.

## Expected behavior

TXT contains:

```text
Participants: Ada Lovelace, Grace Hopper
```

Markdown contains:

```markdown
**Participants:** Ada Lovelace, Grace Hopper
```

## Actual behavior

Both export formats omit the Participants line entirely.

## User impact

Exported meeting notes lose attendee context, making shared or archived transcripts less useful and potentially obscuring who participated.

## Environment

- OpenWhispr: current `main` at `7c06935ac05f07a45ad8e771227a68a8d499a8e4`
- Node.js: 24.9.0
- macOS (the formatter behavior is platform-independent)

## Root-cause hypothesis

Participant extraction reads the legacy `name` field instead of the current `displayName` field and has no email fallback.

## Proposed scope

Update participant metadata extraction to use the stored attendee fields while preserving compatibility with legacy `name` values, and add focused regression coverage for TXT and Markdown exports.

I intend to work on this fix.

hsusul:
## Problem

TXT and Markdown transcript exports omit meeting participants even though the note UI has valid calendar attendees.

**Before:** current attendee objects map to empty values, so the export silently leaves out its Participants header.

**After:** exports list attendee display names; attendees without one use their email address.

## Root cause

The current attendee schema stores names in `displayName`, but `extractMetadata()` read only the legacy `name` property.

## Fix

- read `displayName` from current attendee records
- retain `name` compatibility for legacy records
- fall back to `email` when an attendee has no display name
- cover TXT and Markdown output plus fallback behavior with deterministic formatter tests

## Scope

This changes only participant metadata extraction and its focused tests. It does not change attendee persistence, note UI behavior, SRT/JSON output, export layout, or dependencies.

## Validation

Environment: macOS arm64, Node.js 24.9.0, npm 11.6.0.

- `npm ci` — passed; 912 packages installed
- `node --test test/helpers/transcriptFormatter.test.js` — 7 passed, 0 failed
- `npm test` — 678 tests: 663 passed, 0 failed, 15 skipped
- `npm run lint` — passed
- `node --check src/helpers/transcriptFormatter.js` — passed
- `node --check test/helpers/transcriptFormatter.test.js` — passed
- `prettier --check src/helpers/transcriptFormatter.js test/helpers/transcriptFormatter.test.js` — passed
- `git diff --check` — passed

`npm run format:check` continues to report the pre-existing Prettier warnings in `src/locales/zh-CN/translation.json` and `src/locales/zh-TW/translation.json`. The identical failure was reproduced on clean upstream `main`; both changed files pass Prettier.

Fixes #1273

### DIFF
--- a/src/helpers/transcriptFormatter.js
+++ b/src/helpers/transcriptFormatter.js
@@ -59,7 +59,7 @@ function extractMetadata(note) {
   let participants = [];
   try {
     const parsed = JSON.parse(note.participants || "[]");
-    participants = parsed.map((p) => p.name).filter(Boolean);
+    participants = parsed.map((p) => p.displayName || p.name || p.email).filter(Boolean);
   } catch {}
 
   return { title, dateStr, participants };
--- a/src/helpers/transcriptFormatter.js
+++ b/src/helpers/transcriptFormatter.js
@@ -59,7 +59,7 @@ function extractMetadata(note) {
   let participants = [];
   try {
     const parsed = JSON.parse(note.participants || "[]");
-    participants = parsed.map((p) => p.displayName || p.name || p.email).filter(Boolean);
+    participants = parsed.map((p) => p.displayName || p.email).filter(Boolean);
   } catch {}
 
   return { title, dateStr, participants };
--- a/src/helpers/transcriptFormatter.js
+++ b/src/helpers/transcriptFormatter.js
@@ -70,7 +70,8 @@ function formatTxt(note, segments, speakerMappings) {
   const { title, dateStr, participants } = extractMetadata(note);
 
   const lines = [title, dateStr];
-  if (participants.length) lines.push(`Participants: ${participants.join(", ")}`);
+  if (participants.length)
+    lines.push(`${i18nMain.t("notes.editor.participants")}: ${participants.join(", ")}`);
   lines.push("", "──────────────────────────────────", "");
   for (const seg of merged) {
     lines.push(`[${formatTimestamp(seg.timestamp)}] ${resolveSpeaker(seg, speakerMappings)}:`);
@@ -128,7 +129,8 @@ function formatMd(note, segments, speakerMappings) {
   const { title, dateStr, participants } = extractMetadata(note);
 
   const lines = [`# ${title}`, "", `**Date:** ${dateStr}`];
-  if (participants.length) lines.push(`**Participants:** ${participants.join(", ")}`);
+  if (participants.length)
+    lines.push(`**${i18nMain.t("notes.editor.participants")}:** ${participants.join(", ")}`);
   lines.push("", "---", "");
   for (const seg of merged) {
     lines.push(`**${resolveSpeaker(seg, speakerMappings)}** \`${formatTimestamp(seg.timestamp)}\``);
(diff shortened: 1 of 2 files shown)


gabrielste1n (approved):
Thanks @hsusul — nice catch, and thank you for the thorough validation notes. Confirmed the root cause end to end: `CalendarAttendee` is `{ email, displayName, responseStatus, self }`, and the only two writers of `note.participants` both write `displayName` (calendar sync in `googleCalendarManager.js`, manual add in `NoteParticipants.tsx`), so `parsed.map((p) => p.name)` was always emptied by the `.filter(Boolean)` and the header never rendered.

Approving. I pushed two small commits on top:

**1. Dropped the `p.name` fallback** (`7d55105e`)

Tracing this back, `name` isn't a legacy field — it was a typo present from the day the export feature was written. `git show ab1fe256^:src/helpers/ipcHandlers.js` already has `parsed.map((p) => p.name)` inline before the formatter was extracted, and grepping the desktop, mobile, API, notes, and CLI repos turns up no producer that has ever written a `name` key onto an attendee. So the middle branch was unreachable, and the `{ name: "Legacy Guest" }` case would have locked in a schema that never shipped. Resolution is now `displayName || email`, matching the existing pattern in `ipcHandlers._resolveOneOnOneOtherParticipant`. The second test is retargeted to a real mixed case (one attendee with a display name, one without), so it still covers the email fallback.

**2. Localized the label** (`8f3cd845`)

The `Participants:` / `**Participants:**` labels were hardcoded English even though the same file already resolves speaker labels through `i18nMain`. Turns out `notes.editor.participants` already exists and is translated in all 10 locales — it was orphaned in `27803c78` when the note editor metadata chip was replaced by the folder picker. Reusing it was two lines and no locale edits. Verified live: `de: Teilnehmer`, `ja: 参加者`, `ru: Участники`.

**One thing left deliberately unchanged:** calendar-synced notes include your own attendee record (`self: true`), so you appear in your own export. Elsewhere we strip self via `_parseNonSelfParticipants`, but that's for counting *other* speakers for diarization. An export header is a record of who attended, which conventionally includes you — same as Calendar's own guest list. Flagging it so the decision is on the record.

Full suite green on the final commit: 678 tests, 663 passed, 0 failed, 15 skipped; typecheck, lint, `i18n:check`, and prettier all clean.

PR #1276, opened 2026-07-21
fix(notes): distinguish mirrored notes from transcripts

### LINKED ISSUE #1275: Auto-mirrored notes can be mistaken for transcript files
**Describe the bug**

The optional "Save notes as files" mirror does not reliably distinguish a note Markdown file from its `-transcript.md` companion.

This has two observable effects:

- "Show note file" can reveal the transcript companion instead of the note.
- If a note title ends in `transcript` (for example, `Meeting transcript`), writing the companion transcript deletes the note Markdown file because the note filename also ends in `-transcript.md`.

**To Reproduce**

The behavior is deterministic through the `markdownMirror` interface:

1. Initialize the mirror in a temporary directory.
2. Write note `7` with title `Meeting transcript`.
3. Write its transcript companion.
4. List the mirrored files.

On current `main`, the directory contains only:

```text
7-meeting-transcript-transcript.md
```

The expected note file is missing:

```text
7-meeting-transcript.md
```

With an ordinary title such as `Weekly sync`, both files are written, but `getNotePath(42)` returns `42-weekly-sync-transcript.md` on macOS rather than `42-weekly-sync.md`.

**Expected behavior**

The mirror should keep the note Markdown file and transcript companion separate regardless of note title or filesystem enumeration order. `getNotePath()` should always return the note file.

- OS: macOS arm64
- OpenWhispr revision: `7c06935ac05f07a45ad8e771227a68a8d499a8e4`
- Node.js: `v24.9.0`
- npm: `11.6.0`

**Additional context**

User impact: a valid note title can make the local note mirror disappear, and the reveal action can open the wrong mirrored artifact.

Root-cause hypothesis: `_globNoteFiles()` accepts every matching `.md` file, including transcript companions, while `_globTranscriptFiles()` treats every `-transcript.md` file as a transcript, including notes whose slug ends in `transcript`.

Proposed focused scope: distinguish files using their existing written formats (note YAML frontmatter versus transcript Markdown), add deterministic filesystem regression tests, and leave IPC, UI, database, export formatting, and filenames unchanged.

hsusul:
## Summary

- distinguish auto-mirrored note Markdown from transcript companion files by content
- preserve both files when a note title ends in `transcript`
- keep note lookup and stale-file cleanup pointed at the correct artifact after renames

## Problem

The mirror classified files only by their names. Because a transcript companion also starts with the note ID and ends in `.md`, it could be returned as the note file. Conversely, a legitimate note whose slug ended in `transcript` could be classified as a transcript and deleted when its companion was written.

Before this change:

- `getNotePath()` could return the `-transcript.md` companion instead of the note
- a title such as `Meeting transcript` could cause the note Markdown file to disappear

After this change, note Markdown is identified by the YAML frontmatter marker written by the mirror. Transcript Markdown remains identified by its suffix only when that marker is absent. Legacy `-transcript.txt` handling is unchanged.

## Scope

This is limited to the Markdown mirror helper and focused regression coverage. It does not change filenames, exported formats, IPC contracts, UI behavior, or stored data.

## Tests and validation

- `ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/markdownMirror.test.js test/helpers/transcriptFormatter.test.js` — 7 passed
- `ELECTRON_OVERRIDE_DIST_PATH=/tmp npm test` — 678 tests: 663 passed, 15 skipped, 0 failed
- `npm run lint` — passed (existing module-type warning only)
- `npm run build:renderer` — passed (existing bundler warnings only)
- `npx prettier --check src/helpers/markdownMirror.js test/helpers/markdownMirror.test.js` — passed
- `git diff --cached --check` — passed before commit

`npm run format:check` still reports the two existing locale files `src/locales/zh-CN/translation.json` and `src/locales/zh-TW/translation.json`. I reproduced the identical result in a clean worktree at `upstream/main` (`7c06935a`), so it is not introduced by this patch.

Fixes #1275


gabrielste1n (approved):
Approving with a small follow-up — thanks @hsusul. Good bug and the right fix: classifying by filename alone couldn't work when a transcript companion shares the note's ID prefix and `.md` suffix, and the failure mode was nasty — a note titled "Meeting transcript" could have its markdown deleted when the companion was written.

Discriminating on content is the correct call, and I verified it holds: `_buildFrontmatter` joins with `\n` so mirror-written notes always begin `---\n`, and `formatMd` starts transcripts with `# <title>`, so the two kinds never collide. I also confirmed the frontmatter has been written since the mirror's very first commit, so there are no legacy note files that would suddenly stop being recognised.

**What I pushed** (`c7bf03a`): the probe compared the first 4 bytes to exactly `"---\n"`, so a note file re-saved by an external editor with CRLF endings or a UTF-8 BOM stopped matching. Since these files live in a user-visible folder and are meant to be opened in Obsidian and the like, that's reachable — and once it happens the note is orphaned from the mirror: `getNotePath` returns null, renames leave a duplicate, and `deleteNote` skips it. Now matches an optional BOM followed by `---` and either line ending, with a regression test covering both rewrites.

Worth considering separately, not a blocker: parsing the `id:` line out of the frontmatter and requiring it to equal the note ID would be strictly stronger than a marker sniff, and would also rule out adopting a user's own hand-written file that happens to start with the note's ID and carry YAML frontmatter.

Verified against current main: 723 tests / 0 failures, typecheck, lint and prettier clean.

gabrielste1n:
Merging. Only cleanup on top of the BOM/CRLF fix was aligning a `"utf8"` literal to the `"utf-8"` spelling used everywhere else in the file (`daf0213`). 723 tests / 0 failures, typecheck, lint and prettier clean. Thanks!

PR #1278, opened 2026-07-22
feat: multi-language transcription with switcher in the dictation panel

qqubica:
## Why

Bilingual and multilingual users switch dictation languages constantly. Today every switch means opening Settings, changing the transcription language, and coming back — and Whisper's auto-detect, while helpful, is noticeably less accurate than an explicit language for short dictations and for users who mix languages across contexts (work in English, messages in another language).

This PR lets users pick up to **5 preferred transcription languages** and switch the active one directly from the dictation panel, without opening Settings.

## What

- **Settings → Language** and both onboarding language steps become a multi-select picker (up to 5 languages). The trigger shows the selected count; selections render as chips — clicking a chip makes it the active language, ✕ removes it.
- **Dictation panel**: when more than one language is selected, a small chip (e.g. `EN ▾`) appears next to the mic on hover; clicking it opens a dropdown to switch. The command menu gets the same "Transcription language" section. Chip placement is anchor-aware so the idle mic never shifts: bottom-left anchors place the chip after the mic, centered anchors reserve an invisible chip slot.
- **Switching mid-recording just works**: whisper.cpp and batch cloud transcription read the language at transcription time, so a mid-recording switch applies to the recording in progress, and the chunked live preview is retargeted. The one exception is cloud realtime streaming, which fixes its language at session start — switching during it shows a toast explaining the new language takes effect on the next dictation.
- **Auto detect stays exclusive**: choosing it clears the multi-language set, and it can never be part of the set (invalid persisted values are repaired on load).
- **The transcription pipeline is untouched**: `preferredLanguage` remains the single value passed to whisper.cpp / cloud APIs; `preferredLanguages` is a UI-level set that feeds the switcher.

## Where the diff lands

+982/−88 over 27 files overstates the feature's footprint. Split five ways: half is renderer UI, a quarter is tests and the same strings translated ten times, and the main-process surface is deliberately small (~50 lines).

| Area | ± lines | What it is |
| --- | ---: | --- |
| **UI — dictation panel switcher**<br>`src/App.jsx`, `ui/Toast.tsx`, `ui/tooltip.tsx` | +257 / −20 | Hover chip + dropdown next to the mic, the command-menu language section, menu lifecycle tied to recording state, panel sizing, the streaming-switch toast, plus the two small `Toast`/`Tooltip` tweaks described under reviewer notes. |
| **UI — language pickers**<br>`ui/LanguageSelector.tsx`, `SettingsPage.tsx`, `OnboardingFlow.tsx` | +231 / −33 | `LanguageSelector` multi-select mode (chips, selected-count trigger, keyboard navigation, a11y) and its use in Settings → Language and both onboarding language steps. |
| **State — language-set rules**<br>`helpers/languagePreferences.js`, `stores/settingsStore.ts`, `utils/languageSupport.ts`, `hooks/useSettings.ts` | +168 / −4 | New pure helper owning every set invariant — cap of 5, dedupe, "auto" exclusivity, the active language always a member, repair of invalid persisted values — plus store wiring and language metadata. |
| **Backend — main process**<br>`preload.js`, `helpers/ipcHandlers.js`, `helpers/windowManager.js`, `helpers/windowConfig.js`, `types/electron.ts` | +53 / −0 | Two small IPC additions: `set-main-window-menu-focus` (temporary window focus while a panel menu is open) and `update-dictation-preview-language` (retargets the chunked live preview after a mid-recording switch). Two new panel window sizes. |
| **Tests, locales, docs**<br>`test/helpers/languagePreferences.test.js`, 10× `locales/*/translation.json`, `CLAUDE.md` | +273 / −31 | 12 unit tests covering the language-set invariants; the same new strings translated in all ten locales; one settings-docs line. |

Every rule about the language set lives in `languagePreferences.js` as pure functions with a dedicated test file, so the store and the three UIs consuming it stay thin.

## Focus and window behavior (reviewer notes)

The dictation panel window is intentionally non-focusable so it never steals focus from the app being dictated into. That has two consequences for menus living in it:

- A click landing in *another application* is invisible to the renderer and would leave an open menu dangling. While a menu is open, the main process temporarily makes the window focusable and focused (`set-main-window-menu-focus` IPC); a click elsewhere then blurs the window, and the renderer closes the menu on `blur`. The flag is dropped as soon as the menu closes, restoring the non-focusable behavior. The paste path already guards against the panel holding focus.
- The chip (and any open menu) disappears when dictation stops or enters processing, so the menu state is force-closed at that moment — otherwise the invisible panel would keep window focus and the enlarged menu size, then reopen the menu unprompted afterwards.

On Windows the panel is never click-through, so it rests at the base 96×96 size and only widens for the chip while hovered — keeping the extra width at rest would leave a permanent click-eating strip beside the mic.

Small fixes that fell out of this work: the `LanguageSelector` dropdown also closes on window blur; panel toasts stack above an open menu via a `--toast-viewport-bottom` CSS variable instead of covering it; `Tooltip` accepts `ReactNode` content and skips rendering for empty content.

## Testing

- `npm run lint`, `npm run typecheck`, `npm run i18n:check` all pass; the new strings exist in all 10 locale files with placeholder parity.
- `node --test test/helpers/languagePreferences.test.js`: 12/12 pass.
- `npm test`: the only failures (5, in `systemTar` / `download-utils` / `gpuBinaryManager`) are Windows-environment issues in helpers this PR does not touch; CI runs Ubuntu.
- Manually tested on Windows 11: multi-select in Settings and onboarding, chip switching, panel dropdown + command-menu switching, mid-recording switches (batch and streaming), auto-detect exclusivity, menu dismissal on outside clicks (including clicks landing in other apps), Escape handling, and the hover-only panel widening.
- Manually tested on macOS (MacBook): the menu focus dance — opening the chip or command menu does not activate OpenWhispr (no menu bar/Dock switch), focus returns to the target app on dismiss, and the panel stays always-on-top and non-focusable at rest; menu dismissal re-run per the review notes (clicks landing in other apps, Escape, chip ↔ command-menu swap); mid-recording switching with the paste still landing in the target app; menu force-close when dictation stops; the streaming-switch toast; and no Dock icon flicker from panel menus.

## Re-merged with upstream/main (2026-08-22)

Upstream moved ~300 commits past the last merge point (dictation-pill component extraction, onboarding redesign, dictation-error windows), so the feature was re-ported onto the new structure:

- **App.jsx → extracted components.** The language chip now lives next to `VoicePill` in the pill row (same anchor-aware placement rules), and the command menu's language section moved into `PillCommandMenu` (`menuRef` exposure, language-trigger exclusion from outside-click, close-on-select).
- **Window sizing joined the size ladder.** `WITH_LANGUAGE` and `WITH_MENU_LANGUAGE` are ranked in `windowSizeLadder.js` alongside upstream's new `RECORDING`/`DICTATION_ERROR` keys, and `useMainWindowSizeOwner` threads the language inputs through — the Windows rest-size rule (BASE unless hovered) is preserved.
- **Onboarding adopted upstream's redesigned `LanguageSelectionStep`** (multi-select, search, chips) instead of the fork's own picker UI. The step is capped at 5 selections and now feeds `preferredLanguages` via `resolveLanguageSetChange` on completion, so the panel switcher works right after onboarding. The earlier "auto-detect dead-ends in onboarding" review bug is gone by construction — the new step offers no auto option (auto stays reachable from Settings).
- **Menu focus dance unchanged:** the `hasOpenMenu`-keyed `set-main-window-menu-focus` effect, blur-close, and menu-swap race handling are re-implemented against the new component split.

Post-merge verification: `npm run lint`, `npm run typecheck`, `npm run i18n:check` all pass; `node --test` is 2276/2586 pass with the same 133 failures (and 176 skips) that pristine upstream/main produces on this Windows machine (`debugLogger` electron-mock and tar/gpu helper environment issues) — none in files this PR touches; the 12 `languagePreferences` tests all pass.


src/components/OnboardingFlow.tsx
```diff
@@ -578,12 +584,19 @@ export default function OnboardingFlow({ onComplete }: OnboardingFlowProps) {
                     {t("onboarding.setup.language")}
                   </label>
                   <LanguageSelector
-                    value={preferredLanguage}
-                    onChange={(value) => {
-                      updateTranscriptionSettings({ preferredLanguage: value });
+                    multiple
+                    values={transcriptionLanguages}
+                    onValuesChange={(values) => {
```
qqubica: **Bug: Auto-detect becomes a silent no-op here.** Auto-detect is a selectable option in the list (`languageRegistry.json` includes `auto`), but this handler passes the values straight to `updateTranscriptionSettings({ preferredLanguages })`, and `setPreferredLanguages` filters `auto` out. So once the user has picked any language during onboarding, clicking Auto-detect does nothing — no state change, no feedback, and no way back to auto-detection without visiting Settings later.

`SettingsPage.tsx` handles this case explicitly (checks `values.includes("auto")` and routes to `setPreferredLanguage("auto")`); the same branch is needed here. Applies to both selector instances in this file (this one and the one near line 678) — worth extracting the shared handler.

src/helpers/windowManager.js
```diff
@@ -127,6 +127,28 @@ class WindowManager {
     }
   }
 
+  setMainWindowMenuFocus(open) {
```
qqubica, line 215: This is new main-process focus behavior, and the platforms disagree about what `blur()` hands focus back to:

- **Windows**: after `blur()` there is no guarantee the previously active app is re-activated (Electron activates the next window in Z-order). The `paste-text` guard covers the dictation flow, but for an idle language switch the user's previous app may be left unfocused.
- **macOS**: `setFocusable(true)` + `focus()` on the overlay activates the whole app (menu bar/Dock switch, previous app deactivates). `activateTargetPid` recovers on the paste path, but please verify the idle-switch case doesn't leave OpenWhispr as the active app, and that toggling `setFocusable` doesn't disturb the floating-panel/always-on-top setup.

Also note the language chip is clickable mid-recording, so opening the menu steals focus from the app being dictated into — the paste guard should recover, but that scenario deserves an explicit manual test on each platform.

src/App.jsx
```diff
@@ -364,12 +438,75 @@ export default function App() {
...
           onMouseLeave={() => {
             setIsHovered(false);
-            if (!isCommandMenuOpen) {
+            if (!isCommandMenuOpen && !isLanguageMenuOpen) {
               setWindowInteractivity(false);
             }
           }}
         >
-          {(isRecording || isProcessing) && isHovered && (
+          {showLanguageSwitcher && (isHovered || isLanguageMenuOpen) && !isProcessing && (
```
qqubica: The chip renders during recording, so a mid-recording switch has different semantics per engine: whisper/cloud read the language at transcription time (the new language applies to the recording in progress), while the Parakeet online streaming-commit path fixes it at stream start (the old language applies — unless the commit falls back to full decode, which then uses the new language). Not a blocker, but consider either hiding the switcher while recording or documenting that a switch takes effect from the next dictation for streaming models.

src/components/ui/LanguageSelector.tsx
```diff
@@ -143,8 +165,20 @@ export default function LanguageSelector({
...
   };
 
+  const selectedValues = multiple ? (values ?? []) : value ? [value] : [];
+
   const handleSelect = (languageValue: string) => {
-    onChange(languageValue);
+    if (multiple) {
+      const current = values ?? [];
+      if (current.includes(languageValue)) {
+        // Never allow deselecting the last language.
```
qqubica: Two silent no-ops in `multiple` mode: deselecting the last remaining language, and selecting past `maxValues`. Both do nothing with no feedback. Suggest visually disabling (or hinting on) options that can't be toggled so the cap of 5 is discoverable — right now nothing tells the user why a click had no effect.

src/components/SettingsPage.tsx
```diff
@@ -96,6 +97,10 @@ import type { InferenceModeOption } from "./ui/SettingsSection";
 import { useSettingsLayout } from "./ui/useSettingsLayout";
 import { useUsage } from "../hooks/useUsage";
 import { cn } from "./lib/utils";
+import { MAX_PREFERRED_LANGUAGES } from "../stores/settingsStore";
+import languageRegistry from "../config/languageRegistry.json";
+
+const LANGUAGE_BY_CODE = new Map(languageRegistry.languages.map((l) => [l.code, l]));
```
qqubica: Nit: a `const` declaration in the middle of the import block. Also, `src/utils/languageSupport.ts` already wraps the registry (`getLanguageLabel`) — reusing it (extending it with a `getLanguageOption(code)` returning flag+label if needed) would avoid a second parallel lookup structure built from `languageRegistry.json`.

src/stores/settingsStore.ts
```diff
@@ -1325,7 +1351,39 @@ export const useSettingsStore = create<SettingsState>()((set, get) => ({
...
+      // so the dictation panel switcher always offers it.
+      if (s.preferredLanguages.length > 0 && !s.preferredLanguages.includes(value)) {
+        const languages = [...s.preferredLanguages, value].slice(0, MAX_PREFERRED_LANGUAGES);
+        if (isBrowser) localStorage.setItem("preferredLanguages", JSON.stringify(languages));
+        return { preferredLanguage: value, preferredLanguages: languages };
+      }
+      return { preferredLanguage: value };
+    });
+  },
+  setPreferredLanguages: (languages: string[]) => {
```
qqubica, line 1745: The normalization/repair rules here (auto exclusivity, dedupe + cap, active-must-be-in-set, legacy repair at load) are exactly the kind of pure logic this repo unit-tests via `node --test` (cf. `dockPolicy.js` → `test/helpers/dockPolicy.test.js`). Suggest extracting them into a small pure helper with tests — they are the invariants the panel switcher depends on, and nothing covers them right now.

src/helpers/windowConfig.js
```diff
@@ -30,7 +30,10 @@ const FLOATING_OVERLAY_TYPE =
 
 const WINDOW_SIZES = {
   BASE: { width: 96, height: 96 },
-  WITH_MENU: { width: 240, height: 280 },
+  // Wide enough for the language chip + cancel button next to the mic
+  WITH_LANGUAGE: { width: 160, height: 96 },
+  // Tall enough for the command menu including the language switcher section
+  WITH_MENU: { width: 240, height: 340 },
```
qqubica: Nit: `WITH_MENU` grew to 340 for the language section, but it is also used when the switcher is hidden (single language), leaving ~60px of invisible window above the menu that eats clicks while the menu is open (the click closes the menu but never reaches the app underneath). Fine to keep if sizing by content is more trouble than it's worth.

qqubica (commented):
Reviewed the full diff plus the language-consumer paths it touches. Overall this is in good shape: the core design (multi-select as a UI-level set, `preferredLanguage` stays the single value the pipeline consumes) keeps every engine path untouched, existing single-language users see no behavior change (empty set falls back to `[preferredLanguage]`), cross-window sync works because `preferredLanguages` was added to `ARRAY_SETTINGS` for the storage-event listener, and the active-language invariant (always a member of the set, `auto` exclusive) holds through Settings, onboarding, the panel switcher, and the bulk `applySettings` import path.

Verified locally (Windows): `npm run typecheck` passes, `npm run i18n:check` passes (all 10 locales), eslint is clean on the touched files, and `node --test` is 689/694 — the 5 failures (`systemTar`, `extractArchive`, `gpuBinaryManager`) reproduce identically on upstream `main` on this machine, so they are unrelated to this PR.

One functional bug to fix before undrafting (Auto-detect dead-ends in onboarding — see inline comment); the rest are non-blocking suggestions. Since `setMainWindowMenuFocus` is new main-process focus behavior, please confirm in the PR description which platforms the menu focus dance was manually tested on — macOS app-activation semantics differ from Windows.

src/components/ui/LanguageSelector.tsx
qqubica: Can the ifs here be inverted so they are not nested?

src/components/OnboardingFlow.tsx
qqubica: Fixed in ef3c1d14. Both onboarding selector instances now share `handleTranscriptionLanguagesChange`, which routes through `resolveLanguageSelection()` — the same exclusive-auto rule the Settings page uses (that rule now lives in `src/helpers/languagePreferences.js` and both callers consume it). Picking Auto-detect during onboarding clears the preset list and returns to auto detection.

src/helpers/windowManager.js
qqubica: Addressed in ef3c1d14 as far as it can be from code:

- The mid-recording scenario no longer exists — the chip and the command-menu language section are hidden while recording, and an already-open language menu is force-closed when recording starts, which releases the temporary focus before any dictation is under way (see the App.jsx thread).
- After the menu-focus toggle ends, `enforceMainWindowOnTop()` re-asserts the floating-panel/always-on-top setup, so toggling `setFocusable` cannot leave the panel at a degraded window level.

Idle-switch focus verified on Windows (panel stays interactive, previous app regains keyboard focus on next click; no dictation is in flight in that state so nothing is lost). The macOS checks — idle switch must not leave OpenWhispr as the active app, panel keeps floating across Spaces — can't be run from this machine and are explicit steps in the manual test plan (sections 5–6).

src/App.jsx
qqubica: Went with hiding (ef3c1d14): the switcher — both the hover chip and the command-menu language section — is not rendered while recording, and an open language menu closes the moment recording starts. A language switch therefore always applies from the next dictation, uniformly across whisper/cloud and Parakeet streaming engines, and the mid-recording focus-steal scenario from the windowManager thread is gone as a side effect.

src/components/ui/LanguageSelector.tsx
qqubica: Done in ef3c1d14:

- Non-toggleable options are now visually disabled (dimmed, `cursor-not-allowed`, `disabled` + `aria-disabled`) with a native tooltip saying why — "At least one language must stay selected" / "Limit of {{max}} languages reached" (new `languageSelector.*` keys in all 10 locales).
- `handleSelect` is flattened to early returns per the follow-up comment; the disabled logic lives in one `getDisabledReason()` used by both the click/keyboard path and rendering.
- While here: Auto-detect is treated as an exclusive value that bypasses the cap (`exclusiveValues`, defaults to `["auto"]`). Previously, selecting it with a full set of 5 was itself a silent no-op — same family as the onboarding bug.

src/components/SettingsPage.tsx
qqubica: Done in ef3c1d14: the stray `const` and the parallel map are gone. `src/utils/languageSupport.ts` now exposes `getLanguageOption(code)` (registry-backed `{code, label, flag}`), `getLanguageLabel` reuses it, and SettingsPage consumes it for the language chips.

src/stores/settingsStore.ts
qqubica: Extracted in ef3c1d14 to `src/helpers/languagePreferences.js` — pure transitions (`normalizePreferredLanguages`, `resolveActiveLanguageChange`, `resolveLanguageSetChange`, `resolveLanguageSelection`) consumed by the store, SettingsPage and onboarding, with 12 `node --test` cases in `test/helpers/languagePreferences.test.js` (dockPolicy pattern).

Writing the tests surfaced a real bug in the previous inline code: activating a language while the set already had 5 entries appended-then-sliced, dropping the *new* language — the active language ended up outside the set, so the panel switcher wouldn't offer it. The helper now evicts the last preset instead, and a test pins the invariant.

src/helpers/windowConfig.js
qqubica: Fixed in ef3c1d14: `WITH_MENU` is back at 280 and a new `WITH_MENU_LANGUAGE` (340) is requested only when the command menu actually renders the language section (multi-language set, not recording). Single-language users no longer get the invisible click-eating strip above the menu.

src/App.jsx
qqubica: Follow-up: direction reversed in 99a0c3b4 — the switcher is now available **during** recording after all. Digging into the engine flow showed the divergence this comment worried about doesn't exist in the code: the language hint only ever reaches whisper local and cloud batch decodes (both read it at transcription time, after stop), while Parakeet — offline *and* online streaming — never consumes it (sherpa-onnx auto-detects; `createOnlineStream` takes no language). So a mid-recording switch uniformly applies to the recording in progress wherever the engine honors the hint at all. The chunked whisper live preview was the one thing pinned at session start; a new `update-dictation-preview-language` IPC retargets it on switch. Remaining known limit: an AssemblyAI streaming session keeps its start language until the next dictation.

src/helpers/windowManager.js
qqubica: Follow-up: with mid-recording switching re-enabled in 99a0c3b4, the "chip clickable mid-recording steals focus" scenario from this comment is back in scope. The manual test plan covers it explicitly (dictate into Notepad/TextEdit, open the chip menu mid-recording, switch, keep speaking, stop — paste must land in the target app) on Windows and macOS. The `enforceMainWindowOnTop()` re-assert after the focusable toggle stays.

qqubica (commented):
Third self-review pass over the current head (4d4dcf9d, post-merge with upstream main). Re-verified that the earlier rounds' fixes still hold (languagePreferences invariants + tests, `WITH_MENU_LANGUAGE` sizing, exclusive auto handling, disabled-option hints), and that the transcription pipeline still consumes only `preferredLanguage`, read at transcription time via `getEffectiveSttLanguage` — so the mid-recording switch semantics documented in the earlier threads are unchanged.

Found four issues, all fixed in 329ff196:

1. **i18n:check was failing on the head commit** (would block CI): 4d4dcf9d added `languageSelector.selectedCount_one/_other` to seven locales but only `_other` to ja/zh-CN/zh-TW, and `scripts/check-i18n.js` requires full key parity (exit 1). Added the `_one` keys — inert at runtime (the CJK cardinal rule only ever selects `other`) but required by the checker.

2. **Switching directly between the two panel menus could close the menu that just opened** (race): the menu-focus effect was keyed on the two menu booleans, so a language↔command swap re-ran it even though a menu stayed open — cleanup called `setMainWindowMenuFocus(false)` (which `blur()`s the window) before the new run called `setMainWindowMenuFocus(true)`, and the in-flight blur event landed after the swap, closing the new menu via the blur-close handler. The same failure existed through a second path: mousedown on the *other* menu's trigger closed menu A via the outside-click handler (a transient zero-menus state that releases focus) before the click opened menu B. Fixed by keying the effect on the derived `hasOpenMenu` boolean and excluding both menu triggers from the outside-click close — the triggers' own click handlers swap the menus in a single React commit, so window focus is never released mid-switch.

3. **The Settings language chips were indistinguishable to screen readers**: the shared `aria-label` ("Active language") overrode each chip's visible language name, so all chips announced identically. Dropped the label — the flag span is `aria-hidden`, so the accessible name is now the language itself, with `aria-pressed` marking the active chip — and removed the now-unused key from all locales.

4. **Command-menu language items didn't close the menu on select**, unlike the chip popup and every other item in that menu. They do now.

Also updated CLAUDE.md's localStorage key list while touching it: the `language` entry matched no code; it now documents `preferredLanguage`/`preferredLanguages`.

Verified after the fixes: `npm run i18n:check` passes (was exit 1), `npm run typecheck` passes, eslint clean on touched files, `node --test` 1058/1183 pass with the same 5 pre-existing environment failures as upstream `main` on this machine (systemTar ×2, extractArchive, gpuBinaryManager ×2; 120 skipped).

Note for the manual test plan: after the outside-click change, a mousedown on the mic button or the language chip no longer dismisses an open menu — the click completes the menu swap instead — so the menu-dismissal steps (sections 5–6) should be re-run on each platform when doing the macOS pass.

arquitech-ai:
Love that idea to be able to change the dictation language from the widget!

Maoaii:
Would love to see this merged in the near future. I'm constantly sending messages in Portuguese or English during the day, and the auto-detect on iOS doesn't seem to work that well, so having that toggle on the keyboard would fix all of those issues

PR #1283, opened 2026-07-22
fix(parakeet): preserve newer partial after duplicate final

### LINKED ISSUE #1282: Duplicate Parakeet final clears newer partial transcript
**Describe the bug**

The online Parakeet/Nemotron message accumulator clears the current partial transcript whenever it receives a final message, including a duplicate final for an older segment. If a newer segment's partial has already arrived, that live text disappears from the preview and from the stream's accumulated result until another update arrives.

`createOnlineAccumulator()` already deduplicates finalized segments by `segment`, but `partialText` is reset unconditionally after that dedupe check.

**To Reproduce**

On current `main` (`f9085261761ee9706315e0a785e2a89a7ea29887`), run:

```sh
node - <<'NODE'
const { parseOnlineMessages } = require('./src/helpers/parakeetWsResult');
const message = (value) => JSON.stringify(value);

console.log(parseOnlineMessages([
  message({ text: 'hello', is_final: true, segment: 0 }),
  message({ text: 'new partial', is_final: false, segment: 1 }),
  message({ text: 'hello', is_final: true, segment: 0 }),
]));
NODE
```

The output is:

```text
hello
```

**Expected behavior**

The duplicate final for segment 0 should be ignored without clearing segment 1's newer partial. The output should remain:

```text
hello new partial
```

**Screenshots**

Not applicable; the failure is deterministic at the message-accumulator boundary.

- OS: macOS 15.7.3 (arm64)
- OpenWhispr revision: `f9085261761ee9706315e0a785e2a89a7ea29887`
- Node.js: 24.9.0
- npm: 11.6.0

**Additional context**

The online streaming path calls this accumulator for every websocket message and publishes its current text through `onUpdate`, so the state loss is user-visible in live transcription. The existing accumulator tests cover normal final-to-next-partial ordering but not a repeated older final after a newer partial.

hsusul:
## Summary

- preserve the active Parakeet/Nemotron partial when a duplicate final arrives for an older segment
- continue clearing a partial when its own matching final arrives
- cover both message orderings with focused accumulator regressions

## Problem

The online accumulator deduplicates finalized messages by segment ID, but cleared `partialText` after every final message. When an older final was repeated after a newer segment had begun producing partial text, the newer live transcript disappeared from the preview and accumulated stream result.

## Root cause

The accumulator tracked the partial text but not the segment that owned it, so it could not distinguish a matching final from a stale final for another segment.

## Solution

Track the current partial's segment ID and clear the partial only when a final belongs to that segment. Finals without a segment ID retain the previous conservative behavior and clear the partial.

Fixes #1282

## Validation

- `node --test test/helpers/parakeetWsResult.test.js` — 8 passed
- `ELECTRON_OVERRIDE_DIST_PATH=/tmp npm test` — 662 passed, 16 skipped, 0 failed
- `npm run lint` — passed
- `npx prettier --check src/helpers/parakeetWsResult.js test/helpers/parakeetWsResult.test.js` — passed
- `git diff --check` — passed

`npm run format:check` still reports pre-existing formatting differences in `src/locales/zh-CN/translation.json` and `src/locales/zh-TW/translation.json`; neither file is changed by this PR.

## Risk

Low. The change is confined to the online websocket result accumulator. Existing final-segment deduplication and fallback handling are unchanged, and focused tests cover both stale and matching final messages.

### DIFF
--- a/src/helpers/parakeetWsResult.js
+++ b/src/helpers/parakeetWsResult.js
@@ -13,6 +13,7 @@ function createOnlineAccumulator() {
   const finalizedSegments = new Set();
   let finalizedText = "";
   let partialText = "";
+  let partialSegment = null;
   let fallbackKey = 0;
 
   const text = () =>
@@ -33,6 +34,7 @@ function createOnlineAccumulator() {
 
       if (!parsed.is_final) {
         partialText = finalizedSegments.has(parsed.segment) ? "" : messageText;
+        partialSegment = parsed.segment ?? null;
         return text();
       }
 
@@ -41,7 +43,10 @@ function createOnlineAccumulator() {
         finalizedSegments.add(segment);
         finalizedText = finalizedText ? `${finalizedText} ${messageText}` : messageText;
       }
-      partialText = "";
+      if (partialSegment === null || partialSegment === segment) {
+        partialText = "";
+        partialSegment = null;
+      }
       return text();
     },
     text,
--- a/test/helpers/parakeetWsResult.test.js
+++ b/test/helpers/parakeetWsResult.test.js
@@ -19,6 +19,27 @@ test("online sherpa messages keep finalized segments and latest partial", () => 
   );
 });
 
+test("online sherpa duplicate final keeps a newer segment partial", () => {
+  assert.equal(
+    parseOnlineMessages([
+      JSON.stringify({ text: "hello", is_final: true, segment: 0 }),
+      JSON.stringify({ text: "new partial", is_final: false, segment: 1 }),
+      JSON.stringify({ text: "hello", is_final: true, segment: 0 }),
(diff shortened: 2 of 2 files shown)


gabrielste1n (approved):
Thanks @hsusul for the excellent contribution! Really clean fix — tracking the partial's owning segment is exactly the minimal state needed to tell a matching final from a stale duplicate, and the focused regression tests covering both message orderings are much appreciated. Verified end to end locally: full suite passing with lint/typecheck clean, and the streaming preview + flush paths in parakeetWsServer both pick up the fix. Great write-up in the PR description too.

hsusul:
Thanks for the thorough review and for testing it end to end! Glad the fix worked as intended.

Issue #742, opened 2026-05-08
Feature Proposal: Eliminate ~350ms recording start latency via AudioWorklet Pre-Roll & Renderer Bypass


DavidGP:
Hey OpenWhispr team, first off, massive thanks for OpenWhispr! It has completely replaced other dictation tools for me due to its privacy-first nature and the incredible local/BYOK flexibility.

I use it heavily for dictation all day long, but noticed a consistent friction point: there is a perceptible delay (causing swallowed first syllables or words) between hitting the hotkey and the actual recording start, as well as a slight overhead when using ultra-fast cloud APIs like Groq.

After benchmarking the pipeline, the initial delay seems to cost around ~345ms, broken down into:

~250ms: MediaRecorder API cold start / OS hardware negotiation.

~50ms: Opus/WebM encoder lookahead buffer.

~30ms: React state batching overhead after the windows-key-listener:key-down IPC event is received.

I've successfully bypassed these limitations in my own implementations and wanted to propose a two-phase architectural tweak to make OpenWhispr feel absolutely instantaneous ("zero-latency").

1. Phase 1: Continuous AudioWorklet Pre-Roll (Fixing Swallowed Words)
Instead of instantiating the MediaRecorder on-demand in src/hooks/useAudioRecording.js upon receiving the IPC key-down event, we could transition to a continuous AudioWorkletNode implementation.

Keep the OS microphone stream "warm" (with an opt-in toggle in settings for privacy transparency).

The AudioWorklet continuously overwrites a lightweight 500ms circular buffer (Float32Array) in RAM.

When the hotkey is pressed, the worklet retroactively flushes the 500ms pre-roll buffer into the recording stream. Users can start speaking a fraction of a second before bottoming out the key, and zero audio is ever lost. The effective start latency drops to 0ms.

2. Phase 2: Direct Renderer Fetch for Cloud STT (Bypassing Backend IPC & Disk I/O)
Currently, the audio blob is converted to an ArrayBuffer, sent via IPC, written to a temp .wav file, and then processed by whisper.cpp or a cloud API via audioManager.js.

While necessary for local inference, this disk I/O and IPC memory copying is a massive bottleneck for Cloud STT.

[...]

PR #1284, opened 2026-07-22
fix(audio): re-warm mic driver after idle instead of once per session

### LINKED ISSUE #845: The first 1 to 2 seconds of audio are not recorded and therefore are not transcribed.
**Describe the bug**

With my Jabra speakerphone, I’m having the issue that whenever I use this microphone, the first one or two seconds are not recorded, so the beginning of what I say gets cut off.

A colleague of mine has the same phenomenon with his headset and also with his other microphones.

Is this a known bug, or do we have a local issue with our computer or microphones?

**To Reproduce**
1. Press the hotkey to start dictation and begin speaking immediately.
2. Stop the recording.
3. Check the audio file. The first two seconds are missing from the audio file.

**Expected behavior**
Actually, the entire audio track should be recorded and audible in the audio file.

**Video**

What I said: 1,2,3,4,5,6,7,8,9,10
Recorderd audio: 4,5,6,7,8,9,10

Audio: https://github.com/user-attachments/assets/22f2812a-8704-405f-828e-4cf337898434

Video: https://github.com/user-attachments/assets/2c24dea3-6c8d-4418-8be9-b0cb3f427b35

You can clearly see the OpenWhispr icon, but as long as it is still light blue and not yet dark blue, no voice can be heard for 1–2 seconds, even though the audio file is being recorded.

**Log**

These are the log entries from another test.

[log.txt](https://github.com/user-attachments/files/28184173/log.txt)

 - OS: Windows 11
 - Version 1.7.2

xAlcahest:
## Summary

The mic warm-up was a one-shot: `micDriverWarmedUp` got latched true by whichever of the two warm-up sites in `audioManager.js` ran first, and the only reset was the `devicechange` handler. An idle pause never fires that event, so after the first dictation the warm-up was a permanent no-op and a driver that had gone back to sleep was never re-warmed, which matches the report: clipped on the first dictation, fine back-to-back, clipped again after a pause. The fix turns the latch into a timestamp with a 5s TTL (`isMicWarm` in the new `micWarmState.js`) and routes both sites through one shared `_warmMicDriverOnce()` so they can't disable each other. The warm-up `getUserMedia` is capped by a timeout (a late stream still gets released), the batch path now re-warms after each transcription like the streaming path already did, and the hotkey call sites stay fire-and-forget, so no latency is added and the mic is never held open longer than the existing brief acquire-and-release.

Verified by reading and by the unit tests, not by reproducing on Windows (I don't have a Windows machine), so a confirmation from a Windows user before merge would be worth having. Worth saying too: this makes the clip much less likely but won't fully cover someone who starts speaking the exact instant they hit the hotkey after a long idle; closing that gap needs a pre-roll buffer along the lines of #742. The cause is shared across platforms (the warm-up was originally written for macOS cold starts and the code has no platform branches), so the fix isn't gated to Windows.

Fixes #845

## Changes

- src/helpers/micWarmState.js: new pure helper, `MIC_WARM_TTL_MS` and `isMicWarm(lastWarmedAt, now, ttlMs)`, warm state as a timestamp with expiry instead of a permanent flag.
- src/helpers/audioManager.js: both warm-up sites call a shared `_warmMicDriverOnce()` gated on the TTL; `_micWarmedAt` is initialized in the constructor and reset on `devicechange`; the warm-up `getUserMedia` is capped by a timeout and a late-resolving stream is still stopped.
- src/hooks/useAudioRecording.js: the batch path re-warms the driver after a transcription completes, mirroring what the streaming path already did.
- test/helpers/micWarmState.test.js: unit tests for the TTL logic, including expiry at the boundary and a clock that jumps backwards.

### DIFF
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -15,6 +15,7 @@ import {
   recordLocalSpeechWindow,
 } from "./localSpeechGate";
 import { reacquireIfDead } from "./micTrackHealth";
+import { isMicWarm, MIC_WARM_TTL_MS } from "./micWarmState";
 import { ActiveMicRecoveryController } from "./activeMicRecovery";
 import { followsSystemDefaultMic, reconcileSavedMicSelection } from "./micSelectionRecovery";
 import { isStaleDeviceError } from "./staleMicDevice";
@@ -299,7 +300,7 @@ class AudioManager {
     this._onDeviceChange = () => {
       this.cachedMicDeviceId = null;
       this.validatedSelectedMicDeviceId = null;
-      this.micDriverWarmedUp = false;
+      this._micWarmedAt = 0;
       this.rejectedMicDeviceId = null;
     };
     navigator.mediaDevices?.addEventListener?.("devicechange", this._onDeviceChange);
@@ -322,6 +323,7 @@ class AudioManager {
     this.cachedMicDeviceId = null;
     this.validatedSelectedMicDeviceId = null;
     this.rejectedMicDeviceId = null;
+    this._micWarmedAt = 0;
     this.persistentAudioContext = null;
     this.workletModuleLoaded = false;
     this.workletBlobUrl = null;
@@ -688,21 +690,40 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
     }
   }
 
+  // Warm the mic driver at most once per TTL. Shared by the batch and streaming
+  // sites so they can't drift or latch each other permanently off.
+  async _warmMicDriverOnce(logCategory) {
+    if (isMicWarm(this._micWarmedAt, Date.now())) return;
+    try {
+      const constraints = await this.getAudioConstraints();
+      const streamPromise = navigator.mediaDevices.getUserMedia(constraints);
+      // Stop the mic whenever it resolves, so a stream that arrives after the
+      // timeout is still released instead of leaking open.
+      streamPromise
+        .then((stream) => stream.getTracks().forEach((track) => track.stop()))
+        .catch(() => {});
+      let timer = null;
+      const timeout = new Promise((_, reject) => {
+        timer = setTimeout(() => reject(new Error("mic warmup timed out")), MIC_WARM_TTL_MS);
+      });
+      try {
+        await Promise.race([streamPromise, timeout]);
+      } finally {
+        if (timer !== null) clearTimeout(timer);
+      }
+      this._micWarmedAt = Date.now();
+      logger.debug("Microphone driver pre-warmed", {}, logCategory);
+    } catch (e) {
+      logger.debug("Mic driver warmup failed (non-critical)", { error: e.message }, logCategory);
+    }
+  }
+
   // Briefly acquire and release the mic so the OS audio driver is warm before
   // the first real recording, reducing cold-start empty captures. See #871.
   async warmupMicDriver() {
-    if (this.micDriverWarmedUp) return;
     // Skip while a recording is active so we don't double-acquire the mic. See #871.
     if (this.isRecording || this.isProcessing || this.mediaRecorder?.state === "recording") return;
-    try {
-      const constraints = await this.getAudioConstraints();
-      const tempStream = await navigator.mediaDevices.getUserMedia(constraints);
-      tempStream.getTracks().forEach((track) => track.stop());
-      this.micDriverWarmedUp = true;
-      logger.debug("Microphone driver pre-warmed", {}, "audio");
-    } catch (e) {
-      logger.debug("Mic driver warmup failed (non-critical)", { error: e.message }, "audio");
-    }
+    await this._warmMicDriverOnce("audio");
   }
 
   // Recovers a dead/muted capture: retries the same device, then hops to the OS default,
@@ -3253,21 +3274,7 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
         // Warm up the OS audio driver by briefly acquiring the mic, then releasing.
         // This forces macOS to initialize the audio subsystem so subsequent
         // getUserMedia calls resolve in ~100-200ms instead of ~500-1000ms.
-        if (!this.micDriverWarmedUp) {
-          try {
-            const constraints = await this.getAudioConstraints();
-            const tempStream = await navigator.mediaDevices.getUserMedia(constraints);
-            tempStream.getTracks().forEach((track) => track.stop());
-            this.micDriverWarmedUp = true;
-            logger.debug("Microphone driver pre-warmed", {}, "streaming");
-          } catch (e) {
-            logger.debug(
-              "Mic driver warmup failed (non-critical)",
-              { error: e.message },
-              "streaming"
-            );
-          }
-        }
+        await this._warmMicDriverOnce("streaming");
 
         logger.info(
           "Streaming connection warmed up",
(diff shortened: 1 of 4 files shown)


xlurie:
**Windows test report.** Ran the branch today on Windows 11 (commit `4896475b`, dev build from source; mic pinned to a Voicemeeter virtual device — a fast-opening driver, caveat below), plus the same protocol on the base commit (`f908526`) as a control, with `getUserMedia` timing instrumentation on both. Idle gaps tested on the branch: first-of-session, <5 s, ~30 s, ~10 min; on the base: ~60 s and ~30 s. Fixed test phrases, transcript checked for the first word each time.

**The mechanism works as designed on Windows.** On the branch the warm-up fired again after every past-TTL gap (~30 s and ~10 min), and the <5 s case hit the warm path (`driver still warm, ageMs: 707`; recording `getUserMedia` resolved in **36 ms** vs **54–58 ms** cold). On the base, the one-shot latch behaved exactly as diagnosed in #845: after the session's first dictation, every later one — including after idle — logged only "latch already set" and cold-opened the device itself. No console errors, no first-word loss, no regressions on the branch in any case.

**Honest caveat: no audible clipping reproduced on either build in this protocol.** This virtual device cold-opens in ~35–72 ms — below hearing threshold — so a short-idle A/B can't produce an audible clip here regardless of build. The clipping I've experienced on this same machine (including the 11-day window below) came from real-world usage at longer idle scales. What this test confirms is functional correctness of the TTL re-warm on Windows (fires when it should, warm path works, nothing breaks), not a clipped-vs-clean repro.

One timing detail from the logs that feeds note 1 below: past the TTL, the un-awaited warm-up and the recording open the device **concurrently** (warm-up starts ~19 ms earlier, never resolves first; recording resolves at 57–58 ms) — while the base's single open resolves in 35–36 ms. So on a fast device the double open is measurably slightly slower than no warm-up at all. Harmless here; it matters on slow drivers.

Some production data that might help decide this without a from-scratch repro — I wrote the root-cause comment on #845 this PR builds on, and I've been running a local fix for the same bug on Windows since then.

**What the persistent-stream approach measures out at.** Since 2026-07-11 I've run a local patch implementing the "better option" from that comment: a persistent master `getUserMedia` stream opened on first acquire, `track.clone()` handed to each recording instead of reopening the device, master released after a configurable idle timeout (15 min here). Per-call logging gives **0.1–0.2 ms** to re-acquire, against **52.8 ms** for a fresh `getUserMedia` with the driver already warm, and several hundred ms — roughly 0.5–1 s in my baseline timings — after an idle gap. First-word clipping has been a non-issue in daily dictation since.

**A natural experiment, offered as an observation rather than a measurement.** Around 2026-07-23 an auto-update reinstalled the app over my patched build and silently removed the patch (my patch's problem, not the app's). I didn't notice for 11 days, so from 07-23 to 08-03 the app ran stock, with none of my mic changes. First-word clipping came back during exactly that window; I only connected the two in hindsight, when I found the missing patch on 08-03, reapplied it, and the clipping stopped again. Same machine, same mic, same driver: patch off → clipping, patch on before and after → no clipping. It's retrospective and user-perceived, not instrumented, so I'd read it as directional support that the cold-`getUserMedia`-per-recording root cause is the real one — not as a number.

**Notes from reading the branch.** All non-blocking — the timestamp is clearly right where the permanent latch was wrong, and I'd like to see this land. Only the first one affects whether the fix bites:

1. **The re-warm still races the recording it's meant to protect.** `handleToggle` (and `handleStart`) call `warmupMicDriver()` without `await` and go straight into `performStartRecording()`. Both then do their own `await getAudioConstraints()` before their own `getUserMedia`, so past the TTL there are two opens of the same device in flight with no ordering guarantee that the warm-up is first. That's the un-awaited placement I flagged in #845 rather than anything this PR introduces — but dropping the one-shot latch changes what it costs: before, a session did exactly one warm-up ever; now every past-TTL dictation issues a second concurrent capture of the same device. In today's runs the race is right there in the logs: the warm-up starts ~19 ms before the recording's own `getUserMedia` and never resolves first, so past the TTL the recording always cold-opens for itself anyway (and the concurrent double open measured slightly slower than the base's single open, 57–58 ms vs 35–36 ms). On my fast-opening virtual device that's harmless — but on the slow-opening drivers #845 is actually about, where a cold open costs hundreds of ms, the re-warm won't precede capture and the clip window survives. The window that's genuinely guaranteed warm is the <5 s one created by the new post-transcription re-warm. `await`ing the warm-up before capture, or re-warming off a timer / `devicechange` instead of at press time, would close that.

2. **`MIC_WARM_TTL_MS` does double duty as the warm-up's `getUserMedia` timeout.** 5 s is a reasonable warm window, but as an acquisition deadline it's shorter than the cold-open worst case this file documents elsewhere ("can take 10-15s on cold macOS mic driver", in the streaming start path). When the timeout wins, the late stream is still stopped correctly — but the warm timestamp is never stamped even though the driver did come up, so the slowest machines, the ones that need this most, are the ones that keep missing the warm state. Feels like two independent constants. (Separately: my 0.1–0.2 ms numbers are roughly the ceiling this design reaches with a much longer or unbounded TTL and `devicechange`/idle-release as the release signal instead of a fixed clock — a config knob might be enough, and it's a much smaller change than a persistent-stream rewrite.)

3. **The post-transcription re-warm re-opens the device.** It fires right after a recording whose stream was demonstrably warm, so it costs an extra acquisition and, on Windows, a visible blink of the OS mic indicator after every dictation; stamping the warm timestamp from the just-finished recording stream would buy the same window for free. Minor caveat in the same place: `onTranscriptionComplete` is invoked before `isProcessing` is cleared, and `warmupMicDriver()` returns early while `isProcessing` is true. With auto-paste or keep-in-clipboard enabled the callback awaits first, so the flag is false by the time the re-warm runs; with both disabled the callback runs straight through and the re-warm is silently skipped.

4. **No in-flight dedup.** The warm timestamp is only stamped after the race resolves, so two callers inside one acquisition window (press-time plus the post-transcription one, or the batch plus streaming sites) both see "not warm" and both open the device. Worth a mention mostly because I hit the mirror image of it in my own patch: the warm-up/`startRecording` race leaked a master stream until I made concurrent acquires join a single in-flight promise.

One coordination note, unrelated to the review: #1430 rewrites the same block from the push-to-talk side (it replaces `warmupMicDriver()` with a stream prepared during the hold guard), and neither PR references the other — so whichever lands first, the other needs a rebase rather than a merge.

Happy to share the patch or the measurement setup, and I can run more idle-gap variants on the branch if a particular one would help.

Environment: Windows 11, mic pinned to a Voicemeeter virtual device (self-hosted STT, not relevant to capture timing).

gabrielste1n:
Superseded by #1493, which is now merged as 28cb7f82.

Thanks @xAlcahest — your diagnosis was right and the fix that shipped is built on it. The one-shot `micDriverWarmedUp` latch is gone repo-wide (it never re-warmed a driver that had since gone cold), replaced by the timestamp-plus-TTL model from this PR:

- `micWarmState.js` holds a 5s warm TTL, with the warm-up acquisition deadline decoupled from it (15s) so the slowest machines still register a late-resolving open as warm.
- Warmth is stamped from streams that demonstrably existed — prepare, acquisition, and every release site — rather than from a warm-then-discard open.
- Stamping at stop replaced your post-transcription re-warm along with its extra device open, `isProcessing` no-op, and indicator blink.

Also thanks to @xlurie — the instrumented Windows testing on this PR and #1430 is what showed the raced warm-up never resolved before the recording's own open, and directly shaped the corrections in #1493.

Credit is recorded in the #1493 description. Closing as superseded — the work landed.

PR #1285, opened 2026-07-22
fix(parakeet): fall back to full decode when the streaming commit loses audio

### LINKED ISSUE #857: [Bug] Transcription incomplete on first attempt, Retranscribe of same audio is perfect
**Describe the bug**

When recording a voice input, the first transcription is frequently incomplete:
sentences are cut off, chunks in the middle are missing, or the end is truncated.
Clicking \"Retranscribe\" on the same recording (no model change, no settings change)
produces a perfect, complete transcription every time.

This rules out an audio recording problem — the audio file itself must be complete,
since Retranscribe works on the exact same file. The issue occurs on at least 50%
of recordings.

The likely root cause is a race condition where transcription starts before the
audio file is fully flushed/closed to disk. On retry, the file is complete and
the transcription succeeds.

**To Reproduce**
1. Record a voice input of several sentences using a local model
2. Observe that the transcription is incomplete (missing words, sentences in the middle, or end cut off)
3. Click \"Retranscribe\" — without changing model or any settings
4. Observe that the transcription is now complete and correct

**Expected behavior**
The first transcription should be as complete and accurate as the Retranscribe result.

**Additional context**
- Local model only (no cloud)
- Happens on approximately 50%+ of recordings, not every time
- Retry (Retranscribe) consistently fixes it without any changes
- Likely root cause: transcription triggered before audio file write is fully flushed to disk
- Related: v1.6.2 had a fix for "initial audio being dropped during realtime streaming" — this may be a regression or a related but distinct code path
- Related: Issue #845 (first 1–2 seconds missing from audio file recording)

**Desktop**
- OS: Linux (Fedora, KDE Plasma, Wayland)
- Version: 1.7.2

xAlcahest:
## Summary

First transcriptions could lose sentences in the middle or at the end while Retranscribe of the same recording was always perfect. The reported flush-to-disk race is not the mechanism: the first run transcribes the in-memory blob, and the file Retranscribe reads is written from that same blob after transcription. The real hole is the streaming commit for online-runtime Parakeet models, where the live websocket transcript is committed as the final text and could lose audio without ever tripping `truncated`: a `ws.send` error was only logged, the trailing worklet chunk was gated by a 150ms timeout treated as success in `cleanupPreview`, and the accumulator in `parakeetWsResult.js` dropped refined finals through first-wins dedup by segment id, which is exactly what produces gaps in the middle. Retranscribe always re-decodes the complete file offline, so the retry was always right.

The fix makes every loss vector set `truncated` (send errors, chunks arriving after `finish()`, a flush watchdog that marks failure instead of faking success), switches the accumulator to latest-wins per segment, and lets anything untrusted fall back to the same full decode Retranscribe uses. Empty streamed text also falls through to the decode instead of throwing. The clean path gains no latency.

Fixes #857

## Changes

- src/helpers/parakeetWsServer.js: `sendFloat32` flags `truncated` on send errors and on chunks arriving after `finish()`, and the pending-chunk flush on open got the same error handling.
- src/helpers/parakeetWsResult.js: `createOnlineAccumulator` keeps the latest text per finalized segment id in arrival order instead of discarding refinements.
- src/helpers/audioManager.js: `cleanupPreview` awaits the worklet "flushed" sentinel with a 1s failure-only watchdog and passes the outcome to `stopDictationPreview`. Empty `streamedText` in `processWithLocalParakeet` now falls back to the offline blob decode.
- src/helpers/ipcHandlers.js: `stop-dictation-preview` trusts the streamed transcript only when both the server flush and the renderer flush were clean.
- src/types/electron.ts: optional `flushed` flag on `stopDictationPreview`.
- test/helpers/parakeetOnlineStream.test.js, test/helpers/parakeetWsResult.test.js: cover the three loss vectors and the new accumulator semantics. The loss tests fail on the code before this change.

Not verified live: reproducing needs an online-runtime nemotron model, which I don't have downloaded. Verified through the new unit tests against the mock streaming server and a full suite run. On #845: investigated and independent. That issue is about capture (the file itself missing the first seconds), while here the file is complete and only the streamed transcript comes up short. If they shared a root cause, Retranscribe would come out incomplete too. One more first-run/retry divergence worth a follow-up on the whisper path: the first run passes the custom dictionary as `initialPrompt` while retry-transcription doesn't, so dictionary-heavy setups can still see differences there.


gabrielste1n (approved):
Reviewed end to end: traced the streaming commit, the flush handshake, and the fallback decode path, and verified the new loss-vector tests fail on the pre-PR source exactly as described. Full suite, lint, and typecheck pass. Clean-path latency is unchanged and every distrust trigger converges on the same full-blob decode Retranscribe uses. Great work tracking down the real mechanism here.

Two small follow-ups worth a look later (non-blocking): skipping the flush wait on the dismiss/cancel path (the result is discarded there), and suppressing the untrusted streamed text in the preview overlay while the fallback decode runs so preview and pasted text can't diverge.

PR #1286, opened 2026-07-22
fix(widget): panel-aware overlay placement on Linux X11

### LINKED ISSUE #1110: Linux/X11: dictation overlay mispositioned on multi-monitor mixed-DPI setups
## Environment

- Linux, KDE Plasma (Wayland session); app runs under XWayland (self-relaunches with `--ozone-platform=x11`)
- 2 monitors, each with an always-visible bottom panel (KDE default, reserves space):
  - 4K @170% scale: `DP-2`, primary, right
  - 1080p: `HDMI-A-1`, left
- X11 exposes a single `_NET_WORKAREA` for the whole virtual screen (measured `0, 0, 7104, 1758` in device px)

## Expected

The dictation overlay should rest ~4px above the bottom panel of whichever monitor it is on (bottom-right corner), the way it does on Windows.

## Actual (two separate bugs)

`screen.getAllDisplays()` as the app sees it (Electron under X11, values in DIP, scaleFactor ≈ 1.70):

| Display | bounds (DIP) | workArea (DIP) |
|---|---|---|
| 4K (`DP-2`, primary) | `x1916 y0 2256×1269` | `x1916 y0 2256×`**`1033`** |
| 1080p (`HDMI-A-1`) | `x-1 y0 1918×1079` | `x-1 y0 1918×`**`1079`** |

**Bug 1: overlay sits inside the panel, not above it.**
For the 1080p, `workArea == bounds` (both 1079 tall): the monitor's own bottom panel is not excluded from its work area. The overlay is placed ~4px above the screen edge, overlapping the panel/tray, instead of ~4px above the panel.

**Bug 2: overlay floats on the taller monitor.**
For the 4K, `workArea` is 1033 tall vs `bounds` 1269, i.e. ~236 DIP (~402px) is cut from the bottom. That inset is not a panel on the 4K; it is the 1080p's bottom-panel strut (measured from the tall virtual-screen bottom) bleeding into the 4K via the single `_NET_WORKAREA`. The overlay lands ~402px above the 4K's true bottom, floating in mid-screen.

Root cause: on X11 each display's `workArea` is derived from the single global `_NET_WORKAREA`, and comes out wrong in both directions. The 1080p's own panel is not excluded (Bug 1), while the 4K is cut by the 1080p's panel that does not apply to it (Bug 2). Windows computes per-monitor `rcWork` correctly and shows neither bug.

## Screenshots

**Bug 1 (1080p):** overlay inside the bottom-right panel, should be above it:

Image

**Bug 2 (4K):** overlay floating ~402px above the bottom:

Image

xAlcahest:
## Summary

On Linux the widget position comes from `display.workArea`, and on X11 Chromium builds every work area from the single root `_NET_WORKAREA`: at most one display gets clipped by it (sometimes with another monitor's panel strut) and the rest keep `workArea == bounds`. On a multi-monitor KDE setup that puts the widget on top of the panel on one screen and floating ~400px above the true bottom on the other. KWin doesn't publish per-monitor work areas on X11 and Chromium wouldn't read them anyway, so this can't be fixed upstream. The new `src/helpers/effectiveWorkArea.js` asks plasmashell over D-Bus (`evaluateScript`, via the `@homebridge/dbus-native` dependency already in the tree) for per-screen geometry and panel struts, rebuilds each display's real work area (scale-aware, so the reporter's 170%+100% mixed-DPI layout lands right), and hands it to the existing positioning math through the #977 resolver. Results are cached (10s TTL, invalidated on display changes) with a 400ms timeout, and any failure falls back to current behavior; on Windows/macOS the helper returns the display untouched.

Stacked on #977 and based on its branch, so that PR must merge first.

Verified live on a two-monitor KDE Plasma 6 machine: Electron under XWayland reproduces the bug (the secondary display reports `workArea == bounds` with its panels ignored) and the Plasma query returns the true struts in a few ms. The unit tests encode the reporter's exact numbers. I haven't yet watched the widget land on both monitors in the running app; worst case of the fallback path is exactly today's placement.

Fixes #1110

## Changes

- src/helpers/effectiveWorkArea.js: new helper that queries Plasma for per-screen panel struts and rebuilds per-display work areas; pure parse/match/compute core plus a cached D-Bus layer with timeout and graceful fallback.
- src/helpers/windowManager.js: wraps the resolved target display with `resolveEffectiveDisplay` at the widget positioning chokepoints (creation, reposition, corner change, resize, transcription preview) and inits the helper.
- src/helpers/dragManager.js: drag clamping uses the corrected work area.
- test/helpers/effectiveWorkArea.test.js: covers the reporter's mixed-DPI numbers, the reference machine's layout, matching across scale differences, and malformed replies.


src/helpers/effectiveWorkArea.js
```diff
@@ -0,0 +1,395 @@
...
+// Return a display whose workArea reflects real KDE panel struts, or the SAME object when we
+// can't improve it. On Windows/macOS (correct per-monitor workArea natively) this is identity.
+function resolveEffectiveDisplay(display) {
+  if (!isKDE()) return display;
+  if (!display || !isValidBounds(display.bounds)) return display;
+
+  maybeRefresh();
+  const corrected = cache.byDisplayId.get(display.id);
+  if (!corrected) return display;
+  return { ...display, workArea: { ...corrected } };
```
gabrielste1n: **`workAreaSize` goes stale — low**

This spreads the display and overrides only `workArea`, so `workAreaSize` still describes the uncorrected area. No consumer today (I grepped), but it's a trap for the next caller — one line to set it alongside `workArea`.

src/helpers/effectiveWorkArea.js
```diff
@@ -0,0 +1,395 @@
...
+
+// Return a display whose workArea reflects real KDE panel struts, or the SAME object when we
+// can't improve it. On Windows/macOS (correct per-monitor workArea natively) this is identity.
+function resolveEffectiveDisplay(display) {
+  if (!isKDE()) return display;
+  if (!display || !isValidBounds(display.bounds)) return display;
+
+  maybeRefresh();
+  const corrected = cache.byDisplayId.get(display.id);
+  if (!corrected) return display;
```
gabrielste1n, line 427: **The correction can never land, and nothing re-applies it — medium/high**

`resolveEffectiveDisplay` is synchronous over an async cache. `init()` kicks a prefetch at `WindowManager` construction (`main.js:390`) and `createMainWindow()` runs at `main.js:869` — in production only `environmentManager.init()` and `migrateCookieToBearerToken()` yield in between, so it usually wins, but it's a race with no fallback.

The part that makes it stick: `_moveWidgetToDisplay` returns early when `currentDisplay.id === targetDisplay.id`, so `showDictationPanel` → `_repositionToCursorDisplay()` will never fix a bad `y` on the *same* display. If the snapshot loses the race, the widget sits on the panel for the entire session — no later event recovers it.

Same gap on hot-plug: `invalidate` (line 367) clears the cache without re-fetching, so the reposition that fires on the display change runs against an empty or pre-change cache.

Cheap fix: when `refreshSnapshot` populates a cache that differs from the previous one, call back into the widget positioning path, and give that path a force flag so it re-applies position even when the display id is unchanged.

src/helpers/effectiveWorkArea.js
```diff
@@ -0,0 +1,395 @@
...
+
+const isFiniteNumber = (value) => typeof value === "number" && Number.isFinite(value);
+
+const isValidBounds = (bounds) =>
+  bounds !== null &&
+  typeof bounds === "object" &&
+  isFiniteNumber(bounds.x) &&
+  isFiniteNumber(bounds.y) &&
+  isFiniteNumber(bounds.width) &&
+  isFiniteNumber(bounds.height);
```
gabrielste1n: **Duplicated verbatim — medium**

`isFiniteNumber` and `isValidBounds` here are a byte-for-byte copy of `displaySelection.js:6-14`. Both files are new in this stack and `displaySelection.js` is already Electron-free CommonJS, so importing them is free and keeps the unit tests running bare.

src/helpers/effectiveWorkArea.js
```diff
@@ -0,0 +1,395 @@
...
+// ---- Impure layer: synchronous, best-effort, stale-while-revalidate ------------------
+
+let cache = { byDisplayId: new Map() };
+let lastAttemptAt = 0;
+let fetchInFlight = false;
+let initialized = false;
+
+function isKDE() {
+  if (process.platform !== "linux") return false;
+  return (process.env.XDG_CURRENT_DESKTOP || "").toLowerCase().includes("kde");
```
gabrielste1n: **Duplicates an existing helper — medium**

`KDEShortcutManager.isKDE()` at `kdeShortcut.js:110` is identical logic (`XDG_CURRENT_DESKTOP` lowercased, `.includes("kde")`). Reuse it rather than carrying a second copy.

src/helpers/effectiveWorkArea.js
```diff
@@ -0,0 +1,395 @@
...
+}
+
+function logDebug(message, meta) {
+  try {
+    require("./debugLogger").debug(message, meta, "effectiveWorkArea");
```
gabrielste1n, line 257: **Minor — explain the lazy require**

The lazy `require("./debugLogger")` deviates from every other helper's top-level import. It's necessary here — `debugLogger` requires `electron` and the unit tests load this module bare — but without one comment line saying so it reads as defensive noise and someone will "fix" it into a top-level import and break the tests.

src/helpers/windowManager.js
```diff
@@ -46,18 +53,45 @@ class WindowManager {
     this._floatingIconAutoHide = false;
     this._agentAnimationState = null;
     this._panelStartPosition = "bottom-right";
+    this._panelDisplay = AUTO_DISPLAY;
+    this._onDisplayRemoved = null;
     this._isDictatingToggle = false;
     this._pendingMeetingNoteNavigation = null;
 
+    this._registerDisplayListeners();
+    effectiveWorkArea.init();
```
gabrielste1n, line 72: **Listener ordering — part of the hot-plug gap**

`effectiveWorkArea.init()` registers its `display-added`/`display-removed`/`display-metrics-changed` invalidation listeners *after* `this._registerDisplayListeners()` on the line above. Electron fires them in registration order, so on a display change windowManager's `_onDisplayRemoved` → `_repositionToCursorDisplay()` runs **first**, using work areas computed for the pre-change layout, and only then is the cache wiped. Registering the invalidation before the reposition listener (or re-applying after the refresh lands — see the comment on `resolveEffectiveDisplay`) fixes the ordering.

gabrielste1n (commented):
## Code quality review

Reviewed this PR's own commit (`b4b2fbb5`); the diff vs `main` also carries #977's commit, noted where the two interact.

**Verification done locally:** all 43 tests pass (`node --test` on `effectiveWorkArea.test.js`, `displaySelection.test.js`, `windowPosition.test.js`), files are Prettier-clean against the repo config, `debugLogger.debug(message, meta, scope)` matches usage, and `effectiveWorkArea.init()` is reached from the `WindowManager` constructor at `main.js:390` — after `app.whenReady()` — so the `screen` usage is safe.

### What's good

The pure/impure split is the right shape: parse → match → compute are Electron-free and directly testable, and the tests encode real reported geometry rather than synthetic numbers. Failure handling is genuinely thorough — the single `finish` guard in `fetchPlasmaSnapshot` closes the bus exactly once on reply/error/timeout, `timer.unref()` is there, the `connection.on("error")` listener matches the pattern already used in `gnomeShortcut.js:152` and `kdeShortcut.js:150`, and every path degrades to today's behavior. Returning `null` on an ambiguous match instead of guessing is the correct call.

### Findings that don't fit inline (unchanged lines)

**Same bug left unfixed on the sibling overlays — medium.** The agent overlay clamps to raw `display.workArea` at `windowManager.js` ~1080 and ~1124, the meeting and update notification windows position off `screen.getPrimaryDisplay()`, and `snapControlPanelToMeetingMode` uses primary `workArea`. On the reporter's machine those all land under the KDE panel for the same reason. Fine to defer, but the PR reads as "panel-aware overlay placement" and covers only the widget and preview.

**`getNotificationPosition` still floors at 0 — low** (`windowConfig.js:182-184`). The stacked commit fixed `Math.max(0, …)` → `Math.max(workArea.x, …)` in `getMainWindowPosition` for negative-origin monitors but left `getNotificationPosition` clamping to 0, which snaps it to primary.

**Wayland gating — worth confirming, not necessarily changing.** `isKDE()` doesn't look at `XDG_SESSION_TYPE`, so the override also runs on native-Wayland KDE where Chromium already reports correct per-display work areas. Gating on session type would be wrong (XWayland reports `wayland` and is exactly where you verified the bug), so leaving it ungated is defensible — but it'd be good to confirm it's a no-op under native Wayland rather than silently recomputing.

**Minor.** `org.kde.PlasmaShell.evaluateScript` can be disabled by kiosk restrictions on locked-down setups; the fallback covers it, just worth a note so it isn't debugged twice. Also, CLAUDE.md's Helper Modules list isn't updated with the new file.

### Bottom line

The approach is sound and the implementation is careful — Plasma's `panels()`/`screenGeometry()` over D-Bus is the standard workaround for X11's single `_NET_WORKAREA`, and the fallback discipline means the worst case is today's behavior. The startup/hot-plug race (inline comment on `resolveEffectiveDisplay`) is the one I'd want resolved before merge: as written, the fix depends on a race it can lose permanently, and the note in the description ("haven't yet watched the widget land on both monitors in the running app") is exactly the gap that would surface it. The two DRY findings are quick cleanups.

src/helpers/effectiveWorkArea.js
xAlcahest: Fixed: resolveEffectiveDisplay now overrides workAreaSize alongside workArea.

src/helpers/effectiveWorkArea.js
xAlcahest: Fixed along the lines you suggested. effectiveWorkArea now fires a corrections-changed callback whenever a fresh snapshot differs from the last one it reported, and windowManager re-applies placement via `_repositionToCursorDisplay({ force: true })`, with `force` skipping the same-display early return. `invalidate` refetches immediately instead of only clearing, and `refreshSnapshot` re-queues a fetch when an invalidation lands mid-flight, so hot-plug converges too. The diff is computed against the last notified map rather than the emptied cache, so a metrics-changed event that doesn't actually move panels won't reset a user-dragged widget, and an in-progress drag blocks the re-apply.

One deliberate leftover: a drag completed in the sub-500ms window before the first snapshot lands still gets overwritten by the first correction. Tracking whether the user has ever dragged felt like more state than that window is worth, happy to add it if you disagree.

src/helpers/effectiveWorkArea.js
xAlcahest: Done, both are now imported from displaySelection.js (added them to its exports).

src/helpers/effectiveWorkArea.js
xAlcahest: Reused, with one twist: a top-level require of kdeShortcut would drag debugLogger and therefore electron into the bare unit-test runs, so isKDE() lazy-requires it behind the platform check, with a comment.

src/helpers/effectiveWorkArea.js
xAlcahest: Comment added, plus the same note on the new lazy kdeShortcut require in isKDE().

src/helpers/windowManager.js
xAlcahest: Swapped: effectiveWorkArea.init() now registers its invalidation listeners before windowManager's reposition handler. And since invalidate now refetches immediately, the corrections callback covers whatever window remains.

xAlcahest:
All inline items are addressed in e1363ca5 (2a083bc7 is a prettier-only pass, see below). On the body findings:

- getNotificationPosition now floors at workArea.x/y like getMainWindowPosition does, with a negative-origin test.
- Sibling overlays (agent overlay clamp, meeting and update notification windows, snapControlPanelToMeetingMode): deferring to a follow-up as suggested, so this PR stays widget plus preview.
- Native Wayland: the override does recompute there, but it converges instead of diverging. Chromium's display layout under Wayland comes from wl_output/xdg-output, which is the same layout Plasma reports, so matchDisplayToPlasmaScreen matches and bounds minus panel insets lands on the work area Chromium already reports (within a pixel at fractional scales). If the match ever fails, the display is returned untouched. Worst case is a redundant recompute, not a wrong one. I can verify live under native Wayland if you want empirical confirmation.
- Kiosk restrictions: noted in a comment on the evaluateScript error path.
- CLAUDE.md: effectiveWorkArea.js added to Helper Modules.

About the prettier commit: a handful of lines in the original commit were over the repo's printWidth 100 (prettier --check flags both files in-repo; CI doesn't run prettier, so it never surfaced). Kept it separate so the review diff stays clean.

PR #1288, opened 2026-07-22
fix(settings): commit model selection only on explicit click, not tab browse (#1287)

### LINKED ISSUE #1287: Provider tab switches silently overwrites the persisted model selection in AI models
**Describe the bug**
Switching provider tabs in the AI model picker silently overwrites your persisted model selection with that provider's recommended model. The tab bar is treated as a commit surface when it should be a browse surface.

**To Reproduce**
1. Settings → AI Models → Local. Select Qwen → Qwen3.5 2B (not the recommended 9B).
2. Click the Liquid AI tab to browse LFM2 models. Do not click any card.
3. Check the Active tag switched to LFM2.5 1.2B.
4. Click back to Qwen.
5. Check the Active tag switched to Qwen3.5 9B (the recommended model).

**Expected behavior**
Qwen3.5 2B is still selected. Browsing Liquid AI was a read action — nothing should have changed.

**Actual behavior**
- The Active tag jumped to Liquid AI's recommended model (LFM2.5 1.2B) on step 2 — your explicit Qwen3.5 2B choice was destroyed.
- On step 4, Qwen3.5 9B (recommended) is now selected instead of the 2B you chose — browsing away and back silently pushed you to the default.
Same thing happens within a single provider: select LFM2.5 8B MoE (not the recommended 1.2B) → browse to Llama → browse back to Liquid AI. The Active tag now points to LFM2.5 1.2B. Your explicit non-default choice was replaced by the recommended model.

**Screenshots**

https://github.com/user-attachments/assets/1c4732e8-f312-4acc-a43f-39c786d7f082

- OS: Linux
- Version 1.7.6

**Additional context**
Root cause: `ReasoningModelSelector's` tab-change handlers persist both the provider and its recommended model on every switch. This was an intentional design choice to ensure no provider tab is ever visited without a model selected — but with 8+ cloud and 5+ local providers, the silent loss of explicit user choices to stray browsing is the worse UX outcome.

IdrisGit:
## Summary
Provider tabs in the model pickers are now pure browsing state: the (provider, model) pair commits atomically only on an explicit model click, so switching tabs can never silently overwrite the active model. Closes #1287.

Three fixup commits on top of 90cfe033 on `fix-model-override-in-settings`:

- `983bf58f` fix(settings): harden browse-only provider tabs and drop dead code
- `43cc03a5` test(settings): pin the explicit-click commit sequences for model selection
- `5f268417` test(settings): note the mid-browse policy-flip precondition in the disallowed-browse pin

## Root cause
The base approach (tabs browse, only a model click commits) is correct, but the original head shipped with a ~150-line commented-out cloud panel, a neutered `onCloudProviderSelect` prop still wired by one caller, a race in the local download bootstrap that could commit the browsed tab's provider with another provider's model, a dead Custom STT tab while browsing, prettier drift failing CI, and no tests pinning the new behavior.

## Fix
- **ReasoningModelSelector.tsx**: deleted the commented-out old cloud panel; removed the unused `onCloudProviderSelect` prop and its `_onCloudProviderSelect` plumbing; deleted the superseded `handleLocalModelSelect` and folded its registry lookup into `handleModelSelect` — local commits now derive the provider from `modelRegistry.getModel(modelId)?.provider.id ?? selectedLocalProvider`, so a download finishing while another tab is browsed can never commit a mismatched (provider, model) pair.
- **NotesOnboarding.tsx**: dropped the now-ignored `onCloudProviderSelect={switchCleanupProvider}` wiring and the unused callback.
- **TranscriptionModelPicker.tsx**: the `effectiveCloudSelection` fallback now echoes the browsed input when reconciliation returns null — previously browsing to the Custom tab (always reconciled as valid) never displayed it.
- Prettier on the 7 PR-touched files; `npm run typecheck` verified locally (CI never reached it).

## Tests
- `transcriptionModelMemory.test.js`: pins the picker's click-commit sequence — switch-then-set leaves the clicked model, remembers the outgoing provider's model, and never touches `cloudTranscriptionBaseUrl` (guards #1459/#1463).
- `reasoningModelMemory.test.js`: pins the atomic explicit-click commit — per-provider memory never overrides a click; a plain provider write doesn't drag the model along.
- `policyRules.test.js`: pins `reconcileCloudProviderSelection` over the browsed-provider input path (allowed browse resolves without leaking the committed model; custom stays null; disallowed browse falls back — a state only reachable when policy flips mid-browse, since `handleCloudProviderChange` guards `providerAllowed`).
- Full suite: 2291 tests, 0 failures.

## Limitations
- Component-level tab-click behavior is not reachable under the repo's static-markup test harness; the store/pure-function contracts the components rely on are pinned instead. A DOM-capable component harness would be the real regression net for this class of bug — worth a follow-up issue. Manual repro per #1287 confirmed the behavior: browsing tabs moves nothing, the Active tag stays put, only a model click commits.
- Deliberate trade: while a provider tab is being browsed, a policy-forced provider fallback is displayed but not persisted until the next explicit model click. The window is narrow and keeps tabs strictly commit-free.
- The tests job stays red on prettier until main's own pre-existing drift (PasteToolsInfo.tsx, WorkspacesService.ts, snippets.ts) is fixed on main — out of scope for this PR.
- Per-provider reasoning model memory is intentionally no longer engaged by tab browsing (the store function remains and is test-pinned).


Chadpiha (approved):
Thanks @IdrisGit — the direction here is right and it's a real simplification: browsing is now pure state, and the model click committing provider + model atomically kills the override bug at its root. Typecheck, lint, and the test suite are all green on the branch.

Review turned up three stale-state edge cases in the new code path, and I've pushed a commit on top of your branch that closes them :

1. The download bootstrap decided with click-time state. onSelectAfterDownload fires at download completion — minutes later for multi-GB models — but the callback closed over selectedModel/downloadedModels from the click. Start a download with nothing selected, pick another model while it runs, and the completion would steal your selection — the exact thing the comment above it rules out. The decision now reads current state through a ref.

2. On the Custom/OpenRouter tabs the provider commit could silently never happen. Since provider is only committed via the model click, a fetched list containing the already-persisted model id (LiteLLM/vLLM proxies expose upstream ids like gpt-4.1) rendered that card as Active, suppressed the "select a model" hint, and the user would configure base URL + key believing they're done — while requests kept routing to the previous provider. The panel now only shows a model as selected when that tab is the committed provider.

3. The mode tabs could still wipe a cloud selection. LocalModelPicker's validation effect cleared any selection missing from the local download set, so browsing Cloud → Local in the onboarding flow wiped a committed cloud model — breaking the "browsing never touches the persisted selection" contract this PR establishes. The wipe is now scoped to ids the picker actually owns; deleted-local-model clearing still works.

PR #1291, opened 2026-07-22
fix: stop cloud sync from wiping locally-edited notes with stale empty cloud copies

### LINKED ISSUE #1290: Cloud sync silently wipes locally-edited notes (meeting transcripts destroyed) — root cause + repro
## Summary

Cloud sync can **permanently destroy a note's local** `content`**,** `transcript`**, and** `enhanced_content`, reverting the title to the calendar-event summary. The user sees the note's tabs disappear and an empty note titled with the meeting name. We hit this three times in one morning on a real meeting note (72 KB transcript lost and forensically recovered from the SQLite WAL), traced the mechanism to exact lines, and verified it against both v1.7.6 (latest release) and current `main` — **it is unfixed on both**.

This is not an edge case of one machine: the code path runs for **every signed-in, subscribed user with cloud backup enabled**, on all platforms. Meeting notes are the acute victims because the app auto-creates an empty cloud shell for them.

**Expected behavior:** a locally-edited note (newer in real time, or unsynced) is never overwritten by a staler/emptier cloud copy; sync never replaces non-empty local content with empty remote content.

## Root cause (three interacting defects)

**1. The wipe engine — content-less "migration" push bumps the cloud timestamp without uploading content.**  <br>`src/services/SyncService.ts` `pushPendingNotes()` (~line 510): for any *pending* note that already has a `cloud_id`, the branch PATCHes **only** `{ client_note_id }` — no title/content/transcript — then calls `markNoteSynced`. Server-side, `updateNote` (`openwhispr-api lib/notes-service.ts:404`) stamps `updated_at = now()` while COALESCE-preserving the existing (stale/empty) content. The cloud row is now *genuinely newer but still empty*. The pull in the same `syncNotes()` pass then elects it and destroys the local edit. Content normally reaches the cloud only via the ~2 s debounced `pushNote` while the app is running — quit/relaunch before it fires and the edit is stranded forever.

**2. The amplifier — the pull compares timestamps lexically across two incompatible formats.**  <br>`src/services/SyncService.ts:586`:

```js
if (!local || cloudNote.updated_at > local.updated_at) { … upsertNoteFromCloud(cloudNote, …) }
```

Locally-edited rows carry SQLite `CURRENT_TIMESTAMP` format (`2026-07-22 08:47:00`, space at index 10); cloud rows are ISO (`2026-07-22T08:18:27.790Z`, `T` at index 10). Since `'T'` (0x54) > `' '` (0x20), **any same-UTC-day cloud timestamp beats any local one regardless of actual time** — a cloud copy 30 minutes *older* wins. The `normalizeTimestamp()` helper at the top of the same file (whose comment warns about exactly this) is used by the dictionary and snippet pulls but was never wired into the notes pull. The folders (~483), conversations (~692), and transcriptions pulls share the same raw compare.

**3. Winning is destructive — the pull upsert is a full-row overwrite.**  <br>`src/helpers/database.js` `upsertNoteFromCloud()` (~2699): `title/content/enhanced_content/transcript = excluded.*` unconditionally, while `participants`/`calendar_event_id`/`diarization_*` get `COALESCE(excluded.*, local)`. That asymmetry produces the exact observed corpse: content and transcript gone, calendar identity intact, title reverted to the calendar summary, and `updated_at = excluded.updated_at` stamping the cloud's ISO value into SQLite (the forensic fingerprint that identifies pull-writes).

Server-side hardening gap (separate repo, `openwhispr-api`): the `create`/`batch-create` upsert (`lib/notes-service.ts` ~327–353, `api/notes/batch-create.ts` ~124–150) is `ON CONFLICT … DO UPDATE SET col = EXCLUDED.col` with no COALESCE and no stale-write guard — any omitted column is NULLed — and personal-note `updated_at` is stored verbatim from the client. These semantics date back to Mar–May 2026.

## Real-world kill chain (observed + WAL-verified, times UTC)

| When | Event |
| -- | -- |
| 19:31 | Calendar meeting starts → app creates note shell (title = calendar summary, `content=''`) locally + in cloud |
| 20:36 | 72 KB transcript + generated title saved **locally**; app closed before the debounced push → never reaches cloud |
| 07:48:20 | Next launch: migration PATCH bumps the empty cloud row to `07:48:20.102Z`; same-pass pull elects it → **local transcript NULLed, title reverted** (identical `.102Z` stamp lands on both SQLite and Postgres) |
| 08:12 | User re-types the notes (pending) |
| 08:18:27 | Editor pushes an empty save (secondary trigger, see below); server stamps `.790Z`; same-pass pull copies it back down → **wiped again**, identical millisecond both sides; tabs disappear (they render only when transcript/enhanced are non-empty) |
| 08:57:19 | After a manual local restore (newer, pending): relaunch migration PATCH bumps cloud again *without uploading the restore* → pull overwrites the newer local copy with the older empty row → **wiped a third time** |

Secondary (unproven) trigger for the 08:18 empty save: rapid switching between the Transcript/Notes/Enhanced tabs; plausibly the shared TipTap instance remount emitting `onChange("")` — worth its own look but not required for the main data loss.

## Reproduction — EXECUTED, both levels green

**Unit (deterministic, no app code)** — the comparison at `SyncService.ts:586` is a plain string `>`. Ran 6/6 assertions:

```js
assert.equal("2026-07-22T08:18:00.000Z" > "2026-07-22 08:47:00", true) // PASS: 29-min-OLDER cloud copy judged newer
assert.equal("2026-07-22T00:00:00.000Z" > "2026-07-22 23:59:59", true) // PASS: a whole day staler, still wins
assert.equal("2026-07-21T23:59:59.000Z" > "2026-07-22 08:47:00", false)// PASS (control): earlier DATE loses correctly
// and: wrapping both sides in the file's own normalizeTimestamp() makes the older cloud copy correctly LOSE
```

**Integration (the app's real** `database.js` **from tag v1.7.6, real better-sqlite3, throwaway DB)** — electron stubbed, same pattern as `test/helpers/snippetsDatabase.test.js`. Executed with all assertions green:

1. `saveNote("Gabriel / Joshua", "", "meeting")` — the shell exactly as `meetingDetectionEngine.js:294` creates it
2. `updateNote(id, { title: "Real Title", content: "REAL MEETING NOTES", transcript: <70KB>, cloud_id })` → row is `pending`, `updated_at = "2026-07-22 12:38:06"` (SQLite format ✓)
3. Cloud copy: the still-empty shell stamped `"2026-07-22T00:00:01.000Z"` — **12.6 hours OLDER**
4. Line-586 gate: `"2026-07-22T00:00:01.000Z" > "2026-07-22 12:38:06"` → **true** (older empty shell elected)
5. `upsertNoteFromCloud(cloudNote)` → `content=''`**,** `transcript=NULL`**, title reverted to "Gabriel / Joshua"**, `participants`/`calendar_event_id` preserved (the COALESCE asymmetry), cloud ISO stamp copied into SQLite, row marked `synced` — loss permanent. Identical in shape to the production corpse.

**Manual E2E (stock install, test account, sync flags on — not yet run; the two executed repros + the production forensics cover the mechanism):** start a meeting note → force one sync so the empty shell reaches the cloud → type a long body → within the ~2.8 s save+push debounce window trigger a pull (focus-away/back or "Sync now", or briefly throttle the network so the push is delayed) → note comes back empty with the calendar title.

## Prior art on this tracker (same family)

* OpenWhispr/openwhispr#1088 — calendar-title reversion via auto-titling (maintainer-confirmed, patched in v1.7.5 for that trigger; this issue is the sync-side sibling)
* OpenWhispr/openwhispr#1275 (open) — markdown mirror deletes note files titled `…transcript`
* OpenWhispr/openwhispr#1188 (open) — cloud sync losing notes/transcriptions across devices
* [#499](<https://github.com/OpenWhispr/openwhispr/issues/499>) — lost meeting notes after failed transcription
* PR OpenWhispr/openwhispr#938 (merged, **not in any release yet**) — same bug *class* in dictation: an empty cloud cleanup response overwrote the real transcription; fixed with a one-line non-empty guard. The invariant it establishes — *never replace non-empty local data with an empty remote response* — is exactly what the notes pull is missing.

## Proposed fixes (in necessity order)

1. `pushPendingNotes` **migration branch**: push full note content for `cloud_id`-bearing pending notes (or don't PATCH/mark-synced at all). This alone stops the launch-time wipes.
2. `SyncService.ts:586`: wrap both sides in the existing `normalizeTimestamp()` (and consider the folders/conversations/transcriptions pulls that share the raw compare).
3. `upsertNoteFromCloud`: never overwrite non-empty local `content`/`transcript`/`enhanced_content` with empty cloud values (the OpenWhispr/openwhispr#938 invariant applied to sync).
4. Server (`openwhispr-api`, separate PR): COALESCE omitted columns in the create/batch upsert + a stale-write guard; consider not bumping `updated_at` on content-less PATCHes.
5. Ship a release containing OpenWhispr/openwhispr#938 — the dictation-wipe fix is merged but users (incl. v1.7.6) don't have it.

Happy to open the PRs with regression tests for 1–3 — the test designs above slot into the existing `node --test` suite.

## Environment

* Observed on macOS, v1.7.6 (latest release); defects verified byte-identical on current `main`
* Requires only: signed in + subscribed + cloud backup enabled (all platforms)

greatcoat:
Fixes OpenWhispr/openwhispr#1290

Cloud sync could permanently destroy a note's local `content`, `transcript`, and `enhanced_content` — observed three times in one morning on a real meeting note, root-caused and reproduced in OpenWhispr/openwhispr#1290. Three interacting defects, fixed here in necessity order (the issue's proposed fixes 1–3):

**1.** `pushPendingNotes` **migration branch now pushes the full note content** (`src/services/SyncService.ts`)<br>The branch PATCHed only `{ client_note_id }` for `cloud_id`-bearing pending notes, bumping the cloud row's `updated_at` without ever uploading the local edit — so the pull in the same `syncNotes()` pass elected the *genuinely-newer but still empty* cloud row and destroyed the local edit. It now sends the same full-content payload as the debounced editor push, via a shared `buildNoteUpdatePayload()` helper both call sites use, so the two shapes can't drift apart again. `client_note_id` deliberately stays out of the shared payload and is sent only by the migration branch: pre-`client_note_id`-era cloud notes carry a different backfilled UUID per device, and PATCHing it on every push would flip the cloud row's identity between devices and fork duplicate notes.

**2. Pull gates compare normalized timestamps** (`src/services/SyncService.ts`)<br>The notes/folders/conversations pulls compared SQLite-format local timestamps (`2026-07-22 08:47:00`) against ISO cloud timestamps (`2026-07-22T08:18:27.790Z`) with a raw lexical `>` — since `'T'` > `' '` at index 10, **any same-UTC-day cloud copy beat any local one regardless of actual time**. All three gates now use `isCloudEntryNewer()`, which wraps both sides in the pre-existing `normalizeTimestamp()` (moved unchanged to `src/helpers/cloudSyncGuards.js`; the dictionary/snippet pulls already used it).

**3.** `upsertNoteFromCloud` **never lets an empty cloud value destroy non-empty local content** (`src/helpers/database.js`)<br>The upsert set `content`/`enhanced_content`/`transcript` to `excluded.*` unconditionally. Those three columns now keep the local value when the incoming cloud value is empty/NULL and the local one isn't — the same invariant OpenWhispr/openwhispr#938 established for dictation ("never replace non-empty local data with an empty remote response"), applied to the notes pull. `enhancement_prompt`/`enhanced_at_content_hash` travel with `enhanced_content`, so a preserved enhancement keeps its staleness hash. Non-empty cloud values still overwrite (last-writer-wins is unchanged), and the insert path for brand-new cloud notes is untouched.

Any one of the three stops the observed kill chain; together they also cover the unproven secondary trigger (an empty editor save reaching the cloud).

### Regression tests

`test/helpers/cloudSyncGuards.test.js` + `test/helpers/noteCloudUpsertGuard.test.js` — direct conversions of the executed repro from OpenWhispr/openwhispr#1290 (unit 6/6 + integration on the real `database.js`), assertions flipped to the fixed behavior. The DB tests reuse the `snippetsDatabase.test.js` harness (real `better-sqlite3`, throwaway DB). All were watched failing on the unfixed code — the upsert test fails with exactly the production corpse (`content: '' vs 'REAL MEETING NOTES'`) — and pass with the fix.

**CI change (**`tests.yml`**):** `npm ci --ignore-scripts` skips `better-sqlite3`'s native build, so every DB-backed test in the suite (`snippetsDatabase`, these new ones, …) was silently skipping in CI — a revert of the `database.js` fix would have left CI green. Added `npm rebuild better-sqlite3` so they actually execute, and `npm run typecheck` (previously not run by any workflow) so the TS-side sync wiring is covered too. Verified with a matching-ABI binding locally: full suite 689 tests, **684 pass, 0 fail** (5 remaining skips are unrelated).

### Semantics notes for reviewers

* A 5-lens adversarial review ran over this diff; its two substantive findings (the `client_note_id` tug-of-war, the enhancement-triple split) are folded in above.
* When the fix-3 guard fires, the row still takes the cloud's metadata/stamp and stays `synced` — the preserved local content is not re-pushed until the user's next edit. That's deliberate: passively keeping the data avoids resurrection ping-pong against a device that intentionally cleared a note, and with fix 1 in place, rows still `pending` upload their content before the pull runs. The residual case (deliberate cross-device clears may diverge until the next edit) is the accepted cost of the invariant; the server-side stale-write guard (below) is the systemic answer.
* The guard keys on exactly-empty (`''`/NULL), matching OpenWhispr/openwhispr#938's invariant; the migration PATCH now also writes `folder_id` with the same mapping the debounced push has always used.
* The transcriptions pull needs no gate change — it's insert-only at HEAD and already skips empty `text`.
* Server-side hardening (COALESCE on omitted columns + stale-write guard in `openwhispr-api`, issue fix 4) is a separate PR in that repo. Shipping a release that contains OpenWhispr/openwhispr#938 (issue fix 5) also still stands.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

### DIFF
--- a/.github/workflows/build-and-notarize.yml
+++ b/.github/workflows/build-and-notarize.yml
@@ -25,6 +25,10 @@ jobs:
           node-version-file: .nvmrc
       - name: Install dependencies
         run: npm ci --ignore-scripts
+      - name: Build better-sqlite3 for the runner's node
+        # --ignore-scripts skips the native build; DB-backed tests silently
+        # skip without it. This job is the only test gate on push to main.
+        run: npm rebuild better-sqlite3
       - name: Run tests
         run: npm test
         env:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,6 +13,12 @@ jobs:
           node-version-file: .nvmrc
       - name: Install dependencies
         run: npm ci --ignore-scripts
+      - name: Build better-sqlite3 for the runner's node
+        # --ignore-scripts skips the native build, so every DB-backed test
+        # (snippetsDatabase, noteCloudUpsertGuard, …) silently skips without this.
+        run: npm rebuild better-sqlite3
+      - name: Typecheck
+        run: npm run typecheck
       - name: Run tests
         run: npm test
         env:
--- a/src/helpers/cloudSyncGuards.js
+++ b/src/helpers/cloudSyncGuards.js
@@ -1,0 +1,49 @@
+// Sync-pull ordering and push-payload helpers shared by SyncService and the
+// node --test suite (#1290).
+
+// SQLite `datetime('now')` yields "YYYY-MM-DD HH:MM:SS" (no T, no millis, no Z);
+// the cloud sends ISO 8601 "YYYY-MM-DDTHH:MM:SS.sssZ". Normalize both to
+// millis-precision ISO so the pull loop's lexical greater-than compares
+// correctly — without the ".000" pad a whole-second local value sorts after a
+// sub-second cloud value at the same instant ('Z' > '.').
+export function normalizeTimestamp(value) {
+  if (!value) return "";
+  const iso = value.replace(" ", "T").replace(/Z$/, "");
+  return (/\.\d+$/.test(iso) ? iso : `${iso}.000`) + "Z";
+}
+
+// The pull loops' last-writer-wins gate. Comparing the raw values instead
+// silently hands every same-UTC-day conflict to the cloud copy — 'T' (0x54)
+// sorts after ' ' (0x20) at index 10 — which is how a staler, still-empty
+// cloud shell overwrote fresh local meeting notes (#1290).
+export function isCloudEntryNewer(cloudUpdatedAt, localUpdatedAt) {
+  return normalizeTimestamp(cloudUpdatedAt) > normalizeTimestamp(localUpdatedAt);
+}
+
+// Full-content PATCH payload for an existing cloud note. Both the debounced
+// editor push and the pushPendingNotes migration branch must send this same
+// shape: a content-less PATCH still bumps the cloud row's updated_at, so the
+// next pull elects a copy that never received the local edit (#1290).
+// Deliberately NO client_note_id: pre-client_note_id-era cloud notes carry a
+// different backfilled UUID per device, so PATCHing it from every push would
+// flip the cloud row's identity between devices on each edit and fork
+// duplicate local notes. Only the one-shot migration branch sends it.
+export function buildNoteUpdatePayload(note, folderMap) {
+  return {
+    title: note.title,
+    content: note.content,
+    enhanced_content: note.enhanced_content,
+    enhancement_prompt: note.enhancement_prompt,
+    enhanced_at_content_hash: note.enhanced_at_content_hash,
+    note_type: note.note_type,
+    source_file: note.source_file,
+    audio_duration_seconds: note.audio_duration_seconds,
+    transcript: note.transcript,
+    participants: note.participants,
+    calendar_event_id: note.calendar_event_id,
+    diarization_enabled: note.diarization_enabled,
+    expected_speaker_count: note.expected_speaker_count,
+    folder_id: note.folder_id ? (folderMap.get(note.folder_id) ?? null) : null,
+    updated_at: note.updated_at,
+  };
+}
--- a/src/helpers/cloudSyncGuards.js
+++ b/src/helpers/cloudSyncGuards.js
@@ -1,5 +1,4 @@
-// Sync-pull ordering and push-payload helpers shared by SyncService and the
-// node --test suite (#1290).
+// Sync guards shared by SyncService and the node --test suite (#1290).
 
 // SQLite `datetime('now')` yields "YYYY-MM-DD HH:MM:SS" (no T, no millis, no Z);
 // the cloud sends ISO 8601 "YYYY-MM-DDTHH:MM:SS.sssZ". Normalize both to
@@ -12,22 +11,19 @@ export function normalizeTimestamp(value) {
   return (/\.\d+$/.test(iso) ? iso : `${iso}.000`) + "Z";
 }
 
-// The pull loops' last-writer-wins gate. Comparing the raw values instead
-// silently hands every same-UTC-day conflict to the cloud copy — 'T' (0x54)
-// sorts after ' ' (0x20) at index 10 — which is how a staler, still-empty
-// cloud shell overwrote fresh local meeting notes (#1290).
+// Last-writer-wins gate for the pull loops. A raw compare hands every
+// same-UTC-day conflict to the cloud copy ('T' > ' ' at index 10), which is
+// how a staler empty cloud shell overwrote fresh local notes (#1290).
 export function isCloudEntryNewer(cloudUpdatedAt, localUpdatedAt) {
   return normalizeTimestamp(cloudUpdatedAt) > normalizeTimestamp(localUpdatedAt);
 }
 
-// Full-content PATCH payload for an existing cloud note. Both the debounced
-// editor push and the pushPendingNotes migration branch must send this same
-// shape: a content-less PATCH still bumps the cloud row's updated_at, so the
-// next pull elects a copy that never received the local edit (#1290).
-// Deliberately NO client_note_id: pre-client_note_id-era cloud notes carry a
-// different backfilled UUID per device, so PATCHing it from every push would
-// flip the cloud row's identity between devices on each edit and fork
-// duplicate local notes. Only the one-shot migration branch sends it.
+// Full-content PATCH payload for an existing cloud note — a content-less
+// PATCH still bumps the cloud row's updated_at, so the next pull elects a
+// copy that never received the local edit (#1290). No client_note_id here:
+// legacy cloud notes carry a different backfilled UUID per device, so
+// PATCHing it on every push would fork duplicate notes; only the one-shot
+// migration branch sends it.
 export function buildNoteUpdatePayload(note, folderMap) {
   return {
     title: note.title,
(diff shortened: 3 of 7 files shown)


gabrielste1n (approved):
Reviewed end-to-end. Confirmed all three defects on main (content-less migration PATCH, lexical timestamp compare across SQLite/ISO formats in the notes/folders/conversations pulls, unconditional content overwrite in upsertNoteFromCloud). Verified the regression tests catch the bug: reverted database.js to main's version and the upsert guard test fails with exactly the reported corpse (content '' vs 'REAL MEETING NOTES'); all 13 new tests pass with the fix. Full suite green locally (689 tests, 684 pass, 0 fail) plus lint/typecheck/prettier. Pushed a follow-up adding the better-sqlite3 rebuild to build-and-notarize.yml's tests job — it's the only test gate on push to main, so DB-backed tests were still silently skipping there — and trimmed comments to the load-bearing invariants. Confirmed from CI logs that both test gates now execute the DB-backed tests (686/689 pass, 0 fail, no native-binding skips).

PR #1294, opened 2026-07-22
fix(export): roll SRT millisecond overflow into seconds

### LINKED ISSUE #1293: SRT export emits invalid millisecond values (1000) near second boundaries
## Affected component

`src/helpers/transcriptFormatter.js` — `formatSrtTimestamp()` used by `formatSrt()` for meeting transcript export.

## Reproduction

1. Export a transcript segment whose timestamp rounds to the next whole second, e.g. `0.9995` or `1.9995` seconds (common with Whisper-style fractional timestamps).
2. Inspect the generated SRT cue line.

```js
const { formatSrt } = require("./src/helpers/transcriptFormatter");
console.log(formatSrt([{ speaker: "speaker_0", timestamp: 0.9995, text: "Near boundary." }], {}));
// Actual:   00:00:00,1000 --> 00:00:03,1000
// Expected: 00:00:01,000 --> 00:00:04,000
```

## Actual behavior

`formatSrtTimestamp()` computes milliseconds with `Math.round((seconds - Math.floor(seconds)) * 1000)`, which can produce `1000` instead of rolling into the next second. SubRip/SRT requires milliseconds in the range `000–999`.

## Expected behavior

SRT timestamps should always use valid `HH:MM:SS,mmm` values with `mmm <= 999`, carrying overflow into seconds/minutes/hours as needed.

## Environment

- OpenWhispr `main` (tested on macOS, Node 24)
- Reproducible without credentials, models, or Electron runtime

## User impact

Exported SRT files can contain invalid timestamps that strict players/editors reject or misparse, causing cue sync issues for meeting transcripts.

## Root-cause hypothesis

Millisecond rounding is applied to the fractional part of the floored seconds instead of rounding total elapsed milliseconds and decomposing the result.

## Proposed scope

Fix `formatSrtTimestamp()` only, with deterministic regression tests in `test/helpers/transcriptFormatter.test.js`.

I'll work on a focused fix for this.

hsusul:
## Summary

Fix invalid SRT export timestamps that emit `,1000` millisecond values when segment timestamps round to the next whole second.

Fixes #1293

## Problem

`formatSrtTimestamp()` computed milliseconds from the fractional part of `Math.floor(seconds)`. Values such as `0.9995` and `1.9995` (common with Whisper-style fractional timestamps) rounded to `1000` ms instead of rolling into the next second.

Example before the fix:

```
00:00:00,1000 --> 00:00:03,1000
```

SubRip/SRT requires milliseconds in `000–999`.

## Root cause

Millisecond rounding was applied after flooring seconds, so `(seconds - floor(seconds)) * 1000` could equal `1000` without carrying into the seconds field.

## Fix

Round total elapsed milliseconds once, then decompose into hours/minutes/seconds/milliseconds. Overflow from rounding now rolls into seconds (and higher units when needed).

## Behavior

| Input timestamp | Before | After |
|---|---|---|
| `0.9995` | `00:00:00,1000` | `00:00:01,000` |
| `1.9995` | `00:00:01,1000` | `00:00:02,000` |

Merged same-speaker cues and existing SRT merge behavior are unchanged.

## Scope and non-goals

- In scope: `formatSrtTimestamp()` and regression tests
- Out of scope: out-of-order segment sorting, mic/system merge behavior, TXT/MD export ordering

## Tests added

- `SRT timestamps roll sub-second rounding into the next second`
- `SRT timestamps stay valid for merged cues near second boundaries`

## Validation

```bash
node --version   # v24.9.0
npm --version    # 11.6.0

node --test test/helpers/transcriptFormatter.test.js
# 7 pass, 0 fail

npm run typecheck
# pass

npm run lint
# pass

npm run format:check
# pre-existing failure: locales/zh-CN/translation.json, locales/zh-TW/translation.json (unchanged by this PR)

npm run i18n:check
# pass

npm run build:renderer
# pass

node --check src/helpers/transcriptFormatter.js
# pass

git diff --check
# pass

npm test
# 516 pass, 30 fail (pre-existing Electron install/runtime failures in unrelated tests)
```

## Environment limitations

Full `npm test` still fails in this environment for unrelated Electron-dependent tests (`Electron failed to install correctly`). The focused regression suite and repository checks above all pass for this change.


gabrielste1n (approved):
Approving — thanks @hsusul. Small, exactly right, and it was producing genuinely malformed files.

`00:00:00,1000` is out of spec — SubRip requires milliseconds in 000–999 — so strict parsers reject the cue outright. Computing milliseconds from the fraction *after* flooring seconds is the classic version of this bug, and rounding total elapsed milliseconds once and then decomposing is the standard fix. I verified the carry propagates properly through all four units, including the case where it cascades from milliseconds through seconds into minutes.

Worth noting the fix is broader than the reported symptom: any timestamp whose fractional part rounds to 1.000 hit this, which with Whisper-style fractional timestamps is not rare.

One thing I checked and am explicitly not asking you to change: negative `seconds` produces malformed output both before and after this change (`-500` milliseconds now, `-1:...` hours before). It's pre-existing, not reachable from any current caller, and out of scope here.

gabrielste1n:
Follow-up code-quality pass before merge (`0a825ee`): rewrote the decomposition to use the same all-`const` arithmetic as `formatTimestamp` directly above it, so the two helpers now share identical h/m/s lines instead of one reassigning a running total.

Also tightened the tests. The merged-cue test never actually formatted its second boundary value — the `1.9995` segment merged into the first cue, so only `0.9995` was exercised, duplicating the test above it. Split the speakers so both cues format and the carry is verified twice. Dropped the `doesNotMatch` guard from the single-cue test, where the anchored assertion already pins the millisecond field, and kept it on the multi-cue one where it still covers ground the anchor doesn't.

797 tests / 0 failures, typecheck and lint clean. Merging — thanks!

PR #1300, opened 2026-07-23
fix(permissions): prompt once so macOS lists the app under Accessibility

AIalliAI:
## Problem

macOS only lists an app under **Privacy & Security → Accessibility** once that app has *asked* for the permission. Every check OpenWhispr makes passes `false` to `isTrustedAccessibilityClient`, which queries without registering:

- `main.js:1472`, `main.js:1486`
- `src/helpers/clipboard.js:1957`

So a fresh install never appears in the list at all. `requestAccessibilityPermission` then sends the user straight to System Settings (`src/hooks/usePermissions.ts:255`), which in that state is a dead end — there is no row to enable, and the pane gives no hint that the app has to ask first. The user is told to grant a permission they cannot find.

The IPC that *would* register it already exists and is fully plumbed:

- handler: `src/helpers/ipcHandlers.js:1926` — the one call site that passes `true`
- bridge: `preload.js:237` — `promptAccessibilityPermission`
- type: `src/types/electron.ts:835`

It is never called from the renderer. Dead code.

## Change

Call it once before falling back to System Settings.

The existing comment at `usePermissions.ts:253` is right that the macOS TCC dialog cannot be dismissed programmatically, so this is **not** a revert of that decision — the prompt is one-shot, persisted as `hasPromptedAccessibility` in localStorage. One prompt creates the row; every attempt after that goes straight to System Settings exactly as today. If the user happens to grant it at the prompt, the flow short-circuits and never opens Settings at all.

Decision logic lives in `src/helpers/accessibilityPromptPolicy.js` as a pure function so it is unit-testable, matching the `dockPolicy.js` pattern described in `CLAUDE.md`.

## Note for contributors

In development this registers as **"Electron"**, not "OpenWhispr" — the dev binary is `node_modules/electron/dist/Electron.app`, ad-hoc signed with `Identifier=Electron` and no Team ID. That surprised me while debugging, and is worth knowing when testing this change. Packaged builds register under the product name.

## Testing

- `test/helpers/accessibilityPromptPolicy.test.js` — 4 cases: never-asked prompts, one-shot suppression, already-granted, and non-darwin platforms.
- Full suite: **699 passing, 0 failing, 19 skipped** (`node --test "test/**/*.test.js"`).
- `tsc --noEmit` clean; `eslint` clean.
- Ran the app on macOS 27 (arm64, Node 24.18.0). Clean start with the change applied: no hook-order or ErrorBoundary errors, app reaches `qdrant started successfully` and the usual accessibility notice.
- Confirmed the underlying mechanism separately: invoking `isTrustedAccessibilityClient(true)` against the same ad-hoc `Electron.app` bundle does trigger the macOS dialog and create the Accessibility row, where `(false)` alone never does.

**What I could not automate:** clicking through the TCC dialog itself and confirming the row appears via the real Settings button is a manual step. The policy function, the wiring, and the underlying API behaviour are each verified; the human click-through is not.

### DIFF
--- a/src/helpers/accessibilityPromptPolicy.js
+++ b/src/helpers/accessibilityPromptPolicy.js
@@ -1,0 +1,21 @@
+// macOS Accessibility registration policy.
+//
+// macOS only lists an app under Privacy & Security > Accessibility once that
+// app has *asked* for the permission. Every check OpenWhispr makes passes
+// `false` to isTrustedAccessibilityClient, which queries without registering,
+// so an install that has never prompted does not appear in the list at all.
+// Sending that user to System Settings is a dead end: there is no row to
+// enable, and the pane gives no hint that the app must ask first.
+//
+// The prompt is deliberately one-shot. macOS shows a modal TCC dialog that
+// cannot be suppressed once triggered, so firing it on every attempt would be
+// worse than the dead end it fixes. One prompt is enough — it creates the row,
+// after which System Settings is actionable forever.
+
+// Whether to trigger the registering TCC prompt before opening System Settings.
+// Returns false off macOS, where there is no Accessibility permission to grant.
+export function shouldPromptForAccessibility({ platform, alreadyGranted, hasPromptedBefore }) {
+  if (platform !== "darwin") return false;
+  if (alreadyGranted) return false;
+  return !hasPromptedBefore;
+}
--- a/src/hooks/usePermissions.ts
+++ b/src/hooks/usePermissions.ts
@@ -3,6 +3,7 @@ import type { TFunction } from "i18next";
 import { useTranslation } from "react-i18next";
 import type { PasteToolsResult } from "../types/electron";
 import { useLocalStorage } from "./useLocalStorage";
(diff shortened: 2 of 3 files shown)


AIalliAI:
Closing this — I could not substantiate the premise, and on retest it is wrong.

The claim was that calling `isTrustedAccessibilityClient(true)` registers the app so a row appears under Privacy & Security > Accessibility. Retesting on macOS 27 (arm64) with the one-shot guard cleared and the renderer reloaded, so the prompt branch was genuinely armed, clicking Grant Access produced no TCC dialog and no new row — identical to the existing behaviour of going straight to System Settings.

My earlier check was invalid: `hasPromptedAccessibility` was already `true` from a prior click, so my own guard skipped the prompt. The unit tests, typecheck and lint all passed and verified none of the actual behavioural claim.

Sorry for the noise. The underlying UX problem is real — a dev install never appears in the Accessibility list, and the troubleshooting copy tells users to toggle or re-add a row that isn't there — but this is not the fix, so it should not sit open.

PR #1302, opened 2026-07-23
fix(i18n): cover the case where the app is missing from the Accessibility list

AIalliAI:
## Problem

The accessibility troubleshooting hint (`onboarding.permissions.accessibilityTroubleshooting`, shown by `PermissionsSection.tsx:52` after ~10s of failed permission polls) currently reads:

> In System Settings → Accessibility, switch OpenWhispr's toggle off and back on. If that doesn't work, remove OpenWhispr from the list and add it back.

Both suggestions assume OpenWhispr already has a row under Privacy & Security → Accessibility.

It often doesn't. macOS only creates that row once an app has *asked* for the permission, and OpenWhispr never asks — every check passes `false` to `isTrustedAccessibilityClient`, which queries without registering:

- `main.js:1472`, `main.js:1486`
- `src/helpers/clipboard.js:1957`

(`prompt-accessibility-permission` in `ipcHandlers.js:1926` is the only call that passes `true`, and nothing in the renderer invokes it.)

So the install that most needs this hint is precisely the one with no row to toggle and nothing to remove. The advice sends the user looking for a control that isn't there.

## Change

Point those users at the **+** button, which is the only thing that works when the app isn't listed:

> In System Settings → Accessibility, switch OpenWhispr's toggle off and back on. If OpenWhispr isn't in the list at all, click + and add it from your Applications folder. Still stuck? Email support@openwhispr.com.

Copy-only, updated in all 10 locales. No behaviour change.

## How I ran into this

Setting up a dev environment on macOS 27, auto-paste and the Globe hotkey didn't work, and OpenWhispr was nowhere in the Accessibility list. The in-app hint told me to toggle it off and on — but there was no row. I lost a while to that before realising the app had never registered itself.

Worth noting for anyone testing: a dev build registers as **"Electron"**, not "OpenWhispr", because `node_modules/electron/dist/Electron.app` is ad-hoc signed with `Identifier=Electron` and no Team ID. Packaged builds use the product name, which is why the copy says "OpenWhispr".

I first tried fixing this by calling the existing prompt IPC to force registration. On macOS 27 that does **not** create a row for an ad-hoc-signed binary — no dialog, no entry — so I abandoned that approach and closed it. This copy change is the part I could actually verify.

## Testing

- All 10 locale files parse; key counts unchanged (`en` 2181 keys before and after, and every locale matches its previous count).
- Full suite: **695 passing, 0 failing, 19 skipped** (`node --test "test/**/*.test.js"`).
- `tsc --noEmit` clean.
- Rendered the new string in the running app on macOS 27 (arm64, Node 24.18.0) and confirmed the troubleshooting hint displays the updated text in the onboarding Permissions step.

Non-English strings are my own translations, mirroring the sentence structure and tone of the existing entries — happy to defer to a native speaker on any of them.


gabrielste1n (approved):
thanks!

PR #1304, opened 2026-07-23
fix(dictionary): prefer SQLite over stale cache on startup

### LINKED ISSUE #1295: Custom dictionary entries added directly via the SQLite custom_dictionary table are wiped on app launch
Our team syncs contact names from an external pipeline directly into custom_dictionary via SQLite INSERT, since we couldn't find an API/CLI/import format for bulk-adding entries. Every insert succeeds at write time, but the entries are gone the next time OpenWhispr is launched — the table drops back to just the default OpenWhispr row.

custom_dictionary's schema has cloud_id/sync_status columns (sync_status defaults to 'pending'), suggesting app launch reconciles the table against a cloud account and discards rows it doesn't recognize as already-synced, rather than uploading them. Confirmed by direct DB inspection: inserted 357 rows externally, watched MAX(id) jump to 8678 (proving many prior insert-then-wipe cycles), then reopened the app and the table was back to 1 row within moments.

Is there a supported way to bulk-import dictionary entries (CLI, API, or an "Import a list" format spec) that survives the cloud reconciliation? The in-app "Import a list" button looks like the right path but I haven't found format docs for it.

hsusul:
## Summary

Startup dictionary reconciliation treated a non-empty localStorage cache as
current whenever SQLite was also non-empty. SQLite is the durable store, while
the cache can lag.

`ensureAgentNameInDictionary()` could then call
`setCustomDictionary` → `setDictionary`, replacing the entire SQLite table and
deleting words missing from the stale cache.

## Fix

- Prefer SQLite on startup whenever it has entries
- Skip agent-name dictionary writes when nothing changed
- Skip agent-name ensure if dictionary synchronization failed
- Backfill missing `client_dict_id` values before pending cloud pushes
- Clarify that dictionary imports accept commas or new lines
- Share the import parser between the UI and tests

## Tests

- Startup chooses SQLite when the cache is behind
- Rewriting the adopted database snapshot does not drop words
- Cloud upserts preserve local-only pending rows and avoid duplicates
- Intentional deletions still tombstone or hard-delete correctly
- Imports accept comma-separated, newline-separated, and mixed input

## Validation

- `npm run format`
- `npm run format:check`
- `npm run lint`
- `npm run typecheck`
- `ELECTRON_OVERRIDE_DIST_PATH=/tmp npm test` — 722 passed, 6 skipped
- `npm run i18n:check`
- Targeted dictionary tests — 14 passed

Fixes #1295


gabrielste1n (approved):
Thanks for this — and especially for digging past the reported symptom. The issue's own theory pointed at cloud reconciliation, and it would have been easy to go chase that. The actual path (stale localStorage cache -> `ensureAgentNameInDictionary` -> unconditional `setCustomDictionary` -> full-table replace) is the right diagnosis, and it explains the reporter's jumped-past-8600 row counter exactly.

Verified locally: merges cleanly onto main, typecheck passes, all checks green. The startup policy now matches what snippets already does a few lines down in the same function, which is the right call.

Two things I especially liked:
- The `client_dict_id` backfill isn't scope creep. Externally-inserted rows have a NULL there, and `pushPendingDictionary` keys its response map on that column — several NULLs collide, nothing gets marked synced, and every sync re-creates server rows. That's the second bug waiting behind the first.
- Guarding the agent-name write on `changed` is the load-bearing fix. Good instinct.

Merging as-is. Two follow-ups I'm picking up rather than blocking on:
1. There's no test anywhere for `agentName.ts`, so the `changed` guard — the bit that actually stops the wipe — is unprotected. I'll add one.
2. `getPendingDictionary` doing a write is a contract smell. I'd rather the schema assign `client_dict_id` itself via a trigger so external inserts get one for free, and then this backfill can go away.

Also worth noting for anyone reading later: the deeper issue is that `setDictionary(words[])` is a replace-everything API, so any caller with a partial view can destroy the rest. This PR closes both known triggers; I'll look at retiring the pattern separately.

PR #1310, opened 2026-07-23
fix(api): join paths before query strings on pasted base URLs

### LINKED ISSUE #1309: normalizeBaseUrl/buildApiUrl place paths after query strings on pasted endpoints
## Bug Description

`normalizeBaseUrl`, `buildApiUrl`, and `ensureV1Suffix` treat the entire URL string (including `?...` / `#...`) as a path. Pasted OpenAI-compatible endpoints that include a query string (common with Azure `api-version`, gateways, or auth query params) are not suffix-stripped, and later path joins append after the query — producing invalid URLs.

**Affected component:** `src/config/constants.ts` (`normalizeBaseUrl`, `buildApiUrl`, `ensureV1Suffix`, `getModelListBaseCandidates`)

## To Reproduce

Deterministic (no Electron / API keys / models):

```js
import { normalizeBaseUrl, buildApiUrl, ensureV1Suffix } from "./src/config/constants.ts";

const pasted = "https://api.example.com/v1/chat/completions?api-version=2024-02-01";
console.log(normalizeBaseUrl(pasted));
// actual:   https://api.example.com/v1/chat/completions?api-version=2024-02-01
// expected: https://api.example.com/v1?api-version=2024-02-01

console.log(buildApiUrl(pasted, "/responses"));
// actual:   https://api.example.com/v1/chat/completions?api-version=2024-02-01/responses
// expected: https://api.example.com/v1/responses?api-version=2024-02-01

console.log(ensureV1Suffix("https://gateway.example.com/v1?api-key=secret"));
// actual:   https://gateway.example.com/v1?api-key=secret/v1
// expected: https://gateway.example.com/v1?api-key=secret
```

## Expected behavior

- Path-suffix stripping (`/chat/completions`, `/models`, etc.) applies to the path only.
- Query string and hash are preserved.
- `buildApiUrl` inserts the new path **before** the query/hash.
- `ensureV1Suffix` / `getModelListBaseCandidates` decide `/v1` from the path portion, not the full string.

## Actual behavior

Suffix match fails when a query is present (`$` anchor never sees `/chat/completions`), and `${base}${path}` concatenates after `?…`, so callers request nonsense URLs and get 404/400s.

## Environment

- OS: macOS (repro is pure Node)
- OpenWhispr: `main` @ `ab201b39`
- Node.js 24

## User impact

Anyone who pastes a provider docs URL or Azure/gateway endpoint that includes `?api-version=…` (or similar) into a custom / OpenAI-compatible / LAN base URL field gets broken model listing and inference URL construction.

## Root-cause hypothesis

Helpers were written assuming path-only bases. Query/hash were never split out before `$`-anchored suffix regexes or string concatenation.

## Proposed scope

Small fix in `src/config/constants.ts` plus regression tests in `test/helpers/apiBaseUrl.test.js` / `modelListBaseCandidates.test.js`. No provider API changes.

I intend to work on this.

hsusul:
## Summary

- Pasted OpenAI-compatible base URLs that include a query string (e.g. Azure `?api-version=…` or gateway auth params) no longer break path-suffix stripping or path joining.
- `normalizeBaseUrl`, `buildApiUrl`, `ensureV1Suffix`, and `getModelListBaseCandidates` now operate on the path while preserving `?query` / `#hash`.

Fixes #1309

## Problem

Users often paste a full provider docs URL such as:

`https://api.example.com/v1/chat/completions?api-version=2024-02-01`

The helpers treated the entire string (including the query) as a path. `$`-anchored suffix regexes never matched, and `buildApiUrl` concatenated after the query:

`…/chat/completions?api-version=2024-02-01/responses`

`ensureV1Suffix` also failed to recognize an existing `/v1` when a query was present, appending another `/v1` after the query.

## Root cause

Path-only assumptions in `src/config/constants.ts`: no split of query/hash before suffix stripping or string joins.

## Fix

Split `?` / `#` decorators, normalize/join on the path, then reattach decorators. Same for `/v1` detection and `/api/v0|/api/v1` sibling derivation in `getModelListBaseCandidates`.

## Behavior before / after

| Input | Before | After |
| --- | --- | --- |
| `normalizeBaseUrl(…/v1/chat/completions?api-version=1)` | unchanged full URL | `…/v1?api-version=1` |
| `buildApiUrl(that, "/responses")` | `…?api-version=1/responses` | `…/v1/responses?api-version=1` |
| `ensureV1Suffix(…/v1?api-key=secret)` | `…/v1?api-key=secret/v1` | unchanged |
| `getModelListBaseCandidates(…/api/v1?token=abc)` | bogus `…?token=abc/v1` sibling | `…/api/v1?token=abc`, `…/v1?token=abc` |

Query-free bases behave as before.

## Scope / non-goals

- In scope: URL helper correctness for query/hash on pasted bases.
- Out of scope: Azure-specific deployment URL builders, provider routing refactors, UI copy, dependency upgrades.

## Tests

- `test/helpers/apiBaseUrl.test.js` — query/hash suffix strip, path-before-query joins, `ensureV1Suffix` path check
- `test/helpers/modelListBaseCandidates.test.js` — query-preserving `/v1` and `/api/v1` candidates

## Validation

```text
ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/apiBaseUrl.test.js test/helpers/modelListBaseCandidates.test.js
→ 18 pass

ELECTRON_OVERRIDE_DIST_PATH=/tmp npm test
→ 699 pass, 19 skipped, 0 fail

npm run typecheck  → pass
npm run lint       → pass
npm run i18n:check → pass
npm run build:renderer → pass
git diff --check   → pass
```

`npm run format:check` still warns on `src/locales/zh-CN/translation.json` and `src/locales/zh-TW/translation.json` on clean `upstream/main` (pre-existing; unchanged by this PR). Our touched files pass Prettier.

## Platform / environment

Pure Node unit tests; no Electron, native binaries, API keys, or model downloads required.


gabrielste1n (approved):
Approving with a follow-up commit — thanks @hsusul. The core fix is right and the bug was real: `$`-anchored suffix regexes never matched past a query string, so `buildApiUrl` concatenated after it and produced `…?api-version=2024-02-01/responses`. Splitting `?`/`#`, normalising on the path and reattaching is the correct shape.

**The regression it introduced.** Making the suffix regexes match again has a side effect on Azure. Given the endpoint form the Azure portal hands out:

`https://r.openai.azure.com/openai/deployments/my-deploy/audio/transcriptions?api-version=2024-06-01`

`normalizeBaseUrl` now strips `/audio/transcriptions`, so `buildAzureTranscriptionUrl` no longer matches its "user pasted a full endpoint, respect it as-is" branch. It falls through to the rebuild branch, which discards the path and substitutes the **transcription model setting** as the deployment name:

```
main:      …/deployments/my-deploy/audio/transcriptions?api-version=2024-06-01   (correct)
this PR:   …/deployments/whisper-1/audio/transcriptions?api-version=2024-06-01   (DeploymentNotFound)
```

Before this PR the trailing query accidentally shielded the suffix from stripping — which is precisely what the PR removes. `azureTranscriptionUrl.test.js` only exercises `buildAzureTranscriptionUrl` in isolation with a raw URL, so the composed normalize→build path had no coverage and the suite stayed green.

**What I pushed** (`a81cd11`): `getTranscriptionEndpoint` now builds the Azure URL from the raw base rather than the normalized one, so a deployment the user pinned survives. That also repairs the query-less form, which was *already* broken on main for the same reason — the "respected as-is" contract documented in `urlUtils.ts` was only ever being honoured by accident. Bare resource bases are byte-identical either way, so nothing else moves. Added tests covering the composed path in both directions.

Also folded in the panel hint (`OpenAICompatiblePanel.tsx:282`), which was still built by raw concatenation. Not pre-existing: on main the hint and the request were the same broken string, whereas this PR fixes the request and left the hint showing a URL that's never sent. It now goes through `buildApiUrl`, which the file already imports.

Verified against current main: 726 tests / 0 failures, typecheck, lint and prettier clean.

gabrielste1n:
Follow-up code-quality pass before merge (`2a20307`): `splitUrlDecorators` and `joinUrlDecorators` are exact inverses, so rebuilding the native-API candidate reconstructed the `normalized` string already in scope — now returned directly, which also collapses the array to one line. Removed the `.replace(/\/+$/, "")` added in `buildApiUrl`: `normalizeBaseUrl` already strips trailing slashes from the path before rejoining, so it could never match. And collapsed the `ensureV1Suffix` branches, which differed only in the value handed to the same call.

801 tests / 0 failures, typecheck and lint clean. Merging — thanks again, and sorry for the two rounds of follow-ups here.

PR #1312, opened 2026-07-24
fix(models): clear stale local model selections after model removal

ambbone:
Summary
This fixes stale local model selections that remained in settings after a local model was deleted, causing invalid references in inference scope configuration.

Problem
When a local model was removed, one or more inference scopes could still point to that deleted model ID. This left settings in an inconsistent state and could cause failures or fallback confusion later.

Root cause
Model deletion removed the model artifact but did not fully reconcile all persisted scope-level model references.

What changed

Cleared deleted local model IDs from relevant scope selections at deletion time.
Added startup reconciliation to reset invalid local model references if previously persisted state points to a missing model.
Kept fallback behavior intact and non-destructive.
Validation

Removed a selected local model and confirmed stale references are cleared immediately.
Restarted app with previously stale config and confirmed settings reconcile on startup.
Existing valid model selections are preserved.
Risk and compatibility

Low risk.
Behavior is additive and defensive; only invalid references are touched.


gabrielste1n (approved):
Thanks for this — good catch, and the right diagnosis. A scope left pointing at a deleted model fails at inference time with an error the user has no way to act on, and pairing the deletion-time fix with a startup reconcile is exactly the correct shape: it also covers models removed outside the app.

I pushed a follow-up commit rather than sending you round again:

- **Reconciles off `INFERENCE_SCOPES`** instead of the hand-written five-scope chains. That registry already describes exactly these five scopes, and `reconcileSelectedModels` in `src/models/tinfoilModels.ts` is the same job done generically — worth knowing about for next time. A sixth scope is now covered automatically.
- **Only clears when the delete succeeded.** `model-delete` reports failure by resolving with `{ success: false }` rather than throwing, so a failed unlink left the model on disk but wiped every selection pointing at it (and still toasted "Model deleted"). The whisper and parakeet branches just above already guard on `result.success`.
- **Covers Developer → Remove models**, which deletes every local model at once and left the same stale references behind until the next launch.
- **Pins the local provider for note formatting.** This one is subtle and worth flagging: `noteFormatting` has `fallbackScope: "dictationCleanup"`, so an empty model there doesn't mean "unset" — it resolves to `cleanupModel`. With that scope on a cloud provider, clearing the note-formatting selection would have routed note content to a cloud API while the user's mode still said "local".
- **Added unit tests** for the reconcile rule and the note-formatting routing, following the `dictationRouting.js` pattern.

Net result is smaller than the original and covers more paths. Appreciate you digging into this one.

PR #1314, opened 2026-07-24
fix(hotkeys): ignore native key auto-repeat on key-down to prevent duplicate toggles

ambbone:
Summary
Prevents duplicate/triplicate recording toggles caused by repeated native key-down events from auto-repeat behavior.

Problem
In tap-to-talk flows using native listeners, held keys could emit repeated key-down events, triggering multiple rapid toggles.

Root cause
Native key-down edges were not deduplicated before dispatching toggle logic.

What changed

Added native key edge guard so repeated key-down events for an already-held key are ignored until key-up.
Preserved normal key-down/key-up behavior and existing hotkey pathways.
Validation

Reproduced prior rapid start/stop patterns in logs.
Verified only a single toggle occurs per intentional press cycle after fix.
Risk and compatibility

Low risk.
Narrowly scoped to native listener key-down deduplication.

### DIFF
--- a/main.js
+++ b/main.js
@@ -1512,11 +1512,19 @@ async function startApp() {
   if (process.platform === "win32" || process.platform === "linux") {
     const isWindows = process.platform === "win32";
     const nativeKeyManager = isWindows ? windowsKeyManager : linuxKeyManager;
+    const nativeKeysDown = new Set();
     debugLogger.debug("[Push-to-Talk] Native key listener setup starting");
 
     // Dictation supports push-to-talk and needs the overlay window; agent/meeting
     // drive other windows (matching their globalShortcut callbacks and macOS).
     const dispatchNativeKeyDown = (key) => {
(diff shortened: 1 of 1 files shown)


ambbone:
Follow-up from runtime validation on Windows: I added one more hardening commit to this PR (`2ca5857`) after seeing occasional multi-trigger behavior even with key-down repeat suppression.

What this additional commit does:
- Prevents duplicate native key listener handler bindings in main-process setup.
- Ensures previously-bound `key-down`/`key-up` (and related) handlers are replaced before rebinding.

Why it was needed:
- Logs showed cases with a single native `KEY_DOWN` line but multiple recording starts, which indicates duplicated event routing rather than native auto-repeat.

Result:
- Holding `Control+Super` now behaves consistently (single start/stop cycle), and the runaway multi-trigger behavior is resolved in live testing.

PR #1315, opened 2026-07-24
feat(actions): add per-action temperature override with global fallback

ambbone:
## Summary
This PR adds optional per-action temperature overrides in Manage Actions > Edit Action, while preserving current default behavior when no override is set.

## Why
Some custom actions need more deterministic output, while others benefit from higher creativity. A single global setting is not always ideal for all action types.

## What Changed
- Added optional temperature support to action data:
  - Action type now includes nullable temperature.
  - Database actions table now supports a nullable temperature column.
  - Backward-compatible migration path via add-column logic.
- Extended action create/update APIs:
  - Preload and IPC now pass optional action options for temperature.
  - Create/update flows accept null to mean use global default.
- Updated Manage Actions editor UI:
  - Added a Custom temperature toggle per action.
  - Added a 0.00-1.00 slider shown only when override is enabled.
  - Included unsaved-change detection for temperature fields.
- Applied override at runtime:
  - Action execution uses action temperature when present.
  - Falls back to existing default behavior when absent.
- Added i18n keys for all supported locales:
  - temperatureLabel
  - temperatureHelp
  - temperatureOverride
  - temperatureValue

## Behavior
- If an action has no temperature override, behavior stays unchanged.
- If override is enabled, that value is used for that action only.
- Existing actions remain valid and continue to work without migration steps by users.

## Validation
- Typecheck passes.
- Editor flow tested for:
  - create action with no override
  - create action with override
  - edit action toggle on/off
  - persisted value reload in editor
- Runtime path verified to use override when set and fallback when unset.

## Risk
Low to moderate:
- Scope is focused to action model, action editor, and action execution path.
- Migration is additive (nullable column), preserving existing records.

## Related
- Independent of previously opened PRs and based on upstream main.
- Complements existing global inference configuration by allowing per-action tuning.


ambbone:
Manual smoke test (about 20 seconds):
1. Open Notes > Manage Actions.
2. Create or edit a custom action.
3. Enable Custom temperature and set it to 0.10, then Save.
4. Run that action on any note.
5. Re-open the action and confirm the override persisted at 0.10.
6. Disable Custom temperature, Save, run again, and confirm it still executes normally (fallback path).

Expected:
- Override value persists per action.
- Action execution works with override enabled and disabled.
- No behavior change for actions that do not set an override.

Optional log check:
- Confirm action execution payload uses action temperature when override is enabled, and default path when disabled.

ambbone:
Follow-up runtime validation on Windows is complete for this PR.

What I re-tested after the final updates:
- Per-action override persists after Save (including close/reopen of Manage Actions).
- Turning override off stores fallback behavior correctly.
- Runtime execution uses action-specific temperature when present and global/default path when unset.

Observed outcome:
- Low and high override values are reflected in debug logs as expected.
- No regression in actions that do not define an override.

Issue #494, opened 2026-03-24
[Bug] v1.6.6 paste fails on Linux Wayland — linux-fast-paste --portal times out, uinput fallback doesn't write to clipboard


rearwindowvertigo:
Description
After upgrading from v1.6.2 to v1.6.6 on Linux (Ubuntu 24.04, GNOME Wayland), transcription completes successfully but the result is never written to the clipboard. The previous clipboard contents are pasted instead of the transcription.
This is a separate issue from the mono audio recording regression fixed in v1.6.6 (issue #472). The audio fix works correctly — channelCount: 2 is confirmed in logs — but the new paste pipeline introduced in v1.6.6 is broken.

Root Cause
v1.6.6 attempts linux-fast-paste --portal first, which fails with a timeout. The uinput fallback is then used but does not successfully write the transcription to the Wayland clipboard. The result is that the previous clipboard contents are pasted instead.

v1.6.2 paste behavior (working):
[INFO][clipboard] Paste successful { tool: 'linux-fast-paste', method: 'uinput' }
[INFO][streaming][renderer] Paste timing { pasteMs: 163, source: 'local', textLength: 29 }
v1.6.6 paste behavior (broken):
[WARN][clipboard] linux-fast-paste --portal failed, falling back to uinput {
  error: 'linux-fast-paste --portal exited with code 1: Timeout waiting for portal response'
}
[INFO][clipboard] Paste successful {
  tool: 'linux-fast-paste',
  method: 'uinput',
  detectedWindowClass: null
}
[INFO][streaming][renderer] Paste timing { pasteMs: 10496, source: 'local', textLength: 59 }
Key differences:

v1.6.6 tries --portal first, which times out
Paste takes 10.5 seconds in v1.6.6 vs 163ms in v1.6.2
detectedWindowClass: null in v1.6.6 suggests the target window context is lost during the timeout
wl-paste returns empty after v1.6.6's paste attempt, confirming the transcription never reached the clipboard

Steps to Reproduce

Install OpenWhispr v1.6.6 on Ubuntu 24.04 (GNOME Wayland)
Press F8 to start dictation
Speak a sentence, press F8 to stop
Observe: transcription succeeds in logs (outputTextLength: 59) but previous clipboard content is pasted instead
Run wl-paste in terminal — returns empty, confirming transcription was not written to clipboard

Expected Behavior
Transcription text should be written to the Wayland clipboard and pasted into the active text field, as it works in v1.6.2.
Workaround
[...]

PR #1316, opened 2026-07-24
fix(linux): prefer portal over uinput on GNOME Wayland paste

### LINKED ISSUE #956: fix(linux): GNOME Wayland paste silently fails — Mutter drops uinput key events
## Summary

On GNOME Wayland, auto-paste silently fails after transcription. The `linux-fast-paste --uinput` binary and `ydotool key` both report success (exit 0), but GNOME's Mutter compositor drops the simulated Ctrl+V/Shift+Insert keystrokes. Text is transcribed correctly but never appears in the focused window.

This is a continuation of #240 and #494 — the tool priority reordering in those fixes didn't fully resolve the issue because the underlying problem isn't tool priority, it's that **Mutter silently ignores all uinput-simulated key combos** while allowing direct character injection.

## Root cause

GNOME's Mutter compositor filters out synthesized key events from uinput devices as a security measure. This affects:
- `linux-fast-paste --uinput` (sends Ctrl+V via uinput) → reports success, keystrokes silently dropped
- `ydotool key` (sends raw keycodes via uinput) → reports success, keystrokes silently dropped
- `xdotool key` (XTest) → only works for XWayland windows

However, `ydotool type` (direct character injection via uinput) **does work** — Mutter allows character events, just not key combo events.

## Proposed fix

On GNOME Wayland, try `ydotool type` as the primary paste method before falling back to `linux-fast-paste --uinput`:

```javascript
// In pasteLinux(), the GNOME Wayland branch:
if (ydotoolDaemonRunning && directText) {
  const proc = spawn("ydotool", ["type", "--key-delay", "2", "--file=-"]);
  proc.stdin.end(directText);
  // ... wait for completion + 500ms settling delay
}
```

Key implementation details:
- Use `--file=-` (stdin pipe) instead of command-line arguments — prevents character truncation on longer texts
- Use `--key-delay 2` (2ms between keystrokes) for reliable delivery
- Add 500ms delay after ydotool exits to let the daemon finish injecting buffered keystrokes
- Falls back to existing uinput/portal methods if ydotool type fails

## Secondary fix: YDOTOOL_SOCKET auto-detection

The `_isYdotoolDaemonRunning()` method finds the user's socket at `/run/user/{uid}/.ydotool_socket` but doesn't set `YDOTOOL_SOCKET` in `process.env`. When `ydotool` is spawned, it defaults to `/tmp/.ydotool_socket` (root-owned, inaccessible) and silently fails.

Fix: when a writable socket is found, set `process.env.YDOTOOL_SOCKET` so child processes inherit it. Also use `accessSync` with `W_OK` instead of `statSync` to skip sockets the user can't write to.

## Environment

- Fedora 42 Asahi Remix (GNOME 48, Wayland, aarch64)
- ydotool 1.0.x with systemd user service
- OpenWhispr 1.7.2

## Steps to reproduce

1. Install OpenWhispr on GNOME Wayland
2. Install and start ydotool (`systemctl --user enable --now ydotoold`)
3. Open a text editor
4. Dictate text via OpenWhispr
5. Observe: transcription appears in the app UI but text never appears in the editor
6. Debug logs show `[INFO][clipboard] Paste successful { tool: 'linux-fast-paste', method: 'uinput' }` — false positive

lutfi-zain:
## Problem

On GNOME Wayland, linux-fast-paste --uinput creates a throwaway uinput device for each paste operation. When ydotoold already has its own persistent uinput device running, the two devices race. Mutter/libinput drops the modifier key from the short-lived device, so instead of Ctrl+V the target window receives just V. Paste reports exit 0 but silently fails (false positive — the app never reaches its fallback tools).

This also manifests as portal GDBus errors in the journal:

```
xdg-desktop-portal-gnome: Failed to associate portal window with parent window
Failed to close session implementation: Object does not exist
```

## Fix

Swap the GNOME Wayland paste priority from `uinput → portal` to `portal → uinput`, matching the KDE Wayland approach. uinput remains as a fallback if the portal is denied or unavailable.

**Before (GNOME):** uinput → portal (uinput first, but silently fails)
**After (GNOME):** portal → uinput (one-time dialog, reliable after approval)

## Related

- Closes #956
- Continues the Linux Wayland paste reliability work from #292, #494
- The portal approach was already commented as "reliable once approved" in the existing code (it was the primary path for KDE)
- A simpler workaround (renaming the bundled binary) was confirmed working in the issue comments — this PR makes the fix proper

## Testing

Tested on Ubuntu 24.04 LTS, GNOME Wayland, OpenWhispr 1.7.6:
- One portal dialog on first paste (approve once, never asked again)
- All subsequent pastes work reliably without silent failures
- No regression for KDE or non-GNOME environments (their priority paths are unchanged)

## Changelog

- `src/helpers/clipboard.js`: GNOME Wayland paste priority changed from uinput-first to portal-first
- Updated code comment to document the rationale (uinput device race, #956)

### DIFF
--- a/src/helpers/clipboard.js
+++ b/src/helpers/clipboard.js
@@ -1566,8 +1566,10 @@ class ClipboardManager {
 
         // KDE with XWayland: portal first because clipboard and input are both
         // on X11. uinput causes clipboard desync (X11 clipboard vs Wayland input).
-        // GNOME: uinput first because the portal often times out or shows a
-        // confusing permission dialog, causing a 10s+ delay (issue #494).
+        // GNOME Wayland: portal first (one-time dialog, reliable after approval).
+        // When two uinput devices coexist (linux-fast-paste ephemeral + ydotoold
+        // persistent), Mutter drops the modifier key from the short-lived device,
+        // causing paste to silently fail with exit 0 (issue #956).
         if (isKde && linuxFastPaste && !this.portalDenied) {
           const portalPaste = await tryPortalPaste();
           if (portalPaste) return { method: "portal", ...portalPaste };
@@ -1578,19 +1580,19 @@ class ClipboardManager {
             debugLogger.warn("uinput paste failed", { error: uinputError?.message }, "clipboard");
           }
         } else if (isGnome && linuxFastPaste) {
+          if (!this.portalDenied) {
+            const portalPaste = await tryPortalPaste();
+            if (portalPaste) return { method: "portal", ...portalPaste };
+          }
           try {
             const uinputPaste = await tryUinputPaste();
             return { method: "uinput", ...uinputPaste };
           } catch (uinputError) {
             debugLogger.warn(
-              "uinput paste failed on GNOME, trying portal",
+              "uinput paste failed on GNOME",
               { error: uinputError?.message },
               "clipboard"
             );
-          }
-          if (!this.portalDenied) {
-            const portalPaste = await tryPortalPaste();
-            if (portalPaste) return { method: "portal", ...portalPaste };
           }
         } else {
           // Other compositors (wlroots, etc.): try uinput only


gabrielste1n:
Heads-up: #1629 just landed on main and touches these same portal guards in `pasteLinux()` (new `portalFailed` session flag alongside `portalDenied`, and a 2.5s portal timeout when a restore token exists), so this branch will need a rebase.

Also flagging that #1550 proposes a similar GNOME portal-first reordering, gated on a restore token already existing — maintainers may want to reconcile the two approaches. Either way, any portal-first guard should include `!this.portalFailed` like the KDE one now does, so a stale portal session doesn't stall every paste (see #1614).

greatcoat:
Closing — #956 is fixed in v1.10.0 via #2078, which built directly on your diagnosis of the uinput/Mutter race.

We went with ydotool-before-uinput on GNOME rather than unconditional portal-first: a tokenless portal call can block on the permission dialog for up to 15s (#494), so `main` leads with the portal only once a restore token exists. With uinput falsely reporting success, that ordering could never earn a token in the first place — which is the part your PR surfaced.

Thanks for the clear reproduction and the Ubuntu/GNOME testing.

PR #1318, opened 2026-07-24
fix(security): require IPv4 literals for private HTTP endpoints

### LINKED ISSUE #1317: isSecureEndpoint treats public DNS names like 127.example.com as private HTTP hosts
## Describe the bug

`isSecureEndpoint()` is supposed to allow plain HTTP only for localhost / private-network targets. `isPrivateHost()` implements several of those checks with hostname string prefixes (`startsWith("127.")`, `startsWith("10.")`, `startsWith("192.168.")`, and similar), so **public DNS names** that merely begin with those prefixes are treated as private.

That means a custom / self-hosted endpoint such as `http://127.example.com/v1` is accepted as "secure" and can be used over cleartext HTTP, including paths that attach API credentials.

## Affected component

- `src/utils/urlUtils.ts` — `isPrivateHost()` / `isSecureEndpoint()`
- Callers include custom/self-hosted STT and OpenAI-compatible endpoint validation (`audioManager`, `OpenAICompatiblePanel`, `openaiBase`, retry routing)

## To Reproduce

On current `main` (`ab201b39`):

```js
import { isSecureEndpoint } from "./src/utils/urlUtils.ts";

console.log(isSecureEndpoint("http://127.example.com/v1")); // true  (bug)
console.log(isSecureEndpoint("http://10.example.com/v1"));  // true  (bug)
console.log(isSecureEndpoint("http://192.168.example.com")); // true (bug)
console.log(isSecureEndpoint("http://127.0.0.1:5000"));     // true  (intended)
console.log(isSecureEndpoint("http://api.openai.com/v1"));  // false (intended)
```

Or in the existing unit suite seam: assert `isSecureEndpoint("http://127.example.com/v1") === false`.

## Expected behavior

Only real private/local **address literals** (and existing intentional allowances such as `localhost`, `*.local`, IPv6 ULA/link-local) should bypass the HTTPS requirement. Public DNS names must require HTTPS even when their labels look like private IP prefixes.

## Actual behavior

Any hostname whose string form starts with `127.`, `10.`, `192.168.`, `169.254.`, or a matching CGNAT / `172.16–31` prefix pattern is treated as private, including names like `127.example.com`.

## Environment

- OpenWhispr `main` @ `ab201b39`
- Node.js 24+
- Cross-platform (pure URL/hostname logic)

## User impact

Users (or a crafted config) can point custom STT / reasoning endpoints at a public hostname over HTTP. Credentials and audio may leave the machine in cleartext even though the security gate was meant to block that.

## Root-cause hypothesis

Prefix matching on hostname strings instead of validating dotted-quad IPv4 literals before applying RFC 1918 / loopback / CGNAT / link-local range checks.

## Proposed focused scope

- Parse IPv4 literals strictly (four decimal octets 0–255) before private-range checks
- Keep legitimate `http://127.0.0.1`, `http://10.x`, `http://192.168.x`, CGNAT, link-local, `localhost`, `*.local`, and IPv6 private/ULA behavior
- Add regression tests for `127.example.com`, `10.example.com`, `192.168.example.com`, and incomplete pseudo-addresses such as `127.1`
- No broader URL-normalization or provider-routing changes

I intend to work on this.

hsusul:
## Summary

- Tighten `isPrivateHost()` so private HTTP allowlisting applies only to real IPv4 literals (plus existing localhost / `.local` / IPv6 ULA allowances).
- Prevent public DNS names that merely begin with private-looking prefixes from bypassing the HTTPS requirement for custom/self-hosted endpoints.
- Add regression coverage for the DNS-prefix bypass and for WHATWG URL IPv4 abbreviation expansion.

## Problem

`isSecureEndpoint()` is meant to allow cleartext HTTP only for local/private targets. `isPrivateHost()` used hostname string prefixes (`startsWith("127.")`, `startsWith("10.")`, `startsWith("192.168.")`, etc.), so public DNS names such as `127.example.com` or `10.example.com` were treated as private and accepted over HTTP.

That could let custom STT / OpenAI-compatible endpoints send credentials and audio in cleartext to a public hostname.

## Root cause

Prefix matching on hostname strings instead of validating a dotted-quad IPv4 literal before applying loopback / RFC 1918 / CGNAT / link-local checks.

## Fix

- Add `parseIPv4Literal()` that accepts only four decimal octets in `0–255` (no leading zeros / non-decimal forms).
- Apply private-range checks only when that parse succeeds.
- Preserve intentional allowances: `localhost`, `0.0.0.0`, `::1`, IPv6 ULA/link-local, and `*.local`.

## Behavior before / after

| URL | Before | After |
| --- | --- | --- |
| `http://127.example.com/v1` | allowed | rejected |
| `http://10.example.com/v1` | allowed | rejected |
| `http://192.168.example.com/v1` | allowed | rejected |
| `http://127.0.0.1.nip.io/v1` | allowed | rejected |
| `http://127.0.0.1:5000` | allowed | allowed |
| `http://10.0.0.5:8080` | allowed | allowed |
| `http://100.64.0.1:8080` (Tailscale CGNAT) | allowed | allowed |
| `https://127.example.com/v1` | allowed | allowed |

Note: WHATWG URL expands abbreviated forms like `http://127.1` to `127.0.0.1` before the host check; those remain allowed as real loopback literals.

## Scope / non-goals

- In scope: private-host detection used by `isSecureEndpoint()`.
- Out of scope: URL path/query joining, provider routing, dictionary echo filtering, download/stream cleanup, and unrelated settings migrations.

## Tests added

In `test/helpers/secureEndpoint.test.js`:

- public DNS names resembling private prefixes must require HTTPS
- URL-parser abbreviations of loopback IPv4 remain allowed over HTTP

## Validation

```text
node --version
# v24.9.0

npm ci --ignore-scripts
npm rebuild better-sqlite3

ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/secureEndpoint.test.js
# 9 pass, 0 fail

npm run typecheck
# exit 0

npm run lint
# exit 0

npx prettier --check src/utils/urlUtils.ts test/helpers/secureEndpoint.test.js

npm run format:check
# fails only on pre-existing src/locales/zh-CN/translation.json and src/locales/zh-TW/translation.json
# (also fails when checking those files from upstream/main; not introduced by this change)

npm run i18n:check
# Locale keys and placeholders are consistent.

npm run build:renderer
# exit 0

ELECTRON_OVERRIDE_DIST_PATH=/tmp npm test
# 710 pass, 6 skipped, 0 fail

git diff --check
# clean
```

- Pure hostname/URL logic; no Electron UI, native modules, model downloads, or provider credentials required for reproduction.
- Pre-existing Prettier drift in `src/locales/zh-CN/translation.json` and `src/locales/zh-TW/translation.json` remains on `main` and is untouched here.

Fixes #1317


gabrielste1n (approved):
Approving — excellent catch, @hsusul, and thank you. This was a genuine security hole: prefix matching on hostname strings meant `127.example.com`, `10.example.com` and `127.0.0.1.nip.io` all resolved as private, so a custom STT endpoint could ship credentials and audio in cleartext to a host an attacker controls.

I traced the fix end to end and probed it with 32 URLs covering every branch. All 32 behave correctly. Two things I specifically wanted to confirm before merging:

- **The strict dotted-quad parser can't reject a legitimate literal.** `isSecureEndpoint` runs the URL through WHATWG first, so the hostname is already canonicalised to decimal by the time `isPrivateHost` sees it — `127.1`, `0x7f000001`, `2130706433` and `0177.0.0.1` all arrive as `127.0.0.1` and stay allowed. And where a non-canonical form somehow did arrive, `parseIPv4Literal` returns null and the endpoint falls through to requiring HTTPS, so it fails closed in both directions.
- **`010.0.0.1` is correctly rejected.** That tripped my first probe run until I checked: WHATWG reads the leading zero as octal and normalises it to `8.0.0.1`, a public address. Rejecting it is right.

Boundaries are all exact — `172.15`/`172.32`, `100.63`/`100.128`, `126`/`128` reject; `172.16`–`172.31`, CGNAT `100.64`, `169.254`, `localhost`, `0.0.0.0`, `::1`, ULA/link-local and `*.local` all still pass. I also grepped for the same prefix-matching pattern elsewhere and found none: `isPrivateHost` is module-private with a single consumer, and all five `isSecureEndpoint` call sites simply gate on the boolean, so tightening it can only reject more.

gabrielste1n:
Follow-up code-quality pass before merge (`0d29af9`): dropped `!Number.isInteger(value) || value < 0` — the dotted-quad regex only admits `0` or 1–3 digits starting non-zero, so `Number()` always yields a non-negative integer and only the `> 255` bound can ever fire. Also returned a plain `number[]` instead of a 4-tuple, which removes the `as` assertion (the only consumer destructures two elements, and type assertions are otherwise absent from this file), and removed a test comment that restated the test's own name.

Behaviour is unchanged — re-ran my 32-URL probe plus the suite: 722 tests / 0 failures, typecheck, lint and prettier clean. Merging now. Thanks again for a genuinely valuable security fix.

PR #1322, opened 2026-07-24
Validate LLM API keys before setup

### LINKED ISSUE #1321: LLM API key validation can accept invalid keys or block valid setup
**Describe the bug**
LLM API key validation does not consistently distinguish a definitive authentication rejection from provider, network, quota, or validation-endpoint failures.

Tinfoil is especially affected because its model-list endpoint is public, so an invalid key can appear valid. Other providers can return transient or non-authentication errors that are currently surfaced as validation failures and prevent users from saving otherwise usable credentials.

The UI may alternate between messages such as `The API key could not be validated.` and `The provider rejected this API key.` depending on how the provider or IPC failure is classified.

**To reproduce**
1. Open LLM provider settings.
2. Enter an invalid Tinfoil API key.
3. Attempt to save it.
4. Observe that validation may accept the key because listing models does not require authentication.
5. For any provider, simulate a timeout, rate limit, quota error, upstream 5xx, or unavailable validation endpoint.
6. Observe that setup can be blocked even though the provider did not definitively reject the key.

**Expected behavior**
- Validate Tinfoil through a minimal authenticated request.
- Block saving only when the provider definitively rejects authentication or when local input is invalid.
- Allow credentials to be saved as unverified when validation is inconclusive, with a clear warning so setup can continue.
- Keep secret values out of errors and logs.
- Apply the same classification policy consistently across supported LLM providers.

**Additional context**
A robust validation result should distinguish verified, unverified-but-saved, and definitively invalid states. This prevents false confidence for invalid keys while ensuring transient provider-side validation failures do not lock users out of setup.

boseq:
## Summary
- add atomic API-key validation before provider credentials are accepted
- validate Tinfoil with a minimal authenticated embedding request instead of its public model-list endpoint
- classify validation as verified, unverified-but-saved, or definitively invalid across supported providers
- block only definitive authentication/input failures while allowing setup to continue after transient, quota, rate-limit, network, or upstream validation errors
- show a persistent warning when a key is saved without verification and keep secret values out of errors

## Why
Tinfoil could accept an invalid key because its model-list endpoint is public. At the same time, inconclusive validation failures from Tinfoil and other providers could prevent users from completing setup even when the provider had not rejected their credentials.

Closes #1321

## Verification
- `node --test test/helpers/llmKeyValidation.test.js` — 17 passed
- `node --test --test-concurrency=1 "test/**/*.test.js"` — 734 total, 715 passed, 19 skipped, 0 failed
- `npm run typecheck`
- `npm run i18n:check`
- `npm run format:check`
- `npm run lint`
- `npm run build:renderer`

## Notes
- The default concurrent `npm test` runner can hit a Node 24 test-runner deserialization error in `test/helpers/secretKeys.test.js`; that test passes in isolation and the complete serial suite passes.
- Tinfoil validation performs a tiny authenticated embedding request, so it may consume negligible provider usage.


boseq:
> Case 1: In Case API Key is accepted.

image

> Case 2: In case, the API Key is rejected.
image

> Case 3: In case, the API Key cannot be verified. We do not block the user.

image

boseq:
Closing because this has gone stale. Will re-open if required.

PR #1323, opened 2026-07-24
fix(linux): align Wayland development server with packaged behavior and use XWayland

IdrisGit:
While testing OpenWhispr, whenever I launched the local dev server (i.e  `npm run dev`) on Hyprland, the dictation overlay would open at the center of the screen and the dictation panel wouldn't open until I click the tray icon, for sometime I thought it was an issue with my system because the build/packaged versions worked fine.

Today I decided to debug the issue and found that OpenWhispr depends on launching the app as **XWayland** and it was already handled in the build pipeline for Wayland systems, but the dev launcher only did that for a few Wayland desktops. Hyprland, and any other unlisted compositor, got native Wayland instead.

This updates development launches to force XWayland for every Linux Wayland session, matching packaged behavior instead of maintaining a compositor allowlist.

Explicit `--ozone-platform=...` flags still win, so native-Wayland testing remains available when wanted.

### DIFF
--- a/main.js
+++ b/main.js
@@ -1,5 +1,5 @@
-// Wayland: self-relaunch under XWayland. Chromium picks the display backend
-// before JS runs, so appendSwitch is too late.
+// Chromium picks the display backend before JS runs, so appendSwitch is too
+// late — the flag has to come from a relaunch.
 const { XWAYLAND_FLAG, shouldForceXWayland } = require("./src/helpers/xwayland");
 
 if (shouldForceXWayland(process.argv)) {
--- a/main.js
+++ b/main.js
@@ -1,19 +1,14 @@
-// KDE/GNOME Wayland: self-relaunch with --ozone-platform=x11 to force XWayland.
-// Chromium picks the display backend before JS runs, so appendSwitch is too late.
-if (
-  process.platform === "linux" &&
-  process.env.XDG_SESSION_TYPE === "wayland" &&
-  !process.argv.includes("--ozone-platform=x11")
-) {
-  const desktop = (process.env.XDG_CURRENT_DESKTOP || "").toLowerCase();
-  if (desktop.includes("kde") || /gnome|ubuntu|unity|cosmic/.test(desktop)) {
-    const { spawn } = require("child_process");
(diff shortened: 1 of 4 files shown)


gabrielste1n:
Thanks for digging into this — good catch, and the diagnosis is spot on. The packaged launcher (`scripts/lib/linux-launcher.js`) has always forced XWayland for *any* `XDG_SESSION_TYPE=wayland` with no allowlist, so dev mode was genuinely the odd one out. Your switch from `args.includes("--ozone-platform=x11")` to a prefix check is also a real fix on its own: the old guard would append a second, conflicting flag if you passed `--ozone-platform=wayland`.

I pushed two commits on top that finish the job:

**Deduplicated the rule.** `main.js` still had its own copy of the allowlist for the self-relaunch path, and the two had already drifted — `main.js` matched `cosmic`, `run-electron.js` never did. That's the argument for having one copy. The rule now lives in `src/helpers/xwayland.js` and both `main.js` and `scripts/run-electron.js` call it, so all three launch paths agree with the packaged wrapper. Side effect: a direct binary launch (or `npm start`, which bypasses `run-electron.js`) now covers Hyprland and other wlroots compositors too.

The helper had to go under `src/helpers/` rather than `scripts/` — `scripts/**` isn't in electron-builder's `files` array, so requiring it from `main.js` would be `MODULE_NOT_FOUND` in packaged builds. It's dependency-free so `main.js` can load it before Electron.

**Added `test/helpers/xwayland.test.js`** covering the compositors the old allowlist missed (Hyprland, sway, niri, empty `XDG_CURRENT_DESKTOP`) and the explicit-flag escape hatch.

One thing I deliberately left alone: the helper doesn't check `DISPLAY`, so it will still force XWayland on a Wayland session that has no XWayland server. The packaged wrapper has always behaved that way, and adding the guard to only the JS side would recreate the exact dev/packaged divergence this PR removes. If that's a real failure mode worth handling, it should be a separate change that fixes the bash wrapper and the helper together.

`npm test` (698 pass), `typecheck`, and `lint` are all green.

gabrielste1n (approved):

PR #1327, opened 2026-07-24
fix(translation): route local models through llama.cpp

IdrisGit:
Fix local translation requests failing with `Unsupported reasoning provider: {provider name}` by routing locally selected translation models through the local **llama.cpp** provider instead of treating model-family IDs as inference providers.

### DIFF
--- a/src/models/ModelRegistry.ts
+++ b/src/models/ModelRegistry.ts
@@ -387,6 +387,17 @@ export function getModelProvider(modelId: string): string {
   return model?.provider || "openai";
 }
 
+// Local catalog IDs group models for selection and downloads, but all execute
+// through the single local llama.cpp inference provider.
+export function resolveInferenceProvider(
+  configuredProvider: string | undefined,
+  modelId: string
+): string {
+  const provider = configuredProvider?.trim();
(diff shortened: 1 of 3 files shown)


IdrisGit:
The issue was reported in discord https://discord.com/channels/1463590782689612014/1463591086348832838/1529959521458585621

gabrielste1n:
Note for release notes: this fix is broader than the title suggests — it also repairs the **dictation agent** in local mode.

The fix lands in `processText`, which every scope that forwards a provider goes through. Both the translation route and the agent route pass the stored provider straight down (`audioManager.js`, `resolveReasoningRoute`), so a locally-selected dictation agent hit the identical `Unsupported reasoning provider: <family id>` failure. Dictation cleanup was never affected — it forwards no provider, so it fell through to the model-based heuristic.

Worth exercising both paths when verifying the release:
- local translation (Settings → AI Models → Translation, local mode)
- local dictation agent (Settings → AI Models → Dictation Agent, local mode)

Follow-up #1363 (merged) fixes a related regression from the same root cause — scopes store a catalog family id (`qwen`, `gemma`, …) rather than the literal `"local"` — which had left llama-server pre-warming dead and caused opening the control panel to stop a running local model server.

PR #1335, opened 2026-07-25
fix(export): merge consecutive speaker-less segments within gap threshold

### LINKED ISSUE #1334: Transcript export fails to merge consecutive speaker-less segments
### Component

`src/helpers/transcriptFormatter.js` (`mergeSegments`)

### Reproduction

Pass transcription segments that lack an explicit `speaker` property (e.g. dictations or un-diarized recordings) spaced within 2 seconds of each other to any export formatter (`formatTxt`, `formatSrt`, `formatJson`, `formatMd`):

```javascript
const { formatTxt } = require('./src/helpers/transcriptFormatter');
const segments = [
  { timestamp: 0, text: "Hello world," },
  { timestamp: 1, text: "this is a test." }
];
console.log(formatTxt({ title: "Note" }, segments, {}));
```

### Actual Behavior

`mergeSegments` produces two distinct speaker blocks:

```
[00:00:00] Unknown Speaker:
Hello world,

[00:00:01] Unknown Speaker:
this is a test.
```

Because `last.speaker` is `undefined` while `(seg.speaker || "")` is `""`, the condition `last.speaker === (seg.speaker || "")` evaluates to `false`.

### Expected Behavior

Consecutive segments from the same speaker within 2 seconds of each other—including speaker-less segments—should be merged into a single segment:

```
[00:00:00] Unknown Speaker:
Hello world, this is a test.
```

### Environment

All supported OS platforms (Node 24+, Electron).

### User Impact

Exported transcripts (TXT, SRT, Markdown, JSON) for standard dictations or un-diarized recordings remain fragmented line-by-line every second instead of forming cohesive paragraphs/cues.

### Root Cause Hypothesis

In `src/helpers/transcriptFormatter.js`, `mergeSegments` checks `last.speaker === (seg.speaker || "")`. When `last.speaker` is `undefined` or `null`, the comparison is asymmetric and evaluates to `false`.

### Scope

Normalize both speaker properties before comparison: `(last.speaker || "") === (seg.speaker || "")`.

I intend to work on this issue.

hsusul:
### Summary

Fixes segment merging in transcript export formatters (\`formatTxt\`, \`formatSrt\`, \`formatJson\`, \`formatMd\`) when transcription segments lack an explicit \`speaker\` property (e.g. standard dictation recordings or un-diarized transcriptions).

### Problem

When exporting transcripts to TXT, SRT, JSON, or Markdown format, \`mergeSegments\` checks whether consecutive segments fall within 2 seconds of each other from the same speaker. However, when segments lack a \`speaker\` property (\`seg.speaker\` is \`undefined\` or \`null\`), \`mergeSegments\` checked:

\`\`\`javascript
if (last && last.speaker === (seg.speaker || "") && ts - lastTimestamp < 2)
\`\`\`

Here, \`last.speaker\` is \`undefined\` while \`(seg.speaker || "")\` evaluates to \`""\`. Because \`undefined === ""\` is \`false\`, \`mergeSegments\` treated every segment as belonging to a different speaker and refused to merge them.

### Root Cause

In \`src/helpers/transcriptFormatter.js\`, speaker comparison in \`mergeSegments\` was asymmetric: \`seg.speaker\` was defaulted to \`""\`, but \`last.speaker\` was not.

### Fix

Normalize both speaker properties prior to comparison:

\`\`\`javascript
if (last && (last.speaker || "") === (seg.speaker || "") && ts - lastTimestamp < 2)
\`\`\`

### Behavior Before and After

- **Before**: Dictations and un-diarized transcripts exported to TXT, SRT, Markdown, or JSON were fragmented line-by-line for every segment even when spaced less than 2 seconds apart.
- **After**: Consecutive speaker-less segments within the 2-second threshold are correctly merged into unified transcript blocks and cues.

### Scope & Non-goals

- Only normalizes speaker comparison in \`mergeSegments\` for export formatting.
- Does not alter diarized speaker assignment, timestamp calculation logic, or SRT millisecond rounding.

### Tests Added

Added a unit test in \`test/helpers/transcriptFormatter.test.js\`:
- \`consecutive speaker-less segments within 2s are merged in SRT, TXT, and Markdown export\`

\`\`\`bash
$ ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/transcriptFormatter.test.js
✔ merged same-speaker SRT cues keep the first segment's start time (0.96ms)
✔ same-speaker merging still uses the latest segment for the rolling gap (0.09ms)
✔ same-speaker segments at the merge threshold remain separate cues (0.08ms)
✔ the final merged SRT cue ends relative to its last segment, not its first (0.10ms)
✔ JSON duration reflects the last merged segment's timestamp (10.55ms)
✔ consecutive speaker-less segments within 2s are merged in SRT, TXT, and Markdown export (0.26ms)
ℹ tests 6 | pass 6 | fail 0

$ ELECTRON_OVERRIDE_DIST_PATH=/tmp npm test
ℹ tests 715 | pass 709 | skipped 6 | fail 0

$ npm run typecheck
✓ Passed with 0 errors

$ npm run lint
✓ Passed with 0 errors

$ npm run i18n:check
✓ [i18n] Locale keys and placeholders are consistent.

$ npm run build:renderer
✓ Built successfully

$ git diff --check
✓ Clean (0 errors)
\`\`\`

### Environment & Platform Limitations

- Tested under Node 24.9.0 / npm 11.6.0 on macOS.
- Pre-existing formatting issue noted in \`locales/zh-CN/translation.json\` and \`locales/zh-TW/translation.json\` during \`format:check\`; left untouched per repository guidelines.

Fixes #1334


gabrielste1n (approved):
Approving with a follow-up commit — thanks @hsusul, and apologies for landing a correction on top rather than just merging.

Your diagnosis is exactly right. `last.speaker` was `undefined` while `(seg.speaker || "")` was `""`, so every un-diarized segment looked like a new speaker and plain dictations exported fragmented line by line.

**Where the one-line fix falls short.** Normalising both sides makes merging key on raw `speaker` alone, but `resolveSpeaker` doesn't only read `speaker` — it prefers `speakerName` (when not a placeholder), and falls back to `source` for segments with no speaker at all. Those extra inputs matter in a case that's easy to hit:

`NoteEditor.handleBulkAssignName` sets `speakerName` but never sets `speaker`, and mic segments never get a `speaker` (`assignProvisionalSpeaker` returns early for `source !== "system"`). So after you use "assign name" on one mic segment, it and the un-named mic segment beside it both have `speaker === undefined` — they now compare equal and merge:

```
before this PR:  [00:00:00] Alice:  Alice said this.
                 [00:00:01] You:    And this is me talking.
with the fix:    [00:00:00] Alice:  Alice said this. And this is me talking.
```

The user's own words get exported under Alice's name — in TXT, SRT, JSON, MD, and in the mirrored `-transcript.md` rewritten on every note save. I reproduced this against `origin/main` to confirm main is correct here and it's a regression from the diff, not something pre-existing.

**What I pushed** (`f9aaec4`): merging now keys on a composite of the fields `resolveSpeaker` actually reads — `speaker`, the effective `speakerName`, and `source` when there's no speaker. That still fixes your original `undefined`-vs-`""` asymmetry (both normalise to `""`), so your test passes unchanged, and it adds a regression test for the named-segment case that fails without the change.

This is a good fix and a real bug — the merge key just needed to match the identity the formatter derives downstream. Verified against current main: 722 tests / 0 failures, typecheck, lint and prettier clean.

gabrielste1n:
Merging. Only cleanup on top of the merge-key fix was bringing the test name into the active voice the rest of the file uses (`2f81b4e`).

Verified this composes correctly with #1294, which landed just before and touches the same file — both fixes coexist, 805 tests / 0 failures, typecheck and lint clean. Thanks again for spotting the original asymmetry.

Issue #1280, opened 2026-07-22
[BUG] Note Recording ignores the custom Cloud Providers endpoint — always connects to hardcoded wss://api.openai.com/v1/realtime


joachimleu:
**Describe the bug**

Note Recording (live meeting transcription) ignores the Cloud Providers "Custom" endpoint configured in **Settings → Speech-to-Text**. Regardless of the Endpoint URL / API Key / Model I set, Note Recording always opens a realtime WebSocket to the hardcoded `wss://api.openai.com/v1/realtime?intent=transcription` and authenticates as the built-in OpenAI provider.

The custom endpoint is never consulted. The same Custom endpoint works correctly for **Dictation** and **Audio Upload** (both use the HTTP `/audio/transcriptions` path) — only the realtime Note Recording path is affected.

The meeting engine list appears to be fixed (`local`, `openai-realtime`, `assemblyai-realtime`, `deepgram-realtime`, `corti-realtime`) with no custom / OpenAI-compatible realtime option, so there is no way to point live transcription at an OpenAI-compatible endpoint (e.g. Azure OpenAI's realtime API).

This looks like the same root cause as #750 ("Self-hosted server URL is ignored — STT requests go to api.openai.com instead"), which is closed — the fix appears to have covered the HTTP STT path (Dictation/Audio Upload now honor the custom URL), but the **realtime Note Recording path still hardcodes `api.openai.com`.**

**To Reproduce**

1. Go to **Settings → Speech-to-Text**.
2. Select **Cloud Providers → Custom** and set an OpenAI-compatible Endpoint URL, API Key, and Model (in my case Azure OpenAI: `https://<resource>.cognitiveservices.azure.com`, a valid key, model `gpt-4o-mini-transcribe`). Confirm this Custom config works for **Dictation**.
3. Switch to the **Note Recording** tab and start a live transcription.
4. See error — transcription never starts. The debug log shows it connecting to `api.openai.com` (not the custom endpoint) and failing at the WebSocket handshake.

Debug log excerpt (`~/Library/Application Support/open-whispr/logs/`):

```
[INFO][meeting][renderer] Meeting transcription starting... {}
[DEBUG] OpenAI Realtime connecting { "model": "gpt-4o-mini-transcribe" }
[ERROR] ERROR: OpenAI Realtime WebSocket error { "error": "Unexpected server response: 403" }
[...]

PR #1337, opened 2026-07-25
feat(meetings): allow Self-Hosted for Note Recording via chunked HTTP, opt-in

hannesro:
## Summary

Allow Note Recording's **Self-Hosted** transcription mode, which currently ships disabled with a *Coming soon* badge, as an explicit opt-in using the transport most self-hosted servers actually have: chunked HTTP to `/audio/transcriptions`.

I read @gabrielste1n's 1.8.3 status update on #1280 after opening this. It rules out chunked HTTP as the *universal fallback* for an arbitrary OpenAI-compatible base URL, on accuracy grounds, and I do not dispute that reasoning. This PR is scoped to the narrower case that reasoning leaves open, and is framed accordingly: not a generic fallback, and not the default for anything.

- Nothing silently uses this transport. `self-hosted` is a mode the user selects deliberately, and the panel now states both the transport and the accuracy trade-off.
- It is not a substitute for realtime provider entries. Servers that do speak the realtime protocol are better served by a registry entry, which this does not block or complicate. It covers the servers such an entry cannot reach: the batch-only ones.

Rebased onto current `main` (was 454 commits behind). Refs #1280, Refs #581. **Closes neither**. See *Scope / non-goals*.

## Problem

Note Recording offers four transcription modes: `openwhispr`, `providers`, `local`, `self-hosted`. Since f67f2a6b the fourth one is honest about not working: `resolveMeetingTranscriptionOptions()` throws `"Self-hosted realtime transcription is not supported for Note Recording"`, and `MeetingSettings.tsx` renders the mode `disabled` with a *Coming soon* badge.

That leaves a gap rather than a bug. Dictation, Audio Upload and retry all honor a self-hosted endpoint through `resolveTranscriptionRoute`; Note Recording is the only surface that does not. Users behind a proxy that blocks `api.openai.com`, or who are only permitted to use their own OpenAI-compatible service, can dictate and upload but cannot record a note.

The blocker is the transport. The meeting path is built around realtime WSS engines, and "self-hosted with streaming support" means implementing OpenAI's realtime protocol (`session.update`, `input_audio_buffer.append`, `conversation.item.input_audio_transcription.*`). Some self-hosted servers do: @ianbmacdonald's #1280 comment demonstrates Lemonade doing exactly that against the stock `openai` SDK. Most do not. whisper.cpp's server, speaches / faster-whisper-server, vLLM Whisper, LocalAI and the MLX / Qwen3-ASR shims expose `POST /v1/audio/transcriptions` and nothing else.

So the population splits, and the two halves want different fixes. Realtime-capable servers want a provider entry in the streaming registry, which is what #1280 now tracks. Batch-only servers cannot be served by any registry entry, because there is no socket to open. They are what this PR is for.

## Fix

Reuse the batch path that already exists for the other three surfaces, and the local-mode chunking that already exists for meetings.

**Routing** (`src/helpers/meetingTranscriptionRouting.js`)

- Replace the `self-hosted` throw with a real route carrying the configured URL and model.
- Still fails closed when no URL is configured, so the mode can never fall through to a realtime provider the user did not select. The test that pinned that property (`self-hosted mode never follows a stale Tinfoil provider`) is kept and now asserts the route instead of the throw.

**Main process** (`src/helpers/ipcHandlers.js`, `src/helpers/meetingStreamingProviders.js`)

- Add `NON_STREAMING_MEETING_PROVIDERS` (`local`, `self-hosted`) and derive `ALLOWED_MEETING_PROVIDERS` from it plus the streaming registry, keeping the "an allowed provider can never lack a client class" invariant intact for the streaming half.
- Resolve and validate the endpoint through `resolveTranscriptionRoute` (the same batch router dictation, retry and upload use), so endpoint shape (Azure deployments included), URL validation and fail-closed behavior live in one place rather than a second implementation. Anything but a usable self-hosted route fails the start.
- Skip realtime pre-warm (no socket to warm) and transcribe each chunk over multipart POST with `file`, `model`, `language`, matching the retry/upload self-hosted request.

**Settings / UI**

- Add `meetingRemoteTranscriptionModel`, falling back to the dictation self-hosted model, and include it in `MEETING_TRANSCRIPTION_PAIRS`.
- Re-enable the mode, restore the URL + Model fields via the shared `SelfHostedPanel`, and replace the stale `selfHostedStreamingNote` ("requires a server that supports streaming transcription") with `selfHostedChunkedNote` in all 10 locales.

No warm-socket bypass is needed any more: f67f2a6b added the `meetingConnectionKey === requestedConnectionKey` check on start, and `getMeetingConnectionKey` includes the provider, so a self-hosted start can never reuse a warm cloud socket. That guard from the first version of this PR is now redundant and was dropped.

## Behavior before / after

| Note Recording mode | Before | After |
| --- | --- | --- |
| Self-Hosted | disabled in the UI, throws if reached | chunked POST to `{url}/audio/transcriptions` |
| Self-Hosted, no URL | n/a (unreachable) | start fails with a configuration error |
| Local / realtime engines | unchanged | unchanged |

## Scope / non-goals

The "custom endpoint ignored" symptom in #1280 spans three modes needing three different fixes. This PR is the first only.

| Mode | What it needs | This PR |
| --- | --- | --- |
| `self-hosted` | a non-realtime transport | ✅ chunked HTTP |
| `providers` → **Custom** | thread `meetingCloudTranscriptionBaseUrl` into `OpenAIRealtimeStreaming.connect()`, plus `api-key`-vs-`Bearer` auth | ❌ untouched |
| `providers` → **Tinfoil** | already handled by f67f2a6b | n/a |
| realtime-capable self-hosted (Lemonade, Azure) | a provider entry in the streaming registry | ❌ not blocked by this, and not attempted here |

`openaiRealtimeStreaming.js:102` still hardcodes `wss://api.openai.com/v1/realtime`, and `resolveMeetingTranscriptionOptions` never reads `meetingCloudTranscriptionBaseUrl`, so #1280's original repro still fails. Happy to follow up with that half if you want it split this way; I have no OpenAI-compatible *realtime* endpoint to test it against, which is why it is not here.

## On the accuracy objection

@gabrielste1n's concern is that chunk boundaries clip words mid-utterance, drop cross-chunk context for punctuation and disambiguation, and degrade speaker continuity. That is a real cost of this transport, and the reason the panel copy now says so in all locales rather than only describing the endpoint.

Two things bound it in practice. The same chunking already ships for Local Note Recording, on the same 5s cadence and the same buffers, so this is not a new quality regime for the meeting path, it is the existing one pointed at a different decoder. And the comparison a batch-only user faces is not chunked-versus-realtime, it is chunked-versus-nothing, or in @ianbmacdonald's framing, versus whatever transcript the meeting platform hands out.

If the trade-off is unacceptable even opt-in, that is a legitimate call, and the honest resolution is closing this rather than watering it down. I would rather have that answer than leave the mode visible-but-disabled.

Trade-off of the chunked approach: self-hosted Note Recording is chunk-latency (5s), not word-by-word. Speaker attribution, note linking and the rest of the meeting pipeline are unchanged, since this reuses the local-mode plumbing. Chunk size self-limits to server throughput, exactly as local mode does with a slow on-device model. Dictionary biasing (`prompt`) is deliberately omitted to match the retry/upload self-hosted request; easy to add if you want it everywhere.

## Validation

```text
npm ci            # Node 24, per .nvmrc
npm run lint      # clean
npm run format    # no diff
npm run typecheck # clean
npm test          # 2617 pass, 0 fail (2788 total, 170 skipped)
```

Three test files updated:

- `meetingTranscriptionRouting.test.js`: the self-hosted route, an unset model left to the server default, and fail-closed for empty/whitespace/undefined URLs.
- `meetingStreamingProviders.test.js`: non-streaming providers are allowed *and* have no streaming client, so the no-silent-OpenAI-fallback invariant still covers every streaming provider.
- `dictationStreamingMatrix.test.js`: the "every allowed meeting provider has a token entry" closure now skips non-streaming providers (no socket, no token to mint).

Also exercised against a real batch-only self-hosted server (Qwen3-ASR behind an OpenAI-compatible endpoint, no realtime API), driving the actual shipped functions rather than a mock:

```text
1. renderer -> IPC:  { provider: 'self-hosted', url: 'http://127.0.0.1:8765/v1', model: '…', language: 'en' }
2. main -> route:    { transport: 'http-batch', provider: 'self-hosted',
                       endpoint: 'http://127.0.0.1:8765/v1/audio/transcriptions', … }
3. POST -> 200:      "The quarterly revenue projections are ready for review before the meeting starts."
4. empty URL:        start fails closed with a configuration error
```

## What changed since the first version

The first version was written against `main` before f67f2a6b, when selecting Self-Hosted silently opened a socket to `api.openai.com`. That fail-open behavior is fixed upstream now, so this PR no longer claims it: the remaining gap is the missing transport, not a misrouting bug. The diff is smaller as a result (endpoint validation delegated to `resolveTranscriptionRoute`, warm-socket guard dropped) and now carries tests.

Two claims in the first description and in my earlier comment were wrong, and are corrected above: that essentially no self-hosted server speaks the realtime protocol (Lemonade does), and that the mode was unwired by drift rather than by the deliberate decision recorded on #1280.


IdrisGit:
I am not able to follow the reasoning here, OpenWhispr already supports self-hosted server for Note recordings, given the condition that it supports streaming (I haven't tested it though, I am assuming it works). But your change changes it from streaming to chunked? why?

and this doesn't address the custom cloud provider issue as well.

hannesro:
Fair questions. I have rebased onto current `main` and rewritten the description, since the ground moved: f67f2a6b landed two weeks after your comment and changed what this PR is fixing.

**"OpenWhispr already supports self-hosted for Note recordings, given the server supports streaming."**

It did not, on any code path. At the time of your comment: `getMeetingTranscriptionOptions()` had no self-hosted branch, `meetingRemoteTranscriptionUrl` was read only by the settings input and one unused selector line, `self-hosted` was absent from `ALLOWED_MEETING_PROVIDERS`, and `connectRealtimeStreaming()` had no field through which any URL could reach a streaming client. Selecting Self-Hosted fell through to `catalog?.[0]` and opened `wss://api.openai.com`. So a server supporting streaming would not have helped: nothing pointed at it.

f67f2a6b fixed that half, in the fail-closed direction: the selected mode is now authoritative, `self-hosted` throws instead of substituting a realtime provider, and the UI marks the mode *Coming soon*. The current state is not "works with a streaming server", it is "not implemented yet".

**"Your change changes it from streaming to chunked? Why?"**

Nothing is converted. There was no streaming path to convert. The real question is which transport a given self-hosted server can serve, and the population splits in two.

Some do speak the realtime protocol. @ianbmacdonald's comment on #1280 demonstrates [Lemonade](https://github.com/lemonade-sdk/lemonade) serving an OpenAI-compatible realtime transcription endpoint through the stock `openai` SDK, emitting the same `session.created` and `conversation.item.input_audio_transcription.delta` events this codebase already handles. For those, a provider entry in the streaming registry is the right fix, which is what #1280 now tracks, and this PR neither blocks nor attempts it.

Most do not. whisper.cpp's server, speaches / faster-whisper-server, vLLM Whisper, LocalAI and the MLX / Qwen3-ASR shims expose `POST /v1/audio/transcriptions` and nothing else. No registry entry can serve them, because there is no socket to open. They are what this PR is for, and chunked HTTP is the only transport they have.

So it routes note-recording chunks through `resolveTranscriptionRoute`, the same batch router Dictation, Audio Upload and retry already use, and reuses the existing local-mode buffering and chunking. Endpoint validation, Azure URL shapes and fail-closed behavior come from that router rather than a second implementation. About 200 lines, no new transport.

**On the accuracy objection**

I should have read @gabrielste1n's 1.8.3 status update on #1280 before opening this, and I had not when I first wrote this comment. It rejects chunked HTTP as the *universal fallback* for an arbitrary OpenAI-compatible base URL, because cutting on fixed boundaries clips utterances, drops cross-chunk context for punctuation, and degrades speaker continuity. I do not dispute that reasoning, and I have scoped and reframed this PR to the case it leaves open rather than arguing against it.

- Nothing is routed this way silently. `self-hosted` is a mode the user selects deliberately, and I have pushed a change so the panel states the transport *and* the accuracy trade-off in all ten locales, rather than only describing the endpoint.
- The same chunking already ships for Local Note Recording, on the same 5s cadence and the same buffers. This points that existing path at a different decoder rather than introducing a new quality regime for meetings.
- For a batch-only user the comparison is not chunked versus realtime. It is chunked versus nothing, or in @ianbmacdonald's framing, versus whatever transcript the meeting platform hands out.

**"This doesn't address the custom cloud provider issue as well."**

Correct, and never claimed: it is `Refs`, not `Closes`, with a scope table naming what is left. `openaiRealtimeStreaming.js:102` still hardcodes `wss://api.openai.com/v1/realtime`, and `resolveMeetingTranscriptionOptions` still never reads `meetingCloudTranscriptionBaseUrl`, so #1280's repro still fails. Happy to leave that to the realtime-provider work #1280 tracks.

**Validation**

On current `main`: lint, format and typecheck clean, full suite 2617 passing, three test files updated to cover the new route and keep the provider-closure invariants honest. Also exercised against a real batch-only server (Qwen3-ASR behind an OpenAI-compatible endpoint): chunk POSTs return transcripts, and an unset URL fails the start instead of contacting OpenAI.

**The question, narrowed**

Not "should Note Recording accept arbitrary endpoints", which #1280 has answered, but: is an explicitly opt-in, caveated chunked path acceptable for the batch-only servers a realtime registry entry can never reach? @ianbmacdonald asked for either that or a realtime provider entry and said he would take either, so it is not a one-user request.

If the answer is no even with the opt-in framing, say so and I will close this. That is a reasonable product call, and I would rather have it than leave the mode visible-but-disabled while Dictation and Upload honor the same endpoint.

<sub>Edited after reading the 1.8.3 status update on #1280: the earlier version of this comment claimed that essentially no self-hosted server speaks the realtime protocol, which Lemonade disproves, and attributed the disabled mode to drift rather than to a deliberate decision. Both corrected above.</sub>

xlurie:
Verified this at head `161cc62` on Windows 11, against a real batch-only self-hosted server — the class this PR targets. My daily Dictation backend is whisper.cpp (large-v3) behind a local OpenAI-compatible proxy; its only transcription interface is `POST /audio/transcriptions`, so no realtime provider entry describes it without a translating shim in front. On installed 1.8.3, the Note Recording Self-Hosted tile is disabled ("Coming Soon") while Dictation honors the same endpoint fine (Upload's tile is enabled too).

Build & tests (Windows): lint, typecheck and `build:renderer` clean; the PR's three test files pass 34/34 (run with a local `electron` stub — without it they hit the same pre-existing Windows-only harness defect as the rest of the suite); full `npm test`: 108 failing at the PR head vs 109 on clean `main` on the same machine, every failure the known Windows `require("electron")` artifact, and the two extra failing files on `main` don't exist at this PR's base — no PR-attributable failure.

Live run (`npm run dev`, isolated dev userData, server URL `http://127.0.0.1:8587/v1`, meeting model left blank; audio = a known Russian real-speech recording played into the capture path through the system-audio loop):

- **Chunked path works end-to-end.** My request-level server log shows 26 requests over ~2:00 — request-arrival cadence 4.994–5.012 s across 25 intervals; 21 of 26 payloads are exactly 160044 B (= 5.000 s of 16 kHz mono WAV), the rest within ±0.04 s, plus a 0.5 s flush chunk at stop; per-chunk latency 52–859 ms; no `prompt` field. The note accumulated a coherent Russian transcript (1097 chars returned across the train; one 5-s window came back empty).
- **Fail-closed holds.** With both the meeting and dictation URLs blank (the meeting URL intentionally falls back to the dictation one), start fails with "Self-hosted Note Recording needs a transcription server URL. Add one in Settings." and the server log shows zero requests; no websocket or api.openai.com traffic in the debug log either.
- **Dead server mid-recording** (URL pointed at a closed port): the recording keeps running, each chunk fails non-fatally with `ERR_CONNECTION_REFUSED` surfaced in the store error, zero stray requests elsewhere.
- **Dictation regression:** in the same build, `resolveTranscriptionRoute` still returns the same `http-batch` self-hosted route for Dictation; my daily config is untouched.
- The documented fallback of a blank meeting model to the dictation self-hosted model resolves as described (`whisper-1` in my run) — convenient for single-server setups.

Scope note: this is a transport/config verification from the batch-only user class — a short ~2-min run, not a long-session soak — and it doesn't speak to chunked-vs-realtime quality, which stays #1280's territory for realtime-capable servers.

hannesro:
Rebased onto current `main` (`a77fdce3`) — conflicts resolved, and the PR now follows the conventions that landed since:

- the fail-closed error is a sentinel (`selfHostedUrlMissing`) translated by `MeetingRecordingMount`, replacing the now-unreachable `unsupportedSelfHosted` copy in all 11 locales
- `ar` added, which arrived after my last rebase
- the chunk dispatch sits ahead of `isSherpaLocalProvider`, so NVIDIA and Cohere local meetings are untouched

`npm run lint`, `npm run typecheck` and `npm run format` clean; `npm test` 4714 pass, 0 fail.

@gabrielste1n — what I'm asking for is a decision, not a review. Your 1.8.3 note rules out chunked HTTP as the *universal fallback* for an arbitrary OpenAI-compatible base URL, and I'm not asking you to revisit that. This PR is the narrower case it leaves open: servers that expose only `POST /v1/audio/transcriptions` — whisper.cpp's server, speaches, vLLM Whisper, LocalAI — which no realtime registry entry can reach, behind a mode the user selects deliberately and a panel that states the accuracy trade-off. @xlurie [verified it end to end](https://github.com/OpenWhispr/openwhispr/pull/1337#issuecomment-5371011621) against exactly that class of server.

For what it's worth, I have been running this change myself since I opened the PR, and it has worked perfectly for daily Note Recording against my own server. I would love to finally switch back to an up-to-date `main` build instead of my own.

If the answer is that you don't want chunked audio on the meeting path at all, even opt-in, that's a legitimate call and I'll close this. I'd rather have that answer than leave the mode visible-but-disabled.

Also worth flagging: #2083 covers the same ground independently, and #2176 covers the Custom-provider half I scoped out. Happy to close mine in favour of #2083, or fold its `selfHostedMeetingTranscription.js` split into this one — whichever is less review for you.

PR #1339, opened 2026-07-25
fix(api): extract top-level and string detail fields from API error bodies

### LINKED ISSUE #1338: fix(api): extract top-level and string detail fields from API error bodies
### Affected Component
AI / API Error Handling (`src/services/ai/apiErrorMessage.ts`)

### Reproduction Steps
Call `extractApiErrorMessage` with FastAPI, Pydantic, or Ollama error responses:
1. `extractApiErrorMessage({ detail: "Model 'llama3' not found" }, "fallback")`
2. `extractApiErrorMessage({ detail: [{ loc: ["body", "model"], msg: "field required" }] }, "fallback")`
3. `extractApiErrorMessage({ message: { detail: "Rate limit exceeded" } }, "fallback")`

### Actual Behavior
1. Returns `"fallback"` (top-level `detail` string ignored).
2. Returns `"fallback"` (top-level `detail` validation array ignored).
3. Returns `'{"detail":"Rate limit exceeded"}'` (nested `detail` string stringified instead of extracted).

### Expected Behavior
1. Returns `"Model 'llama3' not found"`.
2. Returns `"model: field required"`.
3. Returns `"Rate limit exceeded"`.

### Root Cause
`extractApiErrorMessage` only passes `errorData.message` to `fromDetailList(message)`, which ignores top-level `errorData.detail`. Furthermore, `fromDetailList` only inspects `Array.isArray(container.detail)` and returns `null` if `detail` is a string.

### User Impact
Users connecting to self-hosted LLM endpoints, FastAPI gateways, Ollama, or custom OpenAI-compatible proxies receive generic fallback error messages ("Unknown API error") or raw JSON blobs instead of clear, human-readable error descriptions.

I intend to submit a fix for this issue.

hsusul:
### Summary
This PR fixes `extractApiErrorMessage` to support top-level `detail` string and array fields commonly returned by FastAPI, Pydantic, Ollama, vLLM, and custom OpenAI-compatible Python backends.

Fixes #1338

### Problem & Root Cause
When an API endpoint returns an error response using FastAPI, Pydantic, or Ollama conventions (such as `{ "detail": "Model 'llama3' not found" }` or `{ "detail": [{ "loc": ["body", "model"], "msg": "field required" }] }`), `extractApiErrorMessage` failed to extract the error description.

Root causes:
1. `extractApiErrorMessage` only passed `errorData.message` to `fromDetailList(message)`, ignoring top-level `errorData.detail`.
2. `fromDetailList` only checked `Array.isArray(container.detail)`, returning `null` if `detail` was a string.
3. If `detail` was nested inside `message` as a string (`{ message: { detail: "Rate limit exceeded" } }`), `extractApiErrorMessage` raw-stringified the object as `'{"detail":"Rate limit exceeded"}'` instead of extracting the string.

### Fix
- Updated `fromDetailList` to accept any object container and check if `container.detail` is a non-empty string or an array of validation objects (`{ loc, msg }`).
- Updated `extractApiErrorMessage` to inspect `fromDetailList` on both top-level `errorData` and `errorData.message`.

### Behavior Before vs After
- **Before**:
  - `{ detail: "Model 'llama3' not found" }` -> returned `"Unknown API error"` (fallback).
  - `{ detail: [{ loc: ["body", "model"], msg: "field required" }] }` -> returned `"Unknown API error"` (fallback).
  - `{ message: { detail: "Rate limit exceeded" } }` -> returned `'{"detail":"Rate limit exceeded"}'`.
- **After**:
  - `{ detail: "Model 'llama3' not found" }` -> returns `"Model 'llama3' not found"`.
  - `{ detail: [{ loc: ["body", "model"], msg: "field required" }] }` -> returns `"model: field required"`.
  - `{ message: { detail: "Rate limit exceeded" } }` -> returns `"Rate limit exceeded"`.

### Scope & Non-Goals
- Scope: Pure error extraction helper (`src/services/ai/apiErrorMessage.ts`).
- Non-goals: No changes to retry policies or network transport layers.

### Tests Added
Added 3 unit tests in `test/helpers/apiErrorMessage.test.js`:
- `extracts a top-level detail string (FastAPI / HTTPException format)`
- `extracts a top-level detail validation array (Pydantic / FastAPI format)`
- `extracts a nested detail string inside message`

- `ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/apiErrorMessage.test.js` -> PASS (14/14 tests)
- `ELECTRON_OVERRIDE_DIST_PATH=/tmp npm test` -> PASS (711/711 tests, 6 skipped)
- `npm run typecheck` -> PASS
- `npm run lint` -> PASS
- `npm run format:check` -> Note: pre-existing formatting warnings in `locales/zh-CN/translation.json` and `locales/zh-TW/translation.json`; `npx prettier --check` on modified files PASSED cleanly with 0 warnings.
- `npm run i18n:check` -> PASS
- `npm run build:renderer` -> PASS
- `git diff --check` -> PASS

### DIFF
--- a/src/services/ai/apiErrorMessage.ts
+++ b/src/services/ai/apiErrorMessage.ts
@@ -12,12 +12,12 @@ function isRecord(value: unknown): value is Record<string, unknown> {
   return typeof value === "object" && value !== null && !Array.isArray(value);
 }
 
-/** FastAPI/Pydantic validation errors or detail strings: {detail: "..."} or {detail: [{loc, msg}]}. */
-function fromDetailList(container: unknown): string | null {
+/** FastAPI/Pydantic error payloads: {detail: "..."} or {detail: [{loc, msg}]}. */
+function fromDetail(container: unknown): string | null {
   if (!isRecord(container)) return null;
 
-  const detailStr = asNonEmptyString(container.detail);
-  if (detailStr) return detailStr;
+  const direct = asNonEmptyString(container.detail);
+  if (direct) return direct;
 
   if (!Array.isArray(container.detail)) return null;
 
@@ -56,22 +56,19 @@ export function extractApiErrorMessage(errorData: unknown, fallback: string): st
   const flat = asNonEmptyString(errorData.message);
   if (flat) return flat;
 
-  const topDetail = fromDetailList(errorData);
-  if (topDetail) return topDetail;
-
-  const messageDetail = fromDetailList(errorData.message);
-  if (messageDetail) return messageDetail;
+  // Nested containers first: a top-level detail usually just restates the HTTP status,
+  // while the wrapped payload carries the upstream error.
+  for (const container of [errorData.message, errorData.error, errorData]) {
+    const detail = fromDetail(container);
+    if (detail) return detail;
+  }
 
   const errorString = asNonEmptyString(errorData.error);
   if (errorString) return errorString;
 
-  if (isRecord(errorData.message) || Array.isArray(errorData.message)) {
-    const stringified = stringify(errorData.message);
-    if (stringified) return stringified;
-  }
-
-  if (isRecord(errorData.error) || Array.isArray(errorData.error)) {
-    const stringified = stringify(errorData.error);
+  for (const value of [errorData.message, errorData.detail, errorData.error]) {
+    if (!isRecord(value) && !Array.isArray(value)) continue;
+    const stringified = stringify(value);
     if (stringified) return stringified;
   }
 
--- a/src/services/ai/apiErrorMessage.ts
+++ b/src/services/ai/apiErrorMessage.ts
@@ -12,12 +12,17 @@ function isRecord(value: unknown): value is Record<string, unknown> {
   return typeof value === "object" && value !== null && !Array.isArray(value);
 }
 
-/** FastAPI/Pydantic validation errors: {detail: [{loc, msg}]}. */
-function fromDetailList(message: unknown): string | null {
-  if (!isRecord(message) || !Array.isArray(message.detail)) return null;
+/** FastAPI/Pydantic validation errors or detail strings: {detail: "..."} or {detail: [{loc, msg}]}. */
+function fromDetailList(container: unknown): string | null {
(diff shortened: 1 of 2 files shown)


gabrielste1n (approved):
Approving — thanks for this, @hsusul! Really well-put-together PR: clear root-cause analysis tying each bad output to the exact line that caused it, a fix scoped tightly to the helper, and unit tests for all three reported shapes. The issue write-up before the PR was much appreciated too.

I pushed one follow-up commit (`ed20423`) on top with a few refinements:

- **Nested detail now takes precedence over top-level `detail`.** A top-level `detail` often just restates the HTTP status (`"Unprocessable Entity"`), while the wrapped payload carries the real upstream error. Probing `message` → `error` → top level also makes the change strictly non-breaking: every body with `message.detail` resolves exactly as it did before.
- **Restored the never-lose-the-body guarantee.** A top-level `detail` that isn't extractable (e.g. `{ detail: { reason: "upstream timeout" } }`) was falling through to `"Unknown API error"`, dropping the payload. The stringify chain only covered `message` and `error`, so `detail` is now included — consistent with the principle the existing "surfaces the raw body over [object Object]" test encodes.
- **Same fix for `error.detail`.** `{ error: { detail: "Rate limit exceeded" } }` returned `'{"detail":"Rate limit exceeded"}'` — the identical defect this PR fixes under `message`, just in the sibling container.
- **Renamed `fromDetailList` → `fromDetail`** now that it handles a string as well as a list, and collapsed the three repeated stringify guards into one loop. Net effect is slightly less code than before despite covering more shapes.

One thing I intentionally left out: array-of-strings `detail` (`{ detail: ["..."] }`). The existing test at `test/helpers/apiErrorMessage.test.js:84` deliberately asserts that bare-string entries in a `detail` array are skipped as junk, and FastAPI only ever emits validation objects there — supporting it would break that assertion and risk leaking noise into user-facing errors.

Verified locally: 720 tests / 0 failures, typecheck, lint and prettier all clean.

hsusul:
Makes sense. Appreciate it.

PR #1344, opened 2026-07-26
feat: add Atlas Cloud inference provider

binyangzhu000-sudo:
## Summary

- add Atlas Cloud as a first-class BYOK reasoning provider with secure API key persistence
- route streaming and AI SDK requests through the existing OpenAI-compatible Chat Completions paths
- register verified Atlas Cloud text models and expose them in the reasoning model selector
- add regression coverage for the key bridge, endpoint, and model registry

## Validation

- `node --test test/helpers/secretKeys.test.js` (6 passed)
- `npm run typecheck`
- ESLint on all changed JavaScript/TypeScript files
- Prettier check on all changed files
- `git diff --check`
- Atlas Cloud live `GET /v1/models`: HTTP 200, 121 models; all three configured model IDs present

## Local environment limitations

- `npm test` could not complete because dependencies were installed with `npm ci --ignore-scripts`; tests importing Electron report `Electron failed to install correctly`
- `npm run build:renderer` remained at the Vite transform phase without progress in this local environment and was terminated after about four minutes

No README, documentation, logo, sponsor, credits, or partner content was changed.


binyangzhu000-sudo:
Rebased onto current `main` (`2e12d32`). Three files conflicted, and in each one main had refactored the surrounding code, so rather than picking a side I kept main's new structure and re-attached Atlas Cloud to it.

**`ReasoningService.ts`** — adopted your `resolveByokAccess()` flow in both streaming paths, replacing the hand-written provider switch this branch had. Two follow-ups were needed to make that work for Atlas Cloud:

- Registered `atlascloud` in `BYOK_STREAM_PROVIDERS`, since `toByokStreamProvider()` throws `Unsupported reasoning provider` for anything outside that list.
- Gave it `API_ENDPOINTS.ATLASCLOUD_BASE` inside `resolveByokAccess`, in the same position openrouter uses. This one matters: `resolveConfiguredOpenAIBase()` returns `OPENAI_BASE` for every non-`custom` provider, so without an explicit branch Atlas Cloud requests would have silently gone to OpenAI with an Atlas key.

**`inferenceProviders/openai.ts`** — kept your `canFallBackToSharedKey` gate around the shared-key lookup while retaining the `atlascloud` key scope inside it, and kept the explicit `ATLASCLOUD_BASE` branch ahead of `resolveConfiguredOpenAIBase()` for the same reason as above.

**`ReasoningModelSelector.tsx`** — `selectedCloudProvider` became `displayedCloudProvider` and the provider blocks were reindented, so I rebuilt the Atlas Cloud API-key block in that shape and placed it between groq and tinfoil to match `CLOUD_PROVIDER_IDS` order.

The other ten files (settingsStore, secretKeys, constants, preload, model registry, providers, electron types, useSettings, the inferenceProviders index and the secretKeys test) merged cleanly; I checked each still carries its atlascloud entry. Net diff is now 13 files / +133, down from +153 — the switch to `resolveByokAccess` removed code this branch used to carry itself.

**Validation**

- `npm run typecheck` — 0 errors.
- `prettier --check` on the three touched files — clean; `eslint` on them — no errors.
- `test/helpers/secretKeys.test.js` — 6 passed, including both atlascloud cases.
- Full `npm test` — 2821 passed / 126 failed. A clean `origin/main` worktree sharing the same `node_modules` gives 2819 passed / **126 failed**: identical failure count, and my run has exactly 2 more passing tests (the atlascloud ones). The 126 are pre-existing here because I installed with `--ignore-scripts`, so the whisper.cpp / sherpa-onnx / GPU-detection binaries those suites need were never downloaded.

CI shows `action_required` on all four workflows — that needs a maintainer to approve the run for a fork.

binyangzhu000-sudo:
Synced with `main` (46 commits). Single conflict in `test/helpers/secretKeys.test.js`, where both sides appended tests. Both are kept.

Upstream's new test is directly relevant to this branch, so I checked it rather than just resolving around it: it pins that the STT accessors keep the spellings their callers use, with a note that deriving them from the `base` field would silently produce `getAssemblyaiKey` and break `realtimeTokenProviders.js` and `stt-canary.mjs`. The `atlascloud` entry states `get` / `save` explicitly in the manifest (`getAtlascloudKey` / `saveAtlascloudKey`) and `preload.js` carries the matching `BYOK_KEY_BRIDGES` tuple, so upstream's other new test — that the preload bridges mirror the manifest exactly — also covers it. Both pass.

The three model ids asserted against `modelRegistryData.json` (`deepseek-ai/deepseek-v4-pro`, `deepseek-ai/deepseek-v4-flash`, `qwen/qwen3.5-flash`) are all still served by Atlas.

Verification:
- `test/helpers/secretKeys.test.js` → **7 passed, 0 failed**
- full suite → 3928 tests, 3564 pass, 143 fail

For the 143: a clean `upstream/main` worktree sharing the same `node_modules` gives **3926 / 3562 / 143**. So this branch adds two passing tests and no failures — the 143 are pre-existing in this environment, not introduced here.

No conflicts remain.

PR #1350, opened 2026-07-27
Fix meeting mic detection when a listener fails to launch

### LINKED ISSUE #1349: Linux meeting detection does not fall back when pactl is unavailable
**Describe the bug**

On Linux systems without `pactl`, enabling Detect meeting does not detect microphone activity. The attempted `pactl subscribe` process fails asynchronously, but the detector remains marked as event-driven and never starts its existing polling fallback.

**To Reproduce**

1. Use a Linux system with PipeWire tools available but without `pactl`.
2. Enable Detect meeting.
3. Start a meeting or another application that uses the microphone.
4. Observe that OpenWhispr does not show the meeting detection prompt.

**Expected behavior**

When `pactl subscribe` cannot start, the detector should use its existing polling path, including the `pw-cli` fallback, and detect microphone activity.

**Screenshots**

Not applicable.

**Desktop**

- OS: Linux (PipeWire, no `pactl`)
- OpenWhispr: current `main` at `ab201b3900caf582e9d70448414c83935fd7c595`

**Additional context**

`child_process.spawn()` reports a missing executable through its asynchronous `error` event. The current Linux startup path returns success before that event, so the fallback handler runs before event-driven mode has been enabled and polling is not started.

fabrizio2210:
Fixes #1349

## What

- Wait for the OS to confirm a mic listener has started before treating event-driven detection as active, on all three platforms.
- Fall back to the existing polling path whenever a listener cannot start.
- Ignore listener events that arrive after detection is stopped or restarted.

## Why

`spawn()` reports a failed launch through the asynchronous `error` event, which Node drains from the nextTick queue *before* the promise microtask awaiting `start()`. The listener therefore reported success, `_eventDriven` was still `false` when the error handler ran, and the polling fallback never started — leaving mic detection silently dead.

This was first found on PipeWire-only Linux systems, where `pactl` is often not installed at all. It is not Linux-specific: the macOS and Windows listeners race the same way whenever their binary resolves but fails to exec (quarantine, wrong architecture, `EACCES`, or deletion between the `existsSync` check and the spawn). Because the defect lives in the shared `start()` contract rather than in one platform branch, the spawn handshake now happens in a single helper that all three platforms use.

Two related problems surfaced while fixing this:

- On a stop → start sequence, the dying listener's `exit` handler nulled the reference to the *new* listener, orphaning a live child process for the lifetime of the app and starting polling alongside it.
- Listener output buffered past `stop()` could still arm the sustained timer and emit `sustained-audio-detected` while the detector was stopped.

## Impact

Any system whose mic listener cannot launch now falls back to polling instead of silently detecting nothing — PipeWire-only Linux via the existing `pw-cli` path, and macOS/Windows via the existing process-name checks. Systems where the listener launches normally are unaffected: the same binaries, arguments, and spawn options are used, including the piped stdin and `windowsHide` the Windows listener depends on.

Refactoring the three platform paths onto the shared helper also removed the line-buffering loop each of them duplicated verbatim, net -121 lines.

## Validation

`npm run lint`, Prettier, and `npm test` under Node.js 24 — 714 → 735 tests, 0 failures, skips unchanged at 19.

New `test/helpers/audioActivityDetector.test.js` covers launch failure, clean launch, stop-during-launch, restart races, and post-stop events, parameterized across darwin/win32/linux. It stubs `child_process` with the repo's existing `Module._load` convention; the fake child emits `spawn`/`error` on `process.nextTick`, reproducing the exact ordering that caused the bug.

Run against each revision to confirm the tests are not vacuous:

| | pre-PR `main` | first commit | this branch |
|---|---|---|---|
| passing | 9 / 21 | 16 / 21 | **21 / 21** |

The five still failing at the first commit are precisely the macOS/Windows launch-failure cases and the post-stop event guard.

Also verified against real spawned processes: no orphaned children across stop/start races, listener killed on stop, real stdout lines parsed, and polling taking over when a running listener dies. Plus a manual check on a PipeWire-only Linux system with a locally built binary.

`test/helpers/windowsHelperSpawnHide.test.js` guards against a console window flashing on Windows via a source-pattern match, which was tied to the old call shape. It is updated for the new shape, and a behavioral assertion now checks that `windowsHide: true` and piped stdin actually reach `spawn`. Both guards were confirmed to fail when `windowsHide` is removed.

### DIFF
--- a/src/helpers/audioActivityDetector.js
+++ b/src/helpers/audioActivityDetector.js
@@ -47,10 +47,10 @@ class AudioActivityDetector extends EventEmitter {
   async start() {
     if (this._running) return;
     this._running = true;
-    const startGeneration = ++this._startGeneration;
+    const generation = ++this._startGeneration;
 
-    const started = await this._tryEventDriven(startGeneration);
-    if (!this._running || startGeneration !== this._startGeneration) return;
+    const started = await this._tryEventDriven(generation);
+    if (this._isStale(generation)) return;
 
     if (started) {
       this._eventDriven = true;
@@ -73,7 +73,6 @@ class AudioActivityDetector extends EventEmitter {
   stop() {
     if (!this._running) return;
     this._running = false;
-    this._startGeneration++;
     this._killListenerProcess();
     this._clearSustainedTimer();
     this._clearResetTimer();
@@ -148,18 +147,23 @@ class AudioActivityDetector extends EventEmitter {
     }
   }
 
+  // True once stop() or a newer start() has superseded the run that owns `generation`.
+  _isStale(generation) {
+    return !this._running || generation !== this._startGeneration;
+  }
+
   // ---------------------------------------------------------------------------
   // Event-driven approach
   // ---------------------------------------------------------------------------
 
-  async _tryEventDriven(startGeneration) {
+  async _tryEventDriven(generation) {
     switch (process.platform) {
       case "darwin":
-        return this._tryEventDrivenDarwin();
+        return this._tryEventDrivenDarwin(generation);
       case "win32":
-        return this._tryEventDrivenWin32();
+        return this._tryEventDrivenWin32(generation);
       case "linux":
-        return this._tryEventDrivenLinux(startGeneration);
+        return this._tryEventDrivenLinux(generation);
       default:
         return false;
     }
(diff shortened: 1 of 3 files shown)


fabrizio2210:
@gabrielste1n, could you please review and approve this minimal Linux meeting-detection fix when you have a chance? It fixes #1349 and has been verified with a locally built binary.

PR #1351, opened 2026-07-27
ci: fail loudly when DB-backed tests cannot run (REQUIRE_DB_TESTS ratchet)

xAlcahest:
## Summary

DB-backed tests skip themselves when the better-sqlite3 native binding can't load, which is the right call locally where the binding is built against Electron's ABI. In CI it's the opposite: `npm rebuild better-sqlite3` runs right after `npm ci --ignore-scripts`, so the binding must load, and a skip there means the rebuild step is gone. This adds `REQUIRE_DB_TESTS` to the test step in both workflows and a shared `skipOrFail` helper in `test/helpers/harness/db.js` that turns that skip into a hard failure when the variable is set. The message names the missing step and says to restore it rather than delete the check, because putting the skip back is the tempting fix for a red build.

It exists because #1285's merge commit silently reverted #1291's rebuild and typecheck steps, every DB test went back to skipping, and CI stayed green with no database coverage. #1299 restored the steps; this makes the next drop fail the build instead of greening it. Phase 0 of the sync test coverage plan.

One behaviour change worth flagging. The three files each carried their own copy of the binding detector and they had drifted: only `translationHistory` checked `error.code`, so locally the first `new Database()` in a process skipped and every later one hard-failed, which is 8 ambient failures on main today. Unifying on the superset detector turns those 8 into clean skips. No test lost assertions and the pass count is identical at 698, but a local `npm test` goes from 8 red to green, so that green isn't the ratchet being defanged.

## Changes

- `test/helpers/harness/db.js`: new shared harness exporting `isNativeBindingUnavailable` and `skipOrFail`, which rethrows non-binding errors, fails with the rebuild-step message when `REQUIRE_DB_TESTS` is set, and skips otherwise
- `test/helpers/snippetsDatabase.test.js`: dropped the local detector, both catch blocks call `skipOrFail`
- `test/helpers/noteCloudUpsertGuard.test.js`: same migration
- `test/helpers/translationHistory.test.js`: same migration, its Electron ABI note now lives in the harness
- `.github/workflows/tests.yml`: `REQUIRE_DB_TESTS: "1"` on the Run tests step env
- `.github/workflows/build-and-notarize.yml`: same, on its Run tests step env

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,3 +24,5 @@ jobs:
         env:
           # keeps require("electron") from throwing when postinstall was skipped
           ELECTRON_OVERRIDE_DIST_PATH: /tmp
+          # a skipped DB test here means the rebuild step above is gone: fail instead
+          REQUIRE_DB_TESTS: "1"
--- a/.github/workflows/build-and-notarize.yml
+++ b/.github/workflows/build-and-notarize.yml
@@ -34,6 +34,8 @@ jobs:
         env:
           # keeps require("electron") from throwing when postinstall was skipped
           ELECTRON_OVERRIDE_DIST_PATH: /tmp
+          # a skipped DB test here means the rebuild step above is gone: fail instead
+          REQUIRE_DB_TESTS: "1"
 
   build-linux:
     needs: tests
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,11 +24,17 @@ jobs:
           node-version-file: .nvmrc
       - name: Install dependencies
         run: npm ci --ignore-scripts
+      - name: Build better-sqlite3 for the runner's node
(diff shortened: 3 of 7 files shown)


.github/workflows/tests.yml
```diff
@@ -24,3 +24,5 @@ jobs:
         env:
           # keeps require("electron") from throwing when postinstall was skipped
           ELECTRON_OVERRIDE_DIST_PATH: /tmp
+          # a skipped DB test here means the rebuild step above is gone: fail instead
+          REQUIRE_DB_TESTS: "1"
```
boseq, line 28: The release workflow still runs tests without rebuilding `better-sqlite3` or setting `REQUIRE_DB_TESTS`, so DB tests can still silently skip there.

boseq (approved):

PR #1352, opened 2026-07-27
test(sync): shared harness, SyncService orchestration tests, and #1290 kill-chain scenarios

xAlcahest:
## Summary

Three layers, each protecting something the current suite does not. The harness in `test/helpers/harness/` wires a real `DatabaseManager` behind a fake `electronAPI` and runs the real `cloudApi.ts`/`NotesService` against an in-memory cloud, so a test drives the actual sync code end to end instead of a mock of it. The orchestration tests assert on what `SyncService` *sends*, not only on what the DB ends up holding, because the primary defect in #1290 was request shape: a migration PATCH carrying `client_note_id` alone leaves the local row looking perfectly synced. The scenarios run whole `syncAll()` passes over real SQLite across relaunches and two devices, and every one of them ends in `assertNoContentRegression`.

The gold acceptance test is the evidence that any of this is worth having. Checking out `SyncService.ts` and `database.js` at `fa47857a^` into an otherwise-current tree makes S1 fail with the production corpse: `AssertionError`, `actual: ''`, where a 70 KB transcript should be, both in the local SQLite row and in the cloud copy. Restoring those two files makes it pass, in the same worktree with the same modules, so there is no environmental delta to hide behind. Reverting only the push-wiring hunk fails `migration push sends the full note body, not just the client id` plus five scenarios; reverting only the timestamp-gate hunk fails `pullNotes applies a strictly newer cloud copy and rejects older and tied ones` and its folders equivalent.

That same revert sweep initially surfaced one gap: reverting the **conversations** pull gate left the entire suite green. Closed now — `syncServiceConversations.test.js` mirrors the notes gate test end to end through `pullConversations` (same-UTC-day older ISO copy rejected, strictly newer applied, ties rejected, plus a tombstone pass), and reverting that gate alone fails exactly that test while everything else stays green. S5's tie-skip case remains a `test.todo`, visible instead of quietly omitted.

The test script now runs `node --import tsx`, because `SyncService.ts` imports `./NotesService.js`-style specifiers that resolve to `.ts` files and Node's type stripping will not remap them. This is Phases 1-3 of the sync test coverage plan, with the API-side suite as Phase 4. `test/helpers/harness/db.js` is also created by #1351; lines 1-30 are byte-identical here, so the two reconcile on a single export line in either merge order.

## Changes

- `test/helpers/harness/electronStub.js`: the `Module._load` electron fake and `NODE_ENV=test`, extracted from the pattern already used by the DB tests
- `test/helpers/harness/db.js`: `createDb(t)` returning a real `DatabaseManager` over a private tmpdir, on top of the binding-detection helpers shared with #1351
- `test/helpers/harness/browserGlobals.js`: Map-backed `localStorage`, and `navigator.locks.request` with real `ifAvailable` semantics including FIFO queueing when the lock is held
- `test/helpers/harness/electronApiBridge.js`: `window.electronAPI` built from a real `DatabaseManager`, 40 methods mirroring the IPC surface, including the dictionary and snippet methods `syncAll()` asserts up front
- `test/helpers/harness/fakeCloud.js`: in-memory cloud with the hardened server semantics each behind its own flag (material-PATCH gate, COALESCE on conflict, stale-write guard, `updated_at` clamp), a request log, one-shot fault injection, and a read-side conversations surface serving seeded rows with tombstones
- `test/helpers/harness/invariants.js`: `assertNoContentRegression`, which encodes the user-visible harm as a single assertion
- `test/helpers/syncServiceNotes.test.js`: migration and fresh push paths, pull gate wired end to end, tombstones, cursor progression, gating and lock behavior, error containment, 401 handling
- `test/helpers/syncServiceFolders.test.js`: folder mapping both directions, default adoption by name, folder push and delete request shapes, folder pull gate and cursor
- `test/helpers/syncServiceConversations.test.js`: conversations pull gate end to end through `pullConversations`, plus a tombstone hard-delete pass
- `test/helpers/syncScenarios.test.js`: S1 the meeting-shell acceptance test in hardened and legacy cloud modes, S2 cross-device convergence, S3 deletes and the delete-vs-edit race, S4 crash and offline, S5 pagination edges, S6 folder lifecycle
- `package.json`: `tsx` devDependency and the loader in the test script
- `src/services/SyncService.ts`: export the class so tests can construct isolated instances; the singleton export is unchanged

### DIFF
--- a/test/helpers/harness/fakeCloud.js
+++ b/test/helpers/harness/fakeCloud.js
@@ -1,0 +1,410 @@
+// In-memory stand-in for the cloud API behind window.electronAPI.cloudApiRequest.
+// Notes and folders are modelled with the hardened server's real write rules;
+// the other entity types exist only so a full syncAll() pass can complete.
+//
+// Each hardening rule sits behind its own flag so a test can flip the server
+// back to the legacy behaviour and prove the client still does not lose data:
+//   materialPatchGate  a PATCH that changes nothing does not bump updated_at
+//   coalesceOnConflict an absent/null incoming field keeps the stored value
+//   staleWriteGuard    a write whose basis predates the stored row is rejected
+//   updatedAtClamp     a stored updated_at never moves backwards
+
+// The server compares timestamps itself; it must not borrow the client's guard,
+// or a client-side regression would be invisible here.
+function normalizeTimestamp(value) {
+  if (!value) return "";
+  const iso = String(value).replace(" ", "T").replace(/Z$/, "");
+  return (/\.\d+$/.test(iso) ? iso : `${iso}.000`) + "Z";
+}
+
+function sameValue(a, b) {
+  return (a ?? null) === (b ?? null);
+}
+
+function compare(a, b) {
+  if (a === b) return 0;
+  return a < b ? -1 : 1;
+}
+
+class HttpError extends Error {
+  constructor(status, message, code) {
+    super(message);
+    this.status = status;
+    this.code = code;
+  }
+}
+
+const NOTE_FIELDS = [
+  "title",
+  "content",
+  "enhanced_content",
+  "note_type",
+  "enhancement_prompt",
+  "source_file",
+  "audio_duration_seconds",
+  "folder_id",
+  "transcript",
+  "enhanced_at_content_hash",
+  "participants",
+  "calendar_event_id",
+  "diarization_enabled",
+  "expected_speaker_count",
+];
+
+const FOLDER_FIELDS = ["name", "is_default", "sort_order"];
+
+const NOTE_DEFAULTS = {
+  title: null,
+  content: "",
+  enhanced_content: null,
+  note_type: "personal",
+  enhancement_prompt: null,
+  source_file: null,
+  audio_duration_seconds: null,
+  folder_id: null,
+  transcript: null,
+  enhanced_at_content_hash: null,
+  participants: null,
+  calendar_event_id: null,
+  diarization_enabled: null,
+  expected_speaker_count: null,
+};
+
+const FOLDER_DEFAULTS = { name: "", is_default: false, sort_order: 0 };
+
+function createFakeCloud(config = {}) {
+  const cfg = {
+    materialPatchGate: true,
+    coalesceOnConflict: true,
+    staleWriteGuard: true,
+    updatedAtClamp: true,
+    ...config,
+  };
+
+  const notes = new Map();
+  const folders = new Map();
+  const log = [];
+  const failures = [];
+  let seq = 0;
+  let clock = () => new Date().toISOString();
+
+  const nextId = (prefix) => `${prefix}-${String(++seq).padStart(4, "0")}`;
+
+  // The server stores timestamps and serves them back as ISO, whatever format
+  // the client sent. Local rows carry SQLite's "YYYY-MM-DD HH:MM:SS", so echoing
+  // the input back would hide the format skew every pull comparison has to
+  // survive.
+  const toIso = (value) => normalizeTimestamp(value);
+
+  // A stored updated_at only ever moves forward under the clamp.
+  function bump(row, proposed) {
+    const candidate = toIso(proposed ?? clock());
+    if (cfg.updatedAtClamp && candidate < toIso(row.updated_at)) {
+      return row.updated_at;
+    }
+    return candidate;
+  }
+
+  function applyWrite(collection, row, incoming) {
+    if (
+      cfg.staleWriteGuard &&
+      incoming.updated_at &&
+      normalizeTimestamp(incoming.updated_at) < normalizeTimestamp(row.updated_at)
+    ) {
+      throw new HttpError(409, "stale write rejected: basis is older than the stored row", "stale");
+    }
+
+    const next = {};
+    let material = false;
+    for (const key of collection.fields) {
+      if (!(key in incoming)) {
+        next[key] = row[key];
+        continue;
+      }
+      const value = incoming[key];
+      if (cfg.coalesceOnConflict && (value === null || value === undefined) && row[key] != null) {
+        next[key] = row[key];
+        continue;
+      }
+      next[key] = value ?? null;
+      if (!sameValue(next[key], row[key])) material = true;
+    }
+
+    // Without the gate a no-op PATCH still bumps updated_at, which is what hands
+    // the next pull a cloud copy that never received the local edit.
+    if (cfg.materialPatchGate && !material) return row;
+
+    Object.assign(row, next);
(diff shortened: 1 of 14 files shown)


gabrielste1n:
## Code quality review

Checked this out and ran it for real — built `better-sqlite3` against Node 24 in a scratch dir so the local Electron-ABI build stayed untouched.

### Verification

| Check | Result |
|---|---|
| Full suite (`test/**/*.test.js`) | **756 tests, 750 pass, 0 fail, 5 skip, 1 todo**, 3.2s |
| The four new sync files | **41 pass, 1 todo, 0 fail** |
| ESLint + Prettier on new files | clean |
| All 48 harness `electronAPI` methods vs `preload.js` | every name exists, arg order matches |

I also ran the central claim as two mutations rather than taking it on faith:

- Dropping `...buildNoteUpdatePayload(note, folderMap)` from the migration PATCH (`SyncService.ts:496`) → **6 failures**: `migration push sends the full note body` plus S1×2, S2, S3, S6. Exactly the set the description claims.
- Replacing `isCloudEntryNewer(...)` with a raw `>` at all three call sites → **4 failures**, one per entity: notes, folders, **and** conversations.

So the kill chain is real, and the conversations gap the description says it closed is genuinely closed. Asserting on the wire payload rather than only on the surviving row is the right call — it's the only way the #1290 shape is visible at all.

---

### Findings

**1. `REQUIRE_DB_TESTS` is never set by any workflow — the guard is dead code.** `test/helpers/harness/db.js:1-29` builds a 28-line skip-or-fail mechanism whose whole purpose is to fail loudly in CI when the `npm rebuild better-sqlite3` step vanishes. Nothing in `.github/workflows/` sets that env var, so it can never fire. Fix: add `REQUIRE_DB_TESTS: "1"` to the existing `env:` block in `.github/workflows/tests.yml:23` and in `build-and-notarize.yml`.

**2. `release.yml` runs `npm test` with no rebuild step.** `.github/workflows/release.yml:26-31` does `npm ci --ignore-scripts` then `npm test`, with no `npm rebuild better-sqlite3`. All 42 new tests silently skip on the release path. Pre-existing gap, but this PR is what makes it expensive.

**3. The bridge overclaims fidelity.** `test/helpers/harness/electronApiBridge.js:1-3` says every method mirrors "its ipcHandlers.js passthrough, so a drift between SyncService and the real IPC surface shows up as a test failure." The four hard-delete handlers aren't passthroughs — `ipcHandlers.js:1519`, `1547`, `1580`, `1604` also fire `_asyncVectorDelete`, `_asyncMirrorDelete`, markdown-mirror removal, and window broadcasts. Every tombstone test here covers the DB row only; an orphaned vector entry would stay green. Worth narrowing the comment to what's actually modelled.

**4. DRY — helpers duplicated across the four test files.** `queryParam` appears verbatim in three (`syncServiceNotes.test.js:58`, `syncServiceConversations.test.js:52`, `syncScenarios.test.js:48`); the same generic `UPDATE <table> SET …` helper is written four times under three names (`updateNote`, `updateConversation`, `setFolderRow`, `setNoteRow`); plus `silenceConsoleErrors`, `shift`, `noteByClientId`, `noteCount`, `folderByName` in two or three each. `harness/` is the obvious home.

**5. Two conventions for the same thing inside one PR.** `syncServiceNotes` / `syncServiceConversations` / `syncScenarios` `require(".../SyncService.ts")` and construct directly; `syncServiceFolders.test.js:19-36` alone uses a top-level `installBrowserGlobals()` plus `const loadSync = () => import(...)`, adding `await (await newSyncService())` to all eleven of its tests for no stated benefit. The three `setup()` bodies also disagree on ordering — folders resets globals *before* `createDb`, the others after. Pick one and hoist `setup()` into the harness.

**6. Dead exports.** `getUserDataDir` (`electronStub.js:37`) has no consumer anywhere. `isNativeBindingUnavailable` / `skipOrFail` (`db.js:78`), `CONTENT_FIELDS` (`invariants.js:64`), `normalizeTimestamp` (`fakeCloud.js:451`), `cloud.config` and `cloud.conversations()` are exported but used only inside their own module. `const api = createElectronApi(...)` is returned from `setup()` in the notes and conversations files and never read.

**7. `syncServiceFolders.test.js` never silences `console.error`** the way its siblings do — its four failure-injection tests dump `CloudApiError` stack traces into the run output. Confirmed in a live run.

**8. Two data-loss behaviours are locked in by assertion with no issue link.** S3's tombstone destroying B's unsynced edit (`syncScenarios.test.js:403-406`) and S1-legacy leaving the cloud copy empty (`syncScenarios.test.js:250-252`). The comments correctly frame these as "documented current behaviour, not a preference", but nothing connects them to a decision to keep or fix. Same for the `test.todo` at `syncScenarios.test.js:652`, which pins a real bug: the `if (next === cursor) break` stall guard at `SyncService.ts:582` abandons the rest of a pull when a full delta page ties on `updated_at`.

---

### Verdict

Approve once **#1** and **#2** are in — without them the CI story is "2,700 lines of tests that can silently skip." The rest is cleanup and shouldn't block.

Production risk is essentially nil: the only source change is `class SyncService` → `export class SyncService`, with the singleton export untouched.

test/helpers/harness/invariants.js
```diff
@@ -0,0 +1,64 @@
...
+  return (cloudRow[field] ?? null) === (after[field] ?? null);
+}
+
+function assertNoContentRegression(dbBefore, dbAfter, cloudTruth) {
+  const before = indexByClientId(dbBefore);
+  const cloud = indexByClientId(
+    cloudTruth instanceof Map ? [...cloudTruth.values()] : (cloudTruth ?? [])
+  );
+
+  for (const after of dbAfter ?? []) {
```
boseq: This only checks notes that still exist after sync. If sync accidentally hard-deletes a note, the invariant passes. I reproduced it with a populated `before` and empty `after`.

test/helpers/harness/fakeCloud.js
```diff
@@ -0,0 +1,451 @@
...
+  function page(collection, query, orderKey) {
+    const limitRaw = Number(query.get("limit"));
+    const limit = Number.isFinite(limitRaw) && limitRaw > 0 ? limitRaw : 50;
+    const since = query.get("since");
+    const before = query.get("before");
+    let rows = [...collection.store.values()];
+
+    if (since) {
+      const bound = normalizeTimestamp(since);
+      rows = rows.filter((r) => normalizeTimestamp(r.updated_at) > bound);
```
boseq: The fake cloud ignores `before_id` and `since_id`. With 51 notes sharing a timestamp, it returns 50 on the first page and none on the second, dropping one note.

boseq (approved):

PR #1354, opened 2026-07-27
fix(sync): close the latent sync defects (mark-synced race, tombstone loss, dead code)

xAlcahest:
## Summary

Three latent defects in the same family as #1290, each landed failing-test-first so the test that proves the bug ships with the fix. `markNoteSynced` was a bare UPDATE, so an edit landing between the push snapshot and the ack got flipped to synced and was never re-pushed: before the fix the new test fails with the row at `synced` where `pending` is expected, after it both the primitive and the end-to-end race pass, with the second pass provably re-pushing the newer content. It now mirrors `markSnippetSynced`: the callers pass the snapshot they pushed (updated_at, content, title, transcript, enhanced_content), the WHERE requires the row to still match, and zero changes leaves the row pending. The new parameters thread through ipcHandlers, preload, the electron types, the three SyncService call sites and the test harness bridge. `pullNotes` lacked the tombstone skip `pullFolders` has, so a newer cloud copy resurrected a locally deleted note and stranded its pending delete; the failing test shows the tombstone flipped to `synced`, the fix is the same one-line skip folders use. The delete loops for notes and conversations now skip rows without a cloud_id; both pending-delete queries already filter these in SQL, so this is defense in depth, and the tests say so. Also removes the dead top-level NoteEditor component (the live one is `notes/NoteEditor.tsx`).

Stacked on #1352, which must merge first: the diff contains its commits until then, and the new tests use its harness.

One edge the new guard creates, worth knowing: a note deleted while its very first push is in flight never gets its cloud_id attached (the ack's WHERE includes `deleted_at IS NULL`), so it becomes a never-synced local tombstone. Before the guard that same race left the tombstone wrongly marked synced instead, so this converges with an existing gap rather than adding one, but it makes the open question below more relevant.

Two remaining defects need design decisions and are left as open questions rather than fixes. First, the pull cursor tie-skip: when a delta page returns exactly BATCH_SIZE rows sharing one updated_at, `next === cursor` breaks the loop and silently skips the rest; notes, conversations and transcriptions are affected, dictionary and snippets are immune via composite cursors, and the durable fix is an API-side composite cursor (updated_at plus id), pinned as the S5 test.todo in `syncScenarios.test.js`. Second, stranded never-synced tombstones: a row with deleted_at set and cloud_id NULL matches neither `getPendingNotes` (deleted_at IS NULL) nor `getPendingNoteDeletes` (cloud_id IS NOT NULL), so it sits in the local database forever; deciding when it is safe to hard-delete local-only tombstones is a purge-policy call.

## Changes

- `src/helpers/database.js`: markNoteSynced guards the ack against the pushed snapshot and reports changes, mirroring markSnippetSynced
- `src/services/SyncService.ts`: the three markNoteSynced call sites pass their pushed snapshot; pullNotes skips locally tombstoned rows; pushNoteDeletes and pushConversationDeletes skip rows without a cloud_id
- `src/helpers/ipcHandlers.js`, `preload.js`, `src/types/electron.ts`: the new markNoteSynced parameters threaded through
- `test/helpers/harness/electronApiBridge.js`: harness bridge mirrors the new signature
- `test/helpers/markNoteSyncedGuard.test.js`: the lost-update regression, primitive and end-to-end
- `test/helpers/syncDeleteGuards.test.js`: tombstone-survives-pull and the never-synced delete filters
- `src/components/NoteEditor.tsx`: removed, dead code


package.json
```diff
@@ -72,7 +72,7 @@
     "format": "npm run format:js",
     "format:check": "eslint . && cd src && eslint . && prettier --check \"**/*.{js,jsx,ts,tsx,json,css,md}\"",
     "lint": "eslint . && cd src && eslint .",
-    "test": "node --test \"test/**/*.test.js\"",
+    "test": "node --import tsx --test \"test/**/*.test.js\"",
```
boseq, line 75: Is this meant to be stacked on #1352? The first four commits are identical, so most of this diff is inherited and difficult to review here.

boseq (approved):

Issue #1356, opened 2026-07-27
feat: add Vercel AI Gateway as a first-class provider


cmglezpdev:
**Is your feature request related to a problem? Please describe.**

Vercel AI Gateway (https://ai-gateway.vercel.sh/v1) routes to hundreds of models (OpenAI, Anthropic, Google, xAI, …) with a single API key and unified billing/observability. It can be used with OpenWhispr today through the **Custom** provider tab, since its chat endpoint is OpenAI Chat Completions-compatible — but that setup has rough edges:

- The base URL and key must be re-entered per scope instead of picking a named provider once; there is no provider tab, icon, or dedicated key management entry.
- For **speech-to-text** the Custom path cannot work at all: the gateway does not expose an OpenAI-compatible `/v1/audio/transcriptions` endpoint. Its transcription API is a different shape (`POST /v4/ai/transcription-model` with base64 audio in a JSON body and the model id in an `ai-model-id` header), so no amount of base-URL configuration reaches it. Today this requires running a local translation proxy.

**Describe the solution you'd like**

Add Vercel AI Gateway as a first-class provider, in two steps:

1. **Reasoning/chat (PR ready):** a `vercel` inference provider following the same pattern as OpenRouter — locked base URL, dynamic model listing via `/models`, BYOK key stored through the existing `secretKeys.js` manifest, and a provider tab in the model selector. Same integration shape as Tinfoil (#875, #944) and Corti (#1111). I have this implemented and tested (typecheck, test suite, and manual testing of both the dictation cleanup and agent chat streaming paths against the live gateway) and will open the PR linked to this issue.

2. **Speech-to-text (follow-up, question below):** a `vercel` transcription provider following the xAI STT precedent (#942), using the gateway's native transcription endpoint.

**Question for maintainers regarding step 2:** the branches `refactor/stt-provider-routing` and `refactor/stt-proxy-provider-registry` look like they are collapsing exactly the per-provider branching a new STT provider would need to extend. Should the STT part wait for those to land and build on the registry/resolver, or is a PR against current `main` acceptable?

**Describe alternatives you've considered**

- **Custom tab:** works for chat completions (this is what I use for text today), but with the per-scope configuration overhead described above, and it cannot cover transcription.
- **Local translation proxy:** what I currently run to bridge the gateway's transcription API — works, but shouldn't be necessary.

**Additional context**

PR #1357, opened 2026-07-27
feat: add Vercel AI Gateway reasoning provider

cmglezpdev:
Part of #1356 (reasoning/chat support; the speech-to-text follow-up is tracked there).

## Why

Vercel AI Gateway routes to hundreds of models with a single API key. It already works through the Custom tab for chat, but as a first-class provider it gets a named tab, its own encrypted key entry, and one-click model selection. It's the same treatment OpenRouter has. Full code in #1356.

## What

- New `vercel` inference provider mirroring the `groq.ts` pattern: fixed base URL (`https://ai-gateway.vercel.sh/v1`) + Chat Completions via the shared helper.
- BYOK key storage: one entry in the `secretKeys.js` manifest (mirrored in `preload.js`), reusing the generic IPC/encryption plumbing.
- Wiring: `ReasoningService` (cleanup + streaming paths), `getAIModel()` (`createOpenAI({ baseURL }).chat(model)` — Chat Completions, not Responses API), chat routing, and `ModelRegistry` (vendor-prefixed model ids like `openai/gpt-4o` resolve `max_tokens` correctly).
- UI: "Vercel AI Gateway" tab in the model selector using `OpenAICompatiblePanel` with a locked base URL and dynamic model listing via `/models` (OpenRouter-style — no static catalog, so no i18n additions), plus provider icon.

## Test plan

- `npm run typecheck`, `npm test` (696 passing, includes a new `chatRouting` case for the provider route), `npm run lint`, `npm run format`
- Manual, against the live gateway: key persistence across app restarts (encrypted storage), dynamic model list loading, dictation cleanup path, and agent chat streaming path
- The gateway sends permissive CORS headers, so direct renderer fetch works like Groq/OpenRouter. No IPC proxy needed

### DIFF
--- a/preload.js
+++ b/preload.js
@@ -14,6 +14,7 @@ const BYOK_KEY_BRIDGES = [
   { base: "openrouter", get: "getOpenrouterKey", save: "saveOpenrouterKey" },
   { base: "tinfoil", get: "getTinfoilKey", save: "saveTinfoilKey" },
   { base: "corti", get: "getCortiKey", save: "saveCortiKey" },
+  { base: "vercel", get: "getVercelKey", save: "saveVercelKey" },
 ];
 const secretKeyApi = {};
 for (const k of BYOK_KEY_BRIDGES) {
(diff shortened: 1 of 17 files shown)


cmglezpdev:
Anyone here to review this?

PR #1361, opened 2026-07-28
refactor(dictionary): add a delta write path and move word edits onto it

gabrielste1n:
**Stacked on #1360** — base that first, then this retargets to `main` automatically.

Retires the pattern behind #1295 rather than guarding it.

## The problem

`setDictionary(words[])` replaces the whole table: anything the caller omits is deleted. Every word-level edit went through it as read-modify-write on a snapshot:

```ts
setCustomDictionary([...customDictionary, ...words]);        // add
setCustomDictionary(customDictionary.filter(w => w !== word)); // remove
```

So any caller holding a stale snapshot silently destroys everything it doesn't know about. That's the mechanism that deleted the reporter's 357 rows. #1304 closed the two known triggers, but the loaded gun stayed on the table.

## The change

`applyDictionaryChanges({ add, remove })` — a caller can only affect the words it names. The wipe stops being expressible rather than merely guarded.

Rather than reimplementing the mutation rules, I extracted the existing ones into shared helpers (`_upsertDictionaryWord`, `_deleteDictionaryRow`, `_dictionaryWriteStatements`) so the delta and whole-list paths physically cannot drift. `setDictionary` is now built from the same pieces — that's why its existing tests pass untouched.

**Moved onto the delta path:** add / import / remove / edit / clear-all in `DictionaryView`, the agent-name sync, and the two auto-learn read-modify-write paths in the main process — the riskiest of the set, since they raced with the renderer.

**`setDictionary` stays** for genuine replace-everything: settings restore, and the first write into an empty database. Both are correct uses.

## Two bugs found on the way

- **Agent rename wrote the dictionary twice.** `DictationAgentSettings` recomputed the list itself and wrote it *after* `setAgentName` had already written — from a snapshot taken before that write, and matching case-insensitively where the store matched case-sensitively. Removed the duplicate; `setAgentName` already owns this.
- **Dead destructures** of `setCustomDictionary` in `useSettings` and `SettingsPage` (pre-existing, unused). Removed, since this PR is specifically about narrowing who can reach that function.

## Compatibility

- No schema change, so nothing to migrate and a clean rollback.
- The cloud contract is untouched — it was already per-record (`batchCreate`/`update`/`delete`).
- `db-set-dictionary` and `electronAPI.setDictionary` are unchanged; the delta channel is additive.
- The CLI bridge has no dictionary surface, so there's no separately-versioned consumer to break. I checked before starting.
- An older preload without the delta channel falls back to the whole-list write rather than dropping the edit — same behavior as today, no worse.

## Validation

Run with a node-ABI `better-sqlite3` so the DB-backed tests actually execute rather than skipping:

- `npm run typecheck`, `npm run lint`, `npm run format:check`, `npm run i18n:check` — all clean
- Full suite — 752 tests, 747 passed, 0 failed, 5 skipped (pre-existing)
- 11 new tests for `applyDictionaryChanges`, covering the delete rules, tombstone restore, source promotion, both-sides-wins, empty delta not re-marking rows pending, and case-variant dedupe
- The headline one: *a caller holding a stale one-word view cannot delete anything*
- `setDictionary`'s existing tests pass unmodified, which is the evidence the extraction preserved semantics

### DIFF
--- a/preload.js
+++ b/preload.js
@@ -79,6 +79,7 @@ contextBridge.exposeInMainWorld("electronAPI", {
   // Dictionary functions
   getDictionary: () => ipcRenderer.invoke("db-get-dictionary"),
   setDictionary: (words) => ipcRenderer.invoke("db-set-dictionary", words),
+  applyDictionaryChanges: (changes) => ipcRenderer.invoke("db-apply-dictionary-changes", changes),
   onDictionaryUpdated: (callback) => {
     const listener = (_event, words) => callback?.(words);
     ipcRenderer.on("dictionary-updated", listener);
--- a/src/components/DictionaryView.tsx
+++ b/src/components/DictionaryView.tsx
@@ -23,7 +23,7 @@ import { parseDictionaryImportText } from "../helpers/dictionaryImport";
 
 export default function DictionaryView() {
   const { t } = useTranslation();
-  const { customDictionary, setCustomDictionary } = useSettings();
+  const { customDictionary, updateCustomDictionary } = useSettings();
   const agentName = getAgentName();
   const { toast } = useToast();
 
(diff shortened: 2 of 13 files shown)


gabrielste1n:
Correction: this merged into its stacked base branch (`fix/dictionary-client-id-trigger`) rather than `main` — I merged it before GitHub had retargeted the base. The content reached `main` via #1365. Nothing was lost; noting it so the merge record isn't misleading.

PR #1362, opened 2026-07-28
feat(dictionary): expose list and bulk update over the CLI bridge

gabrielste1n:
**Stacked on #1361** (which is stacked on #1360). Merge in order; each retargets automatically.

Closes the loop on #1295 by removing the reason anyone writes to the dictionary table by hand.

## Why

The reporter wrote 357 contact names straight into SQLite because there was no supported programmatic path — no CLI command, no local endpoint, no documented import format. Those rows were wiped on launch, and because they had no `client_dict_id` they could never have reached the cloud either. The wipe is fixed; the missing feature that caused it isn't.

## What

Two routes on the existing local bridge:

- `GET /v1/dictionary/list` — current words
- `POST /v1/dictionary/update` — `{ add: [...], remove: [...] }`

The endpoint takes a **delta, not a word list**, so a bulk import cannot delete anything it didn't name — the same property the rest of this stack is built on. Writes go through `applyDictionaryChanges`, so rows carry the sync columns the cloud push needs, and renderers get a `dictionary-updated` broadcast.

```bash
curl -s -X POST localhost:$PORT/v1/dictionary/update \
  -H "Authorization: Bearer $TOKEN" \
  -d '{"add": ["Ada Lovelace", "Grace Hopper"]}'
```

## Note on sync timing

Words added this way are marked pending and upload on the next sync cycle rather than immediately. The renderer broadcast deliberately doesn't trigger a sync — that would loop, since cloud pulls emit the same broadcast. No data is at risk; it's just eventual rather than instant.

## Validation

- `npm run typecheck`, `npm run lint`, `npm run format:check` — clean
- Full suite — 759 tests, 754 passed, 0 failed, 5 skipped (pre-existing)
- 7 new tests, including the one that matters: words added via the endpoint are present in `getPendingDictionary()` **with a `client_dict_id`**, i.e. actually uploadable — the exact thing the reporter's manual inserts failed at

This adds the **first test coverage for the bridge route table**. The constructor builds routes without starting a server, so handlers are exercised directly against a real database with broadcasts stubbed. That harness is reusable for the other routes, none of which are currently covered.

## Follow-up, not included

The `@openwhispr/cli` package needs a `dictionary` command to consume this — separate repo, separate PR. The endpoint is the enabling half and is useful on its own via curl.

### DIFF
--- a/src/helpers/cliBridge.js
+++ b/src/helpers/cliBridge.js
@@ -262,6 +262,16 @@ class CliBridge {
       return id;
     };
 
+    const requireWordList = (value, field) => {
+      if (value === undefined || value === null) return [];
+      if (!Array.isArray(value) || value.some((w) => typeof w !== "string")) {
+        const err = new Error(`'${field}' must be an array of strings`);
+        err.code = "VALIDATION";
+        throw err;
+      }
+      return value;
+    };
+
     const requireSuccess = (result, message) => {
       if (!result?.success) {
         const err = new Error(result?.error || message);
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Superseded by #1366 — same content, re-targeted directly at `main`. This one was stacked on a branch that never reached main.

PR #1371, opened 2026-07-28
fix(privacy): wait for retention sync before cleanup

### LINKED ISSUE #1370: Audio retention cleanup runs before the disabled setting is restored
**Describe the bug**

On startup, the main process runs audio cleanup with the hard-coded default of 30 days before the renderer restores and syncs the persisted Audio Retention setting. A user who selected **Disabled** (`0`) can therefore still lose stored audio older than 30 days whenever the app launches.

**To Reproduce**

1. Keep a saved transcription audio file whose modification time is more than 30 days old.
2. Set **Settings → Privacy → Audio Retention** to **Disabled**.
3. Quit and relaunch OpenWhispr.
4. The startup cleanup runs before `retention-settings-changed` arrives and deletes the file using the default 30-day period.

A deterministic code-level reproduction on current `main` calls `IPCHandlers.prototype._setupRetentionCleanup` with an observed cleanup callback. The callback runs immediately and records `{ audioRetentionDays: 30, transcriptRetentionDays: 0 }` before any renderer sync.

**Actual behavior**

`_setupRetentionCleanup()` invokes `_runRetentionCleanup()` synchronously during `IPCHandlers` construction. Persisted renderer settings are sent later from the `useSettings` mount effect.

**Expected behavior**

Cleanup should wait until the first persisted settings sync. Audio Retention **Disabled** (`0`) must not purge existing files. The first sync should trigger one sweep even when it matches the defaults, while duplicate unchanged syncs from the second renderer window should remain de-duplicated.

**User impact**

Users who explicitly disable audio retention can still have audio older than 30 days deleted at launch.

**Root-cause hypothesis**

The retention lifecycle initializes `_retentionSettings` with defaults and starts cleanup before the renderer-owned persisted values cross the IPC boundary.

**Proposed focused scope**

- Defer startup and periodic cleanup until the first retention settings sync.
- Run cleanup on that first sync or a later value change.
- Add deterministic regression coverage for Disabled-at-startup, default first sync, and duplicate unchanged syncs.

**Desktop**

- Revision: `143971d3aa978403d0110fe82673263f3cdcca7a` (`upstream/main`)
- Node.js: 24.9.0
- OS: macOS (the ordering defect is platform-independent)

hsusul:
## Summary

- wait for persisted retention settings before the main process runs its first cleanup sweep
- preserve one first-sync cleanup while de-duplicating unchanged syncs from the second renderer window
- add deterministic regression coverage for Disabled, default, duplicate, and changed settings

## Problem

`IPCHandlers` ran retention cleanup during construction with the main-process default of 30 audio-retention days. Persisted settings live in renderer `localStorage` and arrive later over `retention-settings-changed`, so a user who selected Audio Retention **Disabled** (`0`) could still lose audio older than 30 days at launch.

## Root cause

`_setupRetentionCleanup()` invoked `_runRetentionCleanup()` before either renderer window could sync its persisted values. The existing change detection also treated an unchanged default first sync as a duplicate, so simply removing the eager call would have skipped the intended initial sweep for default settings.

## Solution

- mark retention settings as unsynchronized during main-process initialization
- defer startup and periodic cleanup until the first settings payload arrives
- apply the incoming settings before the first cleanup
- run the first sync once even when it matches the defaults, while retaining existing de-duplication for later unchanged syncs

The retention values, purge implementation, database behavior, and IPC payload shape remain unchanged.

## Tests

- `ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/retentionSettings.test.js test/helpers/transcriptRetention.test.js test/helpers/discardedRecording.test.js` — 13 passed, 0 failed
- `ELECTRON_OVERRIDE_DIST_PATH=/tmp npm test` — 862 passed, 0 failed, 6 environment-dependent skips

## Validation

- `npm ci --ignore-scripts` — passed
- `npm rebuild better-sqlite3` — passed
- `npm run lint` — passed
- `node --check src/helpers/ipcHandlers.js` — passed
- `node --check test/helpers/retentionSettings.test.js` — passed
- `git diff --check` — passed

The first sandboxed full-suite attempt could not open loopback listeners (`listen EPERM 127.0.0.1`). Re-running the same suite with loopback access passed with the counts above.

## Compatibility and risk

This is cross-platform lifecycle ordering only. It does not change public APIs, stored settings, retention periods, deletion semantics, or error shapes. If renderer settings never sync, cleanup now safely remains inactive instead of applying a destructive guessed default.

## Documentation

No documentation changes are needed because the user-facing retention contract is unchanged.

Fixes #1370.

PR #1368 introduced the renderer-to-main retention sync and states that Audio Retention Disabled must not purge existing files. This PR fixes the startup ordering gap without changing that feature's scope.

### DIFF
--- a/src/helpers/ipcHandlers.js
+++ b/src/helpers/ipcHandlers.js
@@ -514,6 +514,7 @@ class IPCHandlers {
     this.audioStorageManager = new AudioStorageManager();
     this._retentionCleanupInterval = null;
     this._retentionSettings = { ...DEFAULT_RETENTION_SETTINGS }; // Synced from renderer
+    this._retentionSettingsSynced = false;
     this._noteFilesEnabled = false;
     this.speakerDiarizationEnabled = true;
     this.activeMeetingSpeakerConfig = null;
@@ -785,8 +786,17 @@ class IPCHandlers {
 
   _setupRetentionCleanup() {
     const SIX_HOURS_MS = 6 * 60 * 60 * 1000;
-    this._runRetentionCleanup();
-    this._retentionCleanupInterval = setInterval(() => this._runRetentionCleanup(), SIX_HOURS_MS);
+    this._retentionCleanupInterval = setInterval(() => {
+      if (this._retentionSettingsSynced) this._runRetentionCleanup();
+    }, SIX_HOURS_MS);
+  }
+
+  _syncRetentionSettings(incoming) {
+    const firstSync = !this._retentionSettingsSynced;
+    const { changed, settings } = applyRetentionSettings(this._retentionSettings, incoming);
+    this._retentionSettings = settings;
+    this._retentionSettingsSynced = true;
+    if (firstSync || changed) this._runRetentionCleanup();
   }
 
   _runRetentionCleanup() {
@@ -1109,10 +1119,7 @@ class IPCHandlers {
     });
 
     ipcMain.on("retention-settings-changed", (_event, incoming) => {
-      const { changed, settings } = applyRetentionSettings(this._retentionSettings, incoming);
-      if (!changed) return;
-      this._retentionSettings = settings;
-      this._runRetentionCleanup();
+      this._syncRetentionSettings(incoming);
     });
 
     ipcMain.handle("delete-all-audio", async () => {
--- a/test/helpers/retentionSettings.test.js
+++ b/test/helpers/retentionSettings.test.js
@@ -6,6 +6,67 @@ const {
   applyRetentionSettings,
 } = require("../../src/helpers/retentionSettings");
 
+test("waits for persisted settings before running startup cleanup", () => {
+  const IPCHandlers = require("../../src/helpers/ipcHandlers");
+  const cleanupRuns = [];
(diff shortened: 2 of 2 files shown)


gabrielste1n:
Thanks @hsusul — your diagnosis and design are what shipped, in #1558.

I couldn't rebase this one. It was written before `main` replaced the inline `retention-settings-changed` listener with the owner-scoped `createRetentionSettingsHandler` (`retentionSettings.js:28`), so `_syncRetentionSettings` no longer has a place to live and the conflict isn't mechanical.

#1558 keeps your approach exactly: `_retentionSettingsSynced`, no sweep at startup, first sync runs the first cleanup. One addition your version didn't need — because the new handler only fires `onSettingsChanged` when the value *changed*, a user whose settings already match the 30/0 defaults would never trigger that first sweep, so the handler takes a `hasSynced` callback now.

Your three tests carried over in spirit and are in `test/helpers/retentionSettings.test.js`. Closing this in favour of #1558.

Issue #683, opened 2026-04-29
Parakeet dictation fails with EADDRINUSE on port 6006 — sherpa-onnx-ws orphaned across clean Quit after a stalled meeting session (macOS arm64)


beefstew:
**Describe the bug**

After a meeting transcription session stalled (only the first ~60 seconds were captured even though the meeting kept running), I cleanly Quit OpenWhispr from the menu bar and relaunched it. The `sherpa-onnx-ws-darwin-arm64` process spawned during the meeting **survived the clean Quit** and kept holding TCP port 6006. The next Parakeet dictation attempt then failed: the freshly spawned `sherpa-onnx-ws` could not bind, crashed on an uncaught `websocketpp::exception: Address already in use`, and Electron surfaced `parakeet-ws process died during startup`.

There appear to be two issues, probably related:

1. Meeting transcription silently stops capturing after the first minute.
2. `sherpa-onnx-ws` child processes are not terminated on a clean app Quit, so relaunching does not recover.

**To Reproduce**

1. Configure OpenWhispr to use local Parakeet transcription (`parakeet-tdt-0.6b-v3`).
2. Start a meeting that OpenWhispr is set to transcribe locally.
3. Leave the meeting running unattended for an extended period. (At some point in this window, transcript capture silently stops while the meeting itself continues.)
4. Quit OpenWhispr cleanly via the menu bar and relaunch it.
5. Trigger Parakeet dictation via hotkey.
6. See error: dictation fails immediately with `parakeet-ws process died during startup`.

**Expected behavior**

Either (a) the meeting transcription continues for the full meeting, or — if the sidecar got into a bad state — (b) the clean Quit reaps it so the new session spawns a fresh `sherpa-onnx-ws` cleanly.

**Screenshots**

N/A — log excerpts and command output included in *Additional context* below.

**Desktop (please complete the following information):**

- OS: macOS 26.3.1 (a) (25D771280a)
- Browser: N/A — Electron desktop app
- Version: OpenWhispr 1.6.10
- Architecture: Apple Silicon (arm64)

**Smartphone (please complete the following information):**

N/A — desktop only.

**Additional context**

*Log excerpt from the failing dictation:*

```
[DEBUG] FFmpeg conversion complete { "outputSize": 224718 }
[DEBUG] parakeet-ws stdout {
  "data": "[2026-04-29 10:56:11] [info] asio listen error: asio.system:48 (Address already in use)"
}
[DEBUG] parakeet-ws stderr {
[...]

PR #1374, opened 2026-07-28
fix(sidecars): tolerate unreadable PID entries

### LINKED ISSUE #1373: Startup aborts when a stale sidecar PID entry cannot be read
**Describe the bug**

The startup sidecar reaper treats PID metadata as mandatory even though the PID registry is a best-effort recovery mechanism. If any entry under `sidecar-pids` disappears between enumeration and reading, is an unreadable file, or is otherwise not readable as a file, `sidecarPidFile.readAll()` throws before the app creates its managers or windows.

**Affected component**

Electron main-process startup: `src/helpers/sidecarPidFile.js` -> `src/helpers/sidecarReaper.js` -> `startApp()` in `main.js`.

**To Reproduce**

On current `main` (`143971d3`), point the Electron `app.getPath("userData")` test seam at a temporary directory, create `sidecar-pids/stale.pid` as a broken symlink, and call `readAll()`. A minimal reproduction prints:

```text
ENOENT: ENOENT: no such file or directory, open '.../sidecar-pids/stale.pid'
```

The same failure can be produced with a `.pid` entry that is a directory or cannot be read. During a real launch, `reapStaleSidecars()` is the first statement in `startApp()`, so the rejection reaches the fatal startup handler, shows the startup error dialog, and exits with status 1.

**Actual behavior**

One bad PID entry aborts the whole application launch, and valid entries later in the directory are never considered.

**Expected behavior**

Sidecar recovery metadata should remain best-effort: skip unreadable or malformed entries, continue returning valid entries, and treat a missing or unreadable PID directory as an empty registry.

**Desktop**

- OS: macOS
- Node.js: 24.9.0
- OpenWhispr revision: `143971d3`

**User impact**

A torn, externally removed, permission-changed, or otherwise damaged PID entry can make OpenWhispr unable to start until the user manually finds and removes the entry from the application data directory.

**Root-cause hypothesis**

`readAll()` checks that the directory exists, then performs unguarded `readdirSync()` and `readFileSync()` calls. Unlike the module write and clear paths, the read path has no best-effort error boundary.

**Proposed focused scope**

Guard directory enumeration and each individual entry read, retain the existing validation for positive integer PIDs, and add deterministic unit coverage showing that a bad entry cannot hide a valid one. No changes to sidecar signaling, command verification, or shutdown ordering.

hsusul:
## Summary

- Treat the sidecar PID registry as best-effort when its directory cannot be enumerated.
- Skip individual unreadable PID entries while preserving valid entries.
- Add deterministic regression coverage for unreadable, missing, malformed, and valid registry states.

## Problem

The startup reaper calls `sidecarPidFile.readAll()` before managers or windows are initialized. A stale `.pid` entry that disappears, is a directory, or cannot be read throws from `readFileSync()`. That rejects `startApp()`, shows the fatal startup dialog, and exits the app.

Before this change, one damaged recovery entry also prevented later valid entries from being considered. After this change, unreadable recovery metadata is skipped and startup can continue.

## Root cause

`readAll()` checked whether the registry directory existed but left both `readdirSync()` and each `readFileSync()` outside an error boundary. The neighboring write and clear paths already treat this metadata as best-effort.

## Solution

Guard directory enumeration and return an empty registry when it cannot be read. Guard each `.pid` read independently so a bad entry cannot hide valid positive-integer PID entries.

## Scope and non-goals

This PR does not change sidecar signaling, process-command verification, shutdown ordering, PID validation, public APIs, dependencies, or lockfiles.

## Tests

Added `test/helpers/sidecarPidFile.test.js` covering:

- an unreadable entry alongside a valid entry;
- an unreadable registry path;
- missing, malformed, non-positive, fractional, and unrelated entries;
- the existing valid write/read/clear round trip.

The first regression failed before the production change with `EISDIR` from `readAll()`.

## Validation

- `npm ci --ignore-scripts` — passed (locked Node 24 install)
- `npm rebuild better-sqlite3` — passed
- `node --test test/helpers/sidecarPidFile.test.js` — 4 passed, 0 failed
- `ELECTRON_OVERRIDE_DIST_PATH=/tmp npm test` — 869 tests: 863 passed, 6 expected FFmpeg-related skips, 0 failed
- `npm run lint` — passed
- `npx prettier --check src/helpers/sidecarPidFile.js test/helpers/sidecarPidFile.test.js` — passed
- `node --check src/helpers/sidecarPidFile.js` — passed
- `node --check test/helpers/sidecarPidFile.test.js` — passed
- `git diff --check` and `git diff --cached --check` — passed

## Compatibility and risk

The success path and returned entry shape are unchanged. The only new behavior is converting filesystem read failures in best-effort recovery metadata into skipped entries. Local validation ran on macOS with Node 24.9.0; the filesystem logic is platform-neutral, but no packaged application build was run.

## Documentation

No documentation change is needed because this restores the existing best-effort lifecycle contract.

PR #694 introduced startup sidecar reaping for issue #683. That work is adjacent but did not cover unreadable registry entries.

Fixes #1373

### DIFF
--- a/src/helpers/sidecarPidFile.js
+++ b/src/helpers/sidecarPidFile.js
@@ -32,15 +32,25 @@ function clear(name) {
 
 function readAll() {
   const dir = pidDir();
-  if (!fs.existsSync(dir)) return [];
+  let files;
+  try {
+    files = fs.readdirSync(dir);
+  } catch {
+    return [];
+  }
+
   const entries = [];
-  for (const file of fs.readdirSync(dir)) {
+  for (const file of files) {
     if (!file.endsWith(".pid")) continue;
-    const name = path.basename(file, ".pid");
-    const raw = fs.readFileSync(path.join(dir, file), "utf-8").trim();
-    const pid = Number(raw);
-    if (!Number.isInteger(pid) || pid <= 0) continue;
-    entries.push({ name, pid });
+    try {
+      const name = path.basename(file, ".pid");
+      const raw = fs.readFileSync(path.join(dir, file), "utf-8").trim();
+      const pid = Number(raw);
+      if (!Number.isInteger(pid) || pid <= 0) continue;
+      entries.push({ name, pid });
+    } catch {
+      // Best-effort recovery metadata; skip entries that vanished or cannot be read.
+    }
   }
   return entries;
 }
--- a/test/helpers/sidecarPidFile.test.js
+++ b/test/helpers/sidecarPidFile.test.js
@@ -1,0 +1,63 @@
+const test = require("node:test");
+const assert = require("node:assert/strict");
+const fs = require("fs");
+const os = require("os");
+const path = require("path");
+
+let userDataDir;
+const electronPath = require.resolve("electron");
+require.cache[electronPath] = {
+  exports: {
+    app: {
+      getPath: () => userDataDir,
+    },
+  },
+};
+
+const { write, clear, readAll } = require("../../src/helpers/sidecarPidFile");
+
+function createUserDataDir(t) {
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), "ow-sidecar-pids-"));
+  userDataDir = dir;
+  t.after(() => fs.rmSync(dir, { recursive: true, force: true }));
+  return path.join(dir, "sidecar-pids");
+}
+
+test("an unreadable PID entry does not hide valid sidecar entries", (t) => {
+  const pidDir = createUserDataDir(t);
+  fs.mkdirSync(pidDir);
+  fs.writeFileSync(path.join(pidDir, "qdrant.pid"), "1234");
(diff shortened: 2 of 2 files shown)


gabrielste1n (approved):
Verified against `main`: typecheck/lint clean, suite 1689 pass / 0 fail.

Fatal-but-rare bug, correctly scoped: an unreadable `.pid` entry threw out of `readAll()`, which runs as the first statement of `startApp()` (`main.js:894-895`), so the app died at launch with `dialog.showErrorBox` + `app.exit(1)` — no dictation at all until the user found and deleted the file by hand.

Two things I like about the shape: it matches the module's own convention (every sibling function already swallows errors), and replacing `existsSync` with a caught `readdirSync` closes a TOCTOU rather than just moving the guard. The per-entry catch also means one bad file no longer aborts reaping the valid ones — the comment there earns its place by saying why skipping is safe.

PR #1376, opened 2026-07-28
fix(parakeet): normalize non-PCM16 WAV input

### LINKED ISSUE #1375: Parakeet skips normalization for non-PCM16 WAV input
## Component

Local Parakeet audio-file transcription (`src/helpers/parakeetServer.js`, `src/helpers/ffmpegUtils.js`)

## Reproduction

On current `main` (`143971d3aa978403d0110fe82673263f3cdcca7a`), create a valid mono 16 kHz IEEE-float WAV containing one `0.5` sample and pass it through the same WAV decoder used by Parakeet:

```bash
const { parseWavFormat, wavToFloat32Samples, computeFloat32RMS } = require('./src/helpers/ffmpegUtils');

const wav = Buffer.alloc(48);
wav.write('RIFF', 0); wav.writeUInt32LE(40, 4); wav.write('WAVE', 8);
wav.write('fmt ', 12); wav.writeUInt32LE(16, 16);
wav.writeUInt16LE(3, 20); // WAVE_FORMAT_IEEE_FLOAT
wav.writeUInt16LE(1, 22); wav.writeUInt32LE(16000, 24);
wav.writeUInt32LE(64000, 28); wav.writeUInt16LE(4, 32); wav.writeUInt16LE(32, 34);
wav.write('data', 36); wav.writeUInt32LE(4, 40); wav.writeFloatLE(0.5, 44);

const samples = wavToFloat32Samples(wav);
console.log({ format: parseWavFormat(wav), decoded: samples.readFloatLE(0), rms: computeFloat32RMS(samples) });
NODE
```

Current output:

```text
{
  format: { channels: 1, sampleRate: 16000, bitsPerSample: 32 },
  decoded: 0,
  rms: 0
}
```

The full import path reaches `ParakeetServerManager._ensureWav()`. Because the file is already mono/16 kHz, `_ensureWav()` returns it unchanged without checking its encoding or bit depth. `wavToFloat32Samples()` then interprets non-16-bit samples as signed 8-bit values, so valid float audio can become silence and transcription exits with “No audio detected.”

## Expected

Only mono 16 kHz signed PCM16 WAV files should bypass FFmpeg. Other valid WAV encodings/bit depths should be normalized to PCM16 before `wavToFloat32Samples()` runs.

## Actual

Any mono 16 kHz WAV bypasses conversion, including IEEE-float/32-bit WAV input that the downstream decoder cannot read correctly.

## Environment

- macOS 15.7.3
- Node 24.9.0 / npm 11.6.0
- OpenWhispr `main` at `143971d3aa978403d0110fe82673263f3cdcca7a`

## Impact

Imported WAV recordings with a supported container but non-PCM16 encoding can silently produce an empty Parakeet transcription even though they contain audible signal.

## Proposed scope

Expose the WAV audio-format code from `parseWavFormat()`, restrict the no-conversion fast path to format code 1 / 16-bit PCM, and add focused coverage for PCM16 passthrough versus float32 normalization.

I intend to work on this.

hsusul:
## Summary

- normalize mono 16 kHz WAV files unless they are signed PCM16
- cover float32 WAV import through the public Parakeet transcription path

## Problem

A valid mono 16 kHz IEEE-float WAV bypassed FFmpeg because the fast path checked only sample rate and channel count. The PCM16-only decoder then interpreted the float bytes as signed 8-bit samples, which could turn audible input into silence and return an empty transcription.

## Root cause

`parseWavFormat()` exposed the bit depth but not the WAV audio-format code, and `ParakeetServerManager._ensureWav()` ignored both fields when deciding whether the input was already ready for inference.

## Solution

Expose the WAV audio-format code and restrict passthrough to format code 1, 16-bit, mono, 16 kHz input. All other WAV formats use the existing FFmpeg normalization path before sample decoding.

## Tests

- `ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/parakeetWavNormalization.test.js` — 1 passed, 0 failed; confirmed red before the fix (`'' !== 'audible'`)
- `ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/audioSegmentMerge.test.js test/helpers/parakeetModelRegistry.test.js test/helpers/parakeetOnlineStream.test.js test/helpers/parakeetWsResult.test.js` — 26 passed, 0 failed
- `ELECTRON_OVERRIDE_DIST_PATH=/tmp npm test` — 816 passed, 50 skipped, 0 failed; skips are existing `better-sqlite3` Node/Electron ABI exclusions

## Validation

- `npm run lint` — passed
- `node --check src/helpers/ffmpegUtils.js && node --check src/helpers/parakeetServer.js && node --check test/helpers/parakeetWavNormalization.test.js` — passed
- `git diff --check` — passed

## Compatibility and risk

PCM16 mono/16 kHz input keeps the existing zero-conversion fast path. Other WAV encodings now take the existing FFmpeg path already used for non-WAV, non-16 kHz, and non-mono inputs. `parseWavFormat()` gains one additive field; no existing field or caller contract is removed.

## Documentation

No documentation change is needed because supported import formats and user-facing behavior are unchanged; this corrects normalization inside the existing WAV import path.

Fixes #1375.

PR #1238 added sample-rate/channel normalization and the current WAV metadata parser. This patch closes the separate encoding/bit-depth gap in that fast path.

### DIFF
--- a/src/helpers/ffmpegUtils.js
+++ b/src/helpers/ffmpegUtils.js
@@ -196,6 +196,7 @@ function parseWavFormat(wavBuffer) {
 
     if (chunkId === "fmt ") {
       return {
+        audioFormat: wavBuffer.readUInt16LE(offset + 8),
         channels: wavBuffer.readUInt16LE(offset + 10),
         sampleRate: wavBuffer.readUInt32LE(offset + 12),
         bitsPerSample: wavBuffer.readUInt16LE(offset + 22),
--- a/src/helpers/parakeetServer.js
+++ b/src/helpers/parakeetServer.js
@@ -53,10 +53,15 @@ class ParakeetServerManager {
   async _ensureWav(audioBuffer) {
     if (isWavFormat(audioBuffer)) {
       const format = parseWavFormat(audioBuffer);
-      if (format?.sampleRate === SAMPLE_RATE && format?.channels === 1) {
+      if (
+        format?.audioFormat === 1 &&
+        format?.bitsPerSample === 16 &&
+        format?.sampleRate === SAMPLE_RATE &&
+        format?.channels === 1
+      ) {
         return { wavBuffer: audioBuffer, filesToCleanup: [] };
       }
-      debugLogger.debug("WAV input needs resampling", { format });
+      debugLogger.debug("WAV input needs normalization", { format });
     }
 
     const ffmpegPath = getFFmpegPath();
--- a/test/helpers/parakeetWavNormalization.test.js
+++ b/test/helpers/parakeetWavNormalization.test.js
@@ -1,0 +1,125 @@
+const test = require("node:test");
+const assert = require("node:assert/strict");
+const Module = require("node:module");
+const fs = require("node:fs");
+const os = require("node:os");
+const path = require("node:path");
+const { REQUIRED_MODEL_FILES } = require("../../src/helpers/parakeetModelInfo");
+
+const MODEL_NAME = "parakeet-tdt-0.6b-v3";
+
+function createFloat32Wav(sampleValue, sampleCount = 160) {
+  const dataSize = sampleCount * 4;
+  const wav = Buffer.alloc(44 + dataSize);
+
+  wav.write("RIFF", 0);
+  wav.writeUInt32LE(36 + dataSize, 4);
+  wav.write("WAVE", 8);
+  wav.write("fmt ", 12);
+  wav.writeUInt32LE(16, 16);
+  wav.writeUInt16LE(3, 20);
+  wav.writeUInt16LE(1, 22);
+  wav.writeUInt32LE(16000, 24);
+  wav.writeUInt32LE(64000, 28);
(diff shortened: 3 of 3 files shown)


gabrielste1n (approved):
Verified against `main` @ a988d820: merges clean, typecheck/lint clean, full suite (CI env, `REQUIRE_DB_TESTS=1`) 1671 pass / 0 fail.

Confirmed the test actually catches the bug — reverting just the two source hunks makes it fail with `'' !== 'audible'`, which is exactly the reported symptom: a non-PCM16 WAV took the fast path, got reinterpreted as int8 by `ffmpegUtils`, and produced an empty transcript with no error. Gating the fast path on `audioFormat === 1 && bitsPerSample === 16` routes float32/extensible input through `convertToWav` instead. Both changed functions have a single consumer each and FFmpeg is bundled, so the slower path is always available.

Note: the new test file isn't prettier-formatted; I'm folding that into a follow-up formatting PR rather than blocking this.

PR #1379, opened 2026-07-28
fix(reasoning): stop Gemini cleanup truncating dictation at the token limit

### LINKED ISSUE #1341: [Bug] Dictation cleanup silently truncates the last words — Gemini 3 Flash thinking tokens exhaust maxOutputTokens
**Describe the bug**

With cleanup enabled on the Gemini provider using `gemini-3-flash-preview`, the pasted dictation is regularly cut off mid-sentence — anywhere from the last few words to the last couple of sentences are missing. Retranscribing the same recording from history fixes it, usually on the second or third try.

The transcription itself is fine. `raw_text` in `transcriptions.db` is always complete; only `text` (the cleanup output) is truncated. Example from my history (id 167):

```
raw_text  (536 chars, complete):
  ... Parece que tengo créditos, pero revisa todo eso. Cuántos puedo generar. Y si puedes, haz las pruebas ahí.

text      (369 chars, truncated mid-sentence):
  ... Ya tengo una pestaña abierta y he hecho una prueba en
```

15 of 164 cleanup runs in my history end with no terminal punctuation (~9%). The cut point tracks `maxOutputTokens`, not the length of the input:

| `maxOutputTokens` sent | truncated runs | output length at cut |
|---|---|---|
| 2000 | 13 | 267–372 chars (~76–106 tokens) |
| 2120 | 1 | 378 chars |
| 2786 | 1 | 472 chars |

**Root cause**

All references below are against `main` as of today.

1. `src/services/ai/inferenceProviders/gemini.ts` sets the output budget from the input length alone:

   ```ts
   maxOutputTokens:
     config.maxTokens ||
     Math.max(2000, ctx.calculateMaxTokens(text.length, TOKEN_LIMITS.MIN_TOKENS_GEMINI, TOKEN_LIMITS.MAX_TOKENS_GEMINI, TOKEN_LIMITS.TOKEN_MULTIPLIER))
   ```

   With `MIN_TOKENS_GEMINI: 100`, `MAX_TOKENS_GEMINI: 8192`, `TOKEN_MULTIPLIER: 2` (`src/config/constants.ts`), a typical dictation lands on the 2000 floor.

2. In the same file, `thinkingConfig` is only sent when the user opted out **and** the model is flagged:

   ```ts
   if (config.disableThinking === true && getCloudModel(model)?.supportsThinking) {
     generationConfig.thinkingConfig = { thinkingLevel: "minimal", includeThoughts: false };
   }
   ```

   In `src/models/modelRegistryData.json`, the only Gemini model carrying `supportsThinking: true` is `gemini-3.5-flash`. `gemini-3-flash-preview` and `gemini-3.1-pro-preview` do not have it. So no thinking budget is ever sent for them — and `showThinkingToggle` in `src/components/settings/InferenceConfigEditor.tsx` (~line 156) gates the "Disable reasoning output" switch on the same flag, so the user cannot even reach the setting.

3. In the Gemini API, thinking tokens are charged against `maxOutputTokens`. Gemini 3 Flash thinks by default; on this task it consumes roughly 1.9k of the 2k budget, leaving ~100 tokens of visible answer — which matches the measured cut points above.

4. The response handler only raises `MAX_TOKENS` when the candidate is *entirely* empty:

   ```ts
   const candidate = response.candidates?.[0];
   if (!candidate?.content?.parts?.[0]?.text) {
     // ... throws "Gemini reached token limit ..." when finishReason === "MAX_TOKENS"
   }
   const responseText = candidate.content.parts[0].text!.trim();
   return responseText;   // partial text + finishReason MAX_TOKENS returned as if complete
   ```

   A truncated candidate is therefore pasted into the user's editor as a normal result — no error, no toast, no fallback to the raw transcript. Retrying works only because the length of the thinking pass varies between calls, so sometimes the answer fits in what's left.

   Two related gaps in the same block: `parts[0]` drops any additional text parts of a multi-part candidate, and the `GeminiResponse` interface only types `usageMetadata.totalTokenCount`, so `thoughtsTokenCount` / `candidatesTokenCount` are never logged — which is why this is invisible in the debug log.

**To reproduce**

1. Settings → cleanup model: Google Gemini / Gemini 3 Flash (`gemini-3-flash-preview`).
2. Dictate ~30–60 s, enough that the cleaned output needs more than ~120 tokens.
3. The pasted text ends mid-sentence. "View raw transcript" in history shows the full, correct transcription.
4. Retranscribe from history — it completes correctly after one or two attempts.

**Expected behavior**

The cleanup output should be complete, or the app should detect the truncation and fall back to the raw transcript instead of pasting a partial sentence.

**Suggested fix**

- Treat `finishReason === "MAX_TOKENS"` as a failure even when partial text is present: retry once, then fall back to the raw transcript and surface it (cf. #1003, #1045).
- Send `thinkingConfig: { thinkingLevel: "minimal", includeThoughts: false }` for the cleanup route on all Gemini 3.x models, and add `supportsThinking: true` to `gemini-3-flash-preview` and `gemini-3.1-pro-preview` so the existing toggle from #897 actually applies to them.
- Raise the `maxOutputTokens` floor for thinking-capable models — a 2000-token ceiling shared with an unbounded thinking pass cannot hold a normal dictation.
- Join all text parts of the candidate instead of reading `parts[0].text`, and log `thoughtsTokenCount` / `candidatesTokenCount` alongside `totalTokenCount`.

**Additional context**

Same user-visible symptom as #857 (closed), but a different code path: that report is local-model-only on Linux and suspected an audio flush race. Here the audio and the transcription are provably fine — `raw_text` is complete in the database and only the cleanup output is short.

**Desktop**

- OS: Windows 11 Pro 26200
- Version: 1.7.6
- Transcription: local whisper-server (`base` / `medium`, CUDA). Cleanup: Gemini `gemini-3-flash-preview`.

xAlcahest:
## Summary

Cleanup on Gemini 3 Flash regularly pastes the dictation cut off mid-sentence. Gemini charges thinking tokens against `maxOutputTokens` and `gemini-3-flash-preview` thinks at level `high` by default, so the 2000-token budget computed in `gemini.ts` is mostly eaten before any visible answer. The reply then comes back as `finishReason: "MAX_TOKENS"` with partial text, and the handler returned it as a normal result because it only inspected `finishReason` when the candidate was completely empty. On top of that, the registry lacked `supportsThinking` on the Gemini 3 previews, so the existing `thinkingConfig` suppression (and its settings toggle) never applied to them. Same guard family as #938 and #1258, but for partial results instead of empty ones.

A token-limited reply is now treated as a failed cleanup on every path (Gemini `MAX_TOKENS`, chat-completions `finish_reason: "length"` for Groq/LAN/custom endpoints and Tinfoil, Responses API `status: "incomplete"` with reason `max_output_tokens`), so the existing catch keeps the raw transcription and records the failure. The starvation itself is fixed too: cleanup requests the lowest thinking level each model accepts (`minimal`, or `low` for 3.1 Pro which rejects `minimal`) and thinking-capable models get the full 8192 ceiling instead of the 2000 floor. Detection is purely signal-based, no length heuristics, so legitimate cleanups that shorten text are untouched.

Fixes #1341

## Changes

- src/services/ai/inferenceProviders/gemini.ts: classify each candidate via `assessGeminiResponse` and fail truncated ones even with partial text, join all visible parts instead of reading only `parts[0]`, log `thoughtsTokenCount` and `candidatesTokenCount`, raise the output ceiling for thinking-capable models, resolve `thinkingConfig` per model.
- src/helpers/geminiResponse.js: new pure helpers (`assessGeminiResponse`, `extractGeminiCandidateText`, `resolveGeminiThinkingConfig`, `geminiSuppressedThinkingLevel`) so the decisions are unit-testable.
- src/helpers/completionTruncation.js: truncation signal checks for chat completions and the Responses API, plus `truncatedResponseError` so every provider throws the same failure shape.
- src/services/ReasoningService.ts: `callChatCompletionsApi` throws on `finish_reason: "length"`; the streaming path reuses the thinking-level resolver so 3.1 Pro never receives `minimal`.
- src/services/ai/inferenceProviders/openai.ts: throw on truncated chat or Responses payloads; the guard scans every choice, matching how the text extraction picks its content.
- src/services/ai/inferenceProviders/tinfoil.ts: same truncation guard on the Tinfoil SDK response, which bypasses both shared request paths.
- src/services/ai/thinkingSuppressionDialects.ts: the gemini suppression branch picks `low` for Pro models instead of hardcoding `minimal`, which they reject.
- src/models/modelRegistryData.json, src/models/ModelRegistry.ts: mark `gemini-3-flash-preview` and `gemini-3.1-pro-preview` as `supportsThinking` (this also unlocks the existing "Disable reasoning output" toggle for them) and add `minThinkingLevel` for models that reject `minimal`.
- test/helpers/geminiResponse.test.js, test/helpers/completionTruncation.test.js, test/helpers/thinkingSuppressionDialects.test.js: the response classification is unit-tested directly (truncated fails, normal passes, the #938 empty case stays guarded), registry flags asserted, wiring asserts reduced to import-and-usage checks.

## Behavior notes

- On Gemini cleanup the "Disable reasoning output" toggle no longer changes the outcome: cleanup always requests the model's lowest thinking level, mirroring the forced `reasoning_effort: "low"` cleanup already uses for gpt-oss. Honoring the off state would reopen #1341 for every default configuration, since the toggle defaults to off. It still governs the agent and chat paths.
- The chat-completions guard intentionally ships without raising the 4096 budget for Groq/LAN/custom: bumping `max_tokens` can exceed the context window of small local models, and pasting a clipped reply is exactly the failure #1341 describes, so those paths fall back to the raw transcription instead.

## Provider semantics verified

- Gemini 3: `thinking_level` defaults to `high` on `gemini-3-flash-preview`; `minimal` is supported on the Flash family but not on 3.1 Pro ([Gemini 3 developer guide](https://ai.google.dev/gemini-api/docs/gemini-3), [Gemini 3.5 Flash notes](https://ai.google.dev/gemini-api/docs/whats-new-gemini-3.5)). `max_output_tokens` caps thinking plus output combined, measured at ~96% of the budget going to thinking ([googleapis/python-genai#2062](https://github.com/googleapis/python-genai/issues/2062)).
- OpenAI: `finish_reason: "length"` means the reply was cut by the token limit; the Responses API signals it as `status: "incomplete"` with `incomplete_details.reason: "max_output_tokens"` ([reasoning guide](https://platform.openai.com/docs/guides/reasoning)).
- Groq: `finish_reason` is `length` "if the maximum number of tokens specified in the request was reached" ([API reference](https://console.groq.com/docs/api-reference)).
- Mistral: `length` means the response was truncated at `max_tokens`; it reaches us through the OpenAI-compatible custom endpoint path, which is now guarded.

No retry on the token-limit signal: with minimal thinking and the 8192 ceiling the guard only fires on genuinely oversized outputs, where a retry would double the latency and land on the same fallback. Still open elsewhere, worth follow-ups: the Anthropic/local IPC and enterprise SDK paths extract text in the main process without reading their stop reasons, and the chat streaming path shares the bug class (fixed 4096 budget, `finishReason` of the final chunk never read).


Chadpiha (changes requested):
blocking changes requireed do not merge

PR #1381, opened 2026-07-28
fix(translation): route local models with an empty provider through llama.cpp

### LINKED ISSUE #1333: Local Translation Prompt Studio does not support local reasoning providers
Version: 1.7.6

Local models work correctly in Chat (Llama 3.2 3B, Qwen).

However, in Language Models → Translation → Local → Test, Prompt Studio always fails.

Errors:

"Unsupported reasoning provider: llama"

and

"Unsupported reasoning provider: qwen"

Cloud translation works correctly.

Local chat works correctly.

Only Prompt Studio translation testing fails.

xAlcahest:
## Summary

#1327 routes local catalog ids (`llama`, `qwen`, ...) through the llama.cpp provider, but only when the scope's provider field is set. It can legitimately be empty: the local picker's active tab is component state and clicking a model card only wrote the model, so picking a local model without ever touching a provider tab left `translationProvider` empty. `resolveInferenceProvider` then falls back to `getModelProvider`, which consults the cleanup scope first: with cleanup on OpenWhispr Cloud the "local" translation (Prompt Studio test and live dictation alike) silently runs in the cloud, and with cleanup on a custom/OpenRouter endpoint it fails against a server that doesn't have the GGUF model.

The fix works at both ends. `resolveTranslationProviderId` in `dictationRouting.js` now gives the two translate call sites (`resolveReasoningRoute` and the Prompt Studio test) an explicit `local` fallback when translation mode is local, and `ReasoningModelSelector` persists the owning catalog provider whenever a local model is selected, so no scope is left with a model but no provider. Explicit provider ids still pass through untouched, so a genuinely unknown provider keeps failing loudly.

Fixes #1333

## Changes

- `src/helpers/dictationRouting.js`: new `resolveTranslationProviderId`, shared provider resolution for the translate steps with the local-mode fallback.
- `src/helpers/audioManager.js`: `resolveReasoningRoute` builds the translation config through the helper.
- `src/components/ui/PromptStudio.tsx`: the translate test resolves its provider through the same helper as the live path.
- `src/components/ReasoningModelSelector.tsx`: selecting a local model also persists its catalog provider to the scope's store.
- `test/helpers/dictationRouting.test.js`: coverage for the helper: cloud mode, explicit providers, empty provider in and outside local mode.

### DIFF
--- a/preload.js
+++ b/preload.js
@@ -732,6 +732,10 @@ contextBridge.exposeInMainWorld("electronAPI", {
     "meeting-transcription-error",
     (callback) => (_event, data) => callback(data)
   ),
+  onMeetingTranscriptionFatalError: registerListener(
+    "meeting-transcription-fatal-error",
+    (callback) => (_event, data) => callback(data)
+  ),
 
   // Dictation realtime streaming
   dictationRealtimeWarmup: (options) => ipcRenderer.invoke("dictation-realtime-warmup", options),
--- a/src/components/ReasoningModelSelector.tsx
+++ b/src/components/ReasoningModelSelector.tsx
@@ -529,10 +529,8 @@ export default function ReasoningModelSelector({
   };
 
   const handleLocalModelSelect = (modelId: string) => {
-    // Persist the owning catalog provider with the model: an empty provider
-    // sends this scope's routing through another scope's heuristics.
-    const owner = modelId ? modelRegistry.getModel(modelId)?.provider.id : undefined;
-    if (owner) setLocalReasoningProvider(owner);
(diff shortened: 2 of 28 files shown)


gabrielste1n (approved):
Traced live translation, availability checks, local model selection, and Prompt Studio end to end across managed, local, self-hosted, provider, and enterprise modes. Both entry points now share one validated resolver, local selections are pinned to llama.cpp, and incomplete or stale state fails closed instead of inferring a cloud provider. Local lint, typecheck, renderer build, i18n validation, and the full test suite pass; all GitHub checks are green.

PR #1383, opened 2026-07-28
feat(xai): stream Grok STT over websockets

Jakobovski:
## Problem

xAI STT support was batch-only. The whole recording was uploaded to `POST /v1/stt` after the user stopped speaking ([`proxy-xai-transcription`](https://github.com/OpenWhispr/openwhispr/blob/main/src/helpers/ipcHandlers.js)), so dictation latency was one round-trip on the complete file and there was no live preview — while Deepgram, AssemblyAI, Corti and OpenAI Realtime users get partials as they speak.

xAI has offered a streaming websocket endpoint for a while ([docs](https://docs.x.ai/developers/model-capabilities/audio/speech-to-text)); nothing in the app used it.

## What this does

Adds `XaiStreaming` (`src/helpers/xaiStreaming.js`) for `wss://api.x.ai/v1/stt`, following the shape of the existing streaming clients — warmup-and-promote for an instant start, pre-ready audio buffering, keep-alive ping, and a `Finalize` / `audio.done` close handshake.

Wires xAI (BYOK) into the streaming paths:

- **Dictation / notes** — `xai` entry in `STREAMING_PROVIDERS`, plus routing in `getStreamingProviderName()` / `shouldUseStreaming()`
- **Meeting capture** — `xai-realtime` in `STREAMING_CLIENT_BY_PROVIDER` and `ALLOWED_MEETING_PROVIDERS`, with a BYOK branch in `meetingRecordingStore` mirroring Corti's
- `grok-stt` gains `"streaming": true` in the model registry, so it appears in the streaming-only picker

The 25-language allowlist moves out of `ipcHandlers.js` into a shared `xaiSttLanguages` module so the REST and streaming paths can't drift, and gains base-code fallback (`pt-BR` → `pt`).

Endpointing, VAD threshold and Smart Turn are left at xAI's defaults but plumbed through as options, so tuning them later doesn't mean touching the client.

## Protocol notes — verified against the live API, not the docs

Worth flagging for reviewers, since two of these would be easy to reintroduce:

- **`transcript.done` returns an empty `text`**, not the full transcript its JS example prints (`Full transcript: ${event.text}`). Treating it as the source of truth yields empty transcripts. It's used only as a salvage path when no utterance final arrived.
- **Each chunk final is immediately followed by an identical utterance final.** Observed 1 ms apart with the same text. Only `speech_final` commits a segment — accumulating every `is_final` (the natural port of the Deepgram logic) doubles every segment. Chunk finals and interims feed the live preview instead.
- **Utterance finals are scoped to their own utterance** (`start` advances 0 → 2.38 across two sentences), so joining committed segments is correct.
- The docs spell the force-finalize message `{"type": "finalize"}` in prose and `{"type": "Finalize"}` in the JSON examples. The examples are correct.

Raw event trace from a two-utterance clip separated by silence:

```
  344ms created
 1629ms interim         start=0    dur=1    | "First sentence about apple"
 3026ms chunk-final     start=0    dur=2.38 | "First sentence about apples."
 3027ms UTTERANCE-FINAL start=0    dur=2.38 | "First sentence about apples."
 6342ms chunk-final     start=2.38 dur=3.32 | "Second sentence about oranges."
 6888ms UTTERANCE-FINAL start=2.38 dur=3.92 | "Second sentence about oranges."
 7541ms -> audio.done
 7752ms DONE            dur=6.92                | ""
```

## Testing

`test/helpers/xaiStreaming.test.js` — 16 tests against a mock ws server, covering the connection lifecycle (including close-before-ready on both `connect` and `warmup`), warm-connection promotion, chunk-vs-utterance final routing, the `transcript.done` salvage path and its no-duplicate guard, pre-ready audio buffering and its 3 s cap, the close handshake, and URL construction (language allowlist, keyterm limits, opt-in tuning params).

```
ℹ tests 16
ℹ pass 16
ℹ fail 0
```

Also exercised end-to-end against the real endpoint with a 15 s clip: 15 progressive partials, correct final text.

`npm run lint`, `npm run typecheck` and `npx prettier --check` are clean on the changed files. Full suite passes except 4 pre-existing environmental failures unrelated to this change (3 need SQLite FTS5, 1 needs a working local ffmpeg).

## Scope notes

- BYOK only — there's no OpenWhispr-Cloud xAI mode, matching how Corti and Tinfoil are gated.
- No new user-facing strings, so no i18n keys needed.
- The batch REST path is untouched and still serves audio-file upload and the non-streaming fallback.


Jakobovski:
Pushed a follow-up fix after hitting a real failure dictating against a packaged build — worth calling out because the original commit had a silent data-loss path.

**Symptom:** a warm connection promoted after ~10 minutes idle swallowed 17 seconds of speech and returned nothing — no partials, no finals, `text: ""`, then a silent fall back to batch transcription. A cold connect earlier in the same session worked fine.

```
18:47:46.552  xAI connection warmed up
18:58:03.548  xAI using warm connection - instant start   <- 10m17s idle
18:58:20.722  xAI Finalize sent                           <- 17s of speech, zero events
18:58:21.129  Streaming session ended { "text": "" }
18:58:21.130  Streaming produced no text, falling back to batch transcription
```

**Cause:** the socket's `readyState` was still `OPEN`, so `hasWarmConnection()` promoted it, but xAI had ended the transcription session server-side long before. I'd modeled the keep-alive on `cortiStreaming.js` (`ws.ping()`), which keeps the *socket* alive but doesn't stop xAI from finishing the *session* — leaving a live socket that accepts audio and transcribes nothing.

`deepgramStreaming.js` already solves this, and its comment says exactly why:

> 100ms of silence at 16kHz 16-bit mono — sent to warm connections to prevent Deepgram's net0001 idle timeout (which only resets on audio data, not KeepAlive).

**Fix** — three defenses, all mirroring the Deepgram client:

1. Keep warm sockets alive with 100 ms of silence instead of a ping.
2. Cap warm connection age at 2 min, and have the keep-alive **close** the socket when it expires rather than only declining to promote it.
3. Watch a promoted warm session for 2.5 s; if no `transcript.partial` arrives, reconnect and replay the audio captured meanwhile so no speech is lost. xAI emits interims even during silence, so any partial — empty text included — proves the session is alive.

Socket handlers now ignore a socket that's been replaced, so a reconnect can't be torn down by the old socket's `close` event.

**On cost**, since (2) is partly a billing decision. xAI prices STT at **$0.10/hr (REST)** vs **$0.20/hr (streaming)**, per hour of audio. The keep-alive silence *is* billable audio, but at 100 ms every 15 s it's a 0.67% duty cycle — roughly $0.0013 per idle hour, a few cents a year. Not worth engineering around on its own; the age cap is there because holding a paid session nobody claimed is simply pointless. Worth flagging separately that **enabling streaming doubles the per-hour transcription cost** for xAI users versus the batch path, in exchange for live partials.

Tests are up to 20 (from 16), covering the age cap, silence-vs-ping keep-alive, the reconnect-and-replay path, and that a partial cancels the liveness watch. Full suite green; lint, typecheck and prettier clean.

PR #1384, opened 2026-07-28
feat: auto-detect meeting type from transcript content

futuregerald:
## Summary

- Adds a **classify** step to the post-call pipeline (`retranscribe -> title -> classify -> notes`) that auto-detects the meeting type from transcript content
- Uses LLM classification (temperature 0, first 2000 chars of transcript) when a noteFormatting provider is configured, with keyword-heuristic fallback via `keyword_rules`
- Classification is non-fatal: errors are logged and the pipeline continues to the notes step

## How it works

1. After title generation, the pipeline checks if `meeting_type_id` is already set (by calendar auto-map or user selection) -- if so, classification is skipped
2. Fetches all meeting types from the database and attempts LLM classification first
3. If LLM is unavailable or returns no match, falls back to matching `keyword_rules` against transcript content (same keyword matching used by calendar auto-map but applied to transcript text instead of event title)
4. If a match is found, sets the note's `meeting_type_id` so the subsequent notes step uses the correct meeting-type template

## What this does NOT change

- No database schema changes
- No UI changes -- MeetingTypePicker already works reactively with `meeting_type_id`
- No changes to existing calendar auto-map or user selection flows
- Existing pipeline steps (retranscribe, title, notes) are unchanged

## Test plan

- [x] 5 new tests added to `test/helpers/postCallPipelineManager.test.js`:
  - STEP_ORDER includes classify between title and notes
  - classify step skips when `meeting_type_id` already set
  - LLM classification sets correct meeting type
  - Falls back to keyword matching when no LLM configured
  - Classification errors don't halt pipeline (notes still run)
- [x] All 11 tests pass (`node --test test/helpers/postCallPipelineManager.test.js`)
- [ ] Manual: Record a meeting, verify classify step appears in pipeline status
- [ ] Manual: Set meeting type manually before pipeline runs, verify classify is skipped

## i18n

- Added `meetingTypes.picker.autoDetected` key to all 10 locale files
- Added `postCallPipeline.steps.classify` key to all 10 locale files


futuregerald:
Opened against wrong repo — moving to futuregerald/openwhispr

PR #1386, opened 2026-07-29
fix(cli): preserve UTF-8 request bodies

hsusul:
## Summary

- preserve UTF-8 JSON payloads when a code point is split across incoming HTTP chunks
- enforce the CLI bridge 1 MiB request limit against the actual received bytes
- cover split multibyte input, oversized multibyte input, and the exact size boundary

## Root cause

`readJsonBody` appended each incoming `Buffer` directly to a string. That decoded every chunk independently, so a multibyte code point split at a transport boundary was replaced with invalid characters. The same conversion made the byte limit compare JavaScript string length instead of wire bytes.

The bridge now retains raw buffers, counts their byte lengths, and decodes once after the complete body arrives.

## Validation

- `node --test test/helpers/cliBridgeRequestBody.test.js`
- `ELECTRON_OVERRIDE_DIST_PATH=/tmp npm test` (864 passed, 6 skipped)
- `npm run typecheck`
- `npm run lint`
- `npm run format:check`
- `npm run i18n:check`
- `node --check src/helpers/cliBridge.js`
- `node --check test/helpers/cliBridgeRequestBody.test.js`
- `git diff --check`

### DIFF
--- a/src/helpers/cliBridge.js
+++ b/src/helpers/cliBridge.js
@@ -29,20 +29,26 @@ async function findAvailablePort() {
 
 function readJsonBody(req) {
   return new Promise((resolve, reject) => {
-    let raw = "";
-    // Count wire bytes per chunk: string concatenation decodes each Buffer
-    // chunk, so measuring the decoded string would rescan the whole body on
-    // every chunk and miscount sequences split across chunk boundaries.
+    // Buffer the raw chunks and decode once at the end: decoding per chunk
+    // corrupts multibyte sequences split across chunk boundaries.
+    const chunks = [];
     let receivedBytes = 0;
+    let rejected = false;
     req.on("data", (chunk) => {
-      raw += chunk;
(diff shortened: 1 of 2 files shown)


Chadpiha (approved):
Thanks again @hsusul !

We confirmed it on main: a multibyte character split across socket chunks got decoded per-chunk in `readJsonBody`, turned into U+FFFD replacement characters, and was stored with a 201 — silent corruption for CJK/accented text through the CLI bridge. Your `Buffer.concat` single-decode is the right fix.

Since the byte-limit half already landed via your #1777, I rebased this branch down to the decode fix and folded your split-multibyte and exact-boundary tests into the existing `cliBridgeRequestBody.test.js` (same `windowBroadcast` capture pattern as `cliBridgeDictionary.test.js`). Verified the split-chunk test fails against the old code and passes with yours. Approving — thanks for the help!

PR #1396, opened 2026-07-30
fix: recording UI not rendering on correct virtual macOS desktop

### LINKED ISSUE #1391: Recording UI doesn't render on the correct virtual Mac desktop
**Describe the bug**

Recording UI doesn't render on the correct virtual Mac desktop. This leads to a poor user experience because if I'm on a different screen and I start recording, I don't have any visual feedback that the recording actually started because I would have to switch the screen that I'm on in order to see that little recording UI. A more natural, optimal user experience would be for the recording UI to surface on all open desktop screens. This is the behavior of SuperWhisper, for example. 

**To Reproduce**

Enable the recording UI, open multiple macOS virtual desktops, start recording on one, and then switch between them and observe that the recording UI only shows on one of the desktops.

**Expected behavior**

The recording UI should show on all open virtual desktops while recording.

**Screenshots**

This is the "recording UI" that I am referencing: 

Image

- OS: MacOS
- Browser (n.a)
- Version 1.7.6

**Smartphone (please complete the following _information):**_
N.a

**Additional context**
N.a

RakshitKaintura:
Fixes #1391.

## Why
On macOS, users with multiple virtual desktops (Spaces) noticed that if they started recording on one desktop and switched to another, the recording UI (and other floating overlay windows) did not follow them to the active screen.

## What
This PR fixes the macOS window persistence issue across spaces. 

In Electron, defining `visibleOnAllWorkspaces: true` as a boolean inside the initial `BrowserWindow` config can cause a bug where macOS ignores subsequent calls to `.setVisibleOnAllWorkspaces(true, { visibleOnFullScreen: true })` because it caches the state.

- Removed the boolean `visibleOnAllWorkspaces: process.platform !== "win32"` from the configuration objects (`MAIN_WINDOW_CONFIG`, `TRANSCRIPTION_PREVIEW_CONFIG`, etc.) in `windowConfig.js`.
- Enforced the property explicitly inside `WindowPositionUtil.setupAlwaysOnTop()` exactly when `window.isVisible()` is true.
- By cleanly relying on the native method call with `{ visibleOnFullScreen: true, skipTransformProcessType: true }`, the overlays now properly hook into the macOS Space system and follow the active virtual desktop seamlessly.

### DIFF
--- a/src/helpers/windowConfig.js
+++ b/src/helpers/windowConfig.js
@@ -53,7 +53,6 @@ const MAIN_WINDOW_CONFIG = {
   show: false,
   skipTaskbar: true,
   focusable: false,
-  visibleOnAllWorkspaces: process.platform !== "win32",
   fullScreenable: false,
   hasShadow: false,
   acceptsFirstMouse: true,
@@ -116,7 +115,6 @@ const NOTIFICATION_WINDOW_CONFIG = {
     contextIsolation: true,
     sandbox: true,
   },
-  visibleOnAllWorkspaces: process.platform !== "win32",
   type: FLOATING_OVERLAY_TYPE,
 };
 
@@ -147,7 +145,6 @@ const TRANSCRIPTION_PREVIEW_CONFIG = {
     contextIsolation: true,
     sandbox: true,
   },
-  visibleOnAllWorkspaces: process.platform !== "win32",
   type: FLOATING_OVERLAY_TYPE,
 };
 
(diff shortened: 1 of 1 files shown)


RakshitKaintura:
Hi team! Just a quick heads-up that CI workflows are currently awaiting approval to run tests on this PR. Whenever you have a moment to click 'Approve and run', I'd appreciate it so we can verify the macOS Spaces fix passes all checks!

gabrielste1n:
this should be fixed from 1.9+! thanks so much for the contribution!

PR #1411, opened 2026-07-31
fix(renderer): identify control panel by query

### LINKED ISSUE #1410: Dictation window can render the control panel when its install path contains control
**Describe the bug**

The renderer identifies a window as the control panel when any part of its URL pathname contains the substring `control`. In packaged builds, `window.location.pathname` contains the full installation path, so a queryless dictation window installed under a path such as `/Users/controller/...` is routed to the control panel instead of the dictation UI.

The same predicate appears in both `src/AppRouter.jsx` and `src/utils/windowContext.ts`.

I intend to submit a focused fix.

**To Reproduce**

On current `upstream/main` (`bf8b7e0b`), evaluate the shared window-context helper with a packaged file pathname whose parent directory contains `control`:

```bash
node -e 'globalThis.window={location:{pathname:"/Users/controller/Applications/OpenWhispr.app/Contents/Resources/app.asar/src/dist/index.html",search:""}}; import("./src/utils/windowContext.ts").then(({isControlPanelWindow,isDictationPanelWindow})=>console.log({isControlPanelWindow:isControlPanelWindow(),isDictationPanelWindow:isDictationPanelWindow()}))'
```

Observed:

```text
{ isControlPanelWindow: true, isDictationPanelWindow: false }
```

The pathname substring also appears directly in `AppRouter.jsx`, so the main renderer takes the same route.

**Expected behavior**

Only an explicit control-panel marker should select the control panel. Current window creation already supplies `?panel=true` for the real control panel in both development and packaged builds; a queryless packaged URL should remain the dictation panel regardless of its parent directories.

**Desktop**

- OS: macOS 15 (source-level reproduction; the pathname behavior is cross-platform)
- OpenWhispr revision: `bf8b7e0b` (`v1.8.1`, current `main`)
- Node.js: `v24.9.0`

**User impact**

Affected installations can render the settings/control UI in the small dictation window, preventing the normal dictation overlay from loading.

**Root-cause hypothesis**

Window role is inferred with `pathname.includes("control")` instead of the explicit `panel=true` query that `DevServerManager` already assigns to control-panel windows.

**Proposed focused scope**

- make the shared window-context predicate use the explicit query parameter;
- have `AppRouter.jsx` reuse that predicate instead of duplicating it;
- add deterministic regression coverage for packaged paths containing `control`, the genuine control-panel query, and the ordinary dictation URL.

hsusul:
## Summary

- identify the control panel only from the explicit `panel=true` query parameter
- reuse the shared window-role helper in `AppRouter`
- cover packaged paths containing `control` and exact query matching

## Problem

A packaged dictation window can render the control panel when any directory in its file URL contains the substring `control`, such as `/Users/controller/.../index.html`. Window creation already tags the actual control panel with `?panel=true`, so an untagged packaged URL should remain the dictation panel regardless of its install path.

## Root cause

Both `AppRouter` and `windowContext` inferred the window role with `pathname.includes("control")`. Because an Electron file URL contains the full installation path, unrelated directory names could satisfy that heuristic. The existing query check also used substring matching rather than parsing the parameter.

## Solution

Parse the window query and select the control panel only when the exact `panel` parameter equals `true`. `AppRouter` now calls the shared helper instead of maintaining a second copy of the routing predicate. Agent-panel precedence is unchanged.

## Tests

- `node --test test/helpers/windowContext.test.js` — passed (2 tests)
- `ELECTRON_OVERRIDE_DIST_PATH=/tmp npm test` — passed (1,109 tests: 1,103 passed, 6 skipped, 0 failed)

## Validation

- `npm run lint` — passed
- `npm run build:renderer` — passed; Vite reported the existing dynamic-import and chunk-size warnings
- `./node_modules/.bin/prettier --check src/AppRouter.jsx src/utils/windowContext.ts test/helpers/windowContext.test.js` — passed
- `git diff --check` — passed
- Pre-existing failure: `npm run format:check` reports formatting differences in `src/components/notes/overview/OverviewNoteList.tsx`, `src/components/notes/ShareNoteDialog.tsx`, and `src/hooks/useNoteDragAndDrop.ts`. The same three failures reproduce from a clean worktree at upstream `bf8b7e0b`; none is modified here.

## Compatibility and risk

The change is limited to renderer window-role detection and has no public API, persistence, or dependency impact. Control-panel windows created by the application continue to use `?panel=true`. Queryless windows and near matches such as `notpanel=true` or `panel=trueish` no longer route to the control panel.

## Documentation

No documentation change is required because this restores the existing internal window-creation contract and does not change a user-facing workflow or public API.

Fixes #1410.

Repository-wide searches across open and closed issues and pull requests for `windowContext`, control/dictation-panel routing, install paths, `panel=true`, wrong-window behavior, and username paths found no overlapping report or active fix. The superficially related #1227 concerned Windows paste focus, not renderer window-role selection.

### DIFF
--- a/src/AppRouter.jsx
+++ b/src/AppRouter.jsx
@@ -9,6 +9,7 @@ import WindowControls from "./components/WindowControls.tsx";
 import { Card, CardContent } from "./components/ui/card.tsx";
 import { useAuth } from "./hooks/useAuth";
 import { useTheme } from "./hooks/useTheme";
+import { isControlPanelWindow } from "./utils/windowContext.ts";
 
 const ControlPanel = React.lazy(() => import("./components/ControlPanel.tsx"));
 const OnboardingFlow = React.lazy(() => import("./components/OnboardingFlow.tsx"));
@@ -42,9 +43,7 @@ function MainApp() {
   const [postOnboardingSettingsSection, setPostOnboardingSettingsSection] = useState(undefined);
 
   const isAgentPanel = window.location.search.includes("agent=true");
-  const isControlPanel =
-    !isAgentPanel &&
-    (window.location.pathname.includes("control") || window.location.search.includes("panel=true"));
+  const isControlPanel = !isAgentPanel && isControlPanelWindow();
   const isDictationPanel = !isControlPanel && !isAgentPanel;
 
   useEffect(() => {
--- a/src/utils/windowContext.ts
+++ b/src/utils/windowContext.ts
@@ -1,9 +1,8 @@
-// The control panel loads with "control" in the path (packaged) or ?panel=true
-// (dev server); the dictation panel is the plain URL.
+// Window creation tags the control panel with ?panel=true in both development
+// and packaged builds; the dictation panel is the plain URL.
 export const isControlPanelWindow = (): boolean => {
   if (typeof window === "undefined") return false;
-  const { search, pathname } = window.location;
-  return pathname.includes("control") || search.includes("panel=true");
+  return new URLSearchParams(window.location.search).get("panel") === "true";
 };
 
 export const isDictationPanelWindow = (): boolean =>
--- a/test/helpers/windowContext.test.js
+++ b/test/helpers/windowContext.test.js
@@ -1,0 +1,46 @@
+const test = require("node:test");
+const assert = require("node:assert/strict");
+
+const load = () => import("../../src/utils/windowContext.ts");
+
+test("a packaged dictation window ignores control in its install path", async (t) => {
+  const originalWindow = globalThis.window;
+  t.after(() => {
+    globalThis.window = originalWindow;
+  });
+  globalThis.window = {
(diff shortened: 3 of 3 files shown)


gabrielste1n:
Merged as #1563 — your fix, rebased onto current `main`. Thanks @hsusul.

The only conflict was an import collision in `src/AppRouter.jsx` (`main` added `usePolicyStore`, you added `isControlPanelWindow`); both kept, no logic changed. It moved branches because my token can't push to a fork branch when the rebase carries `.github/workflows/` changes.

I verified the two things that made this safe to take: no `control*.html` exists anywhere outside `node_modules` and nothing loads a path containing `control`, so the pathname clause really was vestigial; and `?panel=true` survives both reload and the in-window sign-in callback, so the panel can't lose its identity mid-session. Tightening the substring match to `URLSearchParams.get("panel")` was a good call too.

PR #1419, opened 2026-07-31
fix: correct misleading Settings -> General path in translation hotkey hint

### LINKED ISSUE #1413: Translation tab points users to "Settings → General", a section that does not exist
## What's wrong

The footer line at the bottom of **Settings → Language Models → Translation** reads:

> Set the hotkey under Settings → General → Translation Hotkey

**There is no "General" section in Settings.** The nine sections are Account, Plans & Billing, Workspace, Preferences, Hotkeys, Speech-to-Text, Language Models, Privacy & Data, System.

The Translation Hotkey is in **Settings → Hotkeys**, not Preferences and not any "General".

So the line is wrong twice: it names a section that doesn't exist, and even the closest real match (Preferences, which "General" is the internal name for) is not where the setting lives.

## Why it happens

`General` is the internal key namespace for the Preferences section — the i18n keys are `settingsPage.general.*`. The Hotkeys section was split out of what used to be "General" and the keys never followed, so the hotkeys still live under `settingsPage.general.translationHotkey.*`, `settingsPage.general.meetingHotkey.*`, and `settingsPage.general.voiceAgentHotkey.*` while rendering in **Hotkeys**.

Whoever wrote this footer read the namespace as a user-facing location. That is a reasonable mistake to make from the source, and it is the same mistake the help centre made twice — see below.

## Suggested fix

Change the string to point at **Settings → Hotkeys → Translation Hotkey**, and check whether any sibling strings repeat the "General" wording.

Worth a broader grep: any user-visible copy containing "General" is suspect, because no section by that name is rendered anywhere in the app.

## How it was found

During a help-centre settings-claim verification pass on 2026-07-31 (desktop `bf8b7e0b`, 1.8.1). Josh walked all five settings screens in the running app against 20 published documentation claims; this appeared in a screenshot of the Translation tab.

The same `settingsPage.general.*` namespace confusion produced two separate wrong claims in our own docs, both now corrected:

* **Current version** was documented as being in Preferences (from `settingsPage.general.updates.currentVersion`). It actually renders under **System → Updates** — the block is guarded by `case "system":` in `SettingsPage.tsx:3757`.
* Three hotkeys were documented against Preferences for the same reason.

This one differs from those in that it ships to users rather than sitting in documentation.

## Impact

Low severity, but it sends a user hunting for a section that isn't there, in the one place they've already gone looking for a setting they couldn't find. No functional impact.

RakshitKaintura:
Fixes #1413.

## Why
The translation hotkey hint incorrectly instructed users to look under "Settings → General → Translation Hotkey". Because the Hotkeys section was previously split out from the old "General" (Preferences) tab, the text became outdated and sent users hunting for a section that doesn't exist.

## What
This PR updates the user-facing text in the English translation file to correctly point to: `Settings → Hotkeys → Translation Hotkey`. 

- A broader codebase search was conducted, confirming there are no other stray user-facing references to the non-existent "General" settings section.
- Foreign locale files were left untouched, as standard translation workflows (e.g. Crowdin/i18next) will pick up the English string change and prompt translators to update their respective locales organically.

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -3343,7 +3343,7 @@
       "title": "Translation Prompt",
       "description": "Customize how the model translates your dictation"
     },
-    "hotkeyHint": "Set the hotkey under Settings → General → Translation Hotkey",
+    "hotkeyHint": "Set the hotkey under Settings → Hotkeys → Translation Hotkey",
     "targetsLabel": "Target languages",
     "addTarget": "Add language…",
     "activeTarget": "Active target language",


gabrielste1n:
Thanks @RakshitKaintura — the diagnosis here was right, and it's fixed on `main` in #1564.

Closing as superseded rather than merged, for one reason: this changes the English string only, and the same wrong path ("Settings → General", a section that doesn't exist) was present in all ten locale bundles. #1564 fixes every one, with each locale pointing at its own translated section label.

Your report is what surfaced it, and the issue it closes (#1413) credits that. Sorry not to be able to take the commit itself.

PR #1426, opened 2026-07-31
fix(export): localize the own-voice speaker label in transcripts

### LINKED ISSUE #1422: Transcript exports label your own diarized segments with a hardcoded English "You"
## Component

Transcript export formatting (`src/helpers/transcriptFormatter.js`, `resolveSpeaker`)

## Reproduction

On current `main` (`bf8b7e0b4e1de0c9779c63f4752bd80bdd39ee2c`), export a transcript whose own-voice segments are a mix of diarized (`speaker: "you"`) and un-diarized (`source: "mic"`) rows while the UI language is not English:

```bash
const { changeLanguage } = require("./src/helpers/i18nMain");
const { formatTxt, formatJson } = require("./src/helpers/transcriptFormatter");

changeLanguage("de");

const note = { title: "Standup", created_at: "2026-01-01T00:00:00Z" };
const segments = [
  { source: "mic", speaker: "you", timestamp: 0, text: "Guten Morgen." },
  { source: "mic", timestamp: 3, text: "Noch etwas." },
  { source: "system", speaker: "speaker_0", timestamp: 8, text: "Hallo." },
];

console.log(formatTxt(note, segments, {}));
console.log(JSON.parse(formatJson(note, segments, {})).metadata.speaker_count);
console.log(JSON.parse(formatJson(note, segments, {})).speakers);
NODE
```

Current output (trimmed):

```text
[00:00:00] You:
Guten Morgen.

[00:00:03] Du:
Noch etwas.

[00:00:08] Speaker 1:
Hallo.

3
[ 'You', 'Du', 'Speaker 1' ]
```

## Expected

Both own-voice lines resolve to the same localized label (`Du` under `de`), and the JSON export reports two speakers.

## Actual

`resolveSpeaker` returns the English literal `"You"` for `seg.speaker === "you"`:

```js
if (seg.speaker && speakerMappings[seg.speaker]) return speakerMappings[seg.speaker];
if (seg.speaker === "you") return "You";
...
if (seg.source === "mic") return i18nMain.t("transcript.speaker.you");
```

The sibling branch two lines below already resolves the same concept through `transcript.speaker.you`, which every shipped locale translates (`"Du"` in `de`, `"Vous"` in `fr`, …). Because the two branches disagree, one speaker produces two distinct display names in the same export.

`formatJson` derives `speakers` and `speaker_count` from the resolved names, so the split also inflates the reported speaker count — three speakers for a two-person conversation above.

Both segment shapes reach the same note in practice: `mergeWithTranscript` stamps `speaker: "you"` onto mic segments only when offline diarization produced clusters, while live meeting identification labels system speakers and leaves mic segments without a `speaker`, so a transcript autosaved before (or without) diarization keeps the un-stamped shape.

## Environment

- macOS 15.7.3 (platform-independent — pure formatting logic)
- Node 24.9.0 / npm 11.6.0
- UI language: any non-English locale (`de` above)

## Impact

For every non-English user, TXT/Markdown/SRT transcript exports — and the Markdown mirror written to disk — label their own lines in English, and can label the same person twice under two names in one file. The JSON export's `speakers` list and `speaker_count` are wrong by the same amount.

## Proposed scope

Resolve the `speaker === "you"` branch through the existing `transcript.speaker.you` key so both branches agree. Add focused coverage that a non-English locale yields one label for both own-voice segment shapes and that `formatJson` counts them as one speaker.

No paid API request, provider credential, or microphone is required to reproduce.

I intend to work on this.

hsusul:
## Summary

`resolveSpeaker()` returned the English literal `"You"` for a diarization-stamped own-voice segment while the sibling branch two lines below resolved the same concept through the localized `transcript.speaker.you` key. Under a non-English UI language one person therefore rendered under two names in the same export, and the JSON export counted them as two speakers.

## Affected component

`src/helpers/transcriptFormatter.js` — `resolveSpeaker()`, shared by `formatTxt`, `formatSrt`, `formatJson`, `formatMd`, the transcript export handlers, and the Markdown mirror written to disk.

## Problem

```js
if (seg.speaker && speakerMappings[seg.speaker]) return speakerMappings[seg.speaker];
if (seg.speaker === "you") return "You";
...
if (seg.source === "mic") return i18nMain.t("transcript.speaker.you");
```

Both branches describe the same person. Offline diarization stamps `speaker: "you"` onto mic segments (`DiarizationManager.mergeWithTranscript`), while live meeting identification only labels system speakers, so a transcript autosaved before — or without — a diarization pass keeps the un-stamped shape. A note can contain both shapes, and every shipped locale translates `transcript.speaker.you` (`"Du"` in `de`, `"Vous"` in `fr`, …), so the two branches produced different strings for the same speaker.

`formatJson` derives `speakers` and `speaker_count` from the resolved names, so the split also inflated the reported speaker count.

## Minimal reproduction

```bash
const { changeLanguage } = require("./src/helpers/i18nMain");
const { formatTxt, formatJson } = require("./src/helpers/transcriptFormatter");

changeLanguage("de");

const note = { title: "Standup", created_at: "2026-01-01T00:00:00Z" };
const segments = [
  { source: "mic", speaker: "you", timestamp: 0, text: "Guten Morgen." },
  { source: "mic", timestamp: 3, text: "Noch etwas." },
  { source: "system", speaker: "speaker_0", timestamp: 8, text: "Hallo." },
];

console.log(formatTxt(note, segments, {}));
console.log(JSON.parse(formatJson(note, segments, {})).speakers);
NODE
```

**Actual (before):**

```text
[00:00:00] You:
Guten Morgen.

[00:00:03] Du:
Noch etwas.

[00:00:08] Speaker 1:
Hallo.

[ 'You', 'Du', 'Speaker 1' ]
```

**Expected (after):** both own-voice lines read `Du:`, and `speakers` is `[ 'Du', 'Speaker 1' ]` with `speaker_count` 2.

## Root cause

A hardcoded English string on one of two branches that resolve the same speaker, where the translation key used by the other branch already existed.

## Implementation

```js
if (seg.speaker === "you") return i18nMain.t("transcript.speaker.you");
```

The module already imports `i18nMain` and already uses this exact key.

## Why the fix is minimal

One expression. No new translation keys (`transcript.speaker.you` exists in all ten shipped locales — `npm run i18n:check` passes), no new imports, and no change to resolution order: an explicit `speakerName`, a caller-supplied `speakerMappings` entry, the `speaker_N` numbering, and the `source`-based fallbacks all keep their current precedence.

## Regression tests

Added to `test/helpers/transcriptFormatter.test.js` (3 tests):

- own-voice segments share one label whether or not diarization stamped them — asserts both shapes render as `Du` under `de` and that `You:` never appears — **fails on unmodified `main`**
- the JSON export counts both own-voice segment shapes as one speaker — asserts `speakers` and `speaker_count` — **fails on unmodified `main`**
- an explicit speaker mapping still overrides the own-voice label (guards the branch above the change)

Both locale tests restore `en` through `t.after()`, so ordering with the rest of the file is unaffected. The existing English assertion `[00:00:01] You:` for an un-stamped mic segment is unchanged and still passes, which confirms English output is byte-identical.

Pure formatting logic with no platform branches — identical on macOS, Windows and Linux. English output is unchanged; every non-English UI language gains the corrected label. All four export formats and the Markdown mirror (`markdownMirror.writeTranscript`, which calls `formatMd`) go through the same function.

## Lifecycle and cleanup

No resources, listeners, timers, or files are involved; `resolveSpeaker` stays a pure string lookup and mutates nothing.

## Validation

| Command | Result |
| --- | --- |
| `node --test test/helpers/transcriptFormatter.test.js` (before fix) | 14 tests, 12 pass, **2 fail** |
| `node --test test/helpers/transcriptFormatter.test.js` (after fix) | 14 tests, 14 pass, 0 fail — repeated 3× |
| `node --test` over `transcriptFormatter`, `markdownMirror`, `speakerMerge`, `audioSegmentMerge` | 31 tests, 30 pass, 0 fail, 1 skipped |
| `npm test` | 1110 tests, 1104 pass, 0 fail, 6 skipped |
| `npm run i18n:check` | pass — locale keys and placeholders consistent |
| `npm run lint` | pass |

The 6 skips are environmental and identical before and after: 5 `linuxLauncherSandbox` tests that only run on Linux, and 1 FFmpeg-dependent audio test skipped because `--ignore-scripts` omits the bundled FFmpeg download (that skip is also the one in the four-file run above).

**Not run:** `npm run format:check`. It fails on unmodified `upstream/main` at this commit for three files unrelated to this change (`components/notes/overview/OverviewNoteList.tsx`, `components/notes/ShareNoteDialog.tsx`, `hooks/useNoteDragAndDrop.ts`); reformatting them here would be out of scope. `npm run lint`, the ESLint half of that script, passes. Packaging (`npm run build`) was not run — it needs the platform binary downloads and this change touches no build or packaging config.

## Compatibility

English exports are byte-identical, so no existing test, snapshot, or user-visible output changes for the default locale. No public API, IPC contract, stored-transcript format, or locale-file change; only the rendered label for non-English users is corrected. Previously exported files on disk are not rewritten — the Markdown mirror updates a note's transcript file the next time it is written.

## Non-goals

- The `Speaker {n}` fallback a few lines below is also an English literal. Localizing it would change every diarized speaker label and belongs in its own change.
- `speakerKey()` still keys the two own-voice shapes separately, so consecutive own-voice segments of different shapes remain separate blocks. Changing the merge rule would alter segment grouping and is out of scope here.

Fixes #1422


Chadpiha (approved):
Added translations for additional hardcoded "speaker" and "unknown speaker" labels

PR #1430, opened 2026-08-01
perf: prewarm push-to-talk microphone capture

healthwiredgames:
## Summary
- acquire the microphone during the existing 150 ms push-to-talk hold guard
- reuse the prepared stream for batch and streaming recordings instead of reopening the device
- cancel pending or prepared capture on quick taps, interruptions, device changes, and cleanup
- defer a release that arrives while recording startup is still in flight
- log batch recording startup timing for field verification

## Verification
- npm run typecheck
- npm run lint
- npm test (1111 tests: 995 passed, 116 skipped, 0 failed)
- prepared microphone tests (4 passed)
- Prettier check on all changed files

This preserves the existing 150 ms hold threshold and activation behavior.

### DIFF
--- a/main.js
+++ b/main.js
@@ -1209,6 +1209,7 @@ async function startApp() {
               return;
             }
             windowManager.showDictationPanel();
+            windowManager.sendPrepareDictation();
             const pressTime = now;
             globeKeyDownTime = pressTime;
             globeKeyIsRecording = false;
@@ -1266,6 +1267,9 @@ async function startApp() {
             globeKeyIsRecording = false;
             debugLogger?.debug("[Globe] Stopping dictation (push release)");
             windowManager.sendStopDictation();
+          } else {
+            windowManager.sendCancelDictationPreparation();
+            windowManager.hideDictationPanel();
           }
         }
       }
@@ -1291,6 +1295,7 @@ async function startApp() {
       if (wasRecording) {
         windowManager.sendCancelDictation();
       } else {
+        windowManager.sendCancelDictationPreparation();
         windowManager.hideDictationPanel();
       }
     });
@@ -1329,6 +1334,7 @@ async function startApp() {
         const now = Date.now();
         if (now - rightModLastStopTime < POST_STOP_COOLDOWN_MS) return;
         windowManager.showDictationPanel();
+        windowManager.sendPrepareDictation();
         const pressTime = now;
         rightModActiveKey = modifier;
         rightModDownTime = pressTime;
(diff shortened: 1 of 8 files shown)


xlurie:
Same production data I've posted on #1284 applies here. I've been running a persistent-master-stream mic patch on Windows 11 since 2026-07-11: **0.1–0.2 ms** to re-acquire (logged per call) against **52.8 ms** for a fresh `getUserMedia` with the driver already warm and several hundred ms after an idle gap. There were also 11 days in between where an auto-update silently reverted the patch and the first-word clipping came back — retrospective and user-perceived rather than instrumented, but it lines up with the cold open being the real cost on Windows, not something specific to one device.

Handing `startRecording` the already-prepared stream means the device is opened once instead of twice, which sidesteps the ordering problem I raised on #1284, where the re-warm is fired un-awaited at press time and races the recording's own `getUserMedia`.

Two things worth flagging for whoever sequences the two PRs:

- **They conflict textually.** This PR replaces `warmupMicDriver()` with `prepareMicCapture()` in the same block #1284 rewrites, so whichever lands second needs a rebase rather than a merge.
- **Their coverage differs, and one path could fall between them.** `prepare-dictation` is sent from the push-to-talk paths, and `handleStart` calls `prepareMicCapture()` directly, while this PR also removes the `warmupMicDriver()` call from `handleToggle`. A plain toggle hotkey — tap to start, tap to stop, which is my setup — would then get no warm-up at all if this merges and #1284 doesn't. Sending `prepare-dictation` on the toggle press too, or keeping a re-warm for that path, would cover it.

I haven't run this branch yet (I'm testing #1284's today), but happy to add a Windows datapoint here as well if that's useful.

gabrielste1n:
Superseded by #1493, which is now merged as 28cb7f82.

Thanks @healthwiredgames — the prepared-capture idea from this PR is the core of the fix that shipped. #1493 kept your approach of opening the mic at push-to-talk key-down and handing the same stream to the recording, and built on it:

- Both start paths join a single open via `PreparedMicCapture.take()`, with in-flight dedup and generation-based cancellation.
- The toggle press prepares too, not just push-to-talk.
- A buffering `MediaRecorder` starts the moment the prepared stream resolves, and `createBatchRecorder` adopts it with no gap — so frames from before the recording formally starts land in the transcript, not just the head start.
- A max-age expiry guarantees a prepare the renderer declines can't hold the mic open silently.

Credit is recorded in the #1493 description. Closing this one as superseded — the work landed.

PR #1453, opened 2026-08-03
fix(meeting): align mic reference on helper capture and surface silent system audio

gabrielste1n:
Investigation of a Windows debug log where a full 3.5-minute meeting produced **zero** system-audio transcripts. Both realtime sockets were healthy and received near-identical byte volume, but only one produced text:

```
disconnect  audioBytesSent 3974400  segments 12  textLength 922   ← mic
disconnect  audioBytesSent 3978240  segments  0  textLength   0   ← system
```

The system channel was delivering pure digital silence. Four independent confirmations: `systemSpeaking: false` on all 15 segments (the detector's VAD floor is only 0.004 RMS), `micCorrelation: "0.000"` on every one, `Diarization complete { segmentCount: 0 }` over 83s of audio, and every transcript arriving as `source: "mic"` — the remote party was only ever heard as speaker bleed.

Why the WASAPI process loopback returned silence is still unconfirmed. This PR fixes the real bugs found while tracing it and closes the diagnostic gaps that made it unconfirmable.

## Mic reference alignment was disabled on Windows and Linux

`hasNativeMeetingSystemAudio()` read the capability *mode*, which is `"loopback"` on Windows and Linux for **both** the helper and the renderer capture paths. Alignment therefore only ever engaged on macOS.

A helper process delivers system audio on its own schedule, so mic chunks must wait for the reference to catch up before the echo detector can correlate them. Without alignment, `analyzeMicChunk` runs against a system window that has not arrived yet. It was masked in this log only because the AEC helper happened to be running; whenever AEC is unavailable on Windows or Linux, bleed detection is silently dead.

Now keyed on the strategy capture actually settled on — `native`, `wasapi-loopback`, `pipewire-loopback`. `startMeetingSystemAudio` routes all six exits through a `settle()` helper so the fallback paths (notably the wasapi → renderer-loopback downgrade) record it too.

## Silent system audio was undetectable

The Windows and Linux helpers pad the stream with silence whenever no application is rendering, so a dead capture and a quiet meeting are identical by byte count — `convertRawPcmToWav`'s only guard (`stat.size === 0`) can never fire. A whole meeting gets recorded, FFmpeg and sherpa-onnx burn on 83s of zeros, and nothing reports it.

`SystemAudioSilenceMonitor` warns once when a full 60s grace period of system audio arrives with no audible chunk. It never fires if the far end speaks even once, so a normally quiet meeting stays silent. It reuses the RMS the echo detector already computes rather than rescanning 100 buffers/sec.

## Also fixed

- **`language` was dropped for BYOK realtime sessions.** `connect()` never destructured it and `session.update` omitted it, though callers pass it. Managed mode was fine (server token carries it); BYOK always ran auto-detect. Same failure surface as the hallucinated `"음식"` / `"Поток."` fragments in the log — auto-detect drifting on near-silent bleed audio.
- **`sampleRate` was ignored** by `connect()`, which reads `inputRate`. Harmless today only because both default to 24000.
- **Both sockets logged identically** — mic and system were distinguishable only by byte counts at disconnect. Now tagged `stream: mic|system`.
- **`System Info` omitted the OS build**, which is exactly what decides native capture availability. Now logs `osRelease`, `osVersion`, `arch`.
- **`meetingMicStatsLogCount` was never reset** (`meetingSendCounts` is, per meeting). After ~200 chunks in the app's lifetime the mic-level diagnostic goes dark permanently — likely why there is not a single mic-level line in the 3-minute log that motivated this.
- Removed `getMeetingSystemAudioMode`, a pure alias left with one caller.

## Tests

13 new tests. The realtime ones drive a real socket through the session handshake against a local `WebSocketServer` and assert on the `session.update` payload. Verified they fail against the pre-fix code:

```
✖ pins the requested language on a self-configured session
✖ declares the rate given as sampleRate, and lets inputRate win
```

Full suite: 1114 tests, 998 pass, 0 fail (116 pre-existing skips). Lint and Prettier clean.

## Deliberately out of scope

- **No user-facing warning for dead system audio.** Surfacing it needs i18n keys across 9 locale files plus component work. The `warn` log carries strategy, grace period, and chunk count — what a support log needs.
- **Dictation streaming still does not pin language.** Its call site never passes one and `dictationRealtimeStart`'s option type has no `language` field, so wiring it means touching the renderer, preload, types, and `fetchRealtimeToken`. The client-side fix is in place for whenever it is threaded through.

## Testing notes

- Windows and Linux: run a meeting with the native helper, confirm bleed detection still behaves with the AEC helper both available and unavailable.
- macOS: confirm no regression — alignment now engages once `startMeetingSystemAudio` settles rather than from the first chunk. Mic audio only reaches `sendMeetingAudio` after the start IPC resolves, so there should be no window, but worth a look.
- Confirm the silence warning fires on a meeting with system audio muted, and does **not** fire on a normal meeting where the far end is quiet for over a minute.

### DIFF
--- a/src/helpers/debugLogger.js
+++ b/src/helpers/debugLogger.js
@@ -1,4 +1,5 @@
 const fs = require("fs");
+const os = require("os");
 const path = require("path");
 const { app } = require("electron");
 
@@ -72,6 +73,11 @@ class DebugLogger {
       this.debug("Debug logging enabled", { logFile: this.logFile });
       this.info("System Info", {
         platform: process.platform,
+        // OS build decides which native capture APIs are usable (Windows
+        // process loopback, macOS audio tap), so it belongs in every report.
+        osRelease: os.release(),
+        osVersion: os.version(),
+        arch: process.arch,
         nodeVersion: process.version,
         electronVersion: process.versions.electron,
         appPath: app.getAppPath(),
--- a/src/helpers/ipcHandlers.js
+++ b/src/helpers/ipcHandlers.js
@@ -27,6 +27,11 @@ const TINFOIL_REALTIME_MODEL = "voxtral-mini-4b-realtime";
 const liveSpeakerIdentifier = require("./liveSpeakerIdentifier");
 const MeetingEchoLeakDetector = require("./meetingEchoLeakDetector");
 const { partitionPendingMicFinals, isWithinRetractWindow } = require("./meetingMicHoldback");
+const {
+  requiresMicReferenceAlignment,
+  SystemAudioSilenceMonitor,
+  SYSTEM_AUDIO_SILENCE_GRACE_MS,
+} = require("./meetingSystemAudio");
 const { applySmartSpacing } = require("./smartSpacing");
 const { applyAutoLearnSetting } = require("./autoLearnSetting");
 const { DEFAULT_RETENTION_SETTINGS, applyRetentionSettings } = require("./retentionSettings");
@@ -5384,16 +5389,22 @@ class IPCHandlers {
         if (newSystem) {
           const secrets = await fetchRealtimeToken(tokenEvent, options, { streams: 2 });
           pairs = [
-            { streaming: newMic, secret: secrets[0] },
-            { streaming: newSystem, secret: secrets[1] },
+            { streaming: newMic, secret: secrets[0], source: "mic" },
+            { streaming: newSystem, secret: secrets[1], source: "system" },
           ];
         } else {
-          pairs = [{ streaming: newMic, secret: await fetchRealtimeToken(tokenEvent, options) }];
+          pairs = [
+            {
+              streaming: newMic,
+              secret: await fetchRealtimeToken(tokenEvent, options),
+              source: "mic",
+            },
+          ];
         }
 
         await Promise.all(
-          pairs.map(({ streaming, secret }) =>
-            streaming.connect({ apiKey: secret, token: secret, ...connectOpts })
+          pairs.map(({ streaming, secret, source }) =>
+            streaming.connect({ apiKey: secret, token: secret, ...connectOpts, label: source })
           )
         );
 
@@ -5556,10 +5567,8 @@ class IPCHandlers {
       return "unsupported";
     };
 
-    const getMeetingSystemAudioMode = () => getMeetingSystemAudioCapabilityMode();
-
     const getMeetingSystemAudioPlan = async () => {
-      const mode = getMeetingSystemAudioMode();
+      const mode = getMeetingSystemAudioCapabilityMode();
       if (mode === "unsupported") {
         return { mode, strategy: "unsupported" };
       }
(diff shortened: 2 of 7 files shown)


xAlcahest (approved):
Reviewed with the debug-log story as the spec, checking each claim against the code instead of taking the narrative on faith.

The mode-vs-strategy bug is real and the fix keys on the right signal. `getWindowsSystemAudioAccess` returns mode `"loopback"` with strategy either `"wasapi-loopback"` or `"loopback"`, so the old `hasNativeMeetingSystemAudio()` check could never engage alignment on Windows or Linux. The new `requiresMicReferenceAlignment` reads the settled strategy, and every exit of `startMeetingSystemAudio` routes through `settle()`, including the wasapi to renderer-loopback downgrade and both mic-only fallbacks.

Healthy-path invariance holds. The renderer buffers mic chunks behind `socketReady` in meetingRecordingStore.ts, which only flips after the start IPC resolves, and the strategy settles inside that handler, so by the time any mic chunk reaches `sendMeetingAudio` the alignment decision matches what the old mode check gave on macOS. The AEC path through `processMeetingMicWithAec` is untouched.

The silence monitor reuses the detector's own floor (`AUDIBLE_SYSTEM_RMS` 0.004, same value as `MIN_SYSTEM_RMS` behind `systemSpeaking`), warns once, and the `heardAudio` latch means a normally quiet far end never trips it. The tests pin the once-only behavior, the latch, the clock starting at the first chunk, reset re-arming, and the boundary value.

On the realtime client: the `createSocket` factory path is untouched, `session.update` still goes out only for non-preconfigured sessions (with a test pinning that), and the reconnect path passes the same connectOpts, so language and the stream label survive a session_expired reconnect. Both meeting connectOpts already carried `language` and `sampleRate`; `connect()` just dropped them, exactly as described.

I also checked the premise against the WASAPI docs: a process-loopback capture delivers silence whenever the non-excluded processes have no active render streams, and loopback only sees shared-mode streams, so from inside the helper a dead capture really is indistinguishable from a quiet meeting. Monitoring plus logging the OS build beats guessing at a fix.

Verified locally: full suite 1114 tests, 1112 pass, 0 fail, 2 skipped on Node 24; prettier and eslint clean on the touched files; and the two new realtime tests fail against the pre-fix client exactly as the description says.

One nit, not blocking: the silence warning logs `chunks: meetingSendCounts.system`, which stays 0 in local mode because that counter only increments on the streaming dispatch path. It reads correctly for the realtime scenario the motivating log came from.

No commits pushed; nothing needed fixing.

stantheman0128:
Support. Complementary to #1474, not a duplicate.

#1474 refreshes WASAPI helper capability before recording starts (stale probe cached unavailable). This PR is the next layer: mic-reference alignment never engaged on Windows/Linux because `hasNativeMeetingSystemAudio` keyed on loopback mode, so bleed detection was dead whenever AEC was off. The silence monitor is also the right diagnostic for helper-padded zeros that `stat.size === 0` cannot see.

I will not expand #1474 into the silence monitor or the BYOK language pin. Those belong here.

PR #1457, opened 2026-08-03
fix(linux): keep meeting notifications interactive on Linux

### LINKED ISSUE #1456: meeting notification cannot be dismissed or started after hover on Linux
**Describe the bug**
After hovering a meeting notification and moving the cursor away, the close button or start notes button cannot be accessed again on Linux.

**To Reproduce**

1. Trigger a meeting notification.
2. Hover the notification so the close button appears.
3. Move the cursor away.
4. Move back over the notification.

**Expected behavior**
The popup becomes click-through, so it cannot receive the next hover event.

**Screenshots**

https://github.com/user-attachments/assets/570e3b8c-03f4-460d-a75d-bf88a49f729d

- OS: Linux (Hyprland)
- Version: v1.8.1

IdrisGit:
## Summary
Keep meeting notification windows interactive on Linux so they can receive subsequent hover events and expose the close and start notes button.

macOS and Windows retain their existing click-through behavior.

## Verification Video

https://github.com/user-attachments/assets/c026ce9c-53d8-4b35-9fe0-e32a37ff09ed

fixes #1456

### DIFF
--- a/src/helpers/windowManager.js
+++ b/src/helpers/windowManager.js
@@ -131,6 +131,13 @@ class WindowManager {
     if (!this.notificationWindow || this.notificationWindow.isDestroyed()) {
       return;
     }
+
+    // Electron cannot forward hover events on Linux, so returning to click-through
+    // would prevent this overlay from receiving the next mouse enter.
+    if (process.platform === "linux") {
+      return;
+    }
+
     if (interactive) {
       this.notificationWindow.setIgnoreMouseEvents(false);
     } else {


gabrielste1n:
Thanks @IdrisGit — your diagnosis was correct and it's what shipped, in #1562.

I couldn't take this one as written, for a reason worth explaining rather than just closing:

The patch `return`s early on Linux **before** the `interactive` branch, so it also skips `_notificationDismissTimer.pause()` / `.resume()`. That timer is what stops the auto-dismiss countdown closing the card under the user's pointer while they're reading it — so this would have traded the unclickable-card bug for a closes-while-you-read one.

It also predates `2cb2b2bd`, which added the `sender` parameter and the timer to this method, so it no longer applied cleanly either.

#1562 keeps your insight exactly — `forward` is macOS/Windows only, so a Linux card returned to click-through never sees another `mouseenter` — but skips *only* the hit-testing toggle and leaves the countdown running on every platform. Since the card is only click-through on macOS to begin with (`windowManager.js:1239-1241` is `darwin`-gated), there's nothing to restore on Linux anyway.

Tests in `test/helpers/notificationInteractivity.test.js` pin both halves, including the regression this would have introduced. Closing in favour of #1562.

IdrisGit:
got it 👍️

Issue #1454, opened 2026-08-03
The application loses the entire transcript and injects only words from the dictionary.


edernucci:
**Describe the bug**
Since the last update, it has become impossible to use the application. Whenever I speak a slightly longer text, which would normally break into more than one paragraph, the live transcription even shows that it is understanding everything I am saying. However, when I finish speaking, it loses everything I said and injects only a few words from my dictionary.  This happens 3 to 5 times out of every 10 uses. Now, I have to speak only in very short sentences, because if the sentence is even a little bit longer, it gets completely lost and injects several words from my dictionary in place of the transcription. 

**To Reproduce**
Just use the application normally, using the default conversation Hotkey, and say a sentence that is a little longer than usual, something that lasts around 15 to 20 seconds; it becomes completely unacceptable, as it replaces it with characters. 

**Expected behavior**
It should work the way it worked in previous versions. 

**Screenshots**

<img width="781" height="638" alt="Image" src="https://github.com/user-attachments/assets/14c09ce0-d5d0-4c2e-bb87-c1943c4c9fb6" />

**Desktop (please complete the following information):**

- OS: Windows
- Version: 11

**Additional context**
It is important to notice the screenshot I added, where I even enabled live transcription to be able to identify if I had a problem with my microphone. During my speech, the live transcription shows everything I am saying normally; however, when it finishes, both the transcript and what was processed by the AI become only the dictionary characters.

edernucci:
During the typing process, if you speak quickly and directly in a single sentence, the application works. However, if you reason, think again, take a brief pause and breathe — which is how I used to use the application — and only then press the push-to-talk button to stop the transcription, it no longer works the way the application comes packaged, because it now comes with VAD enabled by default.

So, for anyone having this type of problem, just **disable VAD in all cases, whether for chats, transcription, or meeting detection**, and the application will return to the behavior it had a few months ago. 

Even so, eventually, I've noticed that it has improved a lot, but it's still missing some sentences, especially the initial sentence I said. It kills that and only keeps the final sentence. So, disabling VAD has helped a lot, but it still didn't manage to solve the whole problem.

Chadpiha:
Fix will be in the next release :)

gabrielste1n:
@edernucci — the main problem you reported is fixed on `main`, but I'm keeping this open because of your follow-up.

[...]

PR #1461, opened 2026-08-03
fix(dictation): Make VAD opt-in and rescue dictionary-echo decodes

Chadpiha:
## Problem
Since updating to 1.8.x, long dictations (15–20s, natural pauses) intermittently paste only custom-dictionary words instead of the transcript, while live preview shows correct text during recording (#1454, Windows 11, 3–5 of 10 attempts).

## Root cause
- Dictation decodes ran without VAD through v1.7.4; v1.7.5 turned server-side Silero VAD on by default. On pause-heavy speech, VAD intermittently strips the spoken audio from the final decode. Whisper then decodes near-silence seeded with the custom dictionary (passed as initial_prompt) and does what Whisper does with silence + prompt: continues the prompt. - The failing raw transcripts are literally ( + the first dictionary words, truncated. The existing echo filter (#852) only catches ≥70% dictionary echo, so partial echoes were pasted. Preview is immune because it decodes 1.5s chunks of continuous speech with no prompt — which is why recordings looked healthy while the final result was destroyed.

## Fix (two layers)

- Dictation VAD is now opt-in (default off). Notes/meeting recordings keep VAD (long silences genuinely benefit). Users who explicitly enabled the toggle keep it — only the unset default changes.
- Dictionary-echo rescue: if the final local decode matches the dictionary prompt, retry once without the prompt and with VAD disabled (new narrow skipVad option on the IPC handler). Real speech is recovered; true silence still surfaces "No audio detected". This protects users who keep VAD enabled.


gabrielste1n (approved):

The default flip is the real fix and it's threaded correctly through all four sites that must agree (`settingsStore.ts`, the `ipcHandlers` constructor seed, `_getWhisperVadSettings`, `resolveContextSileroEnabled`). Migration semantics hold up under scrutiny: `readBoolean(key, false)` returns `stored === "true"`, so users who explicitly toggled the setting keep their choice and only the unset default moves — and there's a UI toggle at `SettingsPage.tsx:1539`, so the change is discoverable and reversible. `skipVad` is correctly stripped at the IPC boundary so it never reaches `whisperManager`, and reusing `arrayBuffer` for the second `invoke` is safe (structured clone, no detach).

I checked the PR's framing against the code and it holds: Layer 1 addresses the partial echoes the ≥70% filter lets through, Layer 2 rescues the ≥70% case that previously threw the transcript away entirely.

Two small cleanups pushed: dropped a redundant `resolved` alias in favour of optional chaining, and recomputed `transcriptionProcessingDurationMs` after a rescue so the timing log covers both decodes instead of only the discarded first one.

**Follow-up worth tracking (not blocking):** dictation and notes/meetings now produce different VAD signatures on the shared whisper-server, so alternating a note recording with a dictation costs a process spawn and model reload in the dictation critical path. Partly offset — the pre-warm passes no VAD options, so the first dictation of every launch previously forced a restart and now doesn't.

Typecheck, lint, prettier clean; full suite matches the clean-main baseline exactly.

PR #1466, opened 2026-08-04
fix(widget): keep dictation overlay visible after display changes

vsem-azamat:
## Summary

Reconcile the dictation overlay bounds when display topology or metrics change.

- preserve the overlay on its current display during geometry reconciliation if it is still visible
- clamp partially visible bounds back into the display work area
- move the overlay to the cursor display when it is fully off-screen
- keep the existing show behavior where the overlay follows the cursor display
- stop clamping main overlay placement to `0`, so negative-origin monitor layouts work
- avoid reconciling while the user is dragging the overlay
- add unit tests for negative-origin placement, visible preservation, partial clamping, and off-screen recovery

Related to #1110. Complements #977/#1286, but does not implement panel-aware X11/KDE work-area correction.

## Testing

- `node --test test/helpers/windowPositionUtil.test.js`
- `npm run lint`
- `npm run format`
- `npx prettier --check src/helpers/windowConfig.js src/helpers/windowManager.js test/helpers/windowPositionUtil.test.js`
- Electron runtime smoke-test with real `screen.getAllDisplays()`

Note: a full local `npm test` run is blocked in my environment after `electron-builder install-app-deps` rebuilt `better-sqlite3` for Electron ABI; Node-based database tests then fail with `ERR_DLOPEN_FAILED: Module did not self-register`. The focused geometry tests and lint pass.

### DIFF
--- a/src/helpers/windowConfig.js
+++ b/src/helpers/windowConfig.js
@@ -152,25 +152,105 @@ const TRANSCRIPTION_PREVIEW_CONFIG = {
 };
 
 class WindowPositionUtil {
+  static getWorkArea(display) {
+    return display?.workArea || display?.bounds || null;
+  }
+
+  static clampBoundsToWorkArea(bounds, workArea) {
+    if (!bounds || !workArea) return bounds;
+
+    const maxX = workArea.x + workArea.width - bounds.width;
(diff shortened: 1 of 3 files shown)


vsem-azamat:
Closing this one, no longer pursuing it. Thanks!

PR #1473, opened 2026-08-04
fix(export): normalize meeting transcript timestamps for markdown exports (Fixes #1472)

### LINKED ISSUE #1472: Markdown export: segment timestamps use a raw seconds counter in the hours field, and header times are UTC rendered as local (v1.8.1, Windows)
**Describe the bug**

Two timestamp defects in an exported meeting transcript on Windows, OpenWhispr v1.8.1. They
look related, so I am reporting them together, but they are separable.

**1. Segment timestamps are never offset to the recording start.**

Exported segments render like this:

```
**Speaker A** `496069042:26:40`
**Speaker B** `496069043:11:05`
...
**Speaker B** `496069519:21:00`
```

Across the call the first field moves `496069042` to `496069519`, a span of **477**. The call
ran about 8 minutes, or 480 seconds. So the first field is a raw seconds counter being placed
in the hours position, rather than an offset from the start of the recording.

Expected output for an 8 minute recording is `0:00:00` through roughly `0:07:57`.

The underlying segment data appears correct: the deltas between consecutive segments track the
real conversation, and the total span matches the real duration to within a few seconds. Only
the rendering is wrong.

**2. Header and log times are UTC, displayed as if local.**

The export header read:

```
**Date:** 4 August 2026 13:02
```

for a call that started at approximately 14:02 BST (UTC+1). That is exactly one hour behind
local time.

The same offset appears in debug log filenames, which is what convinced me it is UTC rather
than an unrelated error. This file:

```
debug-2026-08-04T14-03-02-464Z.log
```

was created at 15:03:02 local time. The `Z` suffix says UTC, so the filename is correct and
consistent, but the export header presents the same kind of value as if it were local.

**To Reproduce**

1. Record a meeting note of a known duration in a timezone with a non-zero UTC offset
   (BST, UTC+1, in my case).
2. Export or view the resulting transcript.
3. Segment timestamps show a large constant in the hours field rather than starting at zero.
4. The header date and time are one hour behind local wall-clock time.

**Expected behavior**

1. Segment timestamps should be offsets from the start of the recording, starting at `0:00:00`.
2. The header timestamp should either be rendered in local time, or be explicitly marked as UTC.

**Screenshots**

N/A. Timestamp values quoted above. Transcript content omitted, it is a private business call.

- OS: Windows 11, build 10.0.26200
- Browser: N/A, desktop app
- Version: OpenWhispr 1.8.1
- Locale/timezone: en-GB, Europe/London (BST, UTC+1 at the time of recording)
- STT engine: Local, Parakeet Unified EN 0.6B
- Recording type: meeting note, local diarization enabled

N/A.

**Additional context**

v1.7.6 fixed adjacent timestamp issues, specifically SRT millisecond overflow and start times
for merged segments. This looks like a separate surviving defect on the markdown export path,
since I am on 1.8.1 and still see it.

stantheman0128:
## Summary

- Normalize epoch-millisecond meeting segment timestamps to recording-relative offsets before TXT/SRT/JSON/Markdown export.
- Parse SQLite UTC note timestamps explicitly so export headers render in local time instead of appearing one hour behind BST.

## Root cause

Diarization finalization already subtracts recording start before merge, but export-transcript read raw 
ote.transcript JSON and passed epoch-ms values into ormatTimestamp(), which expects seconds from zero. That produced giant hour fields like `496069042:26:40`.

## Verification

\\\
cd openwhispr-wt/1472-markdown-export-timestamps
node --test test/helpers/transcriptFormatter.test.js
# 13 pass (includes epoch-ms normalization + UTC created_at parsing)
\\\

## AI assistance

AI-assisted (Cursor). Stan Shih reviewed the diff.

Fixes #1472

### DIFF
--- a/src/helpers/ipcHandlers.js
+++ b/src/helpers/ipcHandlers.js
@@ -2045,7 +2045,7 @@ class IPCHandlers {
         if (format === "txt") {
           exportContent = transcriptFormatter.formatTxt(note, segments, speakerMappings);
         } else if (format === "srt") {
-          exportContent = transcriptFormatter.formatSrt(segments, speakerMappings);
+          exportContent = transcriptFormatter.formatSrt(segments, speakerMappings, note);
         } else if (format === "md") {
           exportContent = transcriptFormatter.formatMd(note, segments, speakerMappings);
         } else {
--- a/src/helpers/transcriptFormatter.js
+++ b/src/helpers/transcriptFormatter.js
@@ -21,6 +21,54 @@ function speakerKey(seg) {
   return [seg.speaker || "", named, seg.speaker ? "" : seg.source || ""].join("\u0000");
 }
 
+function parseNoteDate(createdAt) {
+  if (createdAt == null || createdAt === "") {
+    return new Date(NaN);
+  }
+  if (typeof createdAt === "number") {
+    return new Date(createdAt);
+  }
+
+  const value = String(createdAt).trim();
+  if (/[zZ]|[+-]\d{2}:\d{2}$/.test(value)) {
+    return new Date(value);
+  }
+
+  // SQLite CURRENT_TIMESTAMP is UTC without a timezone suffix.
+  return new Date(value.includes("T") ? `${value}Z` : `${value.replace(" ", "T")}Z`);
+}
+
+function resolveRecordingStartMs(segments, note) {
+  const segmentTimestamps = segments
+    .map((seg) => seg.timestamp)
+    .filter((ts) => typeof ts === "number" && Number.isFinite(ts));
+
+  if (segmentTimestamps.length > 0) {
+    return Math.min(...segmentTimestamps);
+  }
+
+  const noteDate = parseNoteDate(note?.created_at);
+  return Number.isNaN(noteDate.getTime()) ? null : noteDate.getTime();
+}
+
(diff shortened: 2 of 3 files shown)


gabrielste1n:
Merged as #1560 — your fix, unchanged, rebased onto current `main`. Thank you @stantheman0128, this was a clean diagnosis of two genuinely separate bugs.

Why it moved branches rather than being rebased in place: my token can't push to a fork branch when the rebase brings `.github/workflows/` changes along with it, which any rebase onto current `main` now does. Nothing to do with your work.

The only conflict was in `test/helpers/transcriptFormatter.test.js` and it was purely additive — `main` picked up i18n speaker-label tests from #1426 while yours added the timestamp tests. Both sets are in; nothing was dropped.

Worth recording that the blast radius was wider than the title suggests: `formatTxt`, `formatSrt`, `formatJson` and `formatMd` all share `mergeSegments`/`formatTimestamp`, and `markdownMirror.js:110` writes through `formatMd` — so every export format and the on-disk mirror were affected, not just manual markdown export. Your fix covers all of them.

stantheman0128:
Thanks @gabrielste1n — glad the diagnosis held. Appreciate you landing it as #1560 when the fork branch was awkward to push.

PR #1474, opened 2026-08-04
fix(meeting): refresh WASAPI helper capability before recording starts (Fixes #1471)

### LINKED ISSUE #1471: "windows-system-audio-helper.exe" is never launched during meeting recording
**Describe the bug**

On Windows 11, OpenWhispr v1.8.1, `windows-system-audio-helper.exe` is never launched during
meeting recording, even though the helper reports full support for native capture on this
machine. The app silently uses the Chromium loopback fallback instead.

This is not a hard failure. Meeting transcription still captures the far end, because the
fallback works when the meeting audio happens to play through the default output device. But
the native path is silently unused, and the fallback's default-device-only limitation applies
without any indication in the UI.

The helper's own probe reports the machine is fully capable:

```
> "C:\Program Files\OpenWhispr\resources\bin\windows-system-audio-helper.exe" probe
{"ok":true,"supportsSystemAudio":true,"supportsNativeCapture":true,"source":"wasapi-process-loopback"}
```

`start` also behaves correctly when invoked directly, emitting `{"type":"start"}` on stderr
before exiting when its parent pipe closes.

**To Reproduce**

1. Confirm the machine supports native capture: run
   `windows-system-audio-helper.exe probe` from `resources\bin`. It returns
   `supportsNativeCapture: true`, `source: wasapi-process-loopback`.
2. Start a meeting recording in OpenWhispr.
3. Poll the process list for `windows-system-audio-helper` for the duration of the recording.
4. The process never appears. `meeting-aec-helper-win32-x64.exe` does appear, so the meeting
   capture path is running.

I sampled the process table every 10 seconds for 57 minutes, spanning a 40-minute meeting
recording. `meeting-aec-helper-win32-x64.exe` was present throughout the recording.
`windows-system-audio-helper.exe` was present in zero samples.

**Expected behavior**

Given `probe` reports `supportsNativeCapture: true`, meeting recording should spawn
`windows-system-audio-helper.exe` and capture via WASAPI process loopback, per the behaviour
described in the v1.7.6 changelog entry and in `TROUBLESHOOTING.md`.

If the native path is deliberately skipped or fails to spawn, the fallback should be visible
somewhere in the UI, because the two paths have materially different behaviour: the fallback
only hears the default output device, so a meeting playing to a non-default device produces a
silent or one-sided note with no error.

**Screenshots**

N/A. Process-level evidence included above.

- OS: Windows 11, build 10.0.26200
- Browser: N/A, desktop app
- Version: OpenWhispr 1.8.1
- Hardware: Razer Blade Stealth 13 (Early 2020), GTX 1650 Ti Max-Q, 16GB RAM
- STT engine: Local, Parakeet Unified EN 0.6B

N/A.

**Additional context**

- The binary is present at
  `C:\Program Files\OpenWhispr\resources\bin\windows-system-audio-helper.exe`,
  171,296 bytes, dated 30/07/2026.
- Authenticode signature status is `Valid`.
- No Mark-of-the-Web / Zone.Identifier stream on the file, so it is not being blocked as a
  downloaded binary.
- Windows build 10.0.26200 is well past the 10.0.19041 (2004) minimum for process loopback.
- I have not seen "Continuing with microphone only" surfaced during recording, which
  `TROUBLESHOOTING.md` describes as the indication that system audio capture failed entirely.
- Debug logging was not enabled at the time of the recording. I can enable
  `OPENWHISPR_LOG_LEVEL=debug` and reproduce with logs if that would help, and would expect
  the `windows-system-audio-helper` entries mentioned in `TROUBLESHOOTING.md` to show why the
  spawn is not happening.

stantheman0128:
## Summary

- Re-probe windows-system-audio-helper.exe at meeting start instead of trusting a stale cached capability result.
- Treat probe spawn/timeout/binary-missing failures as transient (30s TTL) instead of pinning Chromium loopback for the whole session.
- Require supportsNativeCapture from probe JSON, not just ok.

## Root cause

When an early probe failed or returned before the packaged binary was reachable, getCapability() cached vailable: false with infinite TTL. Meeting start then selected renderer loopback (strategy: loopback) and never spawned the native helper, even though manual probe from Program Files succeeded.

## Verification

\\\
cd openwhispr-wt/1471-system-audio-helper-spawn
node --test test/helpers/windowsLoopbackAudioManager.test.js
# skipped here when helper binary is absent from dev checkout; logic covered by unit changes + meeting-start refresh path
\\\

Manual product-path Evidence still needed on an installed OpenWhispr build: start a local meeting recording and confirm windows-system-audio-helper.exe appears in Task Manager while recording.

## AI assistance

AI-assisted (Cursor). Stan Shih reviewed the diff.

Fixes #1471


gabrielste1n (approved):
Thank you @stantheman0128 for the excellent detective work on this one — pinning the root cause to the stale infinite-TTL capability cache and fixing it at the right call site (one forced re-probe at meeting start) was exactly the right shape for this fix.

Before merging we brought the branch up to date with main, restored the debug logging for transient probe failures, and replaced the live-binary test with unit coverage of the caching/TTL/force logic (including a regression test that fails against the old behavior). Full suite, typecheck, and lint are green across all platforms. Much appreciated!

PR #1477, opened 2026-08-04
fix(tools): normalize whitespace and guard nullish inputs in resolveFolderId and resolveSpace

### LINKED ISSUE #1476: fix(tools): normalize whitespace and guard nullish inputs in resolveFolderId and resolveSpace
### Component
`src/services/tools/utils.ts` (`resolveFolderId` and `resolveSpace`)

### Problem
In `src/services/tools/utils.ts`, `resolveFolderId` does not trim leading or trailing whitespace from `folderName` before searching for or creating folders, and both `resolveFolderId` and `resolveSpace` do not guard against nullish or empty/whitespace-only inputs.

Specifically:
1. When `folderName` contains surrounding whitespace (e.g., `"  Work  "`), `folders.find((f) => f.name.toLowerCase() === folderName.toLowerCase())` fails to match an existing folder named `"Work"`. Consequently, it calls `createFolder("  Work  ", spaceId)` and creates a duplicate folder with accidental surrounding spaces.
2. When `folderName` or `spaceName` is `null`, `undefined`, or whitespace-only (`"   "`), calling `.toLowerCase()` or `.trim()` throws a `TypeError`, or causes empty/whitespace folder creation.

### Expected Behavior
1. Both `resolveFolderId` and `resolveSpace` should safely normalize inputs by trimming leading and trailing whitespace.
2. If the trimmed name is missing or empty, they should return a clear error (e.g., `"Folder name cannot be empty"` / `"Space name cannot be empty"`) rather than throwing an exception or creating whitespace folders.
3. Matching against existing folders and creating new folders should use the normalized, trimmed folder name.

I intend to submit a pull request with a fix and unit test coverage.

hsusul:
### Summary
Normalizes leading and trailing whitespace and adds nullish/empty input guards in `resolveFolderId` and `resolveSpace` (`src/services/tools/utils.ts`).

### Problem
In `src/services/tools/utils.ts`:
1. `resolveFolderId` did not trim whitespace from `folderName` prior to case-insensitive lookup against existing folders or prior to calling `createFolder`. When an LLM tool call provided a folder name with surrounding spaces (e.g. `"  Work  "`), `resolveFolderId` failed to match an existing folder named `"Work"` and created a duplicate folder with surrounding spaces.
2. Neither `resolveFolderId` nor `resolveSpace` guarded against `null`, `undefined`, empty (`""`), or whitespace-only (`"   "`) input strings, resulting in uncaught `TypeError` crashes or creation of blank/whitespace folder names.

### Root Cause
`resolveFolderId` directly invoked `.toLowerCase()` on the raw `folderName` without `.trim()`, whereas `resolveSpace` called `.trim()` inline without prior type or empty checks.

### Fix
1. In `resolveFolderId`, normalize `folderName` to a trimmed string (`typeof folderName === "string" ? folderName.trim() : ""`). If empty, return `{ error: "Folder name cannot be empty" }`. Use the normalized name for lookup, error messages, and `createFolder`.
2. In `resolveSpace`, normalize `spaceName` to a trimmed string (`typeof spaceName === "string" ? spaceName.trim() : ""`). If empty, return `{ error: "Space name cannot be empty" }`.

### Behavior Before & After
- **Before**: `resolveFolderId("  Work  ")` created a duplicate folder `"  Work  "` even if `"Work"` existed. `resolveFolderId(null)` or `resolveFolderId("   ")` threw a `TypeError` or created a whitespace folder.
- **After**: `resolveFolderId("  Work  ")` matches existing folder `"Work"`. Nullish or whitespace-only inputs cleanly return `{ error: "Folder name cannot be empty" }`.

### Scope & Non-Goals
- Modifies only `src/services/tools/utils.ts` and adds unit tests to `test/services/toolUtils.test.js`.
- No changes to API contracts or database schema.

### Tests Added
- `test/services/toolUtils.test.js`:
  - `resolveFolderId normalizes whitespace and matches existing folders case-insensitively`
  - `resolveFolderId trims folder name before creating missing folders`
  - `resolveFolderId guards against nullish, empty, and whitespace-only inputs`
  - `resolveSpace normalizes whitespace and guards nullish or empty inputs`

- `node --test test/services/toolUtils.test.js` -> 6/6 passed
- `npm test` -> 1022/1022 passed
- `npm run typecheck` -> Passed
- `npm run lint` -> Passed
- `npm run i18n:check` -> Passed
- `npm run build:renderer` -> Passed
- `git diff --check` -> Passed

Fixes #1476

### DIFF
--- a/src/services/tools/utils.ts
+++ b/src/services/tools/utils.ts
@@ -9,35 +9,45 @@ export async function resolveFolderId(
   options: { createIfMissing?: boolean } = {},
   spaceId: number | null = null
 ): Promise<ResolveFolderResult> {
+  const normalized = typeof folderName === "string" ? folderName.trim() : "";
+  if (!normalized) {
+    return { error: "Folder name cannot be empty" };
+  }
+
   const folders = await window.electronAPI.getFolders(spaceId);
-  const match = folders.find((f) => f.name.toLowerCase() === folderName.toLowerCase());
+  const match = folders.find((f) => f.name.toLowerCase() === normalized.toLowerCase());
   if (match) return { folderId: match.id, created: false };
 
   if (!options.createIfMissing) {
     const available = folders.map((f) => f.name).join(", ");
-    return { error: `Folder "${folderName}" not found. Available folders: ${available}` };
+    return { error: `Folder "${normalized}" not found. Available folders: ${available}` };
   }
 
-  const result = await window.electronAPI.createFolder(folderName, spaceId);
+  const result = await window.electronAPI.createFolder(normalized, spaceId);
   if (result.success && result.folder) {
     return { folderId: result.folder.id, created: true };
   }
 
   const retry = await window.electronAPI.getFolders(spaceId);
-  const reMatch = retry.find((f) => f.name.toLowerCase() === folderName.toLowerCase());
+  const reMatch = retry.find((f) => f.name.toLowerCase() === normalized.toLowerCase());
   if (reMatch) return { folderId: reMatch.id, created: false };
 
-  return { error: result.error || `Failed to create folder "${folderName}"` };
+  return { error: result.error || `Failed to create folder "${normalized}"` };
 }
 
 type ResolveSpaceResult =
   { space: SpaceItem; error?: undefined } | { space?: undefined; error: string };
 
 export function resolveSpace(spaces: SpaceItem[], spaceName: string): ResolveSpaceResult {
-  const match = spaces.find((s) => s.name.toLowerCase() === spaceName.trim().toLowerCase());
+  const normalized = typeof spaceName === "string" ? spaceName.trim() : "";
+  if (!normalized) {
+    return { error: "Space name cannot be empty" };
+  }
+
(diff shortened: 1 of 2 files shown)


gabrielste1n (approved):
Verified against `main`: typecheck/lint clean, suite 1684 pass / 0 fail.

The real defect is the one the issue buries: `resolveFolderId` matched on the untrimmed name, `createFolder` then trimmed it and rejected it as a duplicate, and the retry lookup used the untrimmed name again — so ` Work` aborted the tool call with "already exists" and the note never landed. Normalizing once at the top and threading that value through match/create/retry/errors fixes it.

The `typeof` guards aren't defensive noise here: `folderName`/`spaceName` arrive as `args.folder as string | undefined` (`createNoteTool.ts:39-40`), an unchecked cast of model-generated JSON, so a non-string really can reach this at runtime.

For the record, the issue's other three claims don't hold — `database.js:2155,2174` already trims and rejects empty names, so the duplicate-spaced-folder, whitespace-folder and nullish-TypeError symptoms aren't reachable. Merging on the abort bug alone.

PR #1480, opened 2026-08-05
feat(meetings): route Note Recording through Tinfoil realtime

xAlcahest:
## Summary

Note Recording ignored the Tinfoil selection: `getMeetingTranscriptionOptions` in `meetingRecordingStore.ts` resolves every cloud choice through the server catalog, whose only entry is OpenAI, so a meeting recorded with Tinfoil selected silently streamed to OpenAI. This is the meeting-notes follow-up from the #1280 thread (that issue is about custom endpoints and stays open): Tinfoil ships an OpenAI Realtime-compatible transcription endpoint, and the customer wants meeting notes on it. The fix reuses the attested transport dictation already uses: a `TinfoilRealtimeStreaming` client that dials `wss://inference.tinfoil.sh/v1/realtime?intent=transcription` through tinfoil-js's pinned socket for both meeting streams, on connect and on every reconnect.

Tinfoil's transcription intent has no server VAD (docs.tinfoil.sh/guides/processing-audio, "Our mode does not do that"), so the client owns utterance boundaries: the adapter commits when the delta stream has been idle for 600ms, the same silence window our OpenAI session config uses, with a 30s per-utterance cap. Commits only fire once deltas produced text, so every commit finalizes exactly one segment for diarization and summaries and an empty buffer is never committed mid-meeting.

The invariant this PR enforces: with Tinfoil selected, audio never reaches another provider on any path. The provider registry maps `tinfoil-realtime` to the Tinfoil client and a test fails if an allowed provider would fall through to the OpenAI default class; the client overrides any caller-supplied socket factory with the attested one; socket or key failures abort the recording with a visible error instead of swapping providers, including the missing-key case, which deliberately does not downgrade to the managed OpenAI path.

## Changes

- src/helpers/tinfoilRealtimeStreaming.js: Tinfoil realtime client extending OpenAIRealtimeStreaming, with the attested socket factory and the delta-idle commit adaptation; disconnect drains the in-flight commit, and empty commit replies are retried once then dropped instead of surfacing errors.
- src/helpers/meetingStreamingProviders.js: meeting provider registry and allow-list extracted from ipcHandlers, with tinfoil-realtime added; the allow-list derives from the registry keys so an allowed provider always has a client class.
- src/helpers/ipcHandlers.js: use the extracted registry; TINFOIL_REALTIME_MODEL moves to the client module.
- src/stores/meetingRecordingStore.ts: BYOK Tinfoil selection returns tinfoil-realtime before the catalog fallback, mirroring the Corti branch.
- test/helpers/tinfoilRealtimeStreaming.test.js: factory and connect pins, privacy pins (rogue factory overridden, no fallback socket on failure), commit cadence, in-flight guard and ack-timeout self-heal, 30s cap, cleanup, disconnect drain, empty-reply bounds, empty-delta guards.
- test/helpers/meetingStreamingProviders.test.js: registry and allow-list pins.

No UI or locale changes: the Tinfoil entry already renders in the meeting picker from the model registry with translations in all 10 locales, which is exactly why the silent fallback was reachable.

Unit tests cannot prove, and a live meeting smoke test should cover: repeated commit cycles on one socket over a long meeting (dictation commits once per socket, so multi-utterance longevity is unproven), two concurrent attested sockets on one API key, stop behavior after a silent tail (worst case a 6s delay from the disconnect commit), word quality around the 600ms boundary, and language auto-detection for non-English meetings, since the compat mode has no language field.

Known adjacent issues, deliberately out of scope: Self-Hosted and Custom meeting selections still fall through to the OpenAI catalog default, and the dormant meeting-transcription-prepare warm-reuse path does not compare the warm connections' provider with the requested one (unreachable today because nothing calls prepareTranscription, but it needs fixing before pre-warm is wired into the UI).


gabrielste1n (approved):
Reviewed end to end after updating the branch. Provider selection now follows the active Note Recording mode and fails closed, Tinfoil commits are driven by PCM activity with bounded acknowledgements, and every realtime provider handles connection loss through the same recovery path. Local lint, typecheck, renderer build, i18n validation, and the full 1,438-test suite pass; GitHub tests, lockfile lint, and CodeQL are also green.

PR #1481, opened 2026-08-05
fix(dictation-agent): normalize provider for local and self-hosted modes

### LINKED ISSUE #1434: fix(dictation-agent): normalize provider for local and self-hosted dictation agent modes
## Affected Component
`src/helpers/dictationAgentInference.js`

## Reproduction
Call `resolveDictationAgentInference` with settings where `dictationAgentMode` is `"local"` and `dictationAgentProvider` contains a leftover string like `"openai"`:
```js
resolveDictationAgentInference({
  useDictationAgent: true,
  dictationAgentMode: "local",
  dictationAgentProvider: "openai",
  dictationAgentModel: "qwen2.5-coder",
})
```

## Actual Behavior
`resolveDictationAgentInference` returns `config.provider: "openai"`. When passed to `ReasoningService.processText`, this causes local dictation agent requests to be misrouted to cloud API providers with local model IDs (or fail due to invalid provider config).
Similarly, when `dictationAgentMode` is `"self-hosted"`, `config.provider` retains `dictationAgentProvider` instead of resolving to `undefined`.

## Expected Behavior
- When `dictationAgentMode` is `"local"`, `config.provider` should resolve to `"local"` (matching `buildNoteFormattingOverrides` in `noteFormattingOverrides.js`).
- When `dictationAgentMode` is `"self-hosted"` with a valid remote URL, `config.provider` should resolve to `undefined` so request routing is handled exclusively via `lanUrl`.

## Root Cause
`resolveDictationAgentInference` checked `isCloudAgent` and `isSelfHosted`, but did not check `dictationAgentMode === "local"` or clear `provider` for self-hosted mode.

I intend to submit a pull request fixing this issue.

xAlcahest:
## Summary
`resolveDictationAgentInference` trusted `dictationAgentProvider` even when `dictationAgentMode` was `local` or `self-hosted`, so a provider string left behind by a previous configuration (e.g. `openai`) made `ReasoningService.processText` route a local-mode agent request to that cloud provider with a local model id. The mode now wins over the stored provider: local resolves to the `local` llama.cpp provider and self-hosted clears the provider so routing goes only through `lanUrl`, matching `buildNoteFormattingOverrides` and the same stale-state family as #1381 fixed for translation.

Cloud, providers and enterprise modes resolve exactly as before. Out of scope but same family: the translation route still passes a stale non-empty provider through in local mode (#1381 covers the empty-provider case), and the Prompt Studio test tab info panel still displays the stored provider for a local-mode agent even though the run itself now routes locally.

Fixes #1434

## Changes
- src/helpers/dictationAgentInference.js: resolve the provider from the mode when it is local or self-hosted instead of trusting the stored provider string
- test/helpers/dictationAgentInference.test.js: reproduction from the issue plus the mode matrix (stale and empty provider in local mode, self-hosted with and without remote url, providers and cloud modes unchanged)

### DIFF
--- a/src/components/ui/PromptStudio.tsx
+++ b/src/components/ui/PromptStudio.tsx
@@ -81,6 +81,7 @@ export default function PromptStudio({ className = "", kind = "cleanup" }: Promp
 
   const isCloudDictationAgent = useSettingsStore(selectIsCloudDictationAgentMode);
   const useDictationAgent = useSettingsStore((s) => s.useDictationAgent);
+  const dictationAgentMode = useSettingsStore((s) => s.dictationAgentMode);
   const dictationAgentProvider = useSettingsStore((s) => s.dictationAgentProvider);
   const dictationAgentModel = useSettingsStore((s) => s.dictationAgentModel);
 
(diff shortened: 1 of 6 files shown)


gabrielste1n (approved):
Traced live dictation and Prompt Studio end to end across managed, local, self-hosted, provider, and enterprise modes. Routing now fails closed on stale or incomplete configuration, shares one resolver across both entry points, and is covered by regression tests. Local lint, typecheck, renderer build, i18n validation, and the full test suite pass; all GitHub checks are green.

PR #1486, opened 2026-08-06
fix(calendar): re-arm next meeting timer when resetting provider reminder state

### LINKED ISSUE #1485: fix(calendar): re-arm next meeting timer when resetting provider reminder state
### Affected Component
`CalendarReminderScheduler` (`src/helpers/calendarReminderScheduler.js`)

### Problem
When `CalendarReminderScheduler.prototype.reset(provider)` is called (for example when Google or Apple Calendar completes a background sync), it clears any existing `nextMeetingTimer` to clean up timer references. However, unless `activeMeeting.provider` happened to match the provider being reset, it never calls `this.scheduleNextMeeting()` to reschedule the timer for remaining upcoming events.

As a result, if a user has upcoming events and a calendar background sync finishes (or settings are reloaded for another provider), `nextMeetingTimer` is cleared and nullified, silently disabling upcoming meeting reminder notifications until another calendar mutation occurs.

### Expected Behavior
Resetting state for one calendar provider should clear stale timer state and immediately re-arm `nextMeetingTimer` so upcoming reminders for all configured providers are scheduled reliably.

### Root Cause
`reset(provider)` cleared `this.nextMeetingTimer` on line 150 of `src/helpers/calendarReminderScheduler.js`, but failed to call `this.scheduleNextMeeting()` before returning.

### Proposed Scope
Call `this.scheduleNextMeeting()` at the end of `reset(provider)` and add regression tests. I intend to open a PR for this.

hsusul:
### Problem
When `CalendarReminderScheduler.prototype.reset(provider)` is invoked (such as when Google or Apple Calendar completes a sync), it clears `nextMeetingTimer` to clean up timer references. However, unless `this.activeMeeting?.provider` matched the specific provider being reset, `reset(provider)` returned without calling `this.scheduleNextMeeting()`.

As a result, if upcoming calendar events existed and a background sync or settings refresh completed for another calendar provider, `nextMeetingTimer` was cleared and set to `null`, silently stopping future meeting reminder notifications until another calendar mutation occurred.

### Root Cause
`CalendarReminderScheduler.prototype.reset(provider)` cleared `this.nextMeetingTimer` on line 150 of `src/helpers/calendarReminderScheduler.js`, but failed to call `this.scheduleNextMeeting()` before returning.

### Solution
Added `this.scheduleNextMeeting()` at the end of `reset(provider)` so that upcoming events across all calendar providers are properly re-evaluated and scheduled whenever state for a single provider is reset.

### Changes Made
- `src/helpers/calendarReminderScheduler.js`: Call `this.scheduleNextMeeting()` at the end of `reset(provider)`.
- `test/helpers/calendarReminderScheduler.test.js`: Added regression test verifying `nextMeetingTimer` is re-armed after `reset(provider)`.

### Validation
- `ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/calendarReminderScheduler.test.js` (4/4 passed)
- `npm run typecheck` (Passed)
- `npm run lint` (Passed)
- `npm run format:check` (Passed)
- `npm run i18n:check` (Passed)
- `git diff --check` (Passed)

Fixes #1485

### DIFF
--- a/src/helpers/calendarReminderScheduler.js
+++ b/src/helpers/calendarReminderScheduler.js
@@ -161,6 +161,8 @@ class CalendarReminderScheduler {
     for (const key of this.notifiedMeetings) {
       if (key.startsWith(prefix)) this.notifiedMeetings.delete(key);
     }
+
+    this.scheduleNextMeeting();
   }
 }
 
--- a/test/helpers/calendarReminderScheduler.test.js
+++ b/test/helpers/calendarReminderScheduler.test.js
@@ -86,3 +86,30 @@ test("reconciling a provider clears an active event removed by a snapshot", () =
   assert.equal(promptCount, 1, "a periodic snapshot must not re-fire an old reminder");
   scheduler.stop();
 });
+
+test("resetting a provider re-arms the next meeting timer for upcoming events", () => {
+  const futureEvent = {
+    id: "meeting-future",
+    provider: "google",
+    summary: "Future meeting",
+    start_time: new Date(Date.now() + 600_000).toISOString(),
+    end_time: new Date(Date.now() + 1200_000).toISOString(),
+  };
+  const databaseManager = {
+    getUpcomingEvents: () => [futureEvent],
+    getActiveEvents: () => [],
+  };
+  const scheduler = new CalendarReminderScheduler(databaseManager);
+
+  scheduler.scheduleNextMeeting();
+  assert.notEqual(scheduler.nextMeetingTimer, null);
+
+  scheduler.reset("apple");
+  assert.notEqual(
+    scheduler.nextMeetingTimer,
+    null,
+    "reset(provider) must re-arm the next meeting timer"
+  );
+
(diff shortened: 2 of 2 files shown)


gabrielste1n (approved):
Verified against `main`: typecheck/lint clean, suite 1706 pass / 0 fail.

Defense-in-depth, not a live bug: all seven `reset()` call sites already invoke `scheduleNextMeeting()` on the next line, and the background sync path never calls `reset` at all.

The thing worth checking here was double-invocation, since every caller now calls it twice. It's safe: `scheduleNextMeeting()` clears `nextMeetingTimer` before re-arming, and for the synchronous past-due branch (`delay <= 0` → `onMeetingStart`) the event is added to `notifiedMeetings` before the reminder fires, so the second pass filters it out at the `!notifiedMeetings.has(...)` check. No duplicate prompts.

Taking it so `reset()` is self-consistent rather than depending on every caller remembering the next line.

PR #1493, opened 2026-08-07
fix(audio): eliminate first-words loss from cold mic opens (#845)

### LINKED ISSUE #845: The first 1 to 2 seconds of audio are not recorded and therefore are not transcribed.

Chadpiha:
## Summary

Every dictation used to open the microphone with a fresh `getUserMedia()`, and
`MediaRecorder` only existed after that resolved — so anything spoken during a cold
driver open (hundreds of ms on Windows WASAPI, up to 10–15 s worst case) was never
captured. The existing `warmupMicDriver()` couldn't help: it was one-shot per session
and fired un-awaited, and instrumented field testing on #1284/#1430 showed the raced
warm-up **never** resolved before the recording's own open — it only added a
concurrent double open.

This MR combines both community PRs, corrects the gaps their review surfaced, and
extends them into the complete fix. Four layers:

1. **Prepared capture** (from #1430): every push-to-talk key-down — and now the
   toggle press too — opens the mic immediately; both start paths join that same
   single open via `PreparedMicCapture.take()` and reuse the stream. Quick taps, the
   cancel hotkey, device changes, and cleanup cancel it; a 10 s max-age expiry
   guarantees a prepare the renderer declines can never hold the mic open silently.
2. **Warmth as a TTL** (from #1284, corrected): the one-shot `micDriverWarmedUp`
   latch is gone everywhere, including the streaming warm-up site. Warmth is stamped
   from streams that actually existed (prepare, acquisition, every release site) —
   stamping at stop replaces the post-transcription re-warm and its extra device
   open, `isProcessing` no-op, and indicator blink. The warm-up acquisition deadline
   is decoupled from the TTL (15 s vs 5 s), and a late-resolving stream still stamps
   warmth so the slowest machines finally register it.
3. **Batch pre-roll** (new): the prepared stream starts a buffering `MediaRecorder`
   the moment it resolves; `createBatchRecorder` adopts the running recorder and its
   chunks with zero gap, so every frame the hardware delivers after key-down is in
   the transcript. Cancels discard the chunks without them leaving the renderer.
4. **Opt-in idle-hold** (new — what actually eliminates the after-idle clip):
   `MicStreamHold` keeps a master stream open after a dictation and hands each
   recording a `track.clone()` (~0.1 ms re-acquire, matching the persistent-stream
   patch proven in the field on #845), bounded by a new `micWarmHoldSeconds` setting
   (Off / 10 s / 1 min / 15 min, default **Off** — a hold keeps the OS mic indicator
   lit, so the privacy-first default stays conservative). Recordings own clones, so
   no teardown path changed; a release timer firing mid-recording defers; disabling
   the setting or switching devices releases the master immediately. A new
   `mic-warm-hold-changed` IPC gates the meeting detector's audio evidence during a
   hold — macOS/Linux mic signals are device-global and would otherwise report our
   own held stream as a meeting.

Also fixes a latent push-to-talk bug found while porting: a hotkey **release landing
while `startRecording` was still awaiting the mic open was silently dropped**
(`isRecording` was still false), leaving a runaway recording until the next press.
The deferred-stop now lands it the moment the start resolves.

Fixes #845
Supersedes (with credit) #1430 by @healthwiredgames and #1284 by @xAlcahest — the
instrumented Windows testing by @xlurie on both PRs directly shaped the corrections
in layers 1–2.

## Changes

**New helpers + tests (26 new unit tests):**

- `src/helpers/preparedMicCapture.js` / `test/helpers/preparedMicCapture.test.js`:
  in-flight-dedup'd, generation-cancelled, max-age-expiring prepared capture
- `src/helpers/micWarmState.js` / `test/helpers/micWarmState.test.js`: warm-state
  TTL (5 s) and the decoupled warm-up acquisition deadline (15 s)
- `src/helpers/micStreamHold.js` / `test/helpers/micStreamHold.test.js`: idle-hold
  master stream with clone re-acquire, busy-deferred release, and hold-change events

**Capture pipeline:**

- `src/helpers/audioManager.js`: `prepareMicCapture()`/`take()` in both start paths;
  `_acquireCaptureStream()` choke point (hold clone → fresh open); pre-roll recorder
  start + adoption; warmth stamping at every acquisition/release site; TTL-gated
  `_warmMicDriverIfCold()` replacing the streaming-site latch; prepared/stream leak
  fixes on both start paths' error exits; `Recording start timing` log
  (`usedPreparedCapture`, `micReadyMs`) for field verification
- `src/hooks/useAudioRecording.js`: prepare/cancel IPC listeners; prepare on
  toggle/start handlers; deferred stop during in-flight start; settings subscription
  pushing hold duration and device changes into the audio manager

**Main-process wiring:**

- `main.js`: `sendPrepareDictation()` at Globe/right-modifier/mouse-button key-down;
  `sendCancelDictationPreparation()` on release-below-threshold and interruption
  (also hides the panel on a quick Globe tap, which previously left it visible)
- `src/helpers/windowManager.js`: the two senders; prepare/cancel on macOS compound
  push, Windows native push, the toggle press (about-to-start guess), and the cancel
  hotkey path
- `preload.js`, `src/types/electron.ts`: `onPrepareDictation`,
  `onCancelDictationPreparation`, `micWarmHoldChanged`

**Meeting detection:**

- `src/helpers/audioActivityDetector.js`, `src/helpers/meetingDetectionEngine.js`,
  `src/helpers/ipcHandlers.js`: `setMicWarmHold()` gate — audio evidence during a
  hold is dropped outright (never queued), sustained-detection state resets on both
  transitions; process- and calendar-sourced prompts unaffected

**Setting + UI + i18n:**

- `src/stores/settingsStore.ts`: `micWarmHoldSeconds` (sanitized to 0/10/60/900,
  cross-window synced), `MIC_WARM_HOLD_CHOICES`
- `src/hooks/useSettings.ts`, `src/components/SettingsPage.tsx`,
  `src/components/ui/MicrophoneSettings.tsx`: warm-hold select with privacy note
- `src/locales/*/translation.json` (all 10): `microphoneSettings.warmHold.*`

## Verification

- `npm test`: 1226 tests, 0 failures (was 1200 before; +26 new)
- `npm run typecheck`, `npm run lint`, `npm run i18n:check`, prettier: clean
- `micDriverWarmedUp` is grep-clean repo-wide
- Unit-level only — see limitations for the manual pass this still needs

## Known limitations / follow-ups

- **Manual real-mic verification still needed** (unit tests can't cover it): pre-roll
  words actually appearing in transcripts, indicator behavior during holds, and
  Windows timings — the `Recording start timing` log line discriminates prepared vs
  fresh opens per dictation, so field logs from #845 reporters would confirm.
- **Streaming-commit (online Parakeet) gets no pre-roll splice** — its transcript
  comes from the PCM worklet, which still attaches at recording start. It does get
  the single open, head start, and warmth. Worklet pre-roll is #742 follow-up.
- **The full after-idle fix is opt-in**: with the hold Off (default), a first
  dictation after long idle on a slow driver still loses driver-init time minus the
  head start. Flipping the default is a one-line change in `settingsStore.ts` if we
  decide the indicator trade-off is acceptable.
- **Pre-existing, not addressed here**: `windows-mic-listener.exe` excludes the main
  process PID, but capture runs in Chromium's audio-service process — our own
  capture likely still emits `MIC_START` on Windows and is only masked by gating.
  Tracked separately.


Chadpiha:
@xlurie Since your instrumented runs on [#1284](https://github.com/OpenWhispr/openwhispr/pull/1284) settled the race question, it would be really appreciated if you could give this a test! If you're still willing, a Windows datapoint before merge would be the strongest signal we could get.

What would help most, using your same idle-gap protocol (first-of-session, <5s, ~30s, ~10min, fixed phrases, first-word check):

1. Stock settings — the branch logs a Recording start timing line per dictation (micReadyMs, totalMs, usedPreparedCapture, preparedAgeMs, category audio; enable with --log-level=debug). usedPreparedCapture: true on hotkey dictations and the first-word comparison against your baseline numbers are the interesting bits.
2. Settings → Microphone → "Keep Microphone Warm" = 15 minutes — this should reproduce your persistent-stream patch in stock form: micReadyMs near zero on every re-dictation inside the window. If it doesn't match your 0.1–0.2 ms re-acquire behavior, I want to know.
3. Anything odd around the hold: the OS mic indicator should stay lit for the window and go out on expiry, with no meeting-detection prompts while it's held.

No time pressure of course, and partial data is still useful. Thanks again for the testing on both PRs — it directly shaped this branch.

gabrielste1n (approved):
Approving after the follow-up commit `ac49e4a3`, which closes the six findings from review, plus the merge of `main` (`c205d67c`).

**Fixed in `ac49e4a3`:**

- **Pre-roll no longer discarded as silence.** The speech-gate analyser only attaches at recording start, so a short utterance spoken entirely in pre-roll read as silence and was dropped before transcription. `createBatchRecorder` now fails the gate open (via its existing `unavailable` path) when pre-roll chunks were actually adopted.
- **Warm-hold setting now owned by `AudioManager`.** Every window has its own instance — including the agent overlay — but only `useAudioRecording` subscribed, so turning the setting Off left the agent's master held for up to 15 minutes. Moved the subscription into the constructor next to the existing `api-key-changed`/`devicechange` listeners; `setMicWarmHoldSeconds`/`dropMicWarmHold` are now dead and removed.
- **Toggle during startup honored.** `isRecording` stays false while a start awaits the mic open, so a toggle-off took the start branch and was swallowed. It now routes to the deferred stop under the start lock, and the deferred flag resets in `finally` so early returns can't leave it stale.
- **Orphan prepared captures.** `prepareMicCapture` is guarded on a new `_startInProgress` flag, and adopted pre-roll is bounded by `PRE_ROLL_MAX_AGE_MS` (2s) — a stale capture keeps its stream but can no longer prepend the previous dictation's audio or inflate `durationSeconds`.
- **Prepare window gated for meeting detection.** The gate now reports `micStreamHold.active || preparedMicCapture.active`. Note the subtlety: `cancel()` nulls `pending` so the next prepare can start, but the abandoned `getUserMedia` still owns the device — so `active` counts in-flight acquisitions rather than checking `pending`. A quick tap over a cold open now stays gated for the full time the mic is actually open.
- **Per-`webContents` refcount for the gate**, released on `destroyed`/`did-finish-load` with listeners detached so repeated hold cycles don't accumulate. This subsumes the `windowManager` `did-finish-load` reset (removed) — that only covered `mainWindow` and would have stuck the gate at `true` if the agent window reloaded mid-hold.

Also folded the duplicated detach-stop-clear logic into `discardPreRoll` (three call sites) and made `_warmMicDriverIfCold` skip its open when a hold master is already live.

**From the `main` merge (`c205d67c`):** one add/add conflict in `audioManager.js` (both sides added a module-level helper at the same point — kept both). main's new workspace-policy layer gates `startRecording`/`startStreamingRecording`, but this PR opens the device earlier in `prepareMicCapture` — that now checks `isRecordingAllowedByPolicy()` too, so a policy-blocked key-down can't light the mic or buffer pre-roll.

**Verification:** `npm test` 1397 tests / 0 failures (+3 new covering `discardPreRoll`, the cancel-with-in-flight-acquisition gate case, and `take` deactivation); `lint`, `typecheck`, `i18n:check` and `prettier --check` clean on all changed files.

One naming note, non-blocking: `mic-warm-hold-changed` / `setMicWarmHold` now mean "a renderer has the mic open outside a recording" rather than strictly a warm hold. I widened the comments instead of renaming to keep the diff reviewable — worth a rename later if you'd prefer the names match.

Still unaddressed and worth tracking separately (unchanged from the original review): `micRecovery.acquire` bypasses `_acquireCaptureStream`, so a recovery open neither stamps warmth nor integrates with the hold.

xlurie:
@Chadpiha Here's the Windows datapoint — post-merge, on `main` at the merge commit
(`28cb7f8`), same rig and idle-gap protocol as the #1284 run.

**Rig caveat up front (same as last time):** my mic is a Voicemeeter virtual device —
fresh `getUserMedia` resolves in ~18–53 ms here, so the audible #845 clip can't
reproduce physically. To still probe the loss window I added two sensitivity boosters:
**short low-redundancy first words** (long test tokens give Whisper enough acoustic+LM
material to reconstruct a clipped onset, masking exactly the loss being measured) and
**speech onset co-timed with the key press** instead of after the overlay. Plus a
fault-injection run at the end to simulate a genuinely slow driver.

Setup: Windows 11, dev build from source, `OPENWHISPR_LOG_LEVEL=debug` (env-var form),
isolated `development` profile, batch path via a self-hosted OpenAI-compatible STT on
loopback, **toggle (tap) hotkey mode**, mic pinned to an explicitly selected device
(`preferBuiltInMic=false`).

## 1. Stock settings (warm hold Off)

26 dictations across the session (long-word idle-gap series + short co-timed series).
Every one logged `usedPreparedCapture: true`.

| condition | micReadyMs | totalMs |
|---|---|---|
| first dictation of the session | 53 | 59 |
| warm, gaps <5 s | 18–22 | 22–27 |
| idle ~30 s and ~9.5 min | 18–22 | 22–27 |

- **Idle depth doesn't matter on this rig:** 21 ms after the ~9.5-minute gap vs 20 ms
  warm. No first-word loss anywhere, including the short co-timed probes ("Top one
  two…", "Cat three four…", "Bee five six…"). Isolated short words show only ordinary
  ASR variance — "top"→"stop" (a phoneme *added* at the onset, which clipping can't do)
  and "bee"→"b" (same sound, different spelling). Nothing that looks like a missing
  onset.
- On baselines: I'm not restating the 35–72 ms cold figures from the #1284 run as
  current — fresh opens today measured 18–53 ms on the same machine, so absolute
  numbers aren't directly comparable; first-word integrity is, and there's no loss in
  either run.
- One observation on the toggle path: stock `micReadyMs` (~20 ms) matches the full
  fresh-open cost on this rig, and in the fault injection below `micReadyMs` absorbs
  essentially the entire injected 2 s — so, as I read the flow (start awaits the
  prepared open; correct me if that's off), the key-down head start on a tap press is
  only a few ms, and on a slow driver the stock toggle path would still pay nearly the
  full open (see §4).

## 2. Keep Microphone Warm = 15 minutes

| condition | micReadyMs | totalMs |
|---|---|---|
| first dictation after enabling — establishes the master (held) stream | 20 | 25 |
| inside the window, gaps <1 min (×5, incl. co-timed short probes) | **0** | 5 |
| inside the window after a **~12-minute** idle (1 run) | **0** | 5 |

**Direct answer to your question: nothing here is slower than the out-of-tree
persistent-stream patch I run on my own install, as far as the log resolves.**
`micReadyMs` is integer ms and logs `0` on every re-dictation inside the window —
indistinguishable from the 0.1–0.2 ms clone re-acquire my patch measures directly.
`totalMs` drops 25 → 5, sustained deep into the window. All transcript onsets intact,
including "Toe seven eight…" spoken exactly at the key press.

## 3. Hold behavior

- Hold = 15 min: the OS mic indicator stayed lit for the whole held window, including
  the ~12-minute idle from §2.
- Hold = 10 s (chosen so expiry is observable): the indicator stays lit ~10 s after the
  dictation ends and goes out on its own, on schedule. Consistent in the log: with the
  10 s hold, a dictation 28 s after the previous one pays a fresh ~20 ms open — the
  master had expired.
- Setting switched to Off during an active hold: indicator goes out immediately.
- Device switched in Settings during a hold: master released immediately.
- No meeting-detection prompts during any held period (this is Windows, so probably not
  exercising the device-global case the gate targets).

## 4. Bonus: simulated slow driver (fault injection)

Since nothing on this rig opens slowly, I locally patched the dev build to delay every
fresh `getUserMedia` resolution by 2 s (module-level wrapper; hold clones never call
`getUserMedia`, so the injected delay can't reach them — the same code path a real slow
driver would take, though no claim about real driver timing). Probe phrase: counting
"one two three … ten", starting to speak exactly at the key press.

| condition | micReadyMs | transcript |
|---|---|---|
| stock (hold Off), 5 runs | 2016–2042 | four start at "3" (or "4") — the first ~2 s of speech is gone; the fifth produced no transcript at all |
| hold = 15 min, master established (1 run) | **0** | full "1, 2, 3, … 10" |

Scope note: this validates the pipeline logic under a slow open, not real WASAPI/driver
behavior — no actual hardware here opens slowly.

Two inputs for the "known limitations" paragraph, with different weight. Measured: on a
toggle press the key-down head start is only a few ms, so with the hold Off a slow
driver still pays nearly the whole open. Synthetic: in the simulated 2-second scenario
the stock path lost the first two to three spoken words (in one run, everything), and
the hold eliminated the loss for every dictation once the master existed — the
establishing dictation itself still paid the full open and lost its onset, as expected.
For the flip-the-default question, note this is all **toggle (tap) mode**, the least
favorable mode for a slow open — on push-to-talk the key-hold before speech gives back
most of the head start; I didn't test that path.

What this session directly verifies is prepared capture (`usedPreparedCapture: true` on
every dictation) and the hold — both behaved exactly as described here. Pre-roll's
contribution isn't separable on a ~20 ms-open rig (and the injected open yields no
pre-open frames to adopt, by construction), and I didn't isolate the warm-state TTL —
so the real-mic pre-roll item in your known-limitations list still stands open from my
side. Logs available if useful; heads-up that the tail has a few extra sim runs from
post-test chatter, so raw counts run slightly past the table. Nice to see both
community PRs land combined.

PR #1496, opened 2026-08-07
fix(upload): survive poisoned TLS connections; cap upload diarization speaker clusters

greatcoat:
## Summary

A 73-minute Voice Memo crossed a TLS-intercepting network path. One poisoned HTTP/2 connection failed 15 of 19 upload chunks, yet the app returned a seamless-looking 16-minute transcript; the upload diarization path also reported 46 speakers for a 2-person recording.

### Chunked cloud uploads

- Fatal `net::ERR_SSL_*`, `ERR_HTTP2_*`, and `ERR_QUIC_*` failures force a fresh chunk connection pool.
- Pool teardown happens while the upload slot is still held, and collateral sibling failures receive a bounded retry refund.
- Chunk uploads and inline uploads now use separate Electron sessions, so chunk-pool recovery cannot abort an unrelated inline request.
- `NO_SPEECH_DETECTED` chunks are valid silent outcomes: they do not count as lost audio and do not render missing-audio markers.
- Failed spans render explicit `[missing audio 12:00-16:00]` markers; the final marker uses the real input duration instead of assuming a full segment.
- Partial results carry `failedChunks` and `totalChunks`; more than 50% actual chunk loss throws `CHUNK_LOSS_EXCEEDED`.

### Upload diarization

- The upload path applies `capSpeakerClusters()` using the requested speaker count or `MAX_SPEAKER_COUNT` as a backstop.
- The automatic cluster threshold ramps from 0.55 to 0.8 between 15 and 60 minutes; explicit finite thresholds, including `0`, are clamped and honored.
- Clusters with less than one second of total speech are removed before transcript merging.
- The expected-speaker input has a visible label and guidance in all supported locales.

### Integration and cleanup

- Merged current `main` and preserved its policy-header enforcement on every OpenWhispr multipart request.
- Centralized chunk result classification and diarization threshold normalization in the existing policy modules.
- Removed obsolete failure-code aggregation and completed the IPC result type with chunk counts.
- Restored repository-wide Prettier compliance for two files introduced by the `main` merge.

## Testing

- `npm test`: 1,434 passed, 0 failed, 165 skipped because the local Node runtime could not load the Electron-built `better-sqlite3` binding.
- `npm run quality-check`: ESLint, Prettier, and TypeScript passed.
- `npm run i18n:check`: all locale keys and placeholders passed.
- `npm run build:renderer`: production renderer build passed.

### Scope beyond the upload path

`chunkedCloudTranscribe` is shared by three callers, so the gap markers and the
`CHUNK_LOSS_EXCEEDED` failure apply to all of them, deliberately:

| Caller | Behaviour on chunk loss |
| --- | --- |
| File upload / batch queue | Localized gap markers in the note; over 50% loss fails with a localized message and the source file can be re-uploaded |
| Dictation (`cloud-transcribe`) | Localized gap markers in the pasted text plus the existing partial-transcription toast; over 50% loss fails with a localized toast, and `saveFailedTranscription` keeps the recording so History can re-transcribe it |
| Note recording | Localized gap markers; over 50% loss fails and the stored audio stays re-transcribable |

Every new user-visible string is localized in all 10 locales:
`transcript.missingAudio`, `notes.upload.chunkLossExceeded`, and
`hooks.audioRecording.errorDescriptions.chunkLoss`.

Known pre-existing limitation, unchanged here: the History **Re-transcribe**
toast renders `result.error` raw for every error code, so a repeat chunk loss
shows the English `"12 of 19 audio segments were lost"` there. Localizing that
surface is a separate change.


gabrielste1n (approved):
Traced end to end and fixed the review findings in 1b445f21.

**Verified:** `npm test` 1,438 passed / 0 failed (165 skipped — Electron-built `better-sqlite3` binding can't load under plain Node), `npm run quality-check` (eslint ×2, prettier, tsc) clean, `npm run i18n:check` clean, `npm run build:renderer` clean.

**Fixed**
- `[missing audio …]` marker localized via `i18nMain` — it is persisted into notes and pasted by dictation, so it belonged to the UI language like `[Speaker N]`.
- `CHUNK_LOSS_EXCEEDED` localized on all three surfaces. OpenWhispr Cloud rethrows coded failures through `withSessionRefresh`, so the upload view and batch queue now resolve the code in their catch blocks too — that also fixes `NO_SPEECH_DETECTED`, which was never localized for cloud uploads because the existing check only read the returned result.
- Thrown message no longer starts with `Transcription failed:`, which the dictation toast prefixed a second time.
- Dropped the unreachable no-count branch in `CompleteView` and derived the warning from the chunk counts instead of `warning`.
- Speaker-count label/hint now use this file's design tokens (`/50` label, `/25` hint, `block` label).

**Reviewed and deliberately left as-is**
- Sharing the behaviour across dictation and note recording is correct: the markers make loss visible rather than silent, and `saveFailedTranscription` retains the audio so History can re-transcribe. Documented in the description.
- Auto-detect uploads now cap at `MAX_SPEAKER_COUNT` (8), matching the input's `max`; surplus clusters fold into the dominant speaker via the existing `capSpeakerClusters`.
- `warning` stays in the IPC result — all three consumers use it as a boolean flag with their own localized copy.

Retry handling follows the existing contract: bounded exponential backoff with jitter, teardown-collateral attempts refunded up to 3×, hard ceiling of `MAX_ATTEMPTS + 3` per chunk, and fatal codes aborting the job instead of burning sibling quota.

PR #1501, opened 2026-08-07
fix(meetings): Keep speaker identities stable and manual labels persistent

Chadpiha:
## What & why: 

During meeting recordings, two different voices were shown as the same speaker (sometimes as the recorder), and manually identifying a participant didn't stick for subsequent conversation. 

Root causes: four colliding speaker_N id namespaces that drifted apart, and a speaker cap derived once at recording start (defaulting to a single remote speaker when participants are added mid-meeting — the exact reported repro). This PR unifies the id space and makes the cap follow the roster; 

- src/helpers/liveSpeakerIdentifier.js — cluster ids now double as transcript labels: lowest-free-slot allocation with reuse after recluster merges; mapSpeaker warns instead of silently dropping unknown ids
- src/helpers/ipcHandlers.js — deleted the createSpeakerRemapper layer (source of the drift); stepper configs marked explicit; participant updates on the active meeting note refresh the live cap and broadcast meeting-session-speaker-config-updated; offline _resolveSpeakerExpectation re-reads the note at stop time; removed now-unused parseAttendees
- src/stores/meetingRecordingStore.ts — listens for the config broadcast; locked segments move to the kept cluster id on merges (name preserved); expectedCountIsExplicit start arg; state-only syncSessionExpectedCountFromParticipants for instant stepper feedback
- src/utils/participants.ts — extracted resolveParticipantExpectedSpeakerCount, resolveParticipantSpeakerCountSync, resolveInitialSpeakerCountOverride
- src/components/notes/NoteParticipants.tsx / PersonalNotesView.tsx — roster-edit sync call; pass expectedCountIsExplicit at both start-recording sites
- preload.js / src/types/electron.ts — new meeting-session-speaker-config-updated channel
- test/helpers/liveSpeakerIdentifier.test.js (new, 5 tests) / test/helpers/participants.test.js (new, 4 tests) — written red-first against the pre-fix behavior

## Follow-ups / known limitations: 

mic-bleed attribution to "You" is still heuristic (AEC/echo detection unchanged); local-model chunk timestamps lag decode completion and can miss identification windows (RC5); sub-0.8s speaker turns still share one VAD segment (RC4); Windows loopback live-ID enablement is queued separately pending platform verification. Verified: 1091 tests pass / 0 fail, typecheck and Prettier clean.


gabrielste1n (approved):

The core deletion is the right call. `createSpeakerRemapper` rewrote cluster ids into public labels before they reached the renderer, so when the user later labeled a speaker, `mapSpeaker(speakerId, …)` and `getSpeakerEmbedding(speakerId)` were handed ids the identifier had never minted and silently no-op'd. Removing the layer is a real fix, not a workaround. `parseAttendees` and the remapper are gone with no dangling references. The `_assignOrForceCluster` change — an at-cap overflow labels but no longer pollutes the centroid or steals the host cluster's profile — is a careful correctness improvement, and `_performRecluster` refusing to merge across profiles or user-set names is exactly right.

Three review fixes pushed:

1. **`_refreshMeetingSpeakerConfigFromNote` is now raise-only.** It used the defaulting resolver, so a roster that shrank (or emptied) mid-recording lowered the cap. Below the clusters already discovered that makes `_assignOrForceCluster` fold every later voice onto an existing speaker — reproducing the exact identity collapse this PR exists to fix. Switched to `_noteExpectedSpeakerCountOrNull` and gated on `> config.expectedCount`. Mirrored in `resolveParticipantSpeakerCountSync` so the renderer can't drift below the cap main enforces.
2. **`_resolveSpeakerExpectation`'s fallback returned `cap` in the wrong unit.** Every other branch returns `total - 1` because only system audio reaches the diarizer (verified: `meetingDiarizationStream.write` runs solely inside the `source === "system"` block). Narrowing the first branch from `expectedCount` to `explicit` made that fallback the common path, so a 1-on-1 with no roster could split one remote voice across two labels.
3. **`mapSpeaker` warned on ordinary offline paths.** Labeling a speaker in any past note routes there with no live session; `warn` reaches the persisted production log and said "dropped" when nothing was. Now gated on `this.running`, with the existing test split into running/not-running cases.

Added coverage for the shrinking-roster guard and for the cap unit in all four branches.

**Follow-up worth tracking (not blocking):** with the remapper gone, identifier ids are monotonic and never reused (correct — durable mappings reference them), but both label renderers parse the numeric suffix, so after a recluster merge the live transcript can read "Speaker 1, Speaker 2, Speaker 4". Confined to the live view; `diarization.js` re-densifies on export. The fix is a display-order map in the renderer, not id reuse.

Typecheck, lint, prettier clean; all 22 speaker/participant tests pass; full suite matches the clean-main baseline exactly.

PR #1502, opened 2026-08-07
feat(meeting): enable live speaker identification for Windows loopbac…

Chadpiha:
## Description

Live speaker identification was gated on systemAudioMode === "native", which only macOS reports (Core Audio tap). Windows delivers meeting system audio as mode "loopback" — via the WASAPI process-loopback helper, or Chromium display-media capture as fallback — so the feature was silently disabled there, even though both Windows paths feed the exact same 24 kHz mono s16le PCM into sendMeetingAudio(chunk, "system") → liveSpeakerIdentifier.feedAudio() as the macOS tap.

This replaces the inline mode check with a small, unit-tested predicate. It deliberately takes the platform as a second input: Linux PipeWire capture also reports mode "loopback" (there is no separate "portal" mode), so a mode-only check would have enabled Linux as a side effect. Linux stays disabled until that path is verified.

Verified by code trace: the WASAPI helper is spawned with --sample-rate 24000; the renderer fallback captures through an AudioContext({ sampleRate: 24000 }) worklet emitting mono s16le — so the identifier keeps working across the mid-session wasapi-loopback → renderer-loopback downgrade. Windows builds bundle the required models (prebuild:win downloads silero_vad.onnx + the CAM++ embedding model), and liveSpeakerIdentifier.isAvailable() still guards at runtime if they're missing.

Follow-ups / known limitations
Not yet validated on real Windows hardware — the format/path verification is a careful code trace (as scoped for this change). A smoke test on a Windows machine before release is recommended: start a meeting recording, confirm meeting-speaker-identified events arrive, and check debug logs for identifier start.
Linux (PipeWire loopback) intentionally left disabled — the helper claims the same 24 kHz protocol, but needs separate verification before flipping the predicate.
Split out of the broader meeting speaker-identity work (fix/meeting-speaker-identity)

### DIFF
--- a/src/helpers/ipcHandlers.js
+++ b/src/helpers/ipcHandlers.js
@@ -6180,58 +6180,71 @@ class IPCHandlers {
       }
 
       meetingLiveSpeakerState = null;
-      meetingLiveSpeakerStartedAt = Date.now();
-      const started = await liveSpeakerIdentifier.start(
-        (identification) => {
-          if (!win || win.isDestroyed()) {
-            return;
-          }
-
-          bindOneOnOneAttendeeToSpeaker(identification.speakerId);
-
-          const displayName = meetingOneOnOneAttendee
-            ? meetingOneOnOneAttendee.displayName
-            : identification.displayName;
-
-          const startTime = Math.max(
-            meetingLiveSpeakerStartedAt || 0,
-            (meetingLiveSpeakerStartedAt || 0) + identification.startTime * 1000
-          );
-          const endTime = Math.max(
-            startTime,
-            (meetingLiveSpeakerStartedAt || 0) + identification.endTime * 1000
-          );
-          const enrichedIdentification = {
-            ...identification,
-            displayName,
-            startTime,
-            endTime,
-          };
-
-          win.webContents.send("meeting-speaker-identified", enrichedIdentification);
-
-          for (const seg of meetingDiarizationSegments) {
-            if (
-              seg.source === "system" &&
-              seg.timestamp != null &&
-              seg.timestamp >= startTime &&
-              seg.timestamp <= endTime &&
-              (!seg.speaker || seg.speakerIsPlaceholder)
-            ) {
-              applyConfirmedSpeaker(seg, {
-                speaker: identification.speakerId,
-                speakerName: displayName || seg.speakerName,
-                speakerIsPlaceholder: false,
-              });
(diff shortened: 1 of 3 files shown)


gabrielste1n (approved):

Well-shaped extraction. The `platform` parameter is load-bearing rather than speculative — `getLinuxSystemAudioAccess` also reports `"loopback"`, so a mode-only check would have silently switched on PipeWire. Format and model assumptions check out: both Windows strategies land in `sendMeetingAudio(chunk, "system")` → `feedAudio()` at 24 kHz mono, and `prebuild:win` bundles both ONNX models. The ESM-in-CJS shape looked wrong at first but matches shipping precedent (`localServerPolicy.js` is required the same way from `ipcHandlers.js`); confirmed it resolves under Node 24.

Three review fixes pushed:

1. **Timestamps are now anchored to the first system chunk, not to identifier start.** `identification.startTime` is a sample counter whose zero is the first chunk `feedAudio()` receives, but the wall-clock anchor was stamped before any audio existed. Those coincide on the macOS tap, which is already capturing. On Windows they don't — `startManagedMeetingSystemAudio` can burn the full `START_TIMEOUT_MS` (10 s) and throws into a branch that, unlike the native and pipewire branches, never resets the anchor; the renderer fallback builds its AudioContext later still. `LIVE_WINDOW_PADDING_SECONDS = 0.75` absorbs sub-second skew, not ten. Every label would have landed on the wrong transcript line for the rest of the session — worse than shipping no labels. Now stamped lazily in `sendMeetingAudio`, with the callback guarded and the resulting unreachable `|| 0` fallbacks removed. macOS behavior is unchanged.
2. **An identifier startup throw no longer aborts the recording.** `isAvailable()` only stats the model file, so a corrupt `silero_vad.onnx` or an onnxruntime binding that won't load rejected into the start catch → full rollback, where before the user recorded fine without labels. Now degrades via `.catch`, matching `startMeetingAec`.
3. **The platform-default test was vacuous** — it asserted the function against itself, so both sides collapsed to `false` on CI's Linux runner. I confirmed by mutation that deleting the default parameter left all six tests green. Replaced with a `process.platform` stub (the pattern already in `xwayland.test.js`), and re-ran the mutation to confirm it now fails.

Typecheck, lint, prettier clean; 23/23 speaker tests pass; full suite matches the clean-main baseline exactly.

PR #1506, opened 2026-08-08
fix(policy): Hide restricted options and apply safe fallbacks

Chadpiha:
## Summary
Improves desktop workspace-policy enforcement by hiding unavailable options and selecting permitted runtime fallbacks.

## Changes
Hide policy-denied transcription modes and providers.
Hide restricted LLM providers and voice-agent controls.
Hide unavailable note-sharing actions and visibility options.
Keep privacy-recovery actions available, including making notes private, revoking invitations, and removing access.
Select an allowed fallback when a saved option becomes restricted.
Preserve raw user preferences so they can be restored if policy restrictions are removed.
Apply effective policy settings consistently across dictation, meetings, uploads, cleanup, chat, and retries.
Prevent Custom and self-hosted endpoints from falling through to unintended providers.
Require HTTPS for public endpoints while continuing to allow HTTP for private/local networks.
Backward compatibility
Signed-out, unmanaged, and self-hosted users retain their existing behavior. Managed fallbacks are applied as runtime overlays and do not overwrite persisted preferences.

## Policy refresh timing
Desktop policy updates can take up to approximately five minutes to appear because successful policy responses are cached for 300 seconds.


gabrielste1n (approved):
Reviewed end to end after merging current main. Policy enforcement, UI fallbacks, runtime routing, transcription, cleanup, and regression coverage were verified; local full suite, typecheck, lint, build, and i18n checks pass.

PR #1515, opened 2026-08-09
fix(audio): stop classifying phone microphones as built-in

xpipko:
## Problem

`isBuiltInMicrophone` treats any device label containing "microphone" as built-in unless it matches a deny-list of external keywords. A phone surfaced via macOS Continuity enumerates under the phone's own name (e.g. `Galaxy Cell Microphone`), matches nothing on the list, and is classified built-in — so the app auto-selects it over the real microphone and caches the choice.

The practical failure: dictation silently captures distant phone-mic audio. There is enough energy to trip server VAD, but the streaming ASR returns empty transcripts on every dictation, while the batch fallback still scrapes out text. This cost hours of debugging aimed at the (healthy) streaming pipeline before the device label in the debug log gave it away.

## Fix

Add `cell`, `phone`, and `continuity` to the external indicators. Indicators are matched against the label with the word "microphone" stripped — a bare `phone` indicator would otherwise match the "phone" inside "microphone" itself and flip every generic-branch label (e.g. `Microphone (Realtek High Definition Audio)`) to external. The regression tests caught exactly that on the first attempt.

`phone` also matching "headphones" is intentional: a headphone mic is external too, and the harmful direction is only external-classified-as-built-in — the reverse merely skips an auto-selection fallback.

## Tests

`test/utils/audioDeviceUtils.test.js` (node --test): phone/Continuity labels are never built-in; genuine built-ins (macOS + Windows generic-branch labels) stay built-in; headphone/headset mics are external.

### DIFF
--- a/src/utils/audioDeviceUtils.ts
+++ b/src/utils/audioDeviceUtils.ts
@@ -25,7 +25,7 @@ export function isBuiltInMicrophone(label: string): boolean {
   // must flag "iPhone"/"Cell Phone"/"Headphones" without matching the "phone"
   // inside "microphone" itself.
   if (lowerLabel.includes("microphone")) {
-    const labelWithoutMicrophone = lowerLabel.replace(/microphone/g, "");
+    const labelWithoutMicrophone = lowerLabel.replace(/microphone/g, " ");
     const externalIndicators = [
       "bluetooth",
       "airpods",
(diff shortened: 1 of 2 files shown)


Chadpiha (approved):
Thanks for tracking this down — nice catch on the root cause. Stripping the word
"microphone" before matching is the right trick; it's what makes "phone" usable
as an indicator at all (previously impossible since "micro**phone**" contains it).

Verified before approving:
- All 3 tests pass (Node 24 + tsx), prettier/eslint clean, merges clean with main
- All three callers (`audioManager`, `meetingRecordingStore`, `MicrophoneSettings`)
  use the predicate consistently, so the settings banner stays in sync with what
  actually gets selected
- No existing indicator can only match inside the word "microphone", so nothing
  regresses — Windows generic labels ("Microphone (Realtek…)", "Microphone Array…")
  stay built-in, and your tests pin exactly that
- Misclassification risk runs in the safe direction: a false "external" just skips
  auto-selection and falls back to the system default, while the harmful direction
  (external classified built-in) is what this removes
- The bad selection was only ever cached in-memory, so affected users recover
  automatically on restart — no persisted state to worry about

I pushed two small tweaks to the branch:
1. Comment accuracy: the "Galaxy Cell Microphone" label can't have come via
   Continuity (Apple-only) — reworded to "phones connected to the computer
   (Continuity, Bluetooth)" in both the source and test comments
2. Strip "microphone" to `" "` instead of `""` so word removal can never splice
   adjacent text into a phantom keyword match

Known limitation (fine to ship): keyword matching only catches phones whose names
contain iphone/ipad/cell/phone/continuity — a phone renamed "Zeus" still slips
through. The `rejectedMicDeviceId` no-audio safety net partially covers that at
runtime. I'll open a follow-up issue for a stricter macOS label allowlist, which
would also fix pre-existing false built-ins like "Yeti Stereo Microphone".

PR #1518, opened 2026-08-10
feat(linux): Add launch at login via an XDG autostart entry

edwin-luu:
## Summary

"Launch at login" was hidden on Linux because Electron's `app.setLoginItemSettings()` is a no-op there. This implements it with an XDG autostart entry, so the existing Settings toggle now works on all three platforms instead of two.

No new UI strings: the toggle reuses the existing `settingsPage.general.startup.launchAtLogin` keys, so no locale files change.

| Before (v1.8.1) | After |
| --- | --- |
| Startup section without Launch at login | Startup section with Launch at login |

Settings → Preferences → Startup. The Launch at login row was hidden on Linux; it now appears above Start minimized.

## What changed

- **`src/helpers/linuxAutostart.js`** (new) writes `$XDG_CONFIG_HOME/autostart/openwhispr.desktop`.
- **`ipcHandlers.js`** routes `get-auto-start-enabled` / `set-auto-start-enabled` through it on Linux, leaving macOS and Windows on `setLoginItemSettings()`.
- **`SettingsPage.tsx`** drops the `platform !== "linux"` guard that hid the row.
- **`main.js`** calls `syncAutostartEntry()` from `initializeCoreManagers()`, next to the existing Linux startup cleanup.

## Edge cases handled

Three things make a naive `.desktop` write unreliable, so they're handled explicitly and covered by tests:

- **`process.execPath` is wrong on AppImage.** It points at the ephemeral FUSE mount, which is gone by the next boot, so `Exec` resolves from `$APPIMAGE` first.
- **The entry goes stale when the executable moves** (user renames the AppImage, or an auto-update replaces it with a new filename). The file survives, so the toggle keeps reading as enabled while the session launches nothing. `syncAutostartEntry()` re-points it at startup. It never re-enables an entry the user disabled, and no-ops in development, where `execPath` is the local Electron binary.
- **GNOME Tweaks and KDE's autostart editor disable an entry in place**, writing `X-GNOME-Autostart-enabled=false` or `Hidden=true` rather than deleting the file. `isAutostartEnabled()` parses those instead of just testing for the file, so disabling it outside the app is reflected in the toggle.

Reserved characters in `Exec` are also backslash-escaped per the Desktop Entry spec, so a home directory containing one still resolves.

## Testing

`test/helpers/linuxAutostart.test.js` adds 16 cases covering exec resolution and escaping, both in-place-disable keys, `[Desktop Entry]` group scoping, and every `syncAutostartEntry` path.

`npm run lint`, `npm run format`, `npm run typecheck`, and `npm test` all pass.

Manually verified on Ubuntu 26.04 LTS, GNOME Shell 50.1, Wayland, x86_64:

1. The Launch at login row now renders on Linux, and toggling it on creates `~/.config/autostart/openwhispr.desktop` (mode 644) with the icon installed under `hicolor/256x256/apps/`.
2. Reopening Settings reads the toggle back as on. Reinstating the old `platform !== "linux"` guard reproduces the pre-fix behavior (entry present on disk, toggle reads off), and removing it again restores the correct state.
3. Setting `X-GNOME-Autostart-enabled=false`, and separately `Hidden=true`, each make the toggle read as off. Toggling on from that state rewrites a clean, enabled entry.
4. Toggling off removes the file.
5. With `$APPIMAGE` set and `NODE_ENV` unset, a stale `Exec` is re-pointed at the current executable during startup ("Re-pointed the autostart entry at the current executable"), and `$APPIMAGE` takes precedence over `process.execPath`.
6. The same startup path leaves a user-disabled entry (`X-GNOME-Autostart-enabled=false`) completely untouched.

**Not verified**, to be explicit about the limits of the above:

- **Launch after an actual reboot.** No login cycle was performed.
- **A real packaged AppImage.** Steps 5 and 6 reproduced the packaged environment by setting `$APPIMAGE` and leaving `NODE_ENV` unset, which is what the AppImage runtime does, but no AppImage was built. So icon resolution against the packaged `resourcesPath` layout is untested; those candidate paths mirror the ones `tray.js` already uses in production, and a miss degrades gracefully by omitting the `Icon=` line.
- **Non-GNOME desktops.** `Hidden=true` and `X-GNOME-Autostart-enabled=false` were written by hand to match what KDE's editor and GNOME Tweaks produce; only GNOME was exercised directly.

## Notes

Start-minimized behavior is unchanged. `openAsHidden` is macOS-only in Electron, and the existing `startMinimized` setting is applied app-side in `main.js` on every platform, so the desktop entry deliberately carries no hidden flag.

`CLAUDE.md` is updated with the new helper and a Linux platform note, matching how other helpers are documented there.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -96,7 +96,7 @@ OpenWhispr is an Electron-based desktop dictation application that uses whisper.
   - Only active on Linux + KDE desktop (detected via `XDG_CURRENT_DESKTOP`)
   - D-Bus transport: `@homebridge/dbus-native` (pure JavaScript, no native addons)
 - **linuxAutostart.js**: Launch-at-login on Linux via an XDG autostart entry
-  - `app.setLoginItemSettings()` is a no-op on Linux, so the entry is written directly to `$XDG_CONFIG_HOME/autostart/openwhispr.desktop`
+  - `app.setLoginItemSettings()` is a no-op on Linux, so the entry is written directly to `$XDG_CONFIG_HOME/autostart/open-whispr.desktop`, matching the executable name electron-builder packages under
   - `Exec` resolves from `$APPIMAGE` first: `process.execPath` is the ephemeral AppImage FUSE mount
   - `isAutostartEnabled()` honors `X-GNOME-Autostart-enabled=false` and `Hidden=true`, which GNOME Tweaks and KDE's autostart editor write in place instead of deleting the file
   - `syncAutostartEntry()` runs from `initializeCoreManagers()` in `main.js` and re-points a stale `Exec` after the executable moves (renamed or auto-updated AppImage); it never re-enables an entry the user disabled, and no-ops in development
@@ -775,7 +775,7 @@ const { t } = useTranslation();
 - No standardized URL scheme for system settings (user must open manually)
 - Privacy settings button hidden in UI (not applicable on Linux)
 - Recommend `pavucontrol` for audio device management
-- **Launch at login**: XDG autostart entry at `~/.config/autostart/openwhispr.desktop` (see `linuxAutostart.js`), since Electron's `setLoginItemSettings()` does nothing on Linux
+- **Launch at login**: XDG autostart entry at `~/.config/autostart/open-whispr.desktop` (see `linuxAutostart.js`), since Electron's `setLoginItemSettings()` does nothing on Linux
   - Disabling it from GNOME Tweaks or KDE's autostart editor is reflected in the Settings toggle
   - "Start minimized" is handled app-side by the `startMinimized` setting, not by the desktop entry
 - **Clipboard paste tools** (at least one required for auto-paste):
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -95,6 +95,12 @@ OpenWhispr is an Electron-based desktop dictation application that uses whisper.
   - Converts Electron hotkey format to Qt key codes
   - Only active on Linux + KDE desktop (detected via `XDG_CURRENT_DESKTOP`)
   - D-Bus transport: `@homebridge/dbus-native` (pure JavaScript, no native addons)
+- **linuxAutostart.js**: Launch-at-login on Linux via an XDG autostart entry
+  - `app.setLoginItemSettings()` is a no-op on Linux, so the entry is written directly to `$XDG_CONFIG_HOME/autostart/openwhispr.desktop`
+  - `Exec` resolves from `$APPIMAGE` first: `process.execPath` is the ephemeral AppImage FUSE mount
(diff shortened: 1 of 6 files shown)


gabrielste1n (approved):
Reviewed this end to end and it's solid work — the `$APPIMAGE` handling, the in-place-disable parsing that GNOME Tweaks and KDE actually write, and `syncAutostartEntry()` are all the non-obvious details that make XDG autostart work in practice rather than just on paper. The 16 tests are real behavioural tests, not scaffolding. Thanks @dharampreet's report for surfacing the gap and thanks @edwin-luu for the fix.

I pushed one commit (ab4d054) fixing three things I found while tracing it:

**1. The icon copy could kill the whole feature.** `ensureIconInstalled()` ran `mkdirSync`/`copyFileSync` unguarded, and `writeAutostartEntry()` called it inline as an argument — so an unwritable `$XDG_DATA_HOME` threw *before* the desktop file was written. Verified: with the icon tree unwritable, `setAutostartEnabled(true)` threw `ENOTDIR`, no entry was created, and the toggle reported failure. Since `buildDesktopFileContents()` already accepted a null icon, the intent was clearly for it to be optional — it's now best-effort. Not exotic on locked-down corporate Linux images.

**2. The icon was installed into a theme directory but referenced by absolute path,** which bypasses theme lookup entirely, under a basename (`openwhispr.png`) that no package installs. Net effect on deb/rpm: a duplicate 256px copy in `~/.local/share/icons` shadowing the one the package already ships. Now referenced by theme name using the executable name electron-builder packages under (`open-whispr` — same one `app.setDesktopName()` uses), and copied only when neither the user nor a system theme dir already has it, so deb/rpm does no work and AppImage/tar.gz still get an icon. The autostart entry is renamed to match.

**3. A literal backslash in `Exec` needs four backslashes, not two.** The desktop-entry string rule unescapes the value *before* the quoting rule splits it into arguments, so `\\` collapses to a bare `\` which then escapes the following character. This is the one case the PR body's escaping claim didn't actually cover.

Also made `syncAutostartEntry()` read the file once instead of twice, which incidentally drops a null dereference if the file disappears between the two reads.

Each fix has a test that fails against the original code and passes against the fix (confirmed by reverting the helper and re-running). 18 tests now; full suite 1846 pass / 0 fail, lint, format and typecheck clean against current main.

One gap I'm deliberately leaving for a follow-up rather than expanding scope here: the `Exec` line has no `--hidden`, so launch at login still pops the control panel unless "Start minimized" is separately on. That's the same gap Windows and macOS have (#1541), so it's better fixed once across all three platforms than bolted on here. That PR will need to update the `entry.Exec` comparison in `syncAutostartEntry()` at the same time — once args are part of the written `Exec`, comparing against a bare `quoteExecPath()` never matches and the file gets rewritten on every launch.

PR #1525, opened 2026-08-10
fix(wayland): auto paste not working on non-QWERTY layouts.

### LINKED ISSUE #1523: Auto paste failing on Wayland with non-QWERTY keyboard layouts
**Describe the bug**
This is Wayland fork of the issue #1478 , similar to macOS, the auto paste fails on non-QWERTY keyboards such as **DVORAK**

**To Reproduce**

1. Open OpenWhispr on a DVORAK layout keyboard
2. Use hotkey to dictate
3. Auto paste fails/doesn't happen
4. `ctrl + .` is triggered instead of `ctrl + v`

**Expected behavior**
the transcription should be successfully pasted in the focused input.

- OS: Linux/Hyprland
- Version: 1.8.1

**Additional context**
I was able to successfully reproduce this issue on my local and have a root cause and a rough fix done, I will raise the PR after cleaning up the fix and testing on my system, however this PR will be Linux/Wayland specific only as I don't have access to a mac system currently.

IdrisGit:
### Description
- replace portal keycodes with keysyms
- prefer compositor-native symbolic shortcuts:
  - Hyprland: sendshortcut → wtype → uinput
  - wlroots: wtype → uinput
  - GNOME/KDE: portal keysyms → uinput
  - leave dictated text in the clipboard when automatic paste fails
- preserve X11, XWayland, and clipboard restoration behavior
- use layout-independent Shift+Insert for physical fallbacks

### Testing
Tested it locally with different non-qwerty layouts on hyprland compositor and it's working fine.

fixes #1523


Chadpiha (approved):
Thanks @IdrisGit — this is a really solid fix. The core approach is exactly right: portal NotifyKeyboardKeysym and compositor-level shortcuts are layout-independent where raw evdev keycodes weren't, and pinning the physical fallbacks (uinput/ydotool) to Shift+Insert closes the rest. Dropping the portal-dismissed retry loop also fixes a latent double-paste bug — exit 0 without a token just means the compositor doesn't support restore tokens, so the old retry could paste multiple times. Appreciate the test coverage too.

I pushed a follow-up commit to your branch with a few adjustments from review:

wtype now runs before sendshortcut on Hyprland. sendshortcut has a known stuck-modifier bug ([hyprwm/Hyprland#14099](https://github.com/hyprwm/Hyprland/discussions/14099), still unresolved as of 0.56) where the synthetic modifier stays latched in XKB and silently breaks the user's Ctrl/Shift shortcuts afterward. wtype is equally layout-independent and unaffected, so sendshortcut is now the zero-dependency fallback for setups without wtype (your explicit activewindow target — the reported mitigation — is kept).
Reverted the last-resort tool order on GNOME/KDE to ydotool-first. xdotool key exits 0 even when the focused window is native Wayland and never receives the keystroke, so xdotool-first could silently swallow a paste; ydotool's args are layout-safe Shift+Insert now anyway.
Made the Hyprland shortcut use the same Shift+Insert precedence as the wtype/pasteKeys paths (Konsole/Electron targets), restored the KDE/GNOME portal-ordering rationale comments, cached the blocking --capabilities probe in checkPasteTools, and ran prettier.

Since the ordering change affects which tool fires on your machine, could you re-test on your Hyprland + non-QWERTY setup — ideally both with wtype installed (it should fire first) and without (sendshortcut path)? Full suite is green on my end (1249 passing, including the reworked Hyprland ordering tests).

IdrisGit:
I have just pushed the merge conflict resolution, will test and update today

IdrisGit:
Hi @Chadpiha I have tested locally and both with and w/o `wtype` is working fine 👍.

I have pushed one commit which checks if `wtype` is installed or not shows the user a guide similar to `ydotool` and some code cleanup to the detection system.

### screenshots
#### when `wtype` not installed
image

#### when `wtype` installed
image


### note
it will need new translations in language other than english

Issue #1524, opened 2026-08-10
Persisted copies of retired default prompts permanently shadow every shipped prompt hardening (fix exists unmerged on claude/1-7-5-regression-f9igpg)


greatcoat:
## Summary

A user who ran an older release can be permanently frozen on a **retired default prompt**, so every prompt improvement shipped since is silently inert for them. Updating the app does not fix it, and there is no signal to the user or to us that it is happening.

This is the delivery mechanism behind OpenWhispr/openwhispr#833. OpenWhispr/openwhispr#833 tracks the symptom (cleanup output answers the speaker); this issue tracks why hardening the prompt does not reach the affected users.

## Mechanism

1. Older releases persisted the then-default **unified two-mode prompt** into `localStorage` as `customUnifiedPrompt`.
2. `src/stores/settingsStore.ts:349-362` migrates that legacy key into `customPrompt.cleanup` (and `customPrompt.dictationAgent`).
3. `src/config/prompts/index.ts:18` resolves `useSettingsStore.getState().customPrompts[kind]` **in preference to** the shipped default.

Net effect: a value that was never a deliberate user customisation — it was simply the default at the time — is promoted to an override that outranks every subsequent shipped prompt.

The retired prompt contains, verbatim:

```
OUTPUT RULES
...
7. For direct questions, output just the answer
```

So the prompt itself instructs the model to answer dictated questions. On affected installs this is not a model instruction-following weakness; it is the configured contract.

## Evidence from a real install

Debug log supplied by an affected user, on a release that already contained the hardened cleanup prompt from `87736b85`:

* `systemPromptLength: 3830`, and **no** `<transcript>` **framing** — i.e. the retired unified prompt, not the shipped hardened one.
* Cleanup model `Meta-Llama-3.1-8B-Instruct-Q4_K_M`. Worth stressing: an **8B**, not a small model. The failure is not a capacity problem.
* Confirms `87736b85` (shipped in v1.7.4 / v1.7.5) had no effect on this install, because the persisted override outranked it.

## Impact

* Every prompt hardening we ship is invisible to this cohort, including future ones.
* The cohort is **unmeasurable today**: `user_daily_activity.client_version` is only populated from 1.8.0 onward, so we cannot size how many installs carry a retired override.
* Failure is intermittent and content-dependent (only question-shaped dictations trigger it), so it reads to users as flakiness rather than a setting.

## A fix already exists and has never shipped

[...]

PR #1527, opened 2026-08-10
fix: correct translation hotkey hint path in all locales

### LINKED ISSUE #1413: Translation tab points users to "Settings → General", a section that does not exist

xAlcahest:
## Summary

The footer under Settings → Language Models → Translation pointed users at "Settings → General → Translation Hotkey", a section that does not exist; the hotkey lives under Hotkeys. The string is `dictationTranslation.hotkeyHint` (rendered in `DictationTranslationSettings.tsx`) and it was wrong in all 10 locales, because "general" is the internal i18n namespace of the Preferences section and got read as a user-facing location. Each locale now uses its real Hotkeys section label from `settingsModal.sections.hotkeys.label`.

Completes #1413 together with #1419: that PR fixes the English string only, and the en change here is byte-identical to it, so the two converge whichever merges first. An audit of every path-style string across the 10 locales found no other reference to a nonexistent section: the remaining hits point at OS-level System Settings screens or at Settings → Workspace, which exists, and `settingsPage.workspace.tab.general` is a real tab inside Workspace. Renaming the `settingsPage.general.*` key namespace is deliberately out of scope, it would be key churn with no user-facing change.

Fixes #1413

## Changes

- src/locales/{en,de,es,fr,it,ja,pt,ru,zh-CN,zh-TW}/translation.json: `dictationTranslation.hotkeyHint` now points at Settings → Hotkeys → Translation Hotkey, using each language's existing section and hotkey names.


gabrielste1n:
Merged as #1564 — your change, rebased onto current `main`. Thanks @xAlcahest, that's two of yours in today (this and #1524).

Moved branches only because my token can't push to a fork branch when the rebase carries `.github/workflows/` changes; the rebase itself was clean and nothing was altered.

I cross-checked every locale's replacement against its own `settingsModal.sections.hotkeys.label` rather than taking the translations on trust — all ten match. Good catch on `fr` too: switching *Réglages* → *Paramètres* lines the hint up with `settingsModal.title`, which is what the user actually sees on the window.

PR #1528, opened 2026-08-10
fix(prompts): clear persisted copies of retired default prompts

### LINKED ISSUE #1524: Persisted copies of retired default prompts permanently shadow every shipped prompt hardening (fix exists unmerged on claude/1-7-5-regression-f9igpg)

xAlcahest:
## Summary

Older releases persisted the then-current default prompt into localStorage (an unedited Prompt Studio save wrote `customUnifiedPrompt`, which `migrateCustomPrompts` in `settingsStore.ts` later copied into `customPrompt.cleanup` and `customPrompt.dictationAgent`), and `resolvePrompt` prefers that stored copy over the shipped default forever. Affected installs are frozen on a retired prompt and never receive any prompt hardening shipped since, including `87736b85`; the retired two-mode prompt even instructs the model to answer dictated questions, which is the delivery mechanism behind #833 (#1387 guards the output side, this removes the cause).

A startup sweep now clears an override only when it byte-matches, via SHA-256, one of the retired defaults in `src/config/retiredPrompts.js`, a registry extracted from the full history of every shipped prompt text (`promptData.json`, `locales/*/prompts.json`, the pre-split `prompts.ts` and `ReasoningService.ts` defaults). A prompt the user edited, even by one character, hashes differently and is never touched; swept values are archived under `customPrompt.<kind>.retired` and the sweep is idempotent across restarts. The approach adapts the sweep from unmerged commit `5fe1631c` (branch `claude/1-7-5-regression-f9igpg`), replacing its substring fingerprints, which would also have matched user-edited variants, with exact hashes; that commit's temperature and prewarm changes are left out of scope.

Fixes #1524

## Changes

- src/config/retiredPrompts.js: registry of SHA-256 hashes for every retired default prompt with provenance, a snapshot of the current defaults, and the sweep helper that clears byte-matching overrides and archives them.
- src/stores/settingsStore.ts: run the sweep at startup right after migrateCustomPrompts and clear swept kinds from the store state.
- src/components/ui/PromptStudio.tsx: saving the unedited default no longer persists it as a custom prompt, so this cohort stops re-forming.
- src/config/prompts/registry.ts: note to update the hash registry when the chat agent default changes.
- test/helpers/retiredPrompts.test.js: retired copies are swept and archived, one-character edits are untouched, a stored current default is a no-op, the sweep is idempotent, and a ratchet test fails whenever a shipped default changes without moving its old hash into the retired set.
- test/helpers/retiredPromptFixtures.json: byte-exact retired prompt texts from git history used by the tests.


gabrielste1n:
Merged as #1561 — your work, rebased onto current `main`. Thank you @xAlcahest, this is a genuinely careful piece of engineering.

It moved branches only because my token can't push to a fork branch when the rebase carries `.github/workflows/` changes; the rebase itself was clean and I changed nothing but Prettier formatting on two files.

Three things I specifically verified before taking it:

- **The ratchet isn't vacuous.** I perturbed `en/cleanupPrompt` by one character and confirmed the suite fails as designed. That test is what stops this registry rotting the first time someone edits a default.
- **Non-destructive.** Archiving to `customPrompt.<kind>.retired` instead of deleting is the right call, and the re-read before removal closes the save-during-hash race properly.
- **Hash-equality as the safety property.** A prompt edited by even one character never matches, so a false positive is by definition not a user customization. That's what makes an automatic sweep defensible here.

Two residuals I noted in #1561 rather than blocking on: the sweep is fire-and-forget at module load, so a dictation racing init could use the stale prompt once (self-corrects next launch); and it depends on `crypto.subtle`, which fails closed if absent — `file://` is a secure context in Chromium so it should be fine, but nothing else in the renderer uses WebCrypto today, so it's worth a spot-check in a packaged build.

PR #1529, opened 2026-08-10
feat(updater): add setting to disable auto-install on quit

### LINKED ISSUE #1460: Setting to disable silent install-on-quit after an update has downloaded (`autoInstallOnAppQuit`)
**Labels (suggested):** enhancement

### Is your feature request related to a problem? Please describe.

Once OpenWhispr has downloaded an available update, it installs that update automatically the next time
the app quits — with no confirmation prompt, no visible "installing update..." step, and no way to decline
or postpone at that moment. The only way to avoid it is to never let the app fully quit, which isn't
realistic for a tray/background app across reboots, sleep, or routine restarts.

For anyone running a locally-configured setup — self-hosted STT/LLM endpoints, custom prompts, specific
model settings, or just a version that's known to work well for them — this means the app can change
version out from under you without any notice, and you find out only when something that used to work
doesn't anymore.

**This happened to me concretely.** I run OpenWhispr fully self-hosted (local transcription + local
cleanup endpoints, validated against a specific app version). At some point the app silently reinstalled
itself on quit to a newer version. I didn't see any prompt, dialog, or notification for the install step
itself — the version had simply changed the next time I opened it. I didn't notice for well over a week,
because nothing in the UI flagged that a version change had happened; I only found out when I went looking
for something else and checked the version number. For that entire window I was running a version I had
never explicitly agreed to run, with no record of when or why it changed.

The download step and the notification-of-availability step are both reasonable and already covered by
existing preferences (#943 added a way to silence the "update available" notification). The install step
is the one with no user-facing control at all.

### Describe the solution you'd like

Add a setting (Settings → General, near the existing update-notification preference) that controls what
happens **after** a download completes, independent of whether update *checks*/*notifications* are
enabled. Something like three levels:

1. **Auto (current default, unchanged)** — download available updates, install silently on next quit.
   Existing behavior, nothing changes for users who don't touch the setting.
2. **Download only, ask before installing** — download in the background as today, but keep
   `autoInstallOnAppQuit` off; show a clear prompt/banner ("Update to vX.Y.Z ready — Install & Restart /
   Later") that the user has to act on, instead of installing unattended on quit.
3. **Notify only, no auto-download** — flip `autoDownload` off too, for fully manual control (useful on
   metered connections or when someone wants to review release notes before even fetching the binary).

Even shipping just option 2 (a single "Install updates automatically" checkbox, default ON) would close
the actual gap — the rest is a nice-to-have.

### Implementation sketch

The updater already exposes exactly the right seams (`src/updater.js`, `electron-updater`-based):

```js
autoUpdater.autoDownload = <preference>;          // currently hardcoded false
autoUpdater.autoInstallOnAppQuit = <preference>;   // currently hardcoded true
```

Both are read once in `setupAutoUpdater()`. Wiring a persisted preference (same store the notification
toggle already uses) in and reading it before these two assignments — plus re-reading it if settings can
change at runtime without restart — should be close to the whole change. The existing `update-downloaded`
IPC event already reaches the renderer, so a "ready to install" banner can reuse it directly when
`autoInstallOnAppQuit` is off.

### Additional context

- Prior art: **#943** asked for update control and got a notification-visibility toggle in v1.7.3 — useful,
  but it only silences the "update available" popup. It doesn't touch `autoDownload` or
  `autoInstallOnAppQuit`, so the silent-install-on-quit behavior this issue is about was untouched by that
  fix. **#763** is the same shape (notification preferences, not install behavior).
- Default staying ON means zero behavior change for the overwhelming majority of users who like
  install-and-forget — this is purely additive for people who need predictability: self-hosted/offline
  setups, enterprise change-control, anyone validating a specific version before rolling forward, and
  metered-connection users (via the `autoDownload` half).

xAlcahest:
## Summary

Once an update has downloaded, OpenWhispr installs it on the next quit with no way to decline or postpone: `setupAutoUpdater()` in `src/updater.js` hardcodes `autoUpdater.autoInstallOnAppQuit = true`, so self-hosted or version-pinned setups get silently moved to a new version. This adds an "Install updates automatically" toggle in Settings → System → Software Updates, default ON, so behavior is unchanged for anyone who never touches it.

The preference persists as `UPDATE_AUTO_INSTALL` through `environment.js` rather than the renderer settings store because the flag has to be correct in the main process at boot: a download can be triggered straight from the update notification before the settings page ever mounts and syncs anything. With the gate off, updates still download normally and the explicit Install & Restart button keeps working, since `quitAndInstall()` does not consult the flag ([BaseUpdater source](https://github.com/electron-userland/electron-builder/blob/master/packages/electron-updater/src/BaseUpdater.ts)).

Fixes #1460

## Changes

- src/helpers/updateInstallPolicy.js: gate logic; only the literal string "false" disables, anything else keeps the historical default.
- src/updater.js: read the flag from env in `setupAutoUpdater()`, plus `setAutoInstallOnAppQuit()` for runtime changes, which re-registers electron-updater's quit hook when re-enabling after a download that completed while gated.
- src/helpers/environment.js: `UPDATE_AUTO_INSTALL` in the persisted keys plus get/save accessors.
- src/helpers/ipcHandlers.js, preload.js, src/types/electron.ts: `get-update-auto-install` and `set-update-auto-install` bridge.
- src/components/SettingsPage.tsx: toggle in the Software Updates panel, same load/save pattern as the auto-start toggle.
- src/locales/*/translation.json: `autoInstall` and `autoInstallDescription` keys in all ten languages.
- test/helpers/updateInstallPolicy.test.js, test/helpers/updateManagerInstallGate.test.js, test/helpers/updateAutoInstallSetting.test.js: default and gated flag at setup, runtime flips in both directions, quit-hook re-registration, persistence round-trip through the userData .env across a simulated restart.

### Platform notes

electron-updater (6.8.9 here) reads the flag at two points: `BaseUpdater.addQuitHandler()` registers the install-on-quit hook only when a download completes with the flag on, and the hook checks the flag again at quit. On Windows NSIS and Linux (AppImage, deb, rpm) this means turning the toggle off is honored even for an update that already downloaded, while turning it on after a gated download would normally do nothing until the next download; `setAutoInstallOnAppQuit()` covers that case by calling `addQuitHandler()` directly (protected API, so the call is guarded and degrades to waiting for the next download if a future version drops it). On macOS the flag instead decides at download time whether the update is handed to Squirrel.Mac (see [electron-builder#6418](https://github.com/electron-userland/electron-builder/issues/6418) for the history there): with the gate off the zip downloads but is never staged, so nothing installs on quit; flipping the toggle off after a completed download does not unstage an update Squirrel already has, and flipping it on only affects the next download. The known NSIS install-on-quit flakiness ([electron-builder#7807](https://github.com/electron-userland/electron-builder/issues/7807)) is unaffected since the default path is unchanged.

The actual install-on-quit cannot be exercised without a real signed update artifact, so config-level proof plus gate tests is the honest ceiling here: the tests pin the flag value at setup (before any download can complete, which is when electron-updater first reads it), both runtime directions, and persistence across a simulated restart. The issue's other two levels (a ready-to-install banner when gated, and an `autoDownload` off mode) are left as follow-ups per the issue's "even just option 2" framing.


xlurie:
Author of #1460 here — thanks for picking this up.

I read the diff: the single "Install updates automatically" toggle, default ON, is exactly the "even just option 2" minimum the issue asked for — down to the label — and it closes the issue from my side; the gate sits exactly on the path that changed my version out from under me. The ready-to-install banner and the `autoDownload` level from my sketch are nice-to-haves, fine as separate PRs if ever; I'd rather see this land as-is than wait on them.

Once this is in a release, happy to run the end-to-end check on Windows: toggle off → trigger the download from the update notification → quit → version unchanged, Install & Restart still working — and the same with the toggle flipped off only after the download completes, since that exercises the quit-hook re-check the tests can only simulate.

xlurie:
A scope note now that #1662 has landed on main, changing `src/updater.js`: #1662 gates the automatic update checks but does not touch the install step — `autoInstallOnAppQuit` is still hardcoded `true` on main (`src/updater.js` line 68), so a downloaded update still installs on quit and this PR is not superseded. Merging main (884d9bf) into this branch locally, the only conflict is a single hunk at the top of `src/updater.js` where both sides add a `require`; the rest merges clean.

PR #1532, opened 2026-08-10
fix(meeting): re-evaluate gated mic state and stop prompts expiring early

Chadpiha:
## Description
Meeting detection on macOS was close to non-functional ([user report]): impromptu calls went undetected and calendar prompts vanished before the meeting even started. Three root causes, all confirmed against main:

1. Swallowed mic edges were lost forever. The native mic listeners are edge-triggered — they announce transitions, never steady state (the macOS binary's 5s heartbeat also only emits on change). _onMicStateChanged() discarded the edge outright while any gate was up (user dictating, mic warm-hold, or the 5-minute dismissal cooldown). Since no further event arrives while a call continues, a call starting under a gate was never detected — only the polling fallback self-healed, which made failures look random. macOS is hit hardest: its signal is device-global with no self-PID exclusion, so the user's own dictation both necessitates the gate and consumes the only MIC_ACTIVE edge a concurrent call would produce.

2. An ignored prompt punished the user. An unanswered audio prompt's timeout started the same 5-minute global cooldown as an explicit dismissal — which, combined with (1), made any call starting in the next 5 minutes permanently invisible.
Every prompt died after a hardcoded 30s, regardless of source. Calendar reminders fire 60s before the meeting, so their prompt disappeared ~30s before it started. Hovering only toggled mouse pass-through — the card could close under the pointer mid-click (the renderer also delays the response IPC by a 200ms animation, so a timeout racing a click silently no-oped the Start/Join action).

3. The fix makes the event-driven path level-triggered like the proven polling path: the detector always records _lastKnownMicState and re-evaluates it whenever a gate lifts (recording stop, warm-hold release, cooldown expiry). A false alarm from our own mic self-cancels — the release edge arrives within the 2s sustained window. Timeouts no longer start the cooldown; only explicit dismissal does. And the overlay countdown is now per-source (calendar 2 min, detections 30s) and pauses while hovered, resuming with a 5s minimum runway that also closes the click race.

### DIFF
--- a/src/components/UpdateNotificationOverlay.tsx
+++ b/src/components/UpdateNotificationOverlay.tsx
@@ -49,16 +49,6 @@ export default function UpdateNotificationOverlay() {
     [data]
   );
 
-  const handleMouseEnter = useCallback(() => {
-    setIsHovered(true);
-    window.electronAPI?.setNotificationInteractivity?.(true);
-  }, []);
-
-  const handleMouseLeave = useCallback(() => {
-    setIsHovered(false);
-    window.electronAPI?.setNotificationInteractivity?.(false);
-  }, []);
-
   return (
     <div className="meeting-notification-window w-full h-full bg-transparent p-3">
       <div
@@ -72,8 +62,8 @@ export default function UpdateNotificationOverlay() {
             ? "translate-x-0 opacity-100 scale-100"
             : "translate-x-[120%] opacity-0 scale-95",
         ].join(" ")}
-        onMouseEnter={handleMouseEnter}
-        onMouseLeave={handleMouseLeave}
+        onMouseEnter={() => setIsHovered(true)}
+        onMouseLeave={() => setIsHovered(false)}
       >
         <button
           onClick={() => respond("dismiss")}
--- a/src/helpers/audioActivityDetector.js
+++ b/src/helpers/audioActivityDetector.js
@@ -96,8 +96,7 @@ class AudioActivityDetector extends EventEmitter {
     this._killListenerProcess();
     this._clearSustainedTimer();
     this._clearResetTimer();
-    this._clearCooldownReevalTimer();
-    this._lastKnownMicState = false;
+    this._resetListenerState();
     if (this.checkInterval) {
       clearInterval(this.checkInterval);
       this.checkInterval = null;
@@ -135,9 +134,18 @@ class AudioActivityDetector extends EventEmitter {
     this.consecutiveChecks = 0;
     this.audioActiveStart = null;
     this.hasPrompted = false;
+    this._clearResetTimer();
+  }
+
+  // The pid set and source count mirror what the OS told us is open, not our
+  // own detection state — only losing the listener invalidates them. Clearing
+  // them on dismissal would desync the reference count, so an unrelated app's
+  // mic session ending would report the still-running call as gone.
+  _resetListenerState() {
     this._activeMicPids.clear();
     this._activeSources = 0;
-    this._clearResetTimer();
+    this._lastKnownMicState = false;
+    this._clearCooldownReevalTimer();
   }
 
   _clearSustainedTimer() {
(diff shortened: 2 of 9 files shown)


gabrielste1n (approved):
Traced end to end in an isolated worktree — detector → engine → windowManager → overlay → IPC back — including both native listeners (`macos-mic-listener.swift`, `windows-mic-listener.c`) and the warm-hold wiring.

The diagnosis holds up. Verified independently:

- The `dismissed` flag really was dead on `main` — set to `false` in two places, read in two, never once set `true`.
- The `closed`-handler race is real: `close()` emits asynchronously, so the replaced window's handler was nulling `notificationWindow` (already the replacement) and cancelling the replacement's countdown.
- `MEETING_REMINDER_LEAD_MS` is 60s, so the 2-minute calendar timeout genuinely outlives the meeting start. Only `"audio"`/`"calendar"` ever reach `_showPrompt`, so the default branch is safe.
- The click race is closed. `respond()` slides the card out, mouseleave fires mid-animation, `resume()` re-arms at ≥5s, and the IPC lands at 200ms — safe whether or not mouseleave fires.
- The `handleNotificationTimeout` claim checks out: `hasPrompted` stays `true` for the duration of an ongoing call and only clears via `_startResetTimer()` after 60s of mic-inactive, so an ignored prompt can't re-fire for the same call.

I pushed two fixes for issues the new behaviour surfaced (2cb2b2b), plus a test that fails without the second one:

**The hover pause was attached to a shared channel.** `set-notification-interactivity` ignores `event.sender`, and both `MeetingNotificationOverlay` and `UpdateNotificationOverlay` call it — previously harmless-ish, but this PR gave that channel a countdown to pause. An update toast appearing over a live meeting prompt (same position, same config) paused the *meeting* countdown when hovered, then destroyed itself at its own 5s auto-dismiss without ever firing mouseleave, so `resume()` never came. Result: an always-on-top, content-protected card with no timer left on it. The handler now passes the sender and only the meeting prompt's own window may pause it; the update overlay's two calls are gone, since nothing ever makes that window click-through.

**`dismiss()` was wiping the listeners' OS bookkeeping.** `_reset()` cleared `_activeMicPids`/`_activeSources`, which mirror what the OS says is open rather than our detection state — and the new logic depends on that count being accurate. Windows walkthrough: Teams call ongoing, user dismisses, and during the 5-minute cooldown any unrelated app opens and closes the mic → `MIC_START 22` / `MIC_STOP 22` → size 0 → `_lastKnownMicState` drops to false and the pending re-evaluation is cancelled. The still-running call is never re-detected, i.e. exactly the bug this branch set out to fix, relocated. Same shape on Linux via `_activeSources`; macOS was unaffected (plain boolean). Those counts moved to `_resetListenerState()`, reached only when the listener itself goes away — which also de-duplicates the two lines `stop()` and the polling fallback were already repeating.

Also made `showMeetingNotification` address its own `win` past the load rather than re-reading the shared reference, and bail if a replacement or dismissal took over mid-load. Previously a dev-mode interleave during `waitForDevServer()` would apply one prompt's data and countdown to the other's window, and a dismissal mid-load still armed a countdown that later fired `handleNotificationTimeout()` and cleared `activeDetections` for a prompt that was already gone.

Three things I looked at and deliberately left alone:

- **`pause()` has no upper bound.** With the cross-window leak fixed, the remaining case needs the pointer physically parked on the card, and pausing on hover is the intended feature. Worth revisiting only if it shows up in reports.
- **The 2s self-cancel window is thinner than the description implies.** `setUserRecording(false)` fires from the main process at hotkey release, before the renderer has even received `stop-dictation`, so the sustained timer is armed while our own mic is still open. The CoreAudio property listener normally cancels it in milliseconds — but the Swift binary carries a 5s heartbeat precisely "in case property listeners miss events", and warm-hold defaults to `0`, so nothing else covers that path. Adding an `_isMicActive()` level confirmation would close it; it didn't seem worth an `exec` on the detection path for a speculative failure.
- **Dismissing is now noisier than ignoring.** An explicit dismissal re-prompts every 5 minutes while the mic stays busy; letting the prompt expire leaves you alone. Polling parity is a fair rationale, but the "go away" button nagging more than the no-answer path is worth a follow-up thought.

Local: 28/28 detector, 62/62 across the meeting suite, `tsc --noEmit` and both eslint roots clean.

PR #1539, opened 2026-08-10
fix(meetings): always persist delayed diarization to its owning note (#1495)

### LINKED ISSUE #1495: Back-to-back meetings: delayed diarization can be saved to the wrong note
**Describe the bug**

When two meetings are recorded back-to-back, a delayed diarization result from one recording can be persisted to whichever note is currently rendered instead of the note that owns the recording session. This can cross-contaminate meeting transcripts and speaker embeddings.

This happened in the desktop client with two consecutive meetings. Local/database forensics found a delayed 197-segment diarization result and its speaker embeddings attached to the wrong note. The underlying live transcripts were recoverable, but both local and cloud note records required repair.

**To Reproduce**

1. Start Meeting A with speaker diarization enabled.
2. Stop Meeting A.
3. Before its asynchronous diarization finishes, create/open and begin Meeting B.
4. Allow the delayed diarization callback to complete.
5. Inspect both meeting notes and their speaker assignments.

The issue is timing-dependent: switching the active note while post-meeting diarization is still running opens the race window.

**Expected behavior**

A diarization result, merged transcript, and speaker embeddings must always be saved to the note associated with the completed recording session, regardless of which note is currently open.

**Actual behavior**

The callback validates the diarization `sessionId`, but reads and writes with the current React prop `note.id`. If the active note changes while that diarization session is still current, the validated result can be applied to another note.

In [`NoteEditor.tsx` at this commit](https://github.com/OpenWhispr/openwhispr/blob/8ae216e14c493fa8b237bddaab7a910212dd6794/src/components/notes/NoteEditor.tsx#L528-L570):

- `data.sessionId` is correctly compared with `expectedSession`.
- The persisted note is then fetched with `getNote(note.id)`.
- The merged transcript is saved with `updateNote(note.id, ...)`.
- Speaker embeddings are saved with `saveNoteSpeakerEmbeddings(note.id, ...)`.

The recording store's `recordingNoteId` influences which live segments are selected, but it is not used as the persistence target.

**Suggested fix / regression coverage**

Bind an immutable target note ID to the diarization session when recording starts—either include it in the completion payload or maintain a `sessionId -> noteId` mapping—and use that target for the transcript, embeddings, and cloud sync. Only update the visible React overlay if the rendered note is still that target.

A regression test should stop Meeting A, switch to Meeting B before A's diarization callback completes, and verify that all A results remain attached to A.

**Screenshots**

N/A. Meeting contents are private; the source path and recovered record counts above show the failure without exposing transcript text.

**Desktop**

- OS: Omarchy / Arch Linux, Hyprland
- Browser: N/A — Electron desktop app
- Version: OpenWhispr 1.8.1
- Source inspected: `8ae216e14c493fa8b237bddaab7a910212dd6794`

**Smartphone**

N/A.

**Additional context**

The failure is particularly damaging because the notes can sync successfully after being crossed, making the incorrect assignment appear authoritative on other clients. Preventing stale callbacks from targeting the active note is more important than merely detecting a sync conflict afterward.

Chadpiha:
## Description
Fixes [#1495](https://github.com/OpenWhispr/openwhispr/issues/1495). When two meetings were recorded back-to-back, a delayed diarization result could be saved to whichever note was open instead of the note that owned the recording session — cross-contaminating transcripts and speaker embeddings, and then syncing the corruption to the cloud via the note mirror. A second mode of the same defect silently discarded the result if the next recording started before the callback arrived.

Root cause: the meeting-diarization-complete payload carried only a sessionId, and the renderer's handler — which lived inside the NoteEditor React component — validated that session against mutable global store state but persisted using the rendered note.id prop.

The fix makes completion events self-describing and moves persistence out of the component tree entirely:

The main process now includes the owning noteId in the completion payload (it was already snapshotted at recording stop, just never sent).
Persistence runs in a module-scope listener in meetingRecordingStore, which is guaranteed to be registered before any completion can fire and survives the notes view unmounting — closing the third loss mode where no editor was mounted at all.
NoteEditor only mirrors a published result into its UI (spinner, overlay, speaker mappings) when the rendered note matches, so stale results for a re-recorded note persist without repainting the wrong session's UI.
The routing rules live in a new pure module with node-test coverage, following the repo's dictationRouting/dockPolicy pattern. Guards were added so the write can never resurrect a soft-deleted note's tombstone, and a failed getNote publishes an empty result so a waiting spinner still clears.

## Follow-ups / known limitations
- If the app quits while diarization is still running, the result is lost — the only remaining loss window. Closing it would require main-process-side persistence; worth a separate issue if it shows up in practice.
- updateNote/saveNoteSpeakerEmbeddings remain fire-and-forget without .catch, matching the file's existing style — a logging pass over these IPC writes would make failed persists observable.
- Dev-only: Vite HMR re-evaluating the store module stacks a second listener; matches the pre-existing module-scope listener precedent in the same file.

### DIFF
--- a/src/components/notes/NoteEditor.tsx
+++ b/src/components/notes/NoteEditor.tsx
@@ -72,7 +72,6 @@ import { parseTranscriptSegments } from "../../utils/parseTranscriptSegments";
 import {
   applyTranscriptSpeakerPatch,
   lockTranscriptSpeaker,
-  mergeTranscriptSegments,
   serializeTranscriptSegments,
 } from "../../utils/transcriptSpeakerState";
 import NoteParticipants from "./NoteParticipants";
@@ -365,7 +364,6 @@ export default function NoteEditor({
     Array<{ id: number; display_name: string; email: string | null }>
   >([]);
   const editorRef = useRef<Editor | null>(null);
-  const displaySegmentsRef = useRef<TranscriptSegment[]>([]);
 
   const embeddedChat = useEmbeddedChat({
     noteId: note.id,
@@ -399,10 +397,6 @@ export default function NoteEditor({
     if (diarizedSegments && diarizedSegments.length > 0) return diarizedSegments;
     return parseTranscriptSegments(note.transcript || "");
   }, [diarizedSegments, note.transcript]);
-
-  useEffect(() => {
-    displaySegmentsRef.current = displaySegments;
-  }, [displaySegments]);
 
   const hasChatSegments = displaySegments.length > 0;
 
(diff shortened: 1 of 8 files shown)


gabrielste1n (approved):
Traced end-to-end and verified against the code, not just the diff:

- `noteId` is snapshotted before `resetMeetingLocalState()` nulls it in both `_startOrSkipDiarization` call sites, so the payload always carries the owning note.
- The store listener is genuinely guaranteed to be registered: `meetingLocalWin` is `BrowserWindow.fromWebContents(event.sender)` from `meeting-transcription-start`, so the completion always returns to the window that started the recording — and that window necessarily evaluated `meetingRecordingStore` to start it.
- The soft-delete guard is load-bearing, not defensive noise: `databaseManager.getNote()` has no `deleted_at` filter, and `updateNote` unconditionally appends `sync_status = 'pending'`, which fires `_asyncVectorUpsert` and `_asyncMirrorWrite`. Writing to a tombstone really would resurrect it.
- Only NoteEditor consumed `onMeetingDiarizationComplete`, so nothing else breaks.

Pushed four follow-up fixes in ffe8d68:

1. `let persisted;` was implicitly `any` under `strict: false`, so the tombstone guard had no compiler backstop — a `deletedAtTypo` compiled clean. Now typed as `NoteItem`, and the typo errors.
2. A resolved-`null` note skipped the spinner clear that a thrown `getNote` got. Both paths now publish `[]`, and the read failure is logged instead of swallowed.
3. Dropped the unreachable legacy branch in `resolveDiarizationTarget`. `startRecording` sets `recordingNoteId: args.noteId` and sends `noteId: args.noteId ?? null` from the same value, which main snapshots as `meetingNoteId` — so a null `payloadNoteId` always implies a null `recordingNoteId`, and the branch could only ever return `null`. Main and renderer ship in one bundle, so there's no version skew to guard against. Four tests covering the dead path retired, one added documenting the drop.
4. Trimmed comments that restated the code.

`tsc --noEmit` clean, 12/12 tests pass, `prettier --check` clean, all 9 CI checks green.

One adjacent observation, pre-existing and outside this diff: `isDiarizing` is armed at exactly one place, gated on `prevRecordingForDiarizationRef.current && !isRecording && diarizationSessionId`. `stopRecording` flips `isTranscribing` false in its first `setState` but only sets `diarizationSessionId` two awaits later, so the effect runs once with the session still null — clearing the ref before the render that has the id. If that's right, the hint never appears and the empty-publish plumbing is servicing nothing. Worth a manual check; not a blocker here.

PR #1544, opened 2026-08-11
fix(linux): link libgobject-2.0 for AT-SPI2 text monitor build

### LINKED ISSUE #1531: Linux: auto-learn correction monitoring silently disabled — linux-text-monitor binary was never published
*This issue was AI-generated (investigated and filed by Claude on the reporter's behalf).*

The auto-learn dictionary feature (#327) depends on a native `linux-text-monitor` binary (AT-SPI2-based) that's supposed to be downloaded during packaging via `scripts/download-text-monitor.js`. On Linux this silently no-ops because no `linux-text-monitor-v*` release has ever been published — only `windows-text-monitor-v1.0.0` exists.

**Root cause:** the "Build Linux Text Monitor" Actions workflow has run exactly once (2026-02-26, the same day #327 merged), failed with exit code 1, and was never rerun since. Every Linux release since then (v1.7.0 through the current v1.8.1) ships without the binary — confirmed by inspecting `OpenWhispr-1.8.1-linux-x64.tar.gz` directly: no `linux-text-monitor` and no `.py` fallback present anywhere in the archive.

**Effect:** on Linux, editing pasted transcriptions never triggers dictionary learning, in any application. `textEditMonitor.resolveBinary()` returns `null` for the missing binary, so `startMonitoring()` bails out immediately and no `text-edited` event is ever emitted.

**Ask:** rerun/fix the `build-linux-text-monitor.yml` workflow and publish a `linux-text-monitor-v*` release so `download-text-monitor.js` can pick it up in future builds.

**Workaround** in the meantime: compile `resources/linux-text-monitor.c` locally against `atspi-2` (`gcc -O2 linux-text-monitor.c -o linux-text-monitor $(pkg-config --cflags --libs atspi-2)` — needs `libatspi2.0-dev`/`at-spi2-core` + glib dev headers) and drop the binary at `resources/bin/linux-text-monitor` in the install directory.

iSparsh:
## What this fixes

`pkg-config --libs atspi-2` leaves out `-lgobject-2.0` — it's listed under `Requires.private` in `atspi-2.pc`, not `Requires`. Linkers that default to `--as-needed` (current Ubuntu/Arch, including the GitHub Actions runner) fail with `undefined reference to symbol 'g_object_unref'`.

Fix: use `pkg-config --cflags --libs --static atspi-2`, which pulls in private deps too.

Also tightened the "Verify binary" CI step. It used to only check the file exists. Now it runs `ldd` and checks for unresolved libraries and for `libgobject-2.0.so` specifically, so this class of break gets caught in CI instead of failing silently again.

Tested locally: compiles clean, `ldd` shows no missing libraries.

## Takeaway

Even with the binary working, apps like Firefox won't expose their accessibility tree unless this is set — off by default outside GNOME (niri, Hyprland, Sway, KDE):

```
gsettings set org.gnome.desktop.interface toolkit-accessibility true
```

Restart the app after — it's read once at startup. Without it, auto-learn silently does nothing for that app.

Fixes #1531

## Summary
The ATSPI-2 feature on linux that we use for corrections for the voice models wasn't actually shipped since #327. This PR fixes it by including the c file and then fixing CI as well.

### DIFF
--- a/.github/workflows/build-linux-text-monitor.yml
+++ b/.github/workflows/build-linux-text-monitor.yml
@@ -30,12 +30,16 @@ jobs:
 
       - name: Compile Linux Text Monitor
         run: |
-          gcc -O2 resources/linux-text-monitor.c -o linux-text-monitor $(pkg-config --cflags --libs atspi-2)
+          gcc -O2 resources/linux-text-monitor.c -o linux-text-monitor $(pkg-config --cflags --libs --static atspi-2)
 
       - name: Verify binary
         run: |
           file linux-text-monitor
-          test -f linux-text-monitor || { echo "Binary not found"; exit 1; }
+          test -x linux-text-monitor || { echo "Binary not found or not executable"; exit 1; }
+          ldd linux-text-monitor
+          ldd linux-text-monitor | grep -q "not found" && { echo "Unresolved shared library dependency"; exit 1; }
+          ldd linux-text-monitor | grep -q libgobject-2.0.so || { echo "libgobject-2.0 not linked - pkg-config --static flag may have regressed"; exit 1; }
+          exit 0
 
       - name: Create tar.gz archive
         run: |
--- a/.github/workflows/build-linux-text-monitor.yml
+++ b/.github/workflows/build-linux-text-monitor.yml
@@ -30,7 +30,7 @@ jobs:
 
       - name: Compile Linux Text Monitor
         run: |
-          gcc -O2 resources/linux-text-monitor.c -o linux-text-monitor $(pkg-config --cflags --libs --static atspi-2)
+          gcc -O2 resources/linux-text-monitor.c -o linux-text-monitor $(pkg-config --cflags --libs atspi-2) -lgobject-2.0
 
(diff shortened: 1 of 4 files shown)


gabrielste1n (approved):
Thanks for the excellent investigation and fix — the root-cause analysis was spot on, and the hardened verify step is a great addition.

One correction pushed to your branch: `--static` pulls in the full transitive `Requires.private` closure, which includes dbus-1's `-lsystemd`. On a clean Ubuntu 24.04 (what `ubuntu-latest` runs) that fails with `cannot find -lsystemd` because `libsystemd-dev` isn't installed — it worked on Arch because systemd libs resolve there by default. Swapped it for an explicit `-lgobject-2.0`, which links cleanly with only the deps the workflow installs. Verified the workflow compile line, the build-script path, and the verify step end to end in a clean Ubuntu 24.04 container (including the old flags reproducing the exact `g_object_unref` DSO error you hit).

Merging — the build workflow triggers on this merge and should publish the `linux-text-monitor-v1.0.0` release so future packages pick it up.

PR #1550, opened 2026-08-11
fix(linux): prefer portal paste on GNOME once a restore token exists

wallucas:
## Problem

On GNOME Wayland, `pasteLinux` tries uinput before the RemoteDesktop portal. That order exists for a good reason (#494: the portal's first use shows a permission dialog and can stall pastes 10s+), but the uinput path creates and destroys the `openwhispr-paste` virtual keyboard on **every paste**, forcing mutter to hotplug an input device each time.

Under sustained dictation this destabilized the compositor on my machine (Fedora 44, GNOME/mutter 50.4, Wayland) two days in a row: minutes after a dictation burst, at-spi2 flags gnome-shell as unresponsive, libinput logs `event processing lagging behind`, and the session freezes hard enough to require a power cycle. Kernel logs show one new `input: openwhispr-paste as /devices/virtual/input/inputNN` per paste leading up to each hang; after switching pastes to the portal, the device churn and the hangs are gone.

## Fix

Keep uinput first for the first-ever paste (preserving the #494 behavior — no new dialog, no stall for fresh installs). But once a portal **restore token** has been saved, prefer the portal: with a token it is dialog-free and instant, injects keys compositor-side, and creates no devices. If the tokened portal paste fails, fall back to uinput as before.

The change is confined to the GNOME branch of the Wayland paste ordering in `src/helpers/clipboard.js`; KDE and wlroots ordering are untouched.

## Testing

- New `test/helpers/linuxPasteOrder.test.js` (same `Module._load` mocking pattern as `clipboardRestore.test.js`) covering: portal-first when a token exists, uinput-first when none exists, and uinput fallback when the tokened portal paste fails.
- Full suite: identical pass/fail tally to `main` on my machine, plus the 3 new tests passing.
- Live-verified on GNOME 50.4/Wayland: with a saved token, dictation pastes go through the portal (RemoteDesktop session visible in the journal, no permission dialog, no new kernel input devices); with the token file removed, behavior matches current main.

### DIFF
--- a/src/helpers/clipboard.js
+++ b/src/helpers/clipboard.js
@@ -112,6 +112,7 @@ class ClipboardManager {
     this.linuxFastPastePath = null;
     this.linuxFastPasteChecked = false;
     this.portalDenied = false;
+    this.portalTokenPasteFailed = false;
     this._kwinScriptPath = null;
     this.pasteQueue = Promise.resolve();
 
@@ -1653,16 +1654,16 @@ class ClipboardManager {
           }
         } else if (isGnome && linuxFastPaste) {
           // GNOME: uinput first for the first-ever paste — the portal shows a
-          // permission dialog and can stall for 10s+ (issue #494). But once a
-          // restore token exists the portal is dialog-free and instant, so
-          // prefer it: uinput creates and destroys a virtual keyboard on every
-          // paste, forcing mutter to hotplug an input device each time, which
-          // has driven gnome-shell into unrecoverable hangs under sustained
-          // dictation.
-          const havePortalToken = !this.portalDenied && !!this._readPortalToken();
-          if (havePortalToken) {
+          // permission dialog and can stall for 10s+ (issue #494). A valid
+          // restore token resumes without the dialog, so prefer it until a
+          // tokened attempt fails. This avoids the virtual keyboard hotplug
+          // churn that can hang gnome-shell under sustained dictation.
+          const shouldPreferPortal =
+            !this.portalDenied && !this.portalTokenPasteFailed && !!this._readPortalToken();
+          if (shouldPreferPortal) {
             const portalPaste = await tryPortalPaste();
             if (portalPaste) return { method: "portal", ...portalPaste };
+            this.portalTokenPasteFailed = true;
           }
           try {
             const uinputPaste = await tryUinputPaste();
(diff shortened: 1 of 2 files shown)


Chadpiha (approved):
Approved. GNOME now prefers the portal when a restore token exists, falls back to uinput on failure, and avoids retrying a failed token first on subsequent pastes. Regression coverage includes token, no-token, fallback, and consecutive-paste behavior. Focused tests pass (28/28).

gabrielste1n:
Heads-up: #1629 just landed on main and touches the adjacent portal guards in `pasteLinux()` (new `portalFailed` session flag alongside `portalDenied`, and a 2.5s portal timeout when a restore token exists) — this branch will need a small rebase.

Worth noting the interaction is favorable for this PR: preferring the portal on GNOME once a token exists is now much safer, since a stale/unresponsive portal session costs at most 2.5s once per app session instead of 15s on every paste. When rebasing, the new portal-first guard should include `!this.portalFailed` like the KDE one does.

Chadpiha:
Thanks @wallucas — this was the right call, and it's now on main.

Closing as superseded by [#1525](https://github.com/OpenWhispr/openwhispr/pull/1525)
 (merged 2026-08-17, shipped in v1.9.0). While getting that branch mergeable I folded this ordering into it (a890978b, "Harden Wayland paste fallbacks"), so main now has what this PR proposed:

portal-first on GNOME only when a saved restore token exists (shouldPreferGnomePortal), uinput-first otherwise, so fresh installs still avoid the tokenless permission dialog from #494 a portalTokenPasteFailed session flag so a failed tokened portal attempt isn't retried first on every paste
the !portalFailed / !portalUnavailable guards Gabe flagged above, so a stale portal session costs one paste instead of all of them ([#1614](https://github.com/OpenWhispr/openwhispr/issues/1614))

Coverage for the same scenarios (token → portal first, no token → uinput first, tokened portal failure → uinput, and the consecutive-paste case) lives in test/helpers/clipboardRestore.test.js, so a rebase here would leave only a duplicate test file.

That commit should have referenced this PR and didn't — my miss. The approach and the gnome-shell hang diagnosis are yours. If you see any regression in the tokened portal path on GNOME 50 with 1.9.x, please open an issue and tag me.

PR #1568, opened 2026-08-12
feat(design): global design-language uplift — radius tiers, flat depth, capsule actions, one brand blue

gabrielste1n:
## Summary

A global design-language uplift for the desktop app, derived from the new Figma direction. Everything is done at the **token and primitive layer** — no page-level or inline restyling — so the whole app moves at once.

Net **-69 lines**: the old decoration was more code than the new flatness.

### Token layer (`src/index.css`)

- **Radius tiers fixed by role** instead of ad-hoc values: control 8px / tile 12px / card 16px / floating 20px / window 44px, and every action is a capsule. (Note `--radius-lg`/`--radius-xl` were 8/10px and are now 12/16px, so anything using those classes gets visibly rounder — intended.)
- **Depth is flat.** Card/panel shadows, hover lifts, glassmorphism blurs, and the decorative `accent-bar` gradient are gone. Containment now comes from fill alternation plus at most one hairline per surface, with alpha-based dividers so they sit correctly on any tint. Transient layers that genuinely float (menus, popovers, dialogs, tooltips) share a single restrained `--shadow-transient` token instead of per-component shadow values.
- **Warm paper-neutral surfaces** and alpha hairline borders, in both themes.
- **One brand blue.** `primary`, `accent`, `ring`, focus border, and `link` all resolve to `#2563eb` (blue-600), so there is no second or third blue in the system. Dark-mode links sit slightly lighter (0.72 L) because text needs more contrast against near-black than a fill does.

### Primitives

- **Button** — dense monochrome capsules (h-9/8/7 plus a new `xs` size), fill-step hover/active instead of lifts, contrast-drop disabled states. All existing variant names preserved, so no call sites change.
- **Tabs** — capsule segmented control with a brand-blue active chip.
- **Toggle** — brand-blue checked track, no thumb shadow, no `opacity-50` disabled.
- **Card / SettingsSection** — flat fill plus one hairline at card radius; no blur or translucency.
- **Input / Textarea** — sunken borderless fill with a ring focus state; removes hardcoded `neutral-*`/`indigo-*` palette in favour of tokens.
- **InfoBox / Alert** — quiet tinted strips, fill only, never fill *and* border.
- **Badge** — new `muted` variant for ambient, non-interactive status.

### Settings → Account org banner

The "Some settings are managed by your organization" banner was a one-off inline `<div>` with hardcoded `blue-50/200/800`, which is why the token pass didn't reach it. It broke the grammar four ways: decorative accent colour on a passive notice, fill *and* border, alarm-scale padding for a caption, and `mx-4` stacked on the parent's `p-6` so it sat inset from the content column it was introducing.

It is now a content-hugging status chip with a lock glyph — sunken monochrome fill, muted text, aligned to the content column, 24px instead of ~44px, and no longer colliding with the close button.

It also reuses the existing terse `common.managedByOrg` copy ("Managed by your organization"), already translated in all 9 locales, so no machine-translation guesswork. The now-unused sentence-length `settingsModal.managedByOrg` key is removed from all 10 locale files (one-line deletions, no reformatting).

## Verification

- `npm run typecheck` — passes
- `npm run build:renderer` — passes
- `npm test` — 1825 tests, **1655 pass / 0 fail**, 169 skipped
- `npm run format:check` — eslint + prettier clean
- Ran in dev against the production API and clicked through onboarding and settings

`origin/main` merged in (28 commits, no conflicts); main's incoming UI work was audited and introduces no hardcoded palette, shadows, or off-tier radii.

## Review notes

- **81 files use `primary`-prefixed classes.** Anywhere a page previously used blue-primary to mean "selected", it still renders brand blue — but per-feature selection semantics were deliberately left alone rather than swept inline. Moving true selection states onto `accent` is follow-up work.
- A few tiny chips currently using `rounded-lg` may want re-tiering to `radius-control` during per-feature passes.
- The renderer retries workspace/policy fetches with **no backoff** when the API URL is unset or unreachable, which spins re-renders. Packaged builds always have a channel baked in so users are unlikely to hit it, but it is worth a defensive fix on main — **not touched here**.


gabrielste1n:
## Update: brand blue, amber cleanup, and the welcome/auth screen

### Brand blue restored (`dec369b1`)
The first pass encoded the earlier comps' black pills as the primary action colour, which stripped brand blue out of the product. `primary` is `#2563eb` again, and every blue in the system — `primary`, `accent`, `ring`, focus border, `link` — now collapses onto that one value. Dark-mode links sit slightly lighter (0.72 L) because text needs more contrast against near-black than a fill does.

Figma node `1133:84` (added after that fix) independently confirms the direction: it reworks the welcome card around a brand-blue hero.

### Amber tokenized (`16a08d24`)
The two hand-rolled warning panels in `ControlPanel` and the lone warning caption in `SettingsPage` bypassed the `warning` token entirely. They now route through `--color-warning`, and the fill+border pairing is gone per the depth rule.

Not swept: ~20 further amber usages across notes / history / editor surfaces. Those mix status dots, glow shadows and editor highlight tints, so they want visual checking per feature rather than a blind sweep.

### Welcome/auth screen rebuilt (`a02849ea`)
- Full-bleed brand hero — blue gradient plus a halftone dot screen, white logo glyph at 64px, white title/subtitle. Gradient lives in a `.brand-hero` utility so no hex reaches the component; the band bleeds the 24px padding both call sites apply so their rounded corners clip it.
- Mark rendered inline from `logo.svg` geometry with `currentColor`, rather than the coloured `icon.png` which would clash on blue.
- Email-first order per the comp. The CTA is the standard primary button, so it reads as the comp's grey capsule while empty and brand blue once an address is entered.
- Three stacked social buttons and the SSO text link collapse into one 4-up tile row behind a `ProviderTile` component. Apple stays macOS-only; the row reflows to three.
- Onboarding step-0 column widened to `max-w-md` so four tiles fit without truncating, and the step card's translucency, blur and shadow removed.

Behaviour is unchanged — same handlers, disabled conditions, loading states, protocol gating and error surface. Provider labels are brand names, which the i18n rules exclude from translation, so no new keys.

### Two decisions I did not make unilaterally

1. **Font.** The comp sets the title in **Space Grotesk Medium**, not the Feature Display serif of the earlier comps and not the app's Noto Sans. Adopting it means committing woff2 files (the app bundles fonts locally for offline use) — a brand-level call, so the title currently renders in Noto Sans.
2. **Pill email input.** The comp uses a fully-rounded input on this screen, whereas the grammar puts inputs on the 8px control tier. I followed the comp here since it is a hero surface, but it is a deliberate inconsistency with every other input in the app and is trivial to revert.

### Verification
`typecheck` / `build:renderer` / `lint` clean; `npm test` 1928 tests, **1758 pass / 0 fail**.

**The auth screen has not been visually confirmed** — it only renders during first-run onboarding or the signed-out sign-in dialog, neither reachable in a signed-in dev session, and my attempts at a headless render harness failed (no Chrome available). Worth a look before merge.

iSparsh:
Not sure if its possible but can we see the design board somewhere?

PR #1574, opened 2026-08-12
fix(startup): make launch at login work on Windows and start hidden everywhere

### LINKED ISSUE #1541: Windows startup app not recognized
After I installed openwhispr, it did not appear automatically in the startup apps directory, so I'm having to open it manually

I'd like it to automatically start on startup, so i won't have to open it manually, preferable in hidden mode

gabrielste1n:
Closes #1541. Follows #1518, which made the toggle work on Linux; this fixes what it does once it is on.

## The problem

"Launch at login" was opt-in only, and once enabled it still put the control panel in the user's face at every login. Tracing it turned up three separate defects.

**1. Nothing delivered a hidden start on any platform.** The login item was written with `openAsHidden: true` and the comment `// Start minimized to tray`. Per [Electron's own typings](https://www.electronjs.org/docs/latest/api/app#appsetloginitemsettingssettings-macos-windows), `openAsHidden` is macOS-only *and* deprecated ("does not work on macOS 13 and up"), so it was inert on every platform we ship. Windows has no native equivalent, so the login item now passes `--hidden` and startup reads it back. Linux carries the same flag on the autostart entry's `Exec`. macOS uses `wasOpenedAtLogin`, the only signal left there now that `wasOpenedAsHidden` is also inert — [restored for macOS 13+ in electron#42411](https://github.com/electron/electron/pull/42411), so it is safe on Electron 41.

A login launch skips the control panel *without* writing to the user's `startMinimized` preference, so a manual launch still behaves however they configured it.

**2. Windows read the wrong field — this is the "not recognized" report.** `openAtLogin` is an exact string compare of the `Run` value against `"execPath" + args` ([browser_win.cc:732](https://github.com/electron/electron/blob/v41.2.0/shell/browser/browser_win.cc#L732)) and never consults `StartupApproved\Run`, which is what Task Manager and Settings → Startup Apps write when a user disables a startup app. Only `launchItems[].enabled` and `executableWillLaunchAtLogin` account for it ([:224-260](https://github.com/electron/electron/blob/v41.2.0/shell/browser/browser_win.cc#L224-L260)) — [electron#20122](https://github.com/electron/electron/issues/20122) was only ever half-fixed, on the write side. So a user who switched OpenWhispr off in Task Manager got a toggle that still read "on" while the app never started. State now comes from `executableWillLaunchAtLogin`, and the UI reads the state back after every write instead of assuming the write took.

**3. Uninstalling left a broken startup entry behind.** Electron writes the `Run` value itself, so nothing in the generated NSIS uninstaller removed it and Windows kept listing a startup item pointing at a deleted executable ([electron-builder#2237](https://github.com/electron-userland/electron-builder/issues/2237)). Now removed on uninstall, deliberately *not* on update — guarded by `${isUpdated}` like the existing firewall-rule cleanup, so an update never silently turns the setting off.

Note on what we deliberately did **not** do: pass a friendly `name` to `setLoginItemSettings`. `getLoginItemSettings` has no `name` option and hard-codes the AppUserModelId as the value name it reads ([:729](https://github.com/electron/electron/blob/v41.2.0/shell/browser/browser_win.cc#L729)), so a custom name would make the toggle read false forever. The registry value name stays on the AUMID.

## Also: macOS approval state

On macOS 13+ `setLoginItemSettings()` routes through `SMAppService`, which can register the item and still return [`status: "requires-approval"`](https://developer.apple.com/documentation/servicemanagement/smappservice/status-swift.enum/requiresapproval) while it waits for the user under System Settings → General → Login Items. We ignored `status` entirely, so that state rendered as a toggle that simply would not stick. It now explains itself and links to the pane.

## Structure

Decision logic lives in `autoStartPolicy.js` — no Electron import, fully unit-tested — with `autoStart.js` as the thin shell over `setLoginItemSettings()` and `linuxAutostart.js`. Same split as `dockPolicy.js`/`dockManager.js`. That also collapses the per-platform branching that had started accumulating inside `get-auto-start-enabled` / `set-auto-start-enabled`.

`get-auto-start-enabled` now returns `{ enabled, requiresApproval }` instead of a bare boolean; its one consumer and the type definition are updated with it.

## On defaulting it to ON

Not doing that, and I don't think we should. Apple's review guidelines prohibit apps auto-launching at login without consent, and the convention across macOS utilities is opt-in plus a prompt in onboarding. We ship DMG/ZIP rather than MAS so it isn't a rejection risk, but quietly adding ourselves to startup is the behaviour Windows 11's startup-impact UI exists to shame. The reporter's actual complaint is discoverability, which is better answered by a step in `OnboardingFlow.tsx` — worth doing, but it's a product call and a separate PR.

## Testing

`test/helpers/autoStartPolicy.test.js` adds 13 cases: the Task-Manager-disabled read, the args-mismatch read, all four macOS `status` outcomes, each migration branch (including that launch-at-login merely being *off* is not mistaken for a stale entry, which would resurrect an item the user removed), and hidden-launch detection per platform.

`test/helpers/linuxAutostart.test.js` gains 2 cases pinning the flag into the `Exec` comparison. Without them, an up-to-date entry looks stale and gets rewritten on **every single launch** — the exact trap flagged when reviewing #1518.

`npm test` (1862 pass / 0 fail), `npm run lint`, `npm run typecheck` and Prettier all clean.

**Verified against real Electron 41.2.0 on macOS 26.5.1**, driving the actual helper rather than a mock:

- `getLoginItemSettings()` returns the `status` and `wasOpenedAtLogin` fields the new code reads (`not-found` / `false` for an unregistered binary).
- Enable → `{enabled: true, requiresApproval: false}` with raw `status: enabled`; disable → back to `{enabled: false, requiresApproval: false}`. Round trip is clean and self-reverting.
- `syncAutoStartEntry()` correctly no-ops on darwin.
- The macOS Login Items URL opens **System Settings → Login Items & Extensions**, confirmed on-device rather than taken from a blog post.

**Not verified**, to be straight about the limits of the above:

- **Anything on Windows.** No Windows machine in this environment, so the registry write, the `executableWillLaunchAtLogin` read, the `--hidden` argv round trip, the migration path and the NSIS uninstall macro are all reasoned from Electron's Windows source rather than observed. The NSIS macro in particular is only exercised by an actual uninstall.
- **A real login cycle** on any platform. No logout/login was performed, so hidden start is verified only at the level of "the flag is written and read back".


gabrielste1n:
Traced the whole change end to end before merging. Three things came out of it, pushed in 22f6e4b:

**The Linux entry pointed at the wrong binary.** `scripts/afterPack.js` renames the Linux executable to `<name>-app` and leaves a wrapper script under the original name, so `process.execPath` inside the running app is the *inner* binary. The autostart entry pointed there, which means a login launch skipped everything the wrapper does: XWayland forcing, the `--no-sandbox` fallback for distros with restricted user namespaces, and the user's `~/.config/open-whispr-flags.conf`. It now resolves back to the wrapper when one sits beside the binary, and keeps `execPath` when none does rather than inventing a path that doesn't exist. Both cases are tested; the first fails without the fix.

Worth noting this was latent in #1518 too — it only became worth fixing here because this PR changes what goes on the `Exec` line, and shipping `--hidden` to the wrong binary would have compounded it.

**The login-item probe ran unguarded on the startup path.** `wasLaunchedAtLoginHidden()` reads OS state, and it sat inside `startApp()` with no try/catch — a failure would have surfaced to the user as "Failed to start app" instead of just showing the window. It now falls back to the pre-existing behaviour, matching how `syncAutoStartEntry()` is already guarded a few lines up. Same class of defect as the icon-copy bug found while reviewing #1518: a non-essential step able to kill an essential one.

**One test asserted an impossible state.** Electron derives `openAtLogin` on macOS 13+ from `status == "enabled"` ([browser_mac.mm:432](https://github.com/electron/electron/blob/v41.2.0/shell/browser/browser_mac.mm#L432)), so an item awaiting approval *always* reads as off — `{openAtLogin: true, status: "requires-approval"}` cannot occur. Corrected to the reachable state, which also confirms the UI is right: the toggle reads off and the alert explains why, rather than the two contradicting each other.

Also trimmed comments that restated the same fact in more than one place, and dropped optional chaining on values the single caller always supplies.

Re-verified after the cleanup, driving the real helper under Electron 41.2.0 on macOS 26.5.1: initial read, `syncAutoStartEntry()` no-op, enable → `status: enabled`, disable → `status: not-registered`. Clean round trip. Locale key parity checked across all 10 files. `npm test` 1864 pass / 0 fail, lint, typecheck and Prettier clean.

Windows remains unverified by execution — no Windows machine here — so the registry write, the argv round trip, the migration and the NSIS macro are reasoned from Electron's Windows source. Worth a manual pass on Windows before the next release.

PR #1587, opened 2026-08-12
fix(voice-agent): support extraction when completionMarker is empty or omitted

### LINKED ISSUE #1586: fix(voice-agent): support selection edit replacement extraction when completionMarker is empty or omitted
### Affected Component
`src/helpers/selectionEditing.js` (`extractSelectionEditReplacement`)

### Actual Behavior
`extractSelectionEditReplacement(result, completionMarker)` throws `Error: Model output was incomplete before the selection edit completed` whenever `completionMarker` is falsy (e.g. `""`, `null`, or `undefined`).

Even though `buildSelectionEditSystemPrompt(basePrompt, completionMarker = "")` explicitly permits omitting or passing an empty `completionMarker` (in which case no marker instruction is added to the system prompt), `extractSelectionEditReplacement` fails all marker-free extractions due to `!completionMarker` in its validation guard.

### Expected Behavior
When `completionMarker` is empty, null, or undefined, `extractSelectionEditReplacement` should extract and return the replacement text without requiring or checking for a suffix marker, provided `result` is a non-empty string.

### Environment
- Repository: OpenWhispr/openwhispr
- Branch: main
- Runtime: Node.js 24+ / Electron

### User Impact
Selection edits executed with prompts or models that do not supply a `completionMarker` fail unexpectedly with `SELECTION_EDIT_REASONING_FAILED` ("Model output was incomplete before the selection edit completed"), even when the model returned a valid and complete replacement text.

### Root-Cause Hypothesis
In `src/helpers/selectionEditing.js`:
```javascript
export function extractSelectionEditReplacement(result, completionMarker) {
  if (typeof result !== "string" || !completionMarker || !result.endsWith(completionMarker)) {
    throw new Error("Model output was incomplete before the selection edit completed");
  }
```
The check `!completionMarker` assumes a completion marker is mandatory. It should instead check `completionMarker && !result.endsWith(completionMarker)`.

### Proposed Scope
- Update `extractSelectionEditReplacement` in `src/helpers/selectionEditing.js` to handle empty or omitted `completionMarker`.
- Add deterministic regression tests in `test/helpers/selectionEditing.test.js` covering marker-less, empty-string, nullish, and valid marker extractions.

I intend to submit a pull request with a fix and unit tests for this issue.

hsusul:
## Summary

This pull request fixes a bug in `extractSelectionEditReplacement` where selection edit extractions threw `Error: Model output was incomplete before the selection edit completed` whenever `completionMarker` was falsy (empty string, `null`, or `undefined`).

## Problem

`buildSelectionEditSystemPrompt(basePrompt, completionMarker = "")` explicitly supports omitting or passing an empty `completionMarker`, in which case no marker instruction is added to the system prompt. However, `extractSelectionEditReplacement(result, completionMarker)` evaluated `!completionMarker` in its validation guard as an incomplete output error, causing all marker-less selection edit extractions to fail even when the model produced a complete and valid replacement text.

## Root Cause

`extractSelectionEditReplacement` checked `!completionMarker` as part of its failure condition:
```javascript
if (typeof result !== "string" || !completionMarker || !result.endsWith(completionMarker))
```
This treated `completionMarker` as mandatory.

## Fix

Updated `extractSelectionEditReplacement` to evaluate `(completionMarker && !result.endsWith(completionMarker))` so that extractions without a completion marker succeed and return the non-empty `result` string as expected.

## Behavior Before and After

- **Before**: `extractSelectionEditReplacement("replacement text", "")` threw `Error: Model output was incomplete before the selection edit completed`.
- **After**: `extractSelectionEditReplacement("replacement text", "")` returns `"replacement text"`.

## Scope & Non-Goals

- Scope: `src/helpers/selectionEditing.js` and unit tests in `test/helpers/selectionEditing.test.js`.
- Non-goals: Does not alter system prompt generation or completion marker behavior when a marker is provided.

## Tests Added

- `test/helpers/selectionEditing.test.js`: Added unit tests verifying `extractSelectionEditReplacement` with empty string `""`, `undefined`, `null`, whitespace-only output, and invalid argument types.

## Validation Results

- `ELECTRON_OVERRIDE_DIST_PATH=/tmp node --experimental-strip-types --test test/helpers/selectionEditing.test.js` (Passed 2/2 tests)
- `npm run typecheck` (Passed cleanly)
- `npm run lint` (Passed cleanly)
- `npm run format:check` (Passed cleanly)
- `npm run i18n:check` (Passed cleanly)
- `npm run build:renderer` (Passed cleanly)
- `git diff --check` (Passed cleanly)

Fixes #1586

### DIFF
--- a/src/helpers/selectionEditing.js
+++ b/src/helpers/selectionEditing.js
@@ -43,11 +43,11 @@ export function getSelectionCaptureDisposition(capture, accessibilitySkipped = f
 }
 
 export function extractSelectionEditReplacement(result, completionMarker) {
-  if (typeof result !== "string" || !completionMarker || !result.endsWith(completionMarker)) {
+  if (typeof result !== "string" || (completionMarker && !result.endsWith(completionMarker))) {
     throw new Error("Model output was incomplete before the selection edit completed");
   }
 
-  const replacement = result.slice(0, -completionMarker.length);
+  const replacement = completionMarker ? result.slice(0, -completionMarker.length) : result;
   if (replacement.trim().length === 0) {
     throw new Error("Model returned an empty selection edit");
   }
--- a/test/helpers/selectionEditing.test.js
+++ b/test/helpers/selectionEditing.test.js
@@ -27,14 +27,8 @@ test("builds a structured prompt that keeps instruction and selection separate",
   assert.match(systemPrompt, /Output only the complete replacement text/);
   assert.match(systemPrompt, new RegExp(marker));
 
-  assert.equal(
-    extractSelectionEditReplacement(`Improved text${marker}`, marker),
-    "Improved text"
-  );
-  assert.throws(
-    () => extractSelectionEditReplacement("Truncated text", marker),
-    /incomplete/
-  );
+  assert.equal(extractSelectionEditReplacement(`Improved text${marker}`, marker), "Improved text");
+  assert.throws(() => extractSelectionEditReplacement("Truncated text", marker), /incomplete/);
 
   assert.equal(getSelectionCaptureDisposition({ status: "none" }), "standalone");
   assert.equal(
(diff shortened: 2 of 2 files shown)


Chadpiha (approved):
Thanks @hsusul — approving.

Verified the change end-to-end:

For a truthy marker the new guard is logically identical to the old one, so the production path (the only call site in audioManager.processSelectionEdit always generates a UUID marker) is provably unchanged — including truncated-output detection and the text-only screenshot-retry path.
The marker-less path now works as buildSelectionEditSystemPrompt's optional-marker signature always implied, and the ternary correctly avoids the slice(0, -0) trap an empty marker would have hit.
New tests cover "" / undefined / null, whitespace-only rejection, and non-string rejection. Full suite passes, and I confirmed a trial merge with current main resolves cleanly and keeps the [#1593](https://github.com/OpenWhispr/openwhispr/pull/1593) disposition semantics in the shared test file.

One framing note: since no production code path passes a falsy marker today, this lands as contract-consistency hardening rather than a user-facing bugfix — the SELECTION_EDIT_REASONING_FAILED failures described in [#1586](https://github.com/OpenWhispr/openwhispr/issues/1586) weren't reachable in the shipped app.

PR #1603, opened 2026-08-13
fix(linux): deliver notifications through the desktop notification daemon

### LINKED ISSUE #1599: Linux: Supporting notifications via notify-send instead of hand-rolled electron notifications like they exist currently.
Currently the way notifications work in this app is that they are relayed through the app itself. While that may work on other desktops, it breaks on wayland WMs like Niri and others too because those notifications either get consumed by the open application or they show up randomly in the center, which can be very annoying.

Instead of this, it'd be very sensible if we supported the native notify-send protocol (if detected) on a system. The details for this is here:  https://www.mankier.com/1/notify-send

This will allow us to consume custom inputs (like begin transcription during a meeting), while respecting the user's notification styles and location preferences, making OpenWhispr a more compliant and polished userspace app

No alternatives considered.

**Additional context**

The current notification popping up randomly in the middle and stealing focus.
Image

A proposed notification through notify-send, complying with my notification setting preferences.

Image

I'd be happy to begin work on this - please lmk if this'd be an improvement.

xAlcahest:
## Summary

The meeting prompt and the update notice are frameless always-on-top BrowserWindows, and Wayland compositors ignore client-side positioning, so on WMs like Niri the card pops up centered and steals focus. On Linux, `windowManager` now delivers both through `org.freedesktop.Notifications` on the session bus (the protocol behind notify-send) via the already-bundled `@homebridge/dbus-native`, with the new `linuxNotifier.js` helper mapping the existing semantics onto the daemon: a primary action button that is also the default click, user close as dismiss, expiry as the auto-dismiss timeout, `replaces_id` for the one-prompt-at-a-time rule. Any failure (no session bus, daemon without the `actions` capability, hung call) falls back to the current overlay, with a retry backoff (30s doubling up to 30min) so a daemon that was merely slow at login gets picked up again; one overall deadline bounds the whole native attempt, a broken connection expires live prompts so their owners recover, quitting closes any toast still on screen, and staleness guards keep near-simultaneous detections from double-posting. macOS, Windows and daemon-less Linux setups are untouched, and the existing notification preferences keep applying since they gate upstream of delivery.

One tradeoff to sign off: the meeting-detection docs say prompts use the in-app overlay specifically to survive DND and stay out of screen shares via content protection. Native daemon notifications honor DND and are visible to screen shares, which is what #1599 asks for (respect the user's notification preferences). If that is accepted the doc needs updating; if not, I can gate the native path to update notices only.

Fixes #1599

## Changes

- src/helpers/linuxNotifier.js: new helper speaking org.freedesktop.Notifications (capability check, action and close signal dispatch keyed by notification id, programmatic close, single-deadline calls, failure backoff) plus the localized prompt content builders reusing the overlay i18n keys.
- src/helpers/windowManager.js: showMeetingNotification and showUpdateNotification try the native path first with a staleness guard across the await, and keep the overlay as fallback; the dismiss paths also close the native notification.
- src/updater.js: the notification Update action lives in one method, downloadUpdateFromNotification, used by both the native path and the overlay IPC handler.
- src/helpers/ipcHandlers.js: update-notification-respond calls that shared method.
- src/helpers/linuxAutostart.js: exports LINUX_APP_NAME so the notifier reuses the packaged desktop-entry name instead of re-declaring it.
- main.js: stops the notifier during teardown.
- test/helpers/linuxNotifier.test.js: covers the Linux gating, fallback and backoff-retry paths, the shared deadline, Notify argument marshalling, signal dispatch, orphan recovery on connection loss, close-on-stop, body markup escaping and the content builders.

Verified against a live daemon (KDE Plasma): connection, capabilities, Notify with actions and hints, programmatic close. Clicking the action buttons and the behavior on Niri and GNOME still need a live desktop pass.

### DIFF
--- a/main.js
+++ b/main.js
@@ -259,6 +259,7 @@ process.on("unhandledRejection", (reason, promise) => {
 // Import helper module classes (but don't instantiate yet - wait for app.whenReady())
 const EnvironmentManager = require("./src/helpers/environment");
 const WindowManager = require("./src/helpers/windowManager");
+const { linuxNotifier } = require("./src/helpers/linuxNotifier");
 const DatabaseManager = require("./src/helpers/database");
 const ClipboardManager = require("./src/helpers/clipboard");
 const WhisperManager = require("./src/helpers/whisper");
@@ -1895,4 +1896,5 @@ function performSyncTeardown() {
   if (ipcHandlers) ipcHandlers._cleanupTextEditMonitor();
   if (textEditMonitor) textEditMonitor.stopMonitoring();
   if (updateManager) updateManager.cleanup();
+  linuxNotifier.stop();
 }
--- a/src/helpers/ipcHandlers.js
+++ b/src/helpers/ipcHandlers.js
@@ -9824,11 +9824,7 @@ class IPCHandlers {
     ipcMain.handle("update-notification-respond", async (_event, action) => {
       this.windowManager?.dismissUpdateNotification();
       if (action === "update") {
-        try {
-          await this.updateManager?.downloadUpdate();
-        } catch (error) {
-          console.error("Failed to start update download from notification:", error);
-        }
+        await this.updateManager?.downloadUpdateFromNotification();
       }
       return { success: true };
     });
--- a/src/helpers/linuxAutostart.js
+++ b/src/helpers/linuxAutostart.js
@@ -207,6 +207,7 @@ function syncAutostartEntry() {
 }
 
 module.exports = {
+  LINUX_APP_NAME,
   getDesktopFilePath,
   resolveExecutablePath,
   buildDesktopFileContents,
--- a/src/helpers/linuxNotifier.js
+++ b/src/helpers/linuxNotifier.js
@@ -1,0 +1,292 @@
+const debugLogger = require("./debugLogger");
+
+let dbus = null;
+try {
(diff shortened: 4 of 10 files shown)


Chadpiha (approved):
Approved. I reviewed the Linux notification flow end-to-end, and the native notification behavior, fallback path, dismissal handling, and retry logic all look sound. The full test and quality suites pass.

@iSparsh, since you opened the original ticket, could you also smoke-test this on your Linux setup? Please confirm that meeting notifications are clickable, Start/Join and Dismiss work, and dismissed notifications don’t reappear. It would also help to know which desktop environment and notification daemon you tested with. My approval isn’t blocked on this, but live Linux validation would be valuable.

iSparsh:
@Chadpiha @xAlcahest I did end up testing it for meeting notifications. Here's an image for what shows up by default:
image

seems like there are two take notes buttons which we need to fix. It might have to do with 
["default", actionLabel, actionKey, actionLabel] mapping to two pairs of Take Notes for the meeting notifications
We only need ["default", actionLabel] to be returned in the JS in `linuxNotifier.js`.

The Dismiss button is something that my preferences have, some folks just have a close icon. However we do need  to remove the duplicate Take Notes button (both of which work and do the same thing).

PR #1604, opened 2026-08-13
fix(agent): recognize localized wake words

xAlcahest:
## Problem

Found during the #1264 end to end testing on Linux, flagged in the engineering thread at the time. `detectAgentName` only treats the agent name as an address when it starts the dictation, opens a new sentence, or follows a vocative cue, and the cue set in `src/config/agentDetection.ts` was English only (hey, hi, hello, ok, okay, yo, please). Localized UIs advertise translated wake phrases ("Ehi {{agentName}}" in Italian, "Привет, {{agentName}}" in Russian, "ねぇ {{agentName}}" in Japanese), so a non-English dictation like "stavo pensando, ehi Jarvis scrivi una mail" can never take the agent route by voice: the whole utterance goes to cleanup and gets pasted as text.

## Solution

Additive only. `detectAgentName` gains an optional dictation language parameter and localized vocatives per language, matched with the exact same previous-token rule as the English cues. The English set and the position, distance and casing rules are untouched, and a test loops the same English utterance across languages to prove it.

The language is resolved at the call site by `resolveWakeWordLanguage` (pure, in `dictationRouting.js`): the dictation language when it is explicit, otherwise the UI language, which is the best available hint for what the user actually speaks under auto detect. Gating:

- language with a cue list: English cues plus that language's cues
- language without a cue list (e.g. `ko`): English cues only, same as before
- no usable hint: English cues only, fail closed

A fresh install (language auto, English UI) therefore behaves exactly like main: no localized cue is active, so English dictation containing loanwords ("talk later, ciao Jarvis remind me to call him") keeps getting pasted as text. An Italian-system install gets the Italian cues out of the box because the UI language defaults to the system locale. Translation recordings never route to the agent (`resolveDictationRouteKind` decides translation first), so the detection scan is skipped entirely for them.

Cue lists, taken from what each locale's settings page advertises plus natural vocatives of the language, kept short on purpose:

| Language | Cues | Notes |
|---|---|---|
| de | hallo, servus | UI advertises "Hey", already covered |
| es | oye, hola, oiga | UI advertises "Hey" |
| fr | hé, salut | UI advertises "Hey" |
| it | ehi, ei, ciao, scusa | UI advertises "Ehi {{agentName}}"; "ei" covers the STT spelling variant |
| ja | ねぇ, ねえ, ヘイ | UI advertises "ねぇ {{agentName}}" |
| pt | ei, olá | UI advertises "Ei {{agentName}}" |
| ru | привет, эй, слушай | UI advertises "Привет, {{agentName}}" |
| zh | 嘿, 你好, 喂 | UI advertises "Hey"; zh-CN and zh-TW share the set |

Candidates that double as common words elsewhere were rejected as false-positive magnets: German "bitte" ("frag bitte Jarvis" is not an address), French "dis" ("don't dis Jarvis" in English), German "he" ("is he Jarvis"). A guard test extracts the wake word each locale actually advertises in `howItWorksDescription` and asserts it triggers detection in that locale's language, so a future translation change cannot silently drift away from the cue table.

For ja and zh only, the transcript is normalized before tokenization: fullwidth punctuation becomes its ASCII equivalent plus a space, and CJK/Latin transitions get a space inserted, because CJK STT output is unsegmented ("ねぇJarvis、メールを書いて"). The normalization is gated together with the cues, so a dictation in any other language, including English transcripts that quote CJK text, is processed byte for byte as on main.

## Changes

- src/config/agentDetection.ts: localized cue table, language gating via `getBaseLanguageCode`, ja/zh transcript normalization, optional `language` parameter on `detectAgentName`
- src/helpers/dictationRouting.js: `resolveWakeWordLanguage` resolves the gating language from the dictation and UI languages
- src/helpers/audioManager.js: `resolveReasoningRoute` passes the resolved language and skips the scan for translation recordings
- test/helpers/agentDetection.test.js: localized cues, gating in both directions, false-positive guards, unsegmented CJK shapes, per-locale advertised wake phrase guard
- test/helpers/dictationRouting.test.js: `resolveWakeWordLanguage` cases

### DIFF
--- a/src/config/agentDetection.ts
+++ b/src/config/agentDetection.ts
@@ -1,3 +1,5 @@
+import { getBaseLanguageCode } from "../utils/languageSupport";
+
 function levenshteinDistance(a: string, b: string): number {
   const m = a.length;
   const n = b.length;
@@ -40,14 +42,14 @@ const LOCALIZED_VOCATIVE_CUES: Record<string, readonly string[]> = {
   zh: ["嘿", "你好", "喂"],
 };
 
-const ALL_LOCALIZED_CUES: ReadonlySet<string> = new Set(
-  Object.values(LOCALIZED_VOCATIVE_CUES).flat()
+const LOCALIZED_CUE_SETS: ReadonlyMap<string, ReadonlySet<string>> = new Map(
+  Object.entries(LOCALIZED_VOCATIVE_CUES).map(([lang, cues]) => [lang, new Set(cues)])
 );
 
 const EMPTY_CUES: ReadonlySet<string> = new Set();
 
-// CJK transcripts carry no spaces around punctuation ("ねぇ、Jarvis、メールを"),
-// so fullwidth marks become their ASCII equivalent plus a space before splitting.
+// CJK transcripts carry no spaces ("ねぇ、Jarvis、メールを"), so fullwidth marks
+// become their ASCII equivalent plus a space, and CJK/Latin transitions split.
 const CJK_PUNCTUATION_MAP: Record<string, string> = {
   "、": ", ",
   "。": ". ",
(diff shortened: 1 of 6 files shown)


Chadpiha (approved):
Reviewed and fixed 2 issues:
- Configured Unicode agent names now work correctly in unsegmented Japanese and Chinese transcripts without changing non-CJK matching behavior.
- Automatic language detection now prioritizes the STT-detected language over the UI locale when routing localized wake words.

The new regression and end-to-end tests cover both paths, and the full suite passes. Approved.

PR #1609, opened 2026-08-13
fix(gpu): stop 1.8.3's GPU regression — Vulkan iGPU default, off-PATH nvidia-smi, silent pack deletion (#1606)

### LINKED ISSUE #1606: After upgrading to 1.8.3, Windows GPU detection now fails for all previously-working versions of OpenWhispr
I have been using OpenWhispr with GPU acceleration for local dictation for many months now, over several versions, and it has always worked fine.

However, after updating OpenWhispr to 1.8.3, my GPU is now not detected by it or any previously-working version. It always falls back to CPU. I have tested this with all versions 1.8.3 through 1.7.6.

I have not changed my GPU driver. I have not made any other relevant changes to my Windows installation. The GPU continues to function as usual for llama.cpp, gaming, etc. The only change between past versions of OpenWhispr working and none of them working was updating OpenWhispr to 1.8.3.

I know you have a separate bit of GPU acceleration software that is downloaded when the feature is enabled. I had to re-download that when I updated to 1.8.3. Are you sharing the same GPU acceleration package across all versions of OpenWhispr? If so, a bug in that package would simultaneously break every past version.

Nvidia 4060 Ti 16GB
Windows 11 Pro 25H2

Here is the debug log of the failure for version 1.8.3.
[OpenWhispr_1.8.3_GPU_Detection_Bug.log](https://github.com/user-attachments/files/31028887/OpenWhispr_1.8.3_GPU_Detection_Bug.log)

greatcoat:
**Problem:** On multi-GPU Windows machines, 1.8.3 silently deleted old GPU packs, then offered the wrong replacement pack (nvidia-smi missing from PATH), and Vulkan defaulted to the slow integrated GPU.

**Fix:** Vulkan now restarts once pinned to the discrete GPU (persisted for later launches); NVIDIA detection falls back through System32, DriverStore, and WMI; and a one-time toast tells users to re-download cleared packs.

Fixes #1606.

Three fixes for the 1.8.3 GPU regression (reporter: RTX 4060 Ti + AMD 7900 XTX + Intel UHD 770, Win 11). Full RCA is on the issue (comment 5280781402); the key diagnostic — no `nvidia-smi` anywhere on PATH — was confirmed by the reporter.

## 1. Vulkan: pin transcription to a discrete GPU when device 0 is integrated

The Vulkan spawn path set no device, so ggml defaulted to Vulkan device 0 in loader order — on multi-GPU machines that's typically the iGPU (reporter's debug log: model loaded on `Vulkan0` = Intel UHD 770; a 3.9s clip took 8.2s). Now the server's `ggml_vulkan: N = <name> (<driver>) | uma: <0|1>` startup lines are parsed; if device 0 is UMA and a discrete device exists, the server restarts once pinned and the index is persisted (`WHISPER_VULKAN_DEVICE`) so later launches spawn pinned directly — this matters because Vulkan cold starts can run minutes (#698). A persisted pin that no longer resolves (hardware change) is cleared the same one-shot way. Unparseable stderr, single-GPU, discrete-first, and all-UMA machines all no-op.

**Why `--device N` instead of `GGML_VK_VISIBLE_DEVICES`:** verified against the pinned OpenWhispr/whisper.cpp tag — the env var takes raw *physical* enumeration indices, which diverge from the printed *logical* indices whenever a device is filtered out (lavapipe, dual-driver dedupe). `--device N` selects exactly the N-th printed device, and an out-of-range value degrades to CPU instead of throwing.

## 2. NVIDIA detection: find nvidia-smi off-PATH, fall back to WMI

The CUDA-pack offer gated on a bare `execFile("nvidia-smi")`. DCH/Windows-Update driver installs don't reliably put it on PATH (some ship it only in System32 or only inside the DriverStore), and nothing a user runs day-to-day needs it — so only our gate failed, silently offering the Vulkan pack to NVIDIA machines. Detection now climbs: PATH → `System32\nvidia-smi.exe` → newest `DriverStore\FileRepository\nv*\nvidia-smi.exe` → PowerShell `Get-CimInstance Win32_VideoController` as last resort (works with zero NVIDIA tooling; `wmic` is gone from Win 11). Spawn errors are logged instead of discarded, and `listNvidiaGpus` uses the resolved path so the per-GPU picker works on off-PATH machines too.

**`cudaSupported` on the WMI rung** (no compute cap available): a conservative name allowlist of families provably at or above the 6.1 kernel floor (RTX-branded, GTX 10xx/16xx, post-Maxwell TITANs). A false negative just gets Vulkan, which works on every NVIDIA GPU; a false positive would crash every transcription at the first kernel launch (#1585's floor rationale).

## 3. Notify when the legacy-layout migration clears a GPU pack

`migrateLegacyBinDir` (PR #1577) deliberately deletes lib-carrying legacy packs — but silently, which is why every old version broke at once with no hint. The migration now reports what it cleared; since it runs before any window exists, the names are persisted in a userData sentinel and the control panel shows a one-time toast (localized in all 9 languages) pointing at Settings → re-download, then clears the sentinel.

## Testing

- `npm test`: 1826 pass / 0 fail (169 skipped = the known local Electron-ABI DB skips); `npm run lint` + `npm run typecheck` clean.
- New/extended units: stderr device-line parser + pin decision table, `--device` arg building, every detection-ladder rung with injected exec/fs, the WMI name allowlist, migration return values, sentinel round-trip.
- No multi-GPU Windows hardware here — the reporter should confirm via debug log: first Vulkan start logs the pinned restart, the second launch spawns once with `--device N`, and Settings now offers the CUDA pack. Their reply also tells us which ladder rung catches their machine.

Follow-up (out of scope): a Vulkan device-picker UI matching the CUDA one (`TRANSCRIPTION_GPU_UUID` pattern).


gabrielste1n (approved):
Full investigation and review done: RCA verified against the 1.8.3 code, --device semantics and stderr device-line format verified against the pinned whisper.cpp 0.0.9 tag, and the device parser validated against the reporter's attached debug log (pins device 1, the RTX 4060 Ti). Detection ladder, migration notice, and env-var lifecycle all check out. 1996 tests / 0 failures, lint and typecheck clean.

PR #1610, opened 2026-08-13
fix(notes): persist diarization metadata on upload and URL-ingest notes

greatcoat:
## Problem

A note created through the file-upload / URL-ingest path runs speaker diarization but never persists the diarization metadata. Observed on 1.8.3: a 72.7-minute URL ingest with diarization enabled (`numSpeakers: 2`) produced a fully speaker-labelled note — 633 segments merged, labels covering the whole duration — whose row still read `diarization_enabled: null`, `expected_speaker_count: null`, `audio_duration_seconds: null`.

This is not a 1.8.3 regression: the gap has existed since upload speaker detection shipped in 1.7.6 (#754), and the duration column has been passed as `null` since the upload view was created (#261). 1.8.3 is merely where it was noticed.

Not cosmetic: `PersonalNotesView` derives `diarizationEnabled` from the column, and `isExplicitSpeakerCount()` (`src/utils/participants.ts`) resolves participants from `expected_speaker_count` — so an uploaded note that *was* diarized behaves differently from a meeting recording that went through the identical diarizer. All three columns also cloud-sync, so the CLI, API, and other devices see them.

Root cause: both upload save sites — `batchQueueStore.processItem` and `UploadAudioView.handleTranscribe` — called `saveNote(..., null, folderId)` with the diarization settings in scope and discarded them. The meeting path (`meetingRecordingStore`) persists both diarization columns via `updateNote`.

## Fix

Both upload flows (file **and** URL, batch **and** single) now persist the metadata:

- **`audio_duration_seconds` → initial `saveNote`.** The 5th positional arg and insert column already exist (the CLI bridge uses them); `updateNote` does **not** whitelist this column, so creation is its only writer. No API change.
- **`diarization_enabled` + `expected_speaker_count` → one `updateNote` immediately after creation, only when diarization ran.** These are not `saveNote` insert columns; they are whitelisted `updateNote` fields, the same route the meeting path writes them through. The write is skipped entirely for non-diarized uploads — the meeting path writes these columns only on explicit user action, so `null` retains its meaning of "user never chose". This matters because both consumers (`meetingRecordingStore`'s `initialEnabled ?? …` chain and `_resolveInitialMeetingSpeakerConfig` in main) fall back to the global speaker setting on `null` but treat `0` as a hard off: writing `0` would have silently forced speaker labeling off when recording into an uploaded note. Skipping the write also means non-diarized uploads (the common case) pay no second vector-index embed / mirror write / `note-updated` broadcast — `db-update-note` runs those unconditionally; its meeting-session side effects gate on `updates.participants`, which this payload never carries. The `updateNote` re-queues `sync_status: 'pending'`, which is correct — these fields cloud-sync.
- **Semantics for diarized uploads**: `diarization_enabled` is `1`, and the speaker count is only ever one the diarizer was actually invoked with — auto detection stores `null` so `isExplicitSpeakerCount` never mistakes it for a user choice, and a stale `numSpeakers` lingering in localStorage while its input is hidden can never be stamped onto a non-diarized note.

New helper `src/helpers/uploadNoteMetadata.js` holds the mapping. It carries a renderer twin of `normalizeStoredSpeakerCount` because the original in `src/helpers/speakerCount.js` is CommonJS for the main process (`database.js`, `ipcHandlers.js`) and renderer source cannot load CJS modules under Vite. The twin is **parity-locked by test**: `test/helpers/uploadNoteMetadata.test.js` sweeps both implementations and fails on any divergence.

Duration for picked files: the renderer never knows a local file's duration up front, but the `diarize-audio-file` handler already measured it (converted-wav size) and threw it away. It now returns `durationSeconds`, and `transcribeFileWithSpeakers` surfaces it on the result — a source-known duration (URL downloads) takes precedence. Deliberate non-change: `mergeSpeakerText` still receives only the caller-known duration, so transcript merging behaves exactly as before. Known remaining gap (pre-existing): a file upload whose local diarizer never runs (diarization off, or BYOK-native diarization) still has no duration source.

A meeting recording is unchanged — `meetingRecordingStore` is untouched.

## Tests

- `test/helpers/uploadNoteMetadata.test.js` — field mapping, diarized-only write semantics, auto-detection → null, disabled → no columns at all, clamp, duration degradation, and the parity sweep against `normalizeStoredSpeakerCount`.
- `test/services/uploadDiarizationPersistence.test.js` — regression through the real `batchQueueStore` + `fileTranscription` module graph (renderer test harness, `electronAPI` stubbed at the IPC boundary): a diarized **file upload** and a diarized **URL ingest** both end with `saveNote` receiving the measured duration and `updateNote` receiving `{ diarization_enabled: 1, expected_speaker_count: 2 }`; a **non-diarized upload** ends with no `updateNote` call at all. Each asserted failing on the code it guards against: the diarized tests fail pre-fix (duration `null`, no `updateNote` call), the non-diarized test fails on the intermediate code that wrote `diarization_enabled: 0`.
- Full suite: 1983 tests, 1977 pass, 0 fail, 5 skipped. `npm run typecheck` and `npm run lint` clean.
- Changelog entry added under `[Unreleased]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- **Uploaded and URL-ingested notes remember their speaker detection.** A note created through Upload ran speaker detection but stored none of it — the note now records that diarization ran, the speaker count you chose, and the audio duration, so it behaves like a meeting note when you record into it or resolve participants. An upload with speaker detection off writes nothing, preserving your global speaker setting. Present since upload speaker detection shipped in 1.7.6. (#1610)
+
 ## [1.8.3] - 2026-08-12
 
 GPU acceleration for local transcription gets an overhaul: the status you see is now the truth, enabling it works without a restart, older NVIDIA cards are routed to a backend that actually works on them, and a GPU failure can never cost you a dictation. LLM routing gains the same fail-closed treatment speech-to-text received in 1.8.2. Launch at login arrives on Linux, and transcription errors stop blaming your microphone.
--- a/src/components/notes/UploadAudioView.tsx
+++ b/src/components/notes/UploadAudioView.tsx
@@ -64,6 +64,7 @@ import { isTranscriptionContextAllowed } from "../../stores/policyRules";
 import { usePolicyStore } from "../../stores/policyStore";
 import { usePolicySnapshot, useTranscriptionContextAllowed } from "../../hooks/usePolicy";
 import { resolveTranscriptionRoute } from "../../helpers/transcriptionRoute";
+import { buildUploadNoteMetadata } from "../../helpers/uploadNoteMetadata";
 
 type UploadState = "idle" | "selected" | "downloading" | "transcribing" | "complete" | "error";
 
@@ -656,14 +657,15 @@ export default function UploadAudioView({ onNoteCreated, onOpenSettings }: Uploa
     }
 
     try {
+      const diarization: DiarizationSettings = {
+        enabled: diarizationEnabled,
+        localModelsReady: !!diarizationModelsReady,
+        numSpeakers: diarizationNumSpeakers ? Number(diarizationNumSpeakers) : null,
+      };
       const res: FileTranscriptionResult = await transcribeFileWithSpeakers(
         currentFile.path,
         buildTranscriptionConfig(),
-        {
-          enabled: diarizationEnabled,
-          localModelsReady: !!diarizationModelsReady,
-          numSpeakers: diarizationNumSpeakers ? Number(diarizationNumSpeakers) : null,
-        },
+        diarization,
         currentFile.durationSeconds,
         { requestId }
       ).finally(() => {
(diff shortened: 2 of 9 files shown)


Chadpiha (approved):
Fixed a stale, hidden speaker count that was persisted when diarization is off. Rest looks good!

gabrielste1n (approved):
Reviewed end to end: verified the positional saveNote contract at all four layers, the updateNote whitelist, cloud-sync payload coverage, and meeting-path parity. Confirmed the regression tests fail on pre-fix code. Traced the null/0/1 semantics through both consumers (meetingRecordingStore initialEnabled and _resolveInitialMeetingSpeakerConfig) — the final revision writes the diarization columns only when diarization ran, so a non-diarized upload keeps deferring to the global speaker setting when recorded into later. Full suite 1983/0 fail, typecheck and lint clean, CI green on all platforms.

gabrielste1n:
Thanks @greatcoat — sharp diagnosis, and writing the regression tests first (and showing them failing) made this straightforward to verify end to end. One follow-up landed on top before merge: the diarization columns are now written only when diarization actually ran, so a non-diarized upload keeps a null diarization_enabled and continues to defer to the global speaker setting when recorded into later (a stored 0 would have forced it off). Details in the updated PR description. Also folded in your stale-numSpeakers catch — good spot.

PR #1611, opened 2026-08-13
fix(reasoning): stop gpt-oss on Tinfoil failing every request with a 400

greatcoat:
## Problem

Tinfoil + any `gpt-oss-*` model is 100% broken: every request on all five LLM surfaces (Note Formatting, Dictation Cleanup, Chat, Voice Agent, Translation) returns:

```
400 Harmony does not support reasoning_effort='none'
```

Reproduced live on packaged 1.8.3 (Note Formatting → Tinfoil / `gpt-oss-120b`), and confirmed directly against `https://inference.tinfoil.sh/v1/chat/completions`:

| model | `reasoning_effort` | result |
|---|---|---|
| `gpt-oss-120b` | `none` | **400** |
| `gpt-oss-120b` | `low` | **200** |
| `gpt-oss-safeguard-120b` | `none` | **400** |
| `deepseek-v4-flash` | `none` | 200 |
| `llama3-3-70b` | `none` | 200 |

## Root cause

`suppressThinking()` keyed the gpt-oss exception on `providerKey === "groq"`. The constraint — gpt-oss accepts `low|medium|high` only, no off switch — is a property of the **model family**, not of Groq, so every other provider serving gpt-oss fell into the generic dialect and sent the rejected `"none"`. Tinfoil serves `gpt-oss-120b` and `gpt-oss-safeguard-120b`.

## Fix

Hoist the family check into a shared `isGptOss` predicate: the Groq branch keeps its existing behavior, and the generic dialect now sends `"low"` (the floor) instead of `"none"` for gpt-oss on any provider.

Deliberately untouched: the OpenRouter, `local`, and `lan` dialects. Each uses its own native suppression shape, and flat `reasoning_effort` would be wrong there (the `lan` comment already notes it trips vLLM). No behavior change for any non-gpt-oss model on any provider.

## Tests

Two new cases in `test/helpers/thinkingSuppressionDialects.test.js`, written first and watched fail with exactly the bug's value (`'none' !== 'low'`):
- generic-dialect provider (`tinfoil`) + `gpt-oss-120b` → `reasoning_effort: "low"` + `chat_template_kwargs`
- family match covers `gpt-oss-safeguard-120b` and is case-insensitive

Full suite: 1806 pass / 0 fail. Typecheck, eslint, prettier clean.

Linear: CUS-65

### DIFF
--- a/src/services/ai/thinkingSuppressionDialects.ts
+++ b/src/services/ai/thinkingSuppressionDialects.ts
@@ -32,6 +32,11 @@ export function suppressThinking(
   providerKey: string,
   model: string
 ): void {
+  const modelFamily = (model || "").toLowerCase();
+  // gpt-oss accepts low|medium|high only; it has no off switch. A property of the
+  // model family, not the provider — Tinfoil 400s on "none" just like Groq would.
+  const isGptOss = modelFamily.includes("gpt-oss");
+
   if (providerKey === "gemini") {
     requestBody.reasoning_effort = "minimal";
     return;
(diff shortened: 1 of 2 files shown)


gabrielste1n (approved):
Reviewed with full investigation: root cause verified (model-family constraint scoped to the groq provider branch since 8e394fad), fix is minimal and correct (shared isGptOss predicate, generic dialect floors gpt-oss at 'low'), Groq behavior unchanged, OpenRouter/local/lan dialects correctly left on their native shapes. Tests written fail-first and cover Tinfoil, the safeguard variant, and case-insensitivity. CI green across the board.

PR #1615, opened 2026-08-13
fix(notification): skip Google Calendar time blocks without attendees

IdrisGit:
Skip calendar reminders for Google events without attendees, unless they include a meeting link. This prevents personal time blocks from triggering meeting notifications while preserving reminders for link-based meetings.

This PR is based on the suggestion made by samkay on discord [message link](https://discord.com/channels/1463590782689612014/1463591135254413413/1536487421544370216)

### DIFF
--- a/src/helpers/calendarReminderScheduler.js
+++ b/src/helpers/calendarReminderScheduler.js
@@ -1,9 +1,14 @@
 const debugLogger = require("./debugLogger");
+const { getMeetingJoinUrl } = require("./meetingJoinUrl");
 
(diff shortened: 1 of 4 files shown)


Chadpiha (approved):
Thanks for this! I found one edge case: Google events with no attendees but a meeting link pasted into the location or description were still being filtered out. I pushed commit 8de08f65 to your branch to reuse the existing meeting-link extractor and added regression coverage for both fields. The calendar/reminder test suite passes. Otherwise, the approach looks good.

PR #1618, opened 2026-08-13
fix(translation): treat whitespace-only chain results as empty

### LINKED ISSUE #1616: Whitespace-only translation/cleanup results overwrite the original dictation
## Describe the bug

`resolveTranslatedText()` and `executeTranslationChain()` treat any truthy string as a usable result. `#1258` already guards `""` / `null` / missing `text`, but a whitespace-only result (`"   "`, NBSP) still replaces the original dictation.

The helper comment says an empty/missing result must preserve the input so a dictation is never overwritten with nothing. Whitespace-only output is empty for the user and should use the same guard.

## Affected component

- `src/helpers/translationChain.js` — `resolveTranslatedText`, `executeTranslationChain`
- Callers in `src/helpers/audioManager.js` (`processedText` / `finalText`)

## To Reproduce

On current `main` (`251e1ad0`):

```js
import { resolveTranslatedText, executeTranslationChain } from "./src/helpers/translationChain.js";

resolveTranslatedText("raw dictation", { text: "   " });
// actual: "   "
// expected: "raw dictation"

await executeTranslationChain({
  text: "hello world",
  cleanupReachable: true,
  shouldTranslate: false,
  runCleanup: async () => "   ",
  runTranslate: async () => "translated",
});
// actual: { text: "   ", ... }
// expected: { text: "hello world", ... }
```

## Expected behavior

Whitespace-only cleanup or translation results should be treated as empty: keep the previous text, and fire `onEmptyTranslate` when the translate step returns only whitespace.

## Actual behavior

`"   "` and `"\u00a0"` overwrite the transcript because they are truthy.

## Environment

- OpenWhispr `main` @ `251e1ad0`
- Node.js 24+

## User impact

A cleanup/translate model that returns only spaces (or a blank-looking reply) can paste/store blank text instead of the spoken dictation.

## Root-cause hypothesis

Truthiness checks (`if (cleaned)`, `chainResult?.text ? ...`) instead of a trim/usable-text check, leaving a gap after #1258.

## Proposed focused scope

- Treat trimmed-empty strings as empty in both helpers
- Preserve real non-empty results, including leading/trailing spaces on actual words
- Add regression tests for `"   "` and NBSP
- No routing or audio-pipeline changes

I intend to work on this.

hsusul:
## Summary

- Treat whitespace-only cleanup and chain results as empty so they cannot overwrite a dictation.
- Reuse the existing `normalizeComparableText()` check already used by the translate step.
- Add regression tests for `"   "` and NBSP.

## Problem

`#1258` guards `""` / `null` / missing `text`, but `"   "` and `"\u00a0"` are still truthy. Cleanup used `if (cleaned)` and `resolveTranslatedText()` used `chainResult?.text ? ...`, so a blank-looking model reply replaced the spoken text.

## Root cause

Truthiness checks instead of the same trim/collapse helper the translate step already uses.

## Fix

- Apply `normalizeComparableText()` before accepting a cleanup result.
- Apply the same check in `resolveTranslatedText()`.
- Leave real non-empty results (including padded words) unchanged.

## Behavior before / after

| Input | Before | After |
| --- | --- | --- |
| `resolveTranslatedText("raw", { text: "   " })` | `"   "` | `"raw"` |
| `resolveTranslatedText("raw", { text: "\u00a0" })` | `"\u00a0"` | `"raw"` |
| cleanup `"   "` and `shouldTranslate: false` | `"   "` | original text |
| cleanup `""` | original (unchanged) | original |
| cleanup `"cleaned"` | `"cleaned"` | `"cleaned"` |

## Scope / non-goals

- In scope: empty-result handling in `translationChain.js`.
- Out of scope: provider routing, audio pipeline, and translate-step unchanged-text detection (already normalized).

## Tests added

- cleanup whitespace-only still runs translate on the original text
- translate whitespace-only fires `onEmptyTranslate` and keeps cleaned text
- cleanup-only whitespace keeps the original dictation
- `resolveTranslatedText` treats `"   "` and NBSP as empty

## Validation

```text
node --version
# v24.9.0

npm ci --ignore-scripts

ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/translationChain.test.js
# 27 pass, 0 fail

npx prettier --check src/helpers/translationChain.js test/helpers/translationChain.test.js

git diff --check
# clean
```

`npm test` (`node --import tsx --test`) fails to load named exports from this ESM `.js` helper in this local Node 24.9 / tsx combination, including on unmodified `upstream/main`. The repository test command for this file is therefore exercised with `node --test` (same assertions). Typecheck/lint/i18n/renderer build were not required for this JS-only helper change.

- Pure helper; no Electron, models, or API keys.
- Extends #1258; does not change successful cleanup/translate behavior.

Fixes #1616

### DIFF
--- a/src/helpers/translationChain.js
+++ b/src/helpers/translationChain.js
@@ -34,7 +34,7 @@ export async function executeTranslationChain({
   if (cleanupReachable) {
     try {
       const cleaned = await runCleanup(out);
-      if (cleaned) out = cleaned;
+      if (normalizeComparableText(cleaned)) out = cleaned;
       // Cloud cleanup counts as cloud reasoning once its call succeeds, even if it
       // returned empty text.
       if (cleanupIsCloud) usedCloudReasoning = true;
@@ -64,5 +64,5 @@ export async function executeTranslationChain({
 // Keep the chain result only when it produced text; an empty/missing result
 // preserves the input so a dictation is never overwritten with nothing.
 export function resolveTranslatedText(previousText, chainResult) {
-  return chainResult?.text ? chainResult.text : previousText;
+  return normalizeComparableText(chainResult?.text) ? chainResult.text : previousText;
 }
--- a/test/helpers/translationChain.test.js
+++ b/test/helpers/translationChain.test.js
@@ -81,6 +81,51 @@ test("cleanup throws: onCleanupError fires, translate runs on original text", as
   assert.equal(result.text, "translated(raw)");
 });
 
+test("cleanup returns whitespace-only: text unchanged, translate still runs", async () => {
+  const { executeTranslationChain } = await load();
+
+  const result = await executeTranslationChain(
+    makeOpts({
+      runCleanup: async () => "   ",
+      runTranslate: async (currentText) => `translated(${currentText})`,
+    })
+  );
+
+  assert.equal(result.text, "translated(raw)");
+});
+
+test("translate returns whitespace-only: onEmptyTranslate fires, cleaned text kept", async () => {
+  const { executeTranslationChain } = await load();
+  let emptyCalled = false;
+
+  const result = await executeTranslationChain(
+    makeOpts({
+      runCleanup: async () => "cleaned",
+      runTranslate: async () => "\u00a0",
+      onEmptyTranslate: () => {
+        emptyCalled = true;
+      },
+    })
+  );
+
+  assert.equal(emptyCalled, true);
+  assert.equal(result.text, "cleaned");
+});
+
+test("shouldTranslate false: whitespace-only cleanup keeps the original text", async () => {
+  const { executeTranslationChain } = await load();
+
+  const result = await executeTranslationChain(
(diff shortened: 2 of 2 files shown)


Chadpiha (approved):
Thanks @hsusul! Tight scope, and the fix reuses the exact normalizeComparableText() check the translate step already had instead of inventing a new guard, so all three call sites (batch, streaming, and the Control Panel retry path) now behave consistently.

I verified it end-to-end in a clean worktree: the new tests fail on the pre-fix helper and pass with it, real results with padded whitespace are preserved unmodified, and the full suite is green on the branch (1999 tests, 0 failures). Nice touch pinning the translate step's existing whitespace behavior with a test rather than claiming it as a change — the PR description was honest and easy to review against.

Fixes [#1616](https://github.com/OpenWhispr/openwhispr/issues/1616). Approving — thank you!

Issue #178, opened 2026-02-01
Groq Qwen-32B Model Leaks Internal `<think>` Reasoning in Output


faishal882:
## Describe the Bug
When using a **Groq API key** with the **Qwen-32B** model, the assistant includes internal reasoning / meta text (for example, `<think> ... </think>`) directly in the final response. This internal reasoning should not be exposed to end users.

---

## To Reproduce

1. Configure the application with a **Groq API key**
2. Select the **Qwen-32B** model
3. Submit a standard prompt (e.g., text cleanup or summarization)
4. Observe that the response includes `<think>` blocks containing internal reasoning

---

## Expected Behavior
The assistant should return **only the final user-facing response**, without any internal reasoning, meta commentary, or `<think>` tags.

---

## Actual Behavior
The response includes chain-of-thought style content such as:

```text
<think>
Okay, let's see. The user is talking about the architecture of an agent...
</think>

actual transcript
```<think>
Okay, let's see. The user is talking about the architecture of an agent, planning to make it recursive with two structures: deterministic and non-deterministic. The deterministic agent will verify, and the AI explorer will find all the fields.

First, I need to clean up the transcribed speech. There are a lot of filler words like "so" and "so on." I should remove those. Also, "we are discussing regarding" can be simplified to "discussing the architecture." The sentence structure is a bit run-on, so breaking it into shorter sentences or paragraphs would help. 

The user mentioned "deterministic and non deterministic" – I should add a hyphen between "non" and "deterministic" for correctness. The part about the deterministic agent verifying and the AI explorer finding fields needs to be clear. Maybe rephrase "there is going to be" to "the deterministic agent will verify" for conciseness. 

I should check if the user is addressing me. They said "Hey Ops," but my name is Orb. Since it's a typo, I'll treat it as normal content and correct it to "Orb" if necessary. Wait, the user wrote "Hey Ops," which isn't my name. So it's just part of the speech. I'll remove the "Hey Ops" since it's an address to someone else, not me. 

Also, the user might be using "so on" incorrectly. Maybe they meant "so on and so forth," but in the context, it's better to just remove the filler. 

Putting it all together, the cleaned-up version should be concise, with proper grammar and structure. Let me make sure the technical terms are correct and the flow is logical. The final output should clearly state the two structures and their roles without the filler words.
</think>

[...]

PR #1619, opened 2026-08-13
fix(reasoning): strip nested think blocks without leftover tags

### LINKED ISSUE #1617: Nested <think> blocks leave leftover </think> tags in stripped model output
## Describe the bug

`stripThinkingTags()` removes `<think>...</think>` with a single non-greedy replace. Nested blocks such as `<think>a<think>b</think>c</think>Answer` only strip the inner pair, leaving `c</think>Answer`.

The helper contract is to strip think blocks (including unterminated ones) so reasoning never reaches the user. #178 covered leaking a single well-formed block; nested leftovers are still unhandled. The same regex is duplicated in `localReasoningBridge.js`.

## Affected component

- `src/helpers/stripThinking.js` — `stripThinkingTags`
- `src/services/localReasoningBridge.js` (inline copy)
- Used by `ReasoningService` for note titles / non-streamed reasoning

## To Reproduce

On current `main` (`251e1ad0`):

```js
import { stripThinkingTags } from "./src/helpers/stripThinking.js";

stripThinkingTags("<think>a<think>b</think>c</think>Answer");
// actual: "c</think>Answer"
// expected: "Answer"
```

Existing well-formed and unterminated cases still work.

## Expected behavior

All think blocks, including nested ones, are removed. The visible answer remains. Unterminated trailing `<think>...` continues to be stripped.

## Actual behavior

The first replace eats `<think>a<think>b</think>`. The leftover `c</think>Answer` has no opening tag for the unterminated regex, so `</think>` is pasted into notes/titles.

## Environment

- OpenWhispr `main` @ `251e1ad0`
- Node.js 24+

## User impact

Reasoning models that nest think tags (or emit an extra inner block) can leak `</think>` and inner reasoning fragments into titles, notes, and cleanup output.

## Root-cause hypothesis

A single non-greedy closed-tag replace cannot peel nested pairs. The local-bridge copy has the same pattern.

## Proposed focused scope

- Repeat closed-tag stripping until stable, then strip unterminated tails
- Reuse the shared helper from `localReasoningBridge.js`
- Add a nested-block regression test
- Do not change streaming think-suppression dialects or provider APIs

I intend to work on this.

hsusul:
## Summary

- Strip nested `<think>` blocks by removing innermost closed pairs first.
- Reuse the shared helper from the local reasoning bridge instead of a one-shot regex copy.
- Add regression tests for nested, doubly nested, and sequential think blocks.

## Problem

`stripThinkingTags()` used a single non-greedy `/<think>[\s\S]*?<\/think>/g` replace. For `<think>a<think>b</think>c</think>Answer` that matches the first open tag through the first close tag, leaving `c</think>Answer`. The unterminated-tail regex then does not match, so `</think>` leaks into titles and notes.

#178 covered leaking a well-formed single block. Nested leftovers were still unhandled. The same one-shot regex was duplicated in `localReasoningBridge.js`.

## Root cause

A one-pass non-greedy replace cannot peel nested pairs. Global `/g` also cannot be looped safely because `lastIndex` makes later passes miss remaining tags.

## Fix

- Repeatedly remove the innermost closed `<think>...</think>` pair (a block that does not itself contain `<think>`).
- Then strip a trailing unterminated `<think>...` as before.
- Load `stripThinkingTags` from the local bridge so both paths stay in sync.

## Behavior before / after

| Input | Before | After |
| --- | --- | --- |
| `<think>a<think>b</think>c</think>Answer` | `c</think>Answer` | `Answer` |
| `<think>let me reason</think>Meeting Notes Summary` | `Meeting Notes Summary` | unchanged |
| `Title Here<think>never closed` | `Title Here` | unchanged |
| `<think>one</think>Keep<think>two</think>This` | `KeepThis` | unchanged |

## Scope / non-goals

- In scope: `stripThinkingTags` and the local-bridge call site.
- Out of scope: streaming think-suppression dialects, provider APIs, case-insensitive tag matching, and treating a mention of `<think>` in prose as non-reasoning.

## Tests added

- nested think blocks do not leave `</think>`
- doubly nested think blocks collapse to the answer
- sequential (non-nested) blocks still strip

## Validation

```text
node --version
# v24.9.0

ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/stripThinking.test.js
# 9 pass, 0 fail

npx prettier --check src/helpers/stripThinking.js src/services/localReasoningBridge.js test/helpers/stripThinking.test.js

git diff --check
# clean
```

`npm test` uses `node --import tsx`, which in this local Node 24.9 / tsx combo fails to bind named exports from typeless ESM `.js` helpers (also on unmodified `upstream/main`). This file is exercised with `node --test`.

- Pure string helper; no Electron, models, or API keys.

Fixes #1617

### DIFF
--- a/src/helpers/stripThinking.js
+++ b/src/helpers/stripThinking.js
@@ -1,9 +1,14 @@
 // Strip <think>...</think> blocks (including unterminated ones) emitted by
-// reasoning models. Mirrors the regex in localReasoningBridge.js.
+// reasoning models. Innermost closed pairs are removed first so nested blocks
+// cannot leave a stray </think>, then a trailing unterminated <think> is removed.
 export function stripThinkingTags(text) {
   if (typeof text !== "string") return text;
-  return text
-    .replace(/<think>[\s\S]*?<\/think>/g, "")
-    .replace(/<think>[\s\S]*$/, "")
-    .trim();
+  const innermostClosed = /<think>(?:(?!<think>)[\s\S])*?<\/think>/;
+  let out = text;
+  let next = out.replace(innermostClosed, "");
+  while (next !== out) {
+    out = next;
+    next = out.replace(innermostClosed, "");
+  }
+  return next.replace(/<think>[\s\S]*$/, "").trim();
 }
--- a/src/services/localReasoningBridge.js
+++ b/src/services/localReasoningBridge.js
@@ -49,10 +49,7 @@ class LocalReasoningService {
       const result = await modelManager.runInference(modelId, text, inferenceConfig);
       const stripThinking = config.disableThinking !== false;
       const cleanResult = stripThinking
-        ? result
-            .replace(/<think>[\s\S]*?<\/think>/g, "")
-            .replace(/<think>[\s\S]*$/, "")
-            .trim()
+        ? (await import("../helpers/stripThinking.js")).stripThinkingTags(result)
         : result.trim();
 
       const processingTime = Date.now() - startTime;
--- a/test/helpers/stripThinking.test.js
+++ b/test/helpers/stripThinking.test.js
@@ -35,3 +35,21 @@ test("non-string input is returned as-is", async () => {
   const { stripThinkingTags } = await load();
   assert.equal(stripThinkingTags(undefined), undefined);
 });
+
+test("peels nested think blocks instead of leaving a stray close tag", async () => {
+  const { stripThinkingTags } = await load();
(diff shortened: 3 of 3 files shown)


Chadpiha (approved):
Thanks @hsusul — tight repro, focused fix, good tests. Approving.

Verified end-to-end: reproduced the leak on main and confirmed it's fixed; fuzz-tested the peeling loop against a stack-based reference (20k cases, no mismatches, no pathological regex behavior); confirmed the await import() resolves in the Electron 41 main process and the helper ships in the packaged asar; full suite green under Node 24.

PR #1632, opened 2026-08-14
fix(stt): stop truncating custom dictionary prompts on models without Groq's limit

greatcoat:
**Problem:** The direct-API STT path silently truncates every provider's custom-dictionary prompt at 900 chars — a Groq-only limit — so large dictionaries lose ~half their words on OpenAI's 4o transcribe models.

**Fix:** Per-provider prompt budgets in one helper (Groq 890, Whisper-family 900 on both the direct-API and local whisper-server paths, an 8,000-char context guard for the 4o transcribe family), always keeping the head of the list; DictionaryView warns when the list outgrows the Groq/Whisper window.

## Root cause

`5fb8d90e` (2026-03-09) added the cap to respect Groq's documented 896-char prompt limit, but applied it to **all** providers on the direct-API path because the cap predates the per-provider route resolver — there was nowhere to hang a per-provider limit at the time.

- Groq: 896-char limit is real (890 keeps UTF-16 margin) — cap kept.
- `whisper-1` (and Whisper-family): the decoder reads at most 224 prompt tokens (~900 chars), so the cap is harmless there and keeps head-of-list priority deterministic — cap kept.
- `gpt-4o-transcribe` / `gpt-4o-mini-transcribe`: 16k-token context window, no documented prompt limit. **Verified live** (canary key, real speech clip): HTTP 200 with prompts of 1.6k / 4k / 12k / 40k chars and `usage.text_tokens` confirming the whole prompt is consumed; the only failure is `input_too_large` at ~80k chars (the 16k context). Our own API (`openwhispr-api/lib/providers/openai.ts`) has forwarded the prompt uncapped to `gpt-4o-mini-transcribe` since 2026-02-18. So: Groq-era 900 removed, replaced by a generous 8,000-char guard that leaves context for long dictations.

Reported by a Pro customer (CUS-49) who found in his debug log: `Custom dictionary prompt truncated { originalLength: 1633, truncatedLength: 899, maxChars: 900 }` — ~100 dictionary words, roughly half never reached the recognizer. Every other STT path (local Whisper, OpenWhispr cloud, Tinfoil/Mistral/xAI proxies, realtime keyterms) already sends the dictionary whole or shapes it per-provider; only the direct-API path truncated.

## Fix details

- `src/utils/dictionaryPromptCap.js` (new): the budgets + `trimDictionaryPrompt` (cut at the last comma inside the budget, head kept). Used by the direct-API path, the local whisper-server path, and DictionaryView so all three agree.
- `audioManager.js` direct-API path: `dictionaryPromptLimit({provider, endpoint, model})` → 890 Groq / 900 Whisper-family / 8,000 otherwise.
- `audioManager.js` local Whisper path: previously sent the full list; whisper.cpp keeps only the **last** 224 tokens (`prompt_past.end() - n_take`, "will use only the last N tokens" — same code upstream and in the fork's pinned 0.0.8), so the bottom of the dictionary won locally while the top won on the direct-API path. Now head-trimmed to 900 like everywhere else, so the UI notice is true on every Whisper path.
- `DictionaryView.tsx`: passive notice when the comma-joined hint list (words + snippet triggers — the exact string the request builds) exceeds 890 chars, telling users that Groq/Whisper read only the first ~900 and words nearer the top win. New `dictionary.promptLimitNotice` key added to all 10 locales; `check-i18n` passes.

**Deliberate non-changes:** the OpenWhispr-cloud, Tinfoil/Mistral/xAI proxy and realtime keyterm paths are untouched (cloud already runs uncapped against 4o-mini server-side; proxies shape per-provider). A custom URL pointing at Groq is still caught by the `api.groq.com` endpoint check. Prompt *framing* is unchanged — live probes show the 4o family largely ignores a bare comma list (0/8 spelling hits even for a single word) but follows an instruction-style prompt (8/8 on a short list); that is a separate, larger change that needs its own eval and a check against the dictionary-echo filter.

## Tests

New `test/helpers/audioManagerDictionaryPromptCap.test.js` (renderer harness, captured fetch, asserts on the actual `FormData` prompt field):

- RED before the fix with the customer's exact signature — a 1633-char dictionary reached fetch as 899 chars (47 of 100 terms) on `gpt-4o-mini-transcribe`.
- GREEN after: 4o gets the full prompt; Groq stays ≤890 ending on a whole entry; `whisper-1` stays ≤900 keeping the head.
- Follow-up commit adds two more RED-first cases: a ~13k-char list on `gpt-4o-transcribe` is held to ≤8,000 (and >7,000 — the guard is generous, not the old 900); the local whisper-server path receives ≤900 chars, head of the list, cut on a whole entry.

Full suite: 1891 pass / 0 fail (169 local Electron-ABI skips, expected). `tsc --noEmit`, eslint, prettier, `check-i18n` all clean.


boseq:
Non-blocking: `isWhisperModel` infers the model family from the model/deployment name, which may be arbitrary for Azure or self-hosted endpoints. We can do this as a follow-up.

The direct OpenAI/Groq fix looks good to merge.

boseq (approved):

greatcoat:
@Chadpiha could we also get this one merged in please 🙏

PR #1442, opened 2026-08-02
fix(dictionary): normalize untrimmed agent names in agentNameDictionaryChanges

hsusul:
## Summary

This PR normalizes `newName` and `oldName` by trimming surrounding whitespace before computing dictionary additions and removals in `agentNameDictionaryChanges`.

Fixes #1441

## Problem & Root Cause

`agentNameDictionaryChanges(dictionary, newName, oldName)` trimmed `newName` when building `add`, but evaluated `oldName !== newName` and `words.includes(oldName)` without trimming `oldName` or comparing trimmed values.

This led to two correctness bugs:

1. **Accidental Agent Name Removal:** Updating or saving an agent name with surrounding whitespace (e.g., `oldName = "MyAgent"`, `newName = " MyAgent "`) caused `oldName !== newName` to be `true` and `words.includes("MyAgent")` to be `true`. `add` evaluated to `[]` while `remove` returned `["MyAgent"]`, completely deleting the agent's name from the custom dictionary.
2. **Orphaned Dictionary Entries:** Renaming an agent whose `oldName` had surrounding whitespace (e.g. `oldName = " MyAgent "`, `newName = "Jarvis"`) caused `words.includes(" MyAgent ")` to fail because dictionary entries are stored trimmed (`"MyAgent"`). `remove` returned `[]`, leaving `"MyAgent"` orphaned in SQLite on rename.

## Fix

- Normalize `newName` and `oldName` using `trim()` before equality comparisons and dictionary membership checks in `agentNameDictionaryChanges`.
- Added unit tests covering untrimmed `oldName` and whitespace-only name updates.

## Behavior Before & After

- **Before:**
  - `agentNameDictionaryChanges(["MyAgent"], " MyAgent ", "MyAgent")` returned `{ add: [], remove: ["MyAgent"] }`.
  - `agentNameDictionaryChanges(["MyAgent", "Alice"], "Jarvis", " MyAgent ")` returned `{ add: ["Jarvis"], remove: [] }`.
- **After:**
  - `agentNameDictionaryChanges(["MyAgent"], " MyAgent ", "MyAgent")` returns `{ add: [], remove: [] }`.
  - `agentNameDictionaryChanges(["MyAgent", "Alice"], "Jarvis", " MyAgent ")` returns `{ add: ["Jarvis"], remove: ["MyAgent"] }`.

## Validation

All repository checks executed successfully:

- `node --test test/helpers/agentNameDictionary.test.js` (9/9 passed)
[...]

PR #1639, opened 2026-08-14
fix(dictionary): treat agent names as present ignoring case

### LINKED ISSUE #1637: Agent name dictionary sync adds a case-variant duplicate instead of treating the name as already present
## Describe the bug

`agentNameDictionaryChanges()` decides whether to add/remove the agent name with `Array.includes()`, which is case-sensitive. `DictionaryView` already dedupes imports case-insensitively. If the dictionary has `openwhispr` and the agent is named `OpenWhispr`, startup/`ensureAgentNameInDictionary` asks to **add** `OpenWhispr`, producing two entries that differ only by case.

On rename, the old name is also looked up case-sensitively, so `openwhispr` is **not removed** when `oldName` is `OpenWhispr`.

## Affected component

- `src/helpers/agentNameDictionary.js` — `agentNameDictionaryChanges`
- Callers: `src/utils/agentName.ts` (`syncAgentNameToDictionary`, `ensureAgentNameInDictionary`)

## To Reproduce

On current `main` (`5a6c7f9b`):

```js
const { agentNameDictionaryChanges } = require("./src/helpers/agentNameDictionary.js");

agentNameDictionaryChanges(["openwhispr", "Alice"], "OpenWhispr");
// actual:   { add: ["OpenWhispr"], remove: [] }
// expected: { add: [], remove: [] }

agentNameDictionaryChanges(["openwhispr"], "Jarvis", "OpenWhispr");
// actual:   { add: ["Jarvis"], remove: [] }
// expected: { add: ["Jarvis"], remove: ["openwhispr"] }
```

## Expected behavior

A dictionary word that matches the agent name ignoring case is already present: do not add a second spelling. When renaming away, remove the stored word even if its capitalization differs from `oldName`.

## Actual behavior

Case-sensitive `includes()` treats `openwhispr` and `OpenWhispr` as different words.

## Environment

- OpenWhispr `main` @ `5a6c7f9b`
- Node.js 24+

## User impact

The custom dictionary and Whisper prompt accumulate duplicate agent-name hints. Renaming the agent can leave the old spelling behind if only the case differs.

## Root-cause hypothesis

#1442 trimmed whitespace in this helper but membership stayed `words.includes(trimmedNew)` / `words.includes(trimmedOld)`.

## Proposed focused scope

- Case-insensitive membership for add/remove
- Remove the **stored** spelling, not the `oldName` casing
- Do not recase an existing matching entry
- Regression tests for add-no-op and rename-remove
- No dictionary UI or SQLite schema changes

I intend to work on this.

hsusul:
## Summary

- Treat agent-name dictionary membership as case-insensitive so `openwhispr` and `OpenWhispr` are the same word.
- On rename, remove the **stored** spelling even when it differs only by capitalization from `oldName`.
- Do not recase an existing matching entry.

## Problem

`agentNameDictionaryChanges()` used `Array.includes()`, which is case-sensitive. `DictionaryView` already dedupes imports with `toLowerCase()`. If the table has `openwhispr` and the agent is `OpenWhispr`, `ensureAgentNameInDictionary` adds a second entry. Renaming away with `oldName` `OpenWhispr` also fails to delete `openwhispr`.

#1442 trimmed whitespace in this helper; membership stayed exact-string.

## Root cause

Case-sensitive `includes()` for both add and remove, and remove used `oldName`'s casing instead of the dictionary's stored word.

## Fix

- Look up words with trim + case-insensitive compare.
- Add only when no stored match exists.
- Remove the stored word when renaming to a different name (ignoring case).

## Behavior before / after

| Call | Before | After |
| --- | --- | --- |
| `(["openwhispr"], "OpenWhispr")` | `{ add: ["OpenWhispr"] }` | `{ add: [] }` |
| `(["openwhispr"], "Jarvis", "OpenWhispr")` | `{ add: ["Jarvis"], remove: [] }` | `{ add: ["Jarvis"], remove: ["openwhispr"] }` |
| `(["OpenWhispr"], "openwhispr", "OpenWhispr")` | `{ add: ["openwhispr"], remove: ["OpenWhispr"] }` | no-op |
| `(["OpenWhispr"], "OpenWhispr")` | no-op | no-op |

## Scope / non-goals

- In scope: `agentNameDictionaryChanges` delta logic.
- Out of scope: dictionary UI, SQLite schema, echo filtering.

## Tests added

- no add for a case-variant already present
- rename removes the stored spelling
- capitalization-only change does not recase/remove

## Validation

```text
node --version
# v24.9.0

ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/agentNameDictionary.test.js
# 12 pass, 0 fail

npx prettier --write src/helpers/agentNameDictionary.js test/helpers/agentNameDictionary.test.js
git diff --check
# clean
```

- Pure helper; no Electron, models, or API keys.

Fixes #1637

### DIFF
--- a/src/components/DictionaryView.tsx
+++ b/src/components/DictionaryView.tsx
@@ -37,10 +37,12 @@ export default function DictionaryView() {
 
   const pendingImportCount = useMemo(() => parseDictionaryImportText(bulkText).length, [bulkText]);
 
-  const userWords = useMemo(
-    () => customDictionary.filter((w) => w !== agentName),
-    [customDictionary, agentName]
-  );
+  // Same membership rule as agentNameDictionaryChanges: a stored spelling that
+  // differs only by case is still the agent name's entry, so keep it hidden.
+  const userWords = useMemo(() => {
+    const agentWord = agentName.trim().toLowerCase();
+    return customDictionary.filter((w) => w.trim().toLowerCase() !== agentWord);
+  }, [customDictionary, agentName]);
 
   const searchQuery = newWord.trim().toLowerCase();
   const visibleWords = useMemo(
--- a/src/helpers/agentNameDictionary.js
+++ b/src/helpers/agentNameDictionary.js
@@ -11,14 +11,20 @@
  * @param {string} [oldName]
  * @returns {{ add: string[], remove: string[] }}
  */
+function findStoredWord(words, word) {
+  const needle = word.toLowerCase();
+  return words.find((w) => typeof w === "string" && w.trim().toLowerCase() === needle);
+}
+
 export function agentNameDictionaryChanges(dictionary, newName, oldName) {
   const words = Array.isArray(dictionary) ? dictionary : [];
   const trimmedNew = typeof newName === "string" ? newName.trim() : "";
   const trimmedOld = typeof oldName === "string" ? oldName.trim() : "";
+  const storedNew = trimmedNew ? findStoredWord(words, trimmedNew) : undefined;
+  const storedOld = trimmedOld ? findStoredWord(words, trimmedOld) : undefined;
 
   return {
-    add: trimmedNew && !words.includes(trimmedNew) ? [trimmedNew] : [],
-    remove:
-      trimmedOld && trimmedOld !== trimmedNew && words.includes(trimmedOld) ? [trimmedOld] : [],
+    add: trimmedNew && !storedNew ? [trimmedNew] : [],
+    remove: storedOld && storedOld.toLowerCase() !== trimmedNew.toLowerCase() ? [storedOld] : [],
   };
 }
--- a/src/helpers/agentNameDictionary.js
+++ b/src/helpers/agentNameDictionary.js
@@ -1,3 +1,8 @@
+function findStoredWord(words, word) {
+  const needle = word.toLowerCase();
+  return words.find((w) => typeof w === "string" && w.trim().toLowerCase() === needle);
+}
+
 /**
  * Work out which dictionary changes an agent name requires: drop a
  * renamed-away name, add the current one.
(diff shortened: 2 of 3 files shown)


Chadpiha (approved):
Thanks for the submission @hsusul — nice find on #1637, and the focused scope + regression tests made this easy to review. I traced the fix end-to-end (helper → `agentName.ts` → `updateCustomDictionary` → SQLite `applyDictionaryChanges`) and the core change is correct: startup no longer asks to add a case-variant, and rename now removes the stored spelling.

I pushed 949ddfb3 to your branch with three small follow-ups from review:

1. **Moved `findStoredWord` above the JSDoc block.** As inserted, it sat between the doc comment and `agentNameDictionaryChanges`, so the `@returns {{ add, remove }}` annotation attached to `findStoredWord` instead. TypeScript then typed `storedOld` as a delta object and `npm run typecheck` failed (`agentName.ts(20,61): TS2322`) — CI runs typecheck, so this would have gone red.

2. **Trimmed the stored word in the remove comparison.** `findStoredWord` matches on `w.trim().toLowerCase()`, but the remove check compared the untrimmed stored word, so a padded stored spelling (e.g. `" OpenWhispr "`) matching the agent name produced `{ add: [], remove: [" OpenWhispr "] }` on an unchanged-name save — silently deleting the entry. Now both sides use the same normalization, with two regression tests covering the padded no-op and padded rename-away cases.

3. **Made DictionaryView's agent-name filter case-insensitive.** The word list hides the agent name with `w !== agentName`. Before this fix a case-variant got recased to an exact match on first sync, so the filter held; now the stored spelling persists by design, which left it visible as an ordinary deletable user word (and included in "remove all"). The filter now uses the same trim + case-insensitive rule as the helper.

Verified on the branch: 14/14 helper tests, full suite green

PR #1640, opened 2026-08-14
fix(notes): strip wrapping quotes from generated titles

### LINKED ISSUE #1638: Generated note titles keep wrapping quotes, including curly quotes
## Describe the bug

`generateNoteTitle()` asks the model for title text with no quotes, then strips only ASCII `"` / `'` at the very start **or** end of the string (`/^["']|["']$/g`). Models often wrap the title in curly quotes (`“…”` / `‘…’`) or in more than one layer of quotes. Those survive and become the note title.

## Affected component

- `src/utils/generateTitle.ts` — `generateNoteTitle`
- Callers: note formatting / upload / batch title generation

## To Reproduce

On current `main` (`5a6c7f9b`), the sanitizer is:

```js
raw.trim().replace(/^["']|["']$/g, "")
```

```js
"Weekly standup".trim().replace(/^["']|["']$/g, "")
// "Weekly standup"  (ASCII, OK)

"“Weekly standup”".trim().replace(/^["']|["']$/g, "")
// "“Weekly standup”"  (curly quotes remain)

"\"“Weekly standup”\"".trim().replace(/^["']|["']$/g, "")
// "“Weekly standup”"  (inner curly quotes remain)
```

## Expected behavior

Wrapping quotation marks (ASCII, curly, guillemets), including stacked wrappers, are removed. The stored title is the words the model produced. Inner apostrophes (`Don't stop`) stay.

## Actual behavior

Only a single ASCII quote at each end is removed. Curly quotes and extra wrapping layers remain.

## Environment

- OpenWhispr `main` @ `5a6c7f9b`
- Node.js 24+
- Cross-platform (pure string sanitize)

## User impact

Auto-named notes show `"Weekly standup"` or `“Weekly standup”` in the notes list, despite the prompt forbidding quotes.

## Root-cause hypothesis

The post-process regex is too narrow compared with the prompt contract (“no quotes”).

## Proposed focused scope

- Peel wrapping quote characters from both ends (ASCII + common typographic quotes)
- Keep the length guard (`> 0` and `< 100`)
- Unit-test the sanitizer without calling a model
- No prompt, provider, or routing changes

I intend to work on this.

hsusul:
## Summary

- Peel wrapping quotes from generated note titles, including curly quotes, guillemets, and stacked wrappers.
- Keep inner apostrophes (`Don't`) and the existing empty / over-100-character rejection.

## Problem

`generateNoteTitle()` asks the model for title text with no quotes, then strips only a single ASCII `"` / `'` at the start or end (`/^["']|["']$/g`). Models often wrap titles in `“…”` / `‘…’` or more than one layer of quotes. Those wrappers become the stored title.

## Root cause

The post-process regex is narrower than the prompt contract. It does not treat typographic quotes as wrappers and it only removes one ASCII quote per end.

## Fix

- Extract `sanitizeGeneratedTitle()` and peel ASCII / curly / guillemet wrappers from both ends in a loop.
- Preserve inner apostrophes and the length guard.

## Behavior before / after

| Input | Before | After |
| --- | --- | --- |
| `"Weekly standup"` | `Weekly standup` | `Weekly standup` |
| `“Weekly standup”` | `“Weekly standup”` | `Weekly standup` |
| `"“Weekly standup”"` | `“Weekly standup”` | `Weekly standup` |
| `'Don't stop meeting'` | `Don't stop meeting` | `Don't stop meeting` |

## Scope / non-goals

- In scope: title sanitizer used after the model returns text.
- Out of scope: prompt text, providers, routing, notes UI.

## Tests added

- ASCII wrappers
- curly quotes and guillemets
- stacked mixed wrappers
- inner apostrophes
- surrounding whitespace
- empty / overlong / non-string input

## Validation

```text
node --version
# v24.9.0

ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/generateTitle.test.js
# 7 pass, 0 fail

npx prettier --write src/utils/generateTitle.ts src/utils/sanitizeGeneratedTitle.ts test/helpers/generateTitle.test.js
git diff --check
# clean
```

- Pure string helper; no Electron, models, or API keys.

Fixes #1638

### DIFF
--- a/src/utils/generateTitle.ts
+++ b/src/utils/generateTitle.ts
@@ -1,6 +1,7 @@
 import reasoningService from "../services/ReasoningService";
 import type { ReasoningConfig } from "../services/BaseReasoningService";
 import { getSettings } from "../stores/settingsStore";
+import { sanitizeGeneratedTitle } from "./sanitizeGeneratedTitle";
 
 const TITLE_SYSTEM_PROMPT =
   "Generate a concise 3-8 word title for these notes. Return ONLY the title text, nothing else — no quotes, no prefix, no explanation.";
@@ -18,8 +19,7 @@ export async function generateNoteTitle(
       disableThinking: getSettings().noteFormattingDisableThinking,
       ...config,
     });
-    const cleaned = raw.trim().replace(/^["']|["']$/g, "");
-    return cleaned.length > 0 && cleaned.length < 100 ? cleaned : "";
+    return sanitizeGeneratedTitle(raw);
   } catch {
     return "";
   }
--- a/src/utils/sanitizeGeneratedTitle.ts
+++ b/src/utils/sanitizeGeneratedTitle.ts
@@ -1,18 +1,50 @@
-const WRAPPING_QUOTES = new Set(['"', "'", "\u201c", "\u201d", "\u2018", "\u2019", "«", "»"]);
+// Quote marks models wrap generated titles in despite the prompt: ASCII,
+// typographic (curly, German low-9), guillemets, and CJK title brackets.
+// prettier-ignore
+const WRAPPING_QUOTES = new Set([
+  '"', "'", "“", "”", "‘", "’", "„", "‚", "«", "»", "「", "」", "『", "』", "《", "》",
+]);
 
-// Models often wrap the title in ASCII or typographic quotes despite the prompt.
-// Peel those wrappers from both ends; leave inner apostrophes (Don't) intact.
-export function sanitizeGeneratedTitle(raw: unknown): string {
-  if (typeof raw !== "string") return "";
+// Apostrophe-shaped marks are never stripped one-sided: at a single end they
+// are usually part of the word (Workin’, ’Tis the Season), not a wrapper.
+const APOSTROPHES = new Set(["'", "‘", "’"]);
+
+const hasQuoteBesides = (text: string, index: number): boolean => {
+  for (let i = 0; i < text.length; i += 1) {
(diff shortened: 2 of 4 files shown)


Chadpiha (approved):
Thanks @hsusul — good catch on #1638, and extracting the sanitizer into a tested pure util was the right structure.

I reviewed this end-to-end: all generated titles funnel through `generateNoteTitle` (note formatting via `actionProcessingStore`, single/batch upload via `UploadAudioView` → `batchQueueStore`), so the single choke point covers every path, and all three repro cases from the issue now sanitize correctly.

While verifying edge cases I found that the one-sided strip (the `||` in the loop) regressed some inputs the old code left intact — a quoted phrase at one end (`Feedback on “Project X”` → dangling opener), German `„…“` (only the closer was in the set), and trailing/leading elision apostrophes (`Workin’`, `’Tis the Season`). Rather than round-trip that, I pushed a follow-up commit to your branch that peels only matched pairs (with a safe one-sided rule so unterminated wrappers like `"Weekly standup` still get cleaned), adds `„ ‚` and CJK brackets to the set, and moves the tests to `test/utils/` per repo convention.

One deliberate behavior change from old `main`: ASCII `'Tis the Season` now keeps its leading apostrophe — a lone apostrophe is indistinguishable from an elision, so we no longer strip it.

Verified: new tests 8/8, typecheck clean, full suite at parity with main (the one failure is a pre-existing `modelManagerBridgeDownloadStatus` flake that fails on the merge base too). Approving.

Issue #1633, opened 2026-08-14
Feature: import existing transcripts and notes from markdown/text files (migration path from other tools)


greatcoat:
## Summary

Add an import path for existing meeting transcripts and notes held in other tools, as markdown or plain text carrying dates and titles, including a bulk/folder mode.

## Motivation

Requested via support by a user actively migrating from Granola (tracked customer-side as [CUS-89](https://linear.app/openwhispr/issue/CUS-89/import-existing-meeting-transcripts-granola-bulk-markdown-hyeeyoung)). They have adopted OpenWhispr for new meetings and report the experience as good, but their historical meetings stay behind in the old tool, so search is split across two apps and the migration never actually completes. In their words the gap is that "there is no way to bring my old transcripts with me".

This is a **switching-cost** issue rather than a nice-to-have: the missing import is what keeps a competitor's app open on the machine.

## Current state in the code

The only file import is **audio**. `src/components/notes/UploadAudioView.tsx:1469` accepts `.mp3,.wav,.m4a,.webm,.ogg,.oga,.flac,.aac,.opus` and runs it through transcription. There is no text or markdown ingest anywhere in the app; the only other "import" affordance is bulk **dictionary** words.

Worth noting we already **emit** the target format: the "Save notes as files" setting mirrors notes to disk as real `.md` with frontmatter (`markdownMirror.js`). The export half exists and is deliberately one-way. This request is the missing inverse, so the file shape is already defined by our own mirror.

## Proposal (rough)

* Accept `.md` and `.txt`, with folder or multi-select for bulk.
* Derive title and date from front-matter where present, then filename conventions, then file mtime as a fallback.
* Land them as ordinary notes so existing search and organisation cover them with no further work.
* Prefer a **generic markdown path** over a Granola-specific parser: it covers Obsidian, Notion exports, Otter and plain files in one go, and does not break when a third party changes their export shape. A named-tool mapping can sit on top later if a particular export proves common.

## Related, and deliberately distinct

* OpenWhispr/openwhispr#573 — batch import of multiple **audio** files (Done). Different feature; that transcribes, this ingests already-transcribed text.
* OpenWhispr/openwhispr#460 — bulk **dictionary** import (Done).
* OpenWhispr/openwhispr#1398 — history sync to account (Done); relevant only in that imported notes should participate in it.

## Note for whoever picks this up

[...]

PR #1646, opened 2026-08-14
fix(notes): escape control characters in mirrored note frontmatter

### LINKED ISSUE #1643: Markdown mirror: a newline in a note title corrupts the whole YAML frontmatter block
### Affected component

`src/helpers/markdownMirror.js` — `_buildFrontmatter` / its inner `escYaml` helper (the on-disk markdown mirror that writes each note as a `.md` file with a YAML frontmatter block).

### Summary

`escYaml` is responsible for making a note title / folder name safe to embed in the YAML frontmatter block. It quotes values that contain YAML metacharacters, but it does **not** treat control characters (`\n`, `\r`, `\t`) as needing quoting, and it does not escape them when it does quote. A note whose `title` (or folder name) contains a line break therefore splits the frontmatter mapping onto a bare line, and the **entire** metadata block becomes unparseable YAML — `id`, `title`, `type`, `folder`, `created`, `updated` are all lost.

Note titles can legitimately contain a newline: `generateNoteTitle` (`src/utils/generateTitle.ts`) only trims *outer* whitespace from the model's response, so a model that returns a two-line title stores it verbatim; a user rename or a content-derived title can do the same.

### Reproduction (deterministic, no credentials/native modules)

```js
const markdownMirror = require("./src/helpers/markdownMirror"); // electron stubbed
markdownMirror.init(baseDir);
markdownMirror.writeNote(
  { id: 5, title: "Q3 Review\nconfidential", content: "Body", created_at: "2026-07-21T12:00:00Z" },
  "Personal"
);
```

The written file begins with:

```
---
id: 5
title: Q3 Review
confidential
type: personal
folder: Personal
created: 2026-07-21T12:00:00Z
updated: 2026-08-14T21:39:05.477Z
---

Body
```

### Actual behavior

The bare `confidential` line makes the block invalid YAML. Parsing it with the project's own dependency (`js-yaml`) throws:

```
YAMLException: can not read a block mapping entry; a multiline key may not be an implicit key (4:5)
```

so a consumer of the mirror (external markdown tools, and the markdown-import migration path proposed in #1633) cannot read *any* of the note's metadata.

### Expected behavior

The frontmatter block must remain valid YAML for every note title / folder name. A control character in the value should keep the value on a single line and round-trip back to the original string when parsed (e.g. `title: "Q3 Review\nconfidential"`).

### Root cause

`escYaml`'s quote-trigger regex omits `\n`/`\r`/`\t`, and its escaping only handles `\` and `"`. Control characters slip through both unquoted and unescaped.

### Proposed focused scope

In `escYaml`, treat `\n`/`\r`/`\t` as characters that require quoting, and escape them (`\n` → `\n`, etc.) inside the double-quoted form alongside the existing `\` and `"` escaping. This keeps the value on one line and round-trips through `js-yaml`. No change to the emitted output for values that are already control-character-free. I'll add a regression test that writes a note with a newline title and asserts the frontmatter parses back to the original title.

I intend to work on this.

hsusul:
## Summary

The on-disk markdown mirror (`src/helpers/markdownMirror.js`) writes each note as a `.md` file with a YAML frontmatter block. Its `escYaml` helper quotes values containing YAML metacharacters, but ignores control characters. A note title (or folder name) containing a line break is written into the frontmatter verbatim, which splits the mapping onto a bare line and makes the **entire** block invalid YAML — every metadata field (`id`, `title`, `type`, `folder`, `created`, `updated`) is lost, not just the offending value.

Fixes #1643.

## Problem

Note titles can legitimately contain a newline. `generateNoteTitle` (`src/utils/generateTitle.ts`) only trims *outer* whitespace from the model's response, so a model that answers with a two-line title stores it verbatim; a user rename or a content-derived title can do the same.

Writing such a note produces:

```
---
id: 5
title: Q3 Review
confidential
type: personal
folder: Personal
created: 2026-07-21T12:00:00Z
updated: 2026-08-14T21:39:05.477Z
---
```

The bare `confidential` line is not a `key: value` mapping entry. Parsing this with the project's own `js-yaml` dependency throws:

```
YAMLException: can not read a block mapping entry; a multiline key may not be an implicit key (4:5)
```

so external markdown tools (and the markdown-import migration path proposed in #1633) cannot read any of the note's metadata.

## Root cause

`escYaml`'s quote-trigger regex omits `\n`/`\r`/`\t`, and its escaping only handled `\` and `"`, so control characters slipped through both unquoted and unescaped.

## Fix

`escYaml` now treats `\n`/`\r`/`\t` as characters that require quoting, and escapes them (`\n` → `\n`, `\r` → `\r`, `\t` → `\t`) inside the double-quoted form alongside the existing `\` and `"` escaping. The value stays on one line and round-trips back to the original string when parsed.

## Behavior before / after

For a note titled `Q3 Review\nconfidential`:

| | `title:` line | `js-yaml` parse |
| --- | --- | --- |
| Before | `title: Q3 Review` + a bare `confidential` line | throws — all metadata lost |
| After | `title: "Q3 Review\nconfidential"` | parses back to `Q3 Review\nconfidential` |

Output for values that contain no control characters is unchanged.

## Scope / non-goals

- Only control characters (`\n`/`\r`/`\t`) are addressed — the concrete corruption that makes the block unparseable. Other YAML niceties (e.g. quoting a plain scalar that begins with `-`) are out of scope.
- No change to filenames, the `_slugify` path, or any other behavior.

## Tests

Added a regression test in `test/helpers/markdownMirror.test.js`: it mirrors a note whose title contains a newline and a tab, then parses the frontmatter back and asserts every key survives and the title round-trips exactly. It fails on the pre-fix code (the note file is unrecognizable / the block is not a valid mapping) and passes with the fix.

## Validation

Run from a clean worktree on `upstream/main`:

- `node --test test/helpers/markdownMirror.test.js` → **4 pass / 0 fail** (new test green; verified red without the fix)
- `prettier --check` on both changed files → clean
- `node scripts/check-i18n.js` (`npm run i18n:check`) → consistent
- `cd src && tsc --noEmit` (`npm run typecheck`) → exit 0
- `node --check src/helpers/markdownMirror.js` → OK
- `git diff --check` → clean

Environment note: the full `npm test` uses `node --import tsx`; this environment reused an existing `node_modules` that lacked `tsx`, and a fresh `npm ci` pulls the Electron toolchain/binaries. The changed code is a pure CommonJS helper, so the affected test file was run directly with `node --test` (its dependencies resolved), which is deterministic and needs no native modules, credentials, or network.

### DIFF
--- a/src/helpers/markdownMirror.js
+++ b/src/helpers/markdownMirror.js
@@ -33,8 +33,22 @@ class MarkdownMirror {
   _buildFrontmatter(note, folderName) {
     const escYaml = (str) => {
       if (!str) return '""';
-      if (/[:#{}[\],&*?|>!%@`]/.test(str) || str.includes('"') || str.includes("'")) {
-        return `"${str.replace(/\\/g, "\\\\").replace(/"/g, '\\"')}"`;
+      // A control character (newline, carriage return, tab) in a title splits the
+      // mapping onto a bare line and makes the whole frontmatter block unparseable,
+      // so force quoting and escape them like `"` and `\` — keeping the value on
+      // one line and round-tripping back to the original when parsed.
+      if (
+        /[:#{}[\],&*?|>!%@`]/.test(str) ||
+        /[\n\r\t]/.test(str) ||
+        str.includes('"') ||
+        str.includes("'")
+      ) {
+        return `"${str
+          .replace(/\\/g, "\\\\")
+          .replace(/"/g, '\\"')
+          .replace(/\n/g, "\\n")
+          .replace(/\r/g, "\\r")
+          .replace(/\t/g, "\\t")}"`;
       }
       return str;
     };
--- a/test/helpers/markdownMirror.test.js
+++ b/test/helpers/markdownMirror.test.js
@@ -81,6 +81,53 @@ test("renaming a mirrored note cleans up both stale files and reveals the note",
   assert.match(fs.readFileSync(notePath, "utf-8"), /Updated notes$/);
 });
 
+function readFrontmatter(fileContent) {
+  const match = fileContent.match(/^---\n([\s\S]*?)\n---\n/);
+  assert.ok(match, "file should start with a --- fenced frontmatter block");
+  const map = {};
+  for (const line of match[1].split("\n")) {
+    const kv = line.match(/^([A-Za-z_]+): (.*)$/);
+    assert.ok(kv, `frontmatter line is a "key: value" mapping entry: ${JSON.stringify(line)}`);
+    let value = kv[2];
+    if (value.startsWith('"') && value.endsWith('"')) {
+      value = value
+        .slice(1, -1)
+        .replace(/\\n/g, "\n")
+        .replace(/\\r/g, "\r")
+        .replace(/\\t/g, "\t")
+        .replace(/\\"/g, '"')
+        .replace(/\\\\/g, "\\");
+    }
+    map[kv[1]] = value;
+  }
+  return map;
+}
+
+test("a title with control characters keeps the frontmatter a valid single-line mapping", (t) => {
+  const basePath = fs.mkdtempSync(path.join(os.tmpdir(), "openwhispr-markdown-mirror-"));
+  t.after(() => fs.rmSync(basePath, { recursive: true, force: true }));
+
+  const note = {
+    id: 5,
+    title: "Q3 Review\nconfidential\ttag",
+    content: "Body text",
+    created_at: "2026-07-21T12:00:00Z",
+  };
+
+  markdownMirror.init(basePath);
+  markdownMirror.writeNote(note, "Personal");
+
+  const notePath = markdownMirror.getNotePath(note.id);
+  assert.ok(notePath, "the note file is discoverable, so its frontmatter parsed as a note");
+  const frontmatter = readFrontmatter(fs.readFileSync(notePath, "utf-8"));
+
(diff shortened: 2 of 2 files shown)


Chadpiha (approved):
Thanks @hsusul — excellent find and a really well-scoped fix. A newline in a title silently destroying every metadata field in the mirror is exactly the kind of bug that would have bitten the markdown-import path later, and the issue write-up made it trivial to verify.

I reviewed and tested this end-to-end:

Reproduced the corruption on main with your exact repro (same YAMLException), and found the fix also cures a second, silent variant — a title mixing quotes with a newline previously parsed but folded the newline into a space, so the title read back wrong with no error.
Round-tripped the fixed output through js-yaml for newline, CRLF, tab, literal-backslash (C:\notes), and quote/backslash/newline-mix titles, plus a newline in a folder name — every case parses and round-trips byte-for-byte, and output is unchanged for titles that were already clean.
Full test suite is green on a rebase onto current main (the one CI-visible failure at your branch point was unrelated — it's the test-hermeticity issue fixed in [#1647](https://github.com/OpenWhispr/openwhispr/issues/1647), one commit after your base).

One note for a follow-up (not blocking, tracked separately): other C0 control characters (e.g. \x08) still make the block unparseable — your \n/\r/\t scope covers the realistic cases, so extending to the full C0 range with YAML \xNN escapes can land as its own change.

Approving — thanks again for the careful issue report, the fix, and the regression test.

PR #1649, opened 2026-08-14
fix(notes): isolate per-file unlink failures in mirror deleteNote

### LINKED ISSUE #1648: Markdown mirror: one failing unlink in deleteNote orphans the note's other mirrored files
### Affected component

`src/helpers/markdownMirror.js` — `deleteNote(noteId)` (the on-disk markdown mirror that keeps one `.md` file per note, plus a `-transcript.md` companion).

### Summary

`deleteNote` collects a note's mirrored files and unlinks them in a single loop with **no per-file error handling** — only one `try` around the whole loop. If any single `fs.unlinkSync` throws (e.g. Windows `EPERM` when the file is open in an external markdown editor, or a transient FS error), the loop aborts and every *remaining* file is left orphaned on disk.

This is inconsistent with the mirror's own convention: `writeNote` and `writeTranscript` already wrap each stale-file `unlinkSync` in its own `try/catch` precisely so one un-deletable file cannot block cleaning up the others.

Orphaned mirror files resurface in external tools that sync the mirror folder, and a leftover `-transcript.md` with no owning note is exactly the kind of stray file that #1275 worked to disambiguate.

### Reproduction (deterministic, no credentials/native modules)

```js
markdownMirror.init(baseDir);
markdownMirror.writeNote(note, "Personal");        // 9-team-sync.md
markdownMirror.writeTranscript(note, "Personal", {}); // 9-team-sync-transcript.md

// Simulate the first unlink failing (e.g. the file is open in an editor on Windows).
const realUnlink = fs.unlinkSync;
let n = 0;
fs.unlinkSync = (p) => { if (++n === 1) throw Object.assign(new Error("EPERM"), { code: "EPERM" }); return realUnlink(p); };
markdownMirror.deleteNote(note.id);
fs.unlinkSync = realUnlink;
```

### Actual behavior

The directory still contains **both** files afterwards:

```
before delete: [ '9-team-sync-transcript.md', '9-team-sync.md' ]
after delete:  [ '9-team-sync-transcript.md', '9-team-sync.md' ]
```

The first failing unlink aborts the loop, so the transcript companion — which was perfectly deletable — is orphaned too.

### Expected behavior

`deleteNote` should remove every mirrored file it can, isolating a single failure so it does not prevent deleting the note's other files — the same guarantee `writeNote`/`writeTranscript` already provide. After the reproduction, only the file that genuinely could not be unlinked should remain.

### Root cause

The unlink loop has no per-iteration `try/catch`; a throw on one file escapes to the function-level `catch` and skips the rest.

### Proposed focused scope

Wrap the per-file `unlinkSync` in `deleteNote` in its own `try/catch` (matching `writeNote`/`writeTranscript`), so cleanup continues past an individual failure. No change to which files are targeted or to any other method. I'll add a regression test that makes the first unlink throw and asserts the remaining file is still removed.

I intend to work on this.

hsusul:
## Summary

`markdownMirror.deleteNote` removed a note's mirrored `.md` files in a single loop guarded by one outer `try/catch`. If any single `fs.unlinkSync` threw, the loop aborted and every *remaining* file was left orphaned on disk. This makes cleanup isolate each unlink, matching what `writeNote`/`writeTranscript` already do.

Fixes #1648.

## Problem

```js
for (const f of files) {
  fs.unlinkSync(f); // one throw here aborts the whole loop
}
```

On Windows a mirrored file open in an external markdown editor rejects deletion with `EPERM`; a transient FS error does the same. When that happens on the first file, the note's other file — e.g. its `-transcript.md` companion, which was perfectly deletable — is orphaned. Stray mirror files then resurface in tools that sync the mirror folder, and a transcript with no owning note is exactly the ambiguity #1275 addressed.

This is inconsistent with the mirror's own convention: `writeNote` and `writeTranscript` already wrap each stale-file `unlinkSync` in its own `try/catch`.

## Root cause

The unlink loop had no per-iteration error handling, so a throw on one file escaped to the function-level `catch` and skipped the rest.

## Fix

Wrap each `unlinkSync` in `deleteNote` in its own `try/catch` (logging via the existing `debugLogger`), so a single un-removable file no longer prevents deleting the note's other mirrored files. No change to which files are targeted.

## Behavior before / after

Reproduction: mirror a note (`9-team-sync.md` + `9-team-sync-transcript.md`), then delete it while the first `unlinkSync` throws `EPERM`:

| | Files left after `deleteNote` |
| --- | --- |
| Before | both files orphaned |
| After | only the file that genuinely could not be unlinked remains |

## Tests

Added a regression test in `test/helpers/markdownMirror.test.js`: it mirrors a note with a transcript, stubs `fs.unlinkSync` to throw on the first call, deletes the note, and asserts the remaining file is still removed. It fails on the pre-fix code (both files remain) and passes with the fix.

## Scope / non-goals

- Only `deleteNote`'s error isolation changes. No change to the file-matching globs, `writeNote`/`writeTranscript`, or any other method.

## Validation

Run from a clean worktree on `upstream/main`:

- `node --test test/helpers/markdownMirror.test.js` → **4 pass / 0 fail** (new test green; verified red without the fix)
- `prettier --check` on both changed files → clean
- `node scripts/check-i18n.js` (`npm run i18n:check`) → consistent
- `node --check src/helpers/markdownMirror.js` → OK
- `git diff --check` → clean

Environment note: the full `npm test` uses `node --import tsx`; this environment reused an existing `node_modules` that lacked `tsx`, and a fresh `npm ci` pulls the Electron toolchain/binaries. The changed code is a pure CommonJS helper, so the affected test file was run directly with `node --test` (its dependencies resolved), which is deterministic and needs no native modules, credentials, or network.


Chadpiha (approved):
Thanks again @hsusul !

This change correctly isolates per-file unlink failures in markdownMirror.deleteNote, ensuring one locked or undeletable mirror file does not prevent the remaining files from being deleted. The failed path and error are logged, while enumeration-level error handling remains intact.

PR #1659, opened 2026-08-15
fix(linux): treat Ptyxis and GNOME Console as terminals for paste

### LINKED ISSUE #1656: Linux auto-paste uses Ctrl+V in Ptyxis and GNOME Console
## Describe the bug

Linux terminal paste detection (`LINUX_TERMINAL_CLASSES` in `clipboard.js` and `terminal_classes` in `resources/linux-fast-paste.c`) does not include Ptyxis or GNOME Console (`kgx` / `org.gnome.Console`). Those apps expect Ctrl+Shift+V (or the terminal paste chord). OpenWhispr sends Ctrl+V, which the terminal often swallows or treats as a literal instead of paste.

The two lists are documented as needing to stay in sync. Ghostty/WezTerm/etc. are already present; the current Ubuntu/GNOME defaults are not.

## Affected component

- `src/helpers/clipboard.js` — `LINUX_TERMINAL_CLASSES` / `isTerminalSignature`
- `resources/linux-fast-paste.c` — `terminal_classes`

## To Reproduce

Window class / app id:

- Ptyxis: `org.gnome.Ptyxis` / `ptyxis` (Ubuntu 24.10+ default terminal)
- GNOME Console: `org.gnome.Console` / `kgx`

Current matcher (`normalized.includes(term)` / `strcasestr`):

- `"org.gnome.ptyxis".includes("ptyxis")` → would be true **if** `ptyxis` were in the list (it is not)
- `"org.gnome.console".includes("terminal")` → false
- `"kgx".includes("terminal")` → false

So both apps take the GUI paste path (`ctrl+v`).

## Expected behavior

Ptyxis and GNOME Console are treated as terminals: auto-paste uses Ctrl+Shift+V (and the native XTest helper uses the same classification).

## Actual behavior

Paste is sent as Ctrl+V. Dictation does not land in the terminal.

## Environment

- OpenWhispr `main`
- Linux (X11 or XWayland); native `linux-fast-paste` and JS fallback paths
- Ubuntu 24.10+ / Fedora with Ptyxis or GNOME Console

## User impact

Dictating into the default GNOME/Ubuntu terminal does not paste.

## Root-cause hypothesis

The terminal class lists were not updated when Ubuntu/GNOME switched defaults from GNOME Terminal to Ptyxis / Console.

## Proposed focused scope

- Add `ptyxis`, `kgx`, and `org.gnome.console` to both lists
- Unit-test `isLinuxTerminalWindowClass` / `isTerminalSignature` for those ids
- No paste-tool or Wayland portal changes

I intend to work on this.

hsusul:
## Summary

- Classify Ptyxis and GNOME Console as terminals so auto-paste uses Ctrl+Shift+V.
- Keep the JS and native `linux-fast-paste` class lists in sync.

## Problem

Ubuntu 24.10+ and current GNOME default to Ptyxis (`org.gnome.Ptyxis`) or GNOME Console (`org.gnome.Console` / `kgx`). Neither window class matched `LINUX_TERMINAL_CLASSES` / `terminal_classes`, so dictation paste was sent as Ctrl+V and did not insert in the terminal.

## Root cause

The terminal class lists include Ghostty, WezTerm, Kitty, etc., but not the current GNOME/Ubuntu defaults. Matching is substring-based (`includes` / `strcasestr`), so `org.gnome.Ptyxis` would match `ptyxis` once listed; `org.gnome.Console` does not contain `terminal`.

## Fix

- Add `ptyxis`, `kgx`, and `org.gnome.console` to `src/helpers/clipboard.js` and `resources/linux-fast-paste.c`.

## Behavior before / after

| Window class | Before | After |
| --- | --- | --- |
| `org.gnome.Ptyxis` | GUI paste (Ctrl+V) | terminal paste (Ctrl+Shift+V) |
| `kgx` / `org.gnome.Console` | GUI paste | terminal paste |
| `kitty` / `ghostty` | terminal paste | terminal paste |
| `Code` / `firefox` | GUI paste | GUI paste |

## Scope / non-goals

- In scope: terminal class lists and classification tests.
- Out of scope: Wayland portal paste, ydotool, layout-specific keycodes.

## Tests added

- Ptyxis and GNOME Console ids match
- existing terminals still match
- ordinary GUI apps do not match

## Validation

```text
ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/linuxTerminalClasses.test.js
# 3 pass, 0 fail

npx prettier --write src/helpers/clipboard.js test/helpers/linuxTerminalClasses.test.js
git diff --check
# clean
```

- Native C helper is not compiled in this unit test; the JS matcher is the same class list the C source documents as needing to stay in sync.

Fixes #1656

### DIFF
--- a/resources/linux-fast-paste.c
+++ b/resources/linux-fast-paste.c
@@ -352,7 +352,8 @@ static const char *terminal_classes[] = {
     "konsole", "gnome-terminal", "terminal", "kitty", "alacritty",
     "terminator", "xterm", "urxvt", "rxvt", "tilix", "terminology",
     "wezterm", "foot", "st", "yakuake", "ghostty", "guake", "tilda",
-    "hyper", "tabby", "sakura", "warp", "termius", "waveterm", NULL
+    "hyper", "tabby", "sakura", "warp", "termius", "waveterm",
+    "ptyxis", "kgx", "org.gnome.console", NULL
 };
 
 static int is_terminal(const char *wm_class) {
--- a/src/helpers/clipboard.js
+++ b/src/helpers/clipboard.js
@@ -89,6 +89,9 @@ const LINUX_TERMINAL_CLASSES = [
   "warp",
   "termius",
   "waveterm",
+  "ptyxis",
+  "kgx",
+  "org.gnome.console",
 ];
 
 // macOS reports localized app names rather than window classes, and iTerm2 has
--- a/test/helpers/linuxTerminalClasses.test.js
+++ b/test/helpers/linuxTerminalClasses.test.js
@@ -1,0 +1,54 @@
+const test = require("node:test");
+const assert = require("node:assert/strict");
+const Module = require("node:module");
+
+const clipboardModulePath = require.resolve("../../src/helpers/clipboard");
+const originalLoad = Module._load;
+
+function loadClipboardManager() {
+  delete require.cache[clipboardModulePath];
+  Module._load = function loadWithMocks(request, parent, isMain) {
+    if (request === "electron") {
+      return {
+        clipboard: {
+          readText() {
+            return "";
+          },
+          writeText() {},
+        },
+        systemPreferences: { isTrustedAccessibilityClient: () => true },
+      };
+    }
+    return originalLoad.call(this, request, parent, isMain);
+  };
+  try {
+    return require("../../src/helpers/clipboard");
+  } finally {
+    Module._load = originalLoad;
+  }
+}
+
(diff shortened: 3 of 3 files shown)


Chadpiha (approved):
Approved — no actionable findings.

Ptyxis and GNOME Console are consistently recognized in both the JavaScript and native Linux terminal lists, and the new regression coverage fails on the base branch and passes here. The focused tests, full suite, typecheck, lint, build, and consistency checks all pass.

Thanks for yet another contribution!

PR #1661, opened 2026-08-15
fix(reasoning): make the non-streaming request timeout configurable — fixes #1479

### LINKED ISSUE #1479: Note formatting aborts on local/Self-Hosted models: the 30s timeout is hardcoded
## Problem

`callChatCompletionsApi` in `src/services/ReasoningService.ts` aborts non-streaming reasoning requests after a hardcoded 30 s:

```js
const timeoutId = setTimeout(() => controller.abort(), 30000);
// ...
throw new Error("Request timed out after 30s");
```

**Note formatting**, **dictation cleanup**, and **translation** all use this path. On a local model, formatting a normal meeting transcript reliably exceeds 30 s and the result is lost. (The streaming paths — voice agent, chat — get 60 s.)

## Measurements

Mac Studio, 64 GB, MLX runtime, `qwen/qwen3.6-35b-a3b` 4-bit, Self-Hosted (LM Studio). Real 68-minute transcript truncated to several input sizes, note-formatting system prompt, `max_tokens: 4096`, non-streaming:

```
input   out    gen      wall     vs 30s
4000    3908   36.7s    39.2s    ABORTED
8000    4095   39.8s    42.3s    ABORTED
12000   4095   40.8s    43.6s    ABORTED
16000   3633   37.1s    40.0s    ABORTED
20000   4095   43.1s    46.3s    ABORTED
24000   4095   43.5s    45.0s    ABORTED

generation throughput: ~99 tok/s (stable)
every size exceeded 30s
```

Throughput is healthy — this isn't slow hardware. Producing real notes just takes 35–46 s on a local model.

## Request

**Make the timeout configurable.** `ReasoningService` already imports `getSettings`, so a settings field plus a `REQUEST_TIMEOUTS` constant in `src/config/constants.ts` (next to `TOKEN_LIMITS`) would suffice:

```js
const timeoutMs = getSettings().reasoningTimeoutMs ?? REQUEST_TIMEOUTS.REASONING;
```

Optional follow-up: the streaming transport (`processTextStreaming`) already exists and already handles the Self-Hosted route for chat/agent (via `resolveChatRoute` → `isLanChat`).

## Environment

- OpenWhispr `1.8.1`, Note formatting feature
- macOS `15.7`, Self-Hosted / LM Studio, `qwen3.6-35b-a3b` 4-bit MLX

AdityaPainuli:
Closes #1479

`callChatCompletionsApi` aborted every non-streaming reasoning request at a hardcoded 30s. Note formatting on local models takes 35-46s at healthy throughput (measurements in the issue), so results were reliably lost while the model was still working.

## Changes

- New `REQUEST_TIMEOUTS` in `src/config/constants.ts` (`REASONING: 30000`, `REASONING_STREAMING: 60000`); both literals in `ReasoningService` now read from it (streaming behavior unchanged).
- New `reasoningTimeoutMs` setting (numeric, localStorage-synced, 0 = default). `resolveReasoningTimeoutMs()` guards 0/NaN/negative/Infinity so the timeout can never be disabled by a bad value.
- Timeout error message reports the actual limit instead of a fixed "30s".
- No UI yet — the issue asked for the settings field + constant; a settings row can follow if wanted.

## Test

- `test/utils/reasoningTimeout.test.js`: fails on `main` (3/3, module absent), passes here.
- Full suite: pre-existing failures unchanged vs clean `main`; +3 tests, +3 passes. `npm run typecheck` clean, eslint clean on touched files.
- How to test: `node --import tsx --test test/utils/reasoningTimeout.test.js`; or set `localStorage.reasoningTimeoutMs = "120000"` and format a long note on a local model — the request is no longer cut off at 30s.

### DIFF
--- a/src/config/constants.ts
+++ b/src/config/constants.ts
@@ -141,6 +141,13 @@ export const MODEL_CONSTRAINTS = {
 // List length above which pickers switch to a searchable variant.
 export const LIST_SEARCH_THRESHOLD = 12;
 
+// Reasoning request timeouts (ms). REASONING is the non-streaming default and
+// can be overridden per-install via the reasoningTimeoutMs setting (0 = default).
+export const REQUEST_TIMEOUTS = {
+  REASONING: 30000,
+  REASONING_STREAMING: 60000,
+} as const;
+
 // Token Limits
 export const TOKEN_LIMITS = {
   MIN_TOKENS: 512,
--- a/src/services/ReasoningService.ts
+++ b/src/services/ReasoningService.ts
@@ -9,7 +9,14 @@ import {
 import { BaseReasoningService, ReasoningConfig } from "./BaseReasoningService";
 import { SecureCache } from "../utils/SecureCache";
 import { withRetry, createApiRetryStrategy, httpError } from "../utils/retry";
-import { API_ENDPOINTS, TOKEN_LIMITS, buildApiUrl, ensureV1Suffix } from "../config/constants";
+import {
+  API_ENDPOINTS,
+  REQUEST_TIMEOUTS,
+  TOKEN_LIMITS,
+  buildApiUrl,
+  ensureV1Suffix,
+} from "../config/constants";
+import { resolveReasoningTimeoutMs } from "../utils/reasoningTimeout";
 import logger from "../utils/logger";
 import { getSettings, isCloudCleanupMode } from "../stores/settingsStore";
 import { wrapCleanupTranscript } from "../config/prompts";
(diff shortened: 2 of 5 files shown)


Biotech-Basti:
This PR and #1557 (opened 2026-08-12) fix the same issue and edit the same lines in `ReasoningService.ts`, so whichever lands second will conflict. #1557 also has the settings UI, the locale strings, and the same fix for the hardcoded 30s in the Gemini and OpenAI providers. If the maintainers prefer this PR's approach, the UI from #1557 can be added on top of it. Happy to coordinate.

gabrielste1n:
Thanks @AdityaPainuli for digging into this — your diagnosis of the hardcoded 30s abort was spot on. The configurable timeout has now landed via #1557, which covers the same ReasoningService path plus the OpenAI/Gemini/Tinfoil providers, a Settings UI with i18n, and a clamped helper, so I'm closing this one as superseded. Appreciate the PR!

PR #1662, opened 2026-08-15
fix(updater): stop automatic update checks when the App updates toggle is off — fixes #1605

### LINKED ISSUE #1605: "App updates" switch have no effects and constantly throw errors
**Describe the bug**
Even when `App updates` option is off in Settings:
Image

Then app every start try to connect to internet for checking updates:
Image

Image

And throw error when open Settings window:

Image

**To Reproduce**

1. Use OpenWhisper in offline environment or with internet access blocked on firewall.

**Expected behavior**
1. Don't try forcefully connect to internet if this is set to off in Settings.
2. Don't show connection errors and dialogs when updates are off.

**Screenshots**
See above.

- OS: Windows 11
- Browser Edge
- Version OpenWhispr 1.8.3

AdityaPainuli:
Closes #1605

The "App updates" toggle only gated the update-available popup. The startup check (3s after launch) and the 4-hourly periodic check ran regardless, so offline/firewalled machines got network attempts on every start and the resulting `update-error` events surfaced as error dialogs whenever Settings opened — exactly the screenshots in the issue.

## Changes

- New `src/helpers/updateCheckPolicy.js` (Electron-free, same pattern as `dockPolicy.js`/`autoStartPolicy.js`): automatic checks require `notificationsEnabled !== false && notifyUpdates !== false` — the same tri-state convention the popup gate in `updater.js` already uses.
- `updater.js`: startup and periodic checks go through `_autoCheckForUpdates()`, which reads prefs at fire time, so flipping the toggle takes effect without a restart. With the toggle off, no check runs — no connection attempt, no error event, no dialog.
- Manual "Check for updates" from Settings is intentionally never gated.

## Test

- `test/helpers/updateCheckPolicy.test.js`: fails without the fix (3/3), passes with it.
- Full suite: failures identical to clean `main` (73 pre-existing); +3 tests, +3 passes.
- How to test: turn off Settings → Notifications → App updates, relaunch offline — no update dialog, log shows "update check skipped (app updates disabled)". Toggle back on — startup/periodic checks resume without a restart.

### DIFF
--- a/src/helpers/updateCheckPolicy.js
+++ b/src/helpers/updateCheckPolicy.js
@@ -1,17 +1,8 @@
-// App-update policy, kept free of Electron so it can be unit-tested.
-//
-// One predicate backs both update gates in updater.js so they cannot drift:
-// the automatic checks (startup timer, periodic timer) and the update-available
-// popup. The "App updates" toggle (notifyUpdates) gates the checks themselves,
-// not just the notification they produce: with the toggle off the app must not
-// reach out to the update feed, and a background check that never runs can
-// never surface a connection error dialog on offline or firewalled machines
-// (#1605). A manual "Check for updates" from Settings is an explicit user
-// action and is never gated here.
-//
-// Same tri-state convention as the other notification prefs: only an explicit
-// false disables. An empty or partial prefs object (renderer sync not arrived
-// yet) keeps today's check-by-default behavior.
+// The "App updates" toggle gates the automatic checks themselves, not just the
+// popup: with it off the app must not reach the update feed, so offline or
+// firewalled machines never surface a connection error dialog (#1605).
+// Only an explicit false disables — missing prefs (renderer sync not arrived
+// yet) keep check-by-default behavior.
 function appUpdatesEnabled({ notificationsEnabled, notifyUpdates } = {}) {
   return notificationsEnabled !== false && notifyUpdates !== false;
 }
--- a/src/helpers/updateCheckPolicy.js
+++ b/src/helpers/updateCheckPolicy.js
@@ -1,18 +1,19 @@
-// Automatic update-check decisions, kept free of Electron so they can be
-// unit-tested.
+// App-update policy, kept free of Electron so it can be unit-tested.
 //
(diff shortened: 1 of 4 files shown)


Chadpiha (approved):
Approving — thanks for this, it's a clean fix and the writeup made it easy to review.

**What I verified**
- Traced the whole path from `checkForUpdatesOnStartup()` → `update-error` → `useUpdater`'s module-level `globalState.error` → the `showAlertDialog` in `SettingsPage`. That sticky error is exactly the "dialog when Settings opens" in #1605, and gating the check upstream is the right place to stop it.
- Confirmed the precondition holds at boot: `notifyUpdates` hydrates synchronously from localStorage, and `initializeSettings()` runs in the always-created dictation window via `useTheme → useSettings`, so `sync-notification-preferences` lands well before the 3 s startup timer — no need to open the Control Panel first, and start-minimized/launch-at-login are covered too.
- Ran the full suite on your branch: 0 failures. Manual "Check for Updates" stays ungated, and defaults keep check-by-default until prefs arrive.

**Two small commits pushed onto your branch**:
- `refactor(updater): reuse the app-updates predicate in the popup gate` — the popup gate in `updater.js` still hand-rolled the same `notificationsEnabled !== false && notifyUpdates !== false` check, so the two gates could drift. Both go through the helper now, renamed `appUpdatesEnabled` so it reads right at both call sites. No behavior change.
- `test(updater): cover the automatic-check and popup gates end to end` — drives the real `UpdateManager` with `electron-updater` mocked and fake timers: toggle off → no startup/periodic call and no `update-error` to the renderers; fire-time re-read; manual check ungated; popup gate. The toggle-off cases fail against the pre-fix `updater.js`.

**One design note for the record:** gating on the master "Disable all notifications" switch too is a slightly broader change than the toggle named in the issue — with master off, users lose background checks and the passive "Update" button in the sidebar until they check manually. I think it's the right call, since the sub-toggles are disabled in the UI when master is off (an offline user with master off couldn't otherwise stop the checks), just flagging it so it's a conscious decision.

Issue #851, opened 2026-05-25
Silent input triggers full dictionary paste instead of empty result


KonstantinosPantalos:
Describe the bug
When activating the microphone to start recording and then stopping without providing any audio input (silence), the application automatically pastes the entire dictionary into the text field instead of doing nothing or returning an empty result.
To Reproduce

Open OpenWhisper and focus on a text input to activate the microphone
Do not speak — provide no audio input
Stop the recording
Observe that the entire dictionary is pasted automatically

Expected behavior
When no audio input is detected, the application should return an empty result or do nothing. It should not paste any content.
Screenshots
(optional — attach if you can capture the behavior)
Desktop:

OS: Windows 11 (latest)
Version: OpenWhisper 1.7.2

Additional context
This appears to be an edge case in how the transcription result is handled when the audio input is empty/silent. Instead of returning a null or empty string, the app seems to fall back to outputting the full dictionary content.

xlurie:
I believe I can explain what you saw. Disclosure up front: I filed #1636, the remote-path variant of dictionary echo slipping past the app's echo filter (fix open in PR #1663); while verifying that fix I traced the local path, and the mechanism matches this report. Against v1.8.3 and current main (`41cf1d4`):

1. The local path sends the dictionary join as the `prompt` field uncapped (`src/helpers/audioManager.js:567` builds it, `src/helpers/whisperServer.js:857` forwards it verbatim); the ~900-char cap exists only in `processWithOpenAIAPI`. The open PR #1632 adds a client-side head-trim here — its effect is covered below.
2. whisper.cpp keeps only the tail: the prompt window is `min(n_max_text_ctx, n_text_ctx/2) = 224` tokens (`src/whisper.cpp:6927`), of which the decoder prompt takes the last 223 as a tail slice (`:7128`). The head is discarded silently — no warning fires on this path. The window itself is already documented in PR #1632.
3. On near-silence the model can emit prompt text back. The gate (`matchesDictionaryPrompt` in `src/utils/dictionaryEchoFilter.js`) requires `dictionaryUsage >= 0.7`, computed against the full dictionary's unique words — but an echo can only contain what the engine actually read, the tail. Once fewer than 70% of the dictionary's unique terms fit in those last 223 tokens, the gate is unreachable by construction. That interaction is the part I have not seen documented.

[...]

PR #1663, opened 2026-08-15
fix(dictation): run the dictionary-echo check against the prompt actually sent — fixes #1636

### LINKED ISSUE #1636: Dictionary-echo check compares the transcript against the full hint list while the sent prompt is capped at ~900 chars — an exact echo of the sent prompt passes and pastes
Split out of #1547 (its "truncation mismatch" section; earlier fix attempt: #1229). Orthogonal to #1559, which fixed surfacing of *detected* echoes — here the echo is never detected, so the toast/failed-save path #1559 added never engages. It also survives #1632, which right-sizes the cap per provider but keeps it for Groq (890) and the Whisper family (~900) — and doesn't touch the echo comparison.

**v1.8.3, deterministic live repro** (BYOK-custom batch path, self-hosted OpenAI-compatible endpoint on loopback; a mock stood in for the STT server and echoed back exactly the `prompt` field it received — reproducing deterministically what Whisper's prompt-echo hallucination produces probabilistically):

- Dictionary: 390 terms, hint join 3,937 chars. The request carried the first **898 chars (~95 terms)**: `MAX_PROMPT_CHARS = 900`, `slice(0, 900)` trimmed back to the last comma (`src/helpers/audioManager.js:3089-3110`).
- `isDictionaryEcho` (`src/helpers/audioManager.js:613`) compares against the **full** join — `getCustomDictionaryPrompt()` → `getDictionaryHintWords()`, no cap on that side.
- Echo of the sent prefix vs the full list: textComposition **1.0** (≥ 0.9 passes), dictionaryUsage **0.249** (≥ 0.7 fails) → not flagged. Metrics recomputed offline with the shipped matcher; outcome confirmed live — the 898-char echo pasted as a normal transcript (`status = completed`) and went on through cleanup.

The general shape: dictionaryUsage for a sent-prefix echo is bounded by roughly sentChars / joinChars (≈ 0.23 here), so once a dictionary's join meaningfully exceeds the cap, a perfect echo of everything the recognizer was actually primed with can **never** reach the 0.7 threshold — the bigger the dictionary, the more certain the miss. Post-#1632 this still holds anywhere a cap remains (Groq, the Whisper family, and self-hosted endpoints whose model name reads as Whisper — my case).

Suggested direction (same as the report's suggested fix 3 / #1229's approach): run the check against the prompt that was actually sent — instead of, or in addition to, the full list. The full-list arm still makes sense for paths that send the dictionary whole.

<details><summary>Repro mock (echoes the received prompt field back; contains no dictionary content)</summary>

```js
const http = require("http");
function extractPromptField(bodyBuf, contentType) {
  const m = /boundary=(?:"([^"]+)"|([^;]+))/i.exec(contentType || "");
  if (!m) return null;
  const boundary = "--" + (m[1] || m[2]);
  const body = bodyBuf.toString("latin1");
  for (const part of body.split(boundary)) {
    if (!/name="prompt"/i.test(part)) continue;
    const headerEnd = part.indexOf("\r\n\r\n");
    if (headerEnd === -1) continue;
    return Buffer.from(part.slice(headerEnd + 4).replace(/\r\n$/, ""), "latin1").toString("utf8");
  }
  return null;
}
http.createServer((req, res) => {
  const chunks = [];
  req.on("data", (c) => chunks.push(c));
  req.on("end", () => {
    const prompt = extractPromptField(Buffer.concat(chunks), req.headers["content-type"]);
    res.writeHead(200, { "Content-Type": "application/json" });
    res.end(JSON.stringify({ text: prompt || "no prompt field received" }));
  });
}).listen(8599, "127.0.0.1");
```

Point the custom STT endpoint at `http://127.0.0.1:8599/v1`, dictate anything — the pasted "transcript" is the app's own sent prompt.
</details>

Happy to test a fix on this setup.

AdityaPainuli:
Closes #1636

The request path caps the dictionary hint at ~900 chars (890 for Groq), but `isDictionaryEcho` compared transcripts only against the full dictionary join. An echo of the sent prefix has `dictionaryUsage ≈ sentChars / joinChars` against the full join, so once the dictionary meaningfully exceeds the cap a perfect echo of everything the recognizer was actually primed with could never reach the 0.7 threshold — it pasted as a normal transcript and went on through cleanup (deterministic loopback repro in the issue).

## Changes

- `src/utils/dictionaryEchoFilter.js`: request-side caps moved here (`GROQ_MAX_PROMPT_CHARS = 890`, `DEFAULT_MAX_PROMPT_CHARS = 900`) with the shared `truncateDictionaryPrompt()`; new `matchesSentDictionaryPrompt()` checks the full prompt plus each sent-size prefix. The check reconstructs what was sent instead of threading request state through the pipeline.
- `src/helpers/audioManager.js`: `isDictionaryEcho` uses the sent-prefix-aware matcher for both arms (dictionary-only and merged whisper prompt); request-side truncation calls the shared helper, so the caps have a single source of truth.
- The full-list arm still covers paths that send the dictionary whole (the issue's suggested direction).

## Test

- 4 regression tests in `test/helpers/dictionaryEchoFilter.test.js`, including the issue's exact shape (390 terms, ~3.9k-char join, 900-char echo): fail without the fix, pass with it. False-positive guard included — dictation that merely uses a few dictionary terms stays unflagged.
- Full suite: failures identical to clean `main` (73 pre-existing); +4 tests, +4 passes.
- How to test: the issue's loopback mock (echoes the received `prompt` field back) with a 390-term dictionary — the echo is now discarded via the dictionary-echo path instead of pasting.

### DIFF
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -85,8 +85,11 @@ import { evaluateFinishedRecording, withSalvageWarning } from "./recordingValida
 import { isEmptyRecording } from "./recordingGuard";
 import {
   DICTIONARY_ECHO_CODE,
+  DEFAULT_MAX_PROMPT_CHARS,
+  GROQ_MAX_PROMPT_CHARS,
   dictionaryEchoError,
-  matchesDictionaryPrompt,
+  matchesSentDictionaryPrompt,
+  truncateDictionaryPrompt,
 } from "../utils/dictionaryEchoFilter.js";
 import { getDictionaryHintWords } from "../utils/snippets";
 import { normalizeAgentSelectionContext } from "../utils/agentSelectionContext";
@@ -649,10 +652,13 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
   // Check the dictionary on its own as well as the full prompt: the echo filter needs
   // 70% of the prompt's words to appear, and a Chinese script bias counts as one more
   // word, which alone pushes a one- or two-term dictionary under the threshold.
+  // Each is also compared against its sent-size prefixes, since the request path
+  // caps the prompt at ~900 chars and an echo of that prefix can never clear the
+  // usage threshold against a much larger full join (#1636).
   isDictionaryEcho(text) {
     return (
-      matchesDictionaryPrompt(text, this.getCustomDictionaryPrompt()) ||
-      matchesDictionaryPrompt(text, this.getWhisperPrompt())
+      matchesSentDictionaryPrompt(text, this.getCustomDictionaryPrompt()) ||
+      matchesSentDictionaryPrompt(text, this.getWhisperPrompt())
     );
   }
 
@@ -3325,16 +3331,15 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
       const endpoint = this.getTranscriptionEndpoint(model);
 
       // Groq rejects prompts > 896 chars (incl. when reached via "custom" provider).
-      // 890 leaves margin for UTF-16 vs codepoint counting drift.
+      // Caps and truncation live in dictionaryEchoFilter.js so the echo check can
+      // reconstruct exactly what was sent (#1636).
       const isGroqEndpoint = provider === "groq" || endpoint.includes("api.groq.com");
-      const MAX_PROMPT_CHARS = isGroqEndpoint ? 890 : 900;
+      const MAX_PROMPT_CHARS = isGroqEndpoint ? GROQ_MAX_PROMPT_CHARS : DEFAULT_MAX_PROMPT_CHARS;
       let dictionaryPrompt = this.getWhisperPrompt(apiSettings);
       if (dictionaryPrompt) {
         if (dictionaryPrompt.length > MAX_PROMPT_CHARS) {
           const originalLength = dictionaryPrompt.length;
-          const truncated = dictionaryPrompt.slice(0, MAX_PROMPT_CHARS);
-          const lastComma = truncated.lastIndexOf(",");
-          dictionaryPrompt = lastComma > 0 ? truncated.slice(0, lastComma) : truncated;
+          dictionaryPrompt = truncateDictionaryPrompt(dictionaryPrompt, MAX_PROMPT_CHARS);
           logger.debug(
             "Custom dictionary prompt truncated",
             {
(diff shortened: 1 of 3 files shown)


xlurie:
Author of #1636 here. **Confirmed at the matcher level** — this branch's filter and `main`'s, side by side offline, same dictionary and echo shapes as the issue (390 terms, 3,937-char join → sent prompt 898 chars / 95 terms). Branch tests 21/21.

| case | `main` | this PR | `dictionaryUsage` vs full join |
|---|---|---|---|
| echo of the sent 898-char prompt (the issue's self-hosted / custom-endpoint path) | not flagged | **flagged** | 0.249 |
| same at the Groq 890 cap | not flagged | **flagged** | 0.247 |
| verbatim echo of the full join | flagged | flagged | 1.000 |

0.249 is the issue's exact number. Size sweep (uniform synthetic terms): `main` goes blind past ~1,290-char joins, this branch detects everywhere; the truncated Chinese-bias merged prompt is caught on both arms; real dictation using dictionary terms stays at usage 0.01–0.04 (no false positives).

Two scope notes:

1. **Partial echoes** were previously undetectable at *any* coverage (an 898-char prompt can't reach 0.7 against the full join); now caught from ~70% of the sent prompt (66/95 terms, head-anchored; a tail run of that length still slips at 0.695). The residue is the threshold's ceiling, not this change.
2. **Local whisper.cpp** — `audioManager.js:1847` passes `initialPrompt` uncapped: no 890/900 prefix exists. The ~224-token cap is documented (#1632's comment); which *end* is kept I haven't verified. If it's the tail, an echo of the join's last 898 chars stays unflagged on both builds (0.235) — if so, I'll file it separately with the same rig.

Heads-up: #1632 rewrites the same `audioManager.js` lines (per-provider caps) — one of the two rebases; no semantic conflict, surviving caps are still 890/900. Happy to share the harness (synthetic terms, no private data).

xlurie:
Re-checked after today's rebase (`852eff29`): `src/utils/dictionaryEchoFilter.js` and its test file are byte-identical to the head I measured above (`1f351411`) — same blob SHAs — and the `audioManager.js` hunk is the same patch text with only its line offsets moved. The matcher-level results above carry over by identity, not by a re-run.

Unrelated to your change: the red `tests` (Build and Notarize) check is Prettier drift in `helpers/windowConfig.js`, the only file that run flags — this branch doesn't touch it, and its blob is identical to `main`'s (background: #1735). The standalone `Tests` check is green.

The local-whisper tail question from scope note 2 is since resolved on the app's own binaries: the engine keeps the prompt tail, so that echo stays under the threshold. I wrote it up on #851, which already reports the same symptom, rather than filing a new issue — nothing pending on this PR from it.

PR #1664, opened 2026-08-15
feat(hyprland): support both Lua and legacy configs

### LINKED ISSUE #1675: Dictation hotkey silently fails on Hyprland with Lua config provider (hyprctl keyword rejected)
## Problem
On Hyprland running the Lua config provider (`configProvider: lua` — e.g. Omarchy 4.0 "quattro", Hyprland 0.56), the dictation hotkey never fires, even when the app window is focused.

## Cause
`src/helpers/hyprlandShortcut.js` registers the key with `hyprctl keyword bind/unbind`. Hyprland rejects `keyword` under the Lua provider:

> keyword can't work with non-legacy parsers. Use eval.

`execFileSync` exits 0, so the app believes registration succeeded, but no bind appears (`hyprctl binds` shows none). The persisted fallback (`openwhispr-binds.conf` sourced from `hyprland.conf`) is also ignored, because Hyprland uses `hyprland.lua` instead. `getHyprlandConfigStatus()` only checks file writability, so the Settings UI reports OK.

## Repro
- OpenWhispr 1.8.3 (AUR `openwhispr-bin`), Hyprland 0.56.2, Omarchy 4.0.0, Wayland
- Set dictation hotkey to Super+F1 → press it with any window focused → nothing happens
- `hyprctl binds | grep F1` → empty

## Suggested fix
In `registerKeybinding()`: detect the Lua provider (e.g. `hyprctl systeminfo` → `configProvider: lua`, or presence of `hyprland.lua`) and write the bind into the Lua config, or use the GlobalShortcuts portal / `hyprctl eval`.

---
_Disclaimer: this issue was filed by an AI agent on behalf of a user._

IdrisGit:
## Summary

Adds shortcut persistence support for both `hyprland.lua` and the legacy `hyprland.conf`.

Starting with Hyprland 0.55, the legacy Hyprlang configuration format is deprecated in favor of Lua. The default config is now `~/.config/hypr/hyprland.lua`, with bindings defined through APIs such as `hl.bind()`.

OpenWhispr previously persisted dictation shortcuts only through `hyprland.conf`. On modern installations such as Omarchy Quattro, the shortcut could therefore disappear after a Hyprland reload.

## Changes

- Detect the active Hyprland config provider
- Support both `hyprland.lua` and `hyprland.conf`
- Use `hl.bind()` for Lua configs
- Preserve existing legacy config behavior
- Add focused tests for both formats

## References

- [Hyprland configuration introduction](https://wiki.hypr.land/Configuring/Start/)
- [Hyprland Lua binding documentation](https://wiki.hypr.land/Configuring/Basics/Binds/)

fixes #1675


Chadpiha (approved):
Approved. The original blocking issue is resolved.

Lua providers now use `hyprctl eval` with matching `hl.bind`/`hl.unbind`
key strings, and command output is validated instead of trusting exit
status 0. Persistence now uses protected absolute `pcall(require, ...)`,
provider detection is lazy and memoized, legacy migration/header handling
is fixed, and the Hyprland documentation is updated.

The new tests cover Lua registration, hotkey changes, unregistering,
Control+Super, provider-detection failures, `HYPRLAND_CONFIG`, legacy
behavior, and migration.

Thanks for the fix @IdrisGit !

PR #1668, opened 2026-08-15
feat(reasoning): log finish_reason and a normalized truncated flag on non-streaming success paths

xlurie:
### Problem

The non-streaming success logs drop the finish reason. A `*_RESPONSE` entry
records `responseLength`, `tokensUsed` and `success: true`, but not *why*
generation stopped — so a cleanup that hit its token cap
(`finish_reason: "length"`) is pasted shortened and logged as an ordinary
success, indistinguishable from a complete response. The codebase already
treats the finish reason as meaningful — the selection-edit guard
(`requireCompleteOutput`) throws on it at seven call sites
(`ReasoningService.ts`, `openai.ts`, `tinfoil.ts`, `gemini.ts`, the anthropic
and enterprise handlers in `ipcHandlers.js`, and `llamaServer.js`) — but on
the success path the field is discarded everywhere.

A second, smaller problem: the field comes back in five provider dialects
(`"length"`/`"max_tokens"` on chat-completions, `"MAX_TOKENS"` on Gemini,
`incomplete_details.reason: "max_output_tokens"` on the OpenAI Responses API,
`stop_reason: "max_tokens"` on Anthropic, and the AI SDK's `"length"`/
`"max-tokens"` on the enterprise path), so logging the raw value alone would
still leave diagnosis provider-specific.

### Fix

Pure observability — no behavior change, nothing user-visible:

- The raw `finishReason` **and** a normalized `truncated: boolean` (computed
  with each provider's own dialect) are added to the success log of every
  non-streaming path where the provider's finish reason is reachable:
  - the shared chat-completions path (LAN/self-hosted, Groq, Corti) in
    `ReasoningService.ts`;
  - `openai.ts` — both the Responses API and chat-completions arms;
  - `tinfoil.ts` and `gemini.ts`;
  - `anthropic.ts` and `enterprise.ts` — both main-process handlers already
    read the finish reason for the selection-edit guard, and now return it in
    the IPC result so `ANTHROPIC_SUCCESS` / `ENTERPRISE_SUCCESS` can log it
    (typed). Anthropic returns the raw `stop_reason` and the renderer
    normalizes it against a single literal; enterprise returns the normalized
    flag alongside the raw value, because its dialect list lives in the main
    process next to the guard and I'd rather not copy it across the process
    boundary.
- Wherever the guard and the log would otherwise be two independent
  definitions of "truncated" — `openai.ts`, `tinfoil.ts`, the enterprise
  handler — the guard's expression is hoisted into one `const` that both use,
  so they cannot drift apart.
- Deliberately not covered, with reasons rather than omissions:
  - the **OpenWhispr-cloud** path — its response carries `model`, `provider`,
    `promptMode` and `matchType`, but no finish reason at all, so exposing one
    is a server-side API change, not a client one;
  - the **local llama-server** path — the guard is there, but `runInference`
    resolves a bare string, so surfacing the finish reason means changing that
    promise's contract and every caller, which is more than an observability
    PR should carry;
  - the **streaming agent path**, which reports through a different channel.

With this, a shorter-than-dictated paste is diagnosable from the logs alone:
one grep for `"truncated": true`, regardless of provider.

### Testing

- New test (existing renderer SSR harness + stubbed `fetch`):
  `finish_reason: "length"` on the cleanup path still returns the text
  unchanged (pins the zero-behavior-change claim), `"stop"` likewise, and the
  selection-edit guard still throws on truncated output. The logged
  `finishReason`/`truncated` pair is asserted on the truncated response, on
  the complete one, and on the enterprise IPC result.
- `npm run typecheck`, `npm run lint`, `npm run build:renderer` — clean.
- Full `npm test`: the only failures (`extractArchive`/`gpuBinaryManager`/
  `systemTar`) reproduce identically on unpatched `main` on my machine
  (Windows tar), so they are not from this change.

### Notes

I run OpenWhispr against a self-hosted llama.cpp/llama-swap backend, where a
capped `n_predict` makes `finish_reason: "length"` easy to produce on demand
— that's how I noticed the field being dropped. (Unrelated to #1636, which is
a different silent-truncation class on the STT-prompt side; I ran into this
gap while instrumenting that repro.)

Possible follow-up, if there's appetite: a user-visible notice when a cleanup
comes back truncated. That one has real UX questions (which providers, and
the default cloud path doesn't expose a finish reason today), so I'd rather
propose it separately than bundle it here.

### DIFF
--- a/src/helpers/ipcHandlers.js
+++ b/src/helpers/ipcHandlers.js
@@ -4508,14 +4508,15 @@ class IPCHandlers {
             }
             const { text: generated, finishReason } = result;
 
-            if (
-              config?.requireCompleteOutput &&
-              ["length", "max-tokens", "max_tokens"].includes(finishReason)
-            ) {
+            // Normalize once so the selection-edit guard and the renderer's
+            // success log cannot disagree about what counts as truncated.
+            const truncated = ["length", "max-tokens", "max_tokens"].includes(finishReason);
+
+            if (config?.requireCompleteOutput && truncated) {
               throw new Error("Model output was truncated before the selection edit completed");
             }
 
-            return { success: true, text: (generated || "").trim() };
+            return { success: true, text: (generated || "").trim(), finishReason, truncated };
           } finally {
             if (controller) {
               sender.removeListener("destroyed", cancelSenderRequests);
@@ -4905,7 +4906,7 @@ class IPCHandlers {
           if (outputText === null) {
             throw new Error(describeMissingAnthropicText(data));
           }
-          return { success: true, text: outputText };
+          return { success: true, text: outputText, stopReason: data.stop_reason };
         } catch (error) {
           debugLogger.error("Anthropic reasoning error:", error);
           return { success: false, error: error.message };
--- a/src/services/ReasoningService.ts
+++ b/src/services/ReasoningService.ts
@@ -404,7 +404,8 @@ class ReasoningService extends BaseReasoningService {
     }
 
     const choice = response.choices[0];
-    if (config.requireCompleteOutput && isTruncatedFinishReason(choice?.finish_reason)) {
+    const finishReason = choice?.finish_reason;
+    if (config.requireCompleteOutput && isTruncatedFinishReason(finishReason)) {
       throw new Error("Model output was truncated before the selection edit completed");
     }
     // Reasoning models leak <think> blocks into non-streamed output; strip them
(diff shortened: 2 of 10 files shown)


xlurie:
Closing this: #2130 (shipped in 1.10.1) fixed the problem it was written to expose. Cleanup now passes `requireCompleteOutput`, so a `finish_reason: "length"` reply fails over to the raw transcript with a toast instead of being pasted and logged as a success (`src/services/ReasoningService.ts:423-424` on `main`). What's left here is log-only, and that doesn't justify rebasing a branch this far behind `main`. Thanks for the fix.

PR #1672, opened 2026-08-16
fix(snippets): handle nullish and partial settings in getDictionaryHintWords

### LINKED ISSUE #1671: getDictionaryHintWords crashes on nullish, empty, or partial settings
### Describe the bug

`getDictionaryHintWords` in `src/utils/snippets.ts` accesses `settings.snippets.length` and spreads `settings.customDictionary` without verifying that `settings`, `settings.snippets`, or `settings.customDictionary` are present and valid arrays.

When called with `null`, `undefined`, an empty object `{}`, or partial settings (such as `stSettings` during streaming transcription in `audioManager.js` or `effectiveSettings` in `PromptStudio.tsx`), it throws uncaught `TypeError` exceptions (`Cannot read properties of undefined/null` or `settings.customDictionary is not iterable`).

Additionally, when `settings.snippets` is empty, it returns the internal `settings.customDictionary` array reference directly instead of a new array copy, risking accidental mutation of store state by callers.

### To Reproduce

1. Call `getDictionaryHintWords(null)` or `getDictionaryHintWords({})`.
2. Observe `TypeError: Cannot read properties of undefined (reading 'snippets')` or `TypeError: Cannot read properties of undefined (reading 'length')`.
3. Call `getDictionaryHintWords({ customDictionary: ["term"] })` without `snippets`.
4. Observe `TypeError: Cannot read properties of undefined (reading 'length')`.
5. Call `getDictionaryHintWords({ snippets: [{ trigger: "term", replacement: "def" }] })` without `customDictionary`.
6. Observe `TypeError: settings.customDictionary is not iterable`.

### Expected Behavior

`getDictionaryHintWords` should safely handle nullish, empty, or partial settings objects, defaulting missing dictionary or snippet lists to empty arrays and always returning a fresh array of string hint words.

hsusul:
Fixes #1671

### Problem
`getDictionaryHintWords` in `src/utils/snippets.ts` assumed `settings`, `settings.snippets`, and `settings.customDictionary` are always present and populated arrays. When called with `null`, `undefined`, empty object `{}`, or partial settings (e.g. `stSettings` during streaming transcription or `effectiveSettings` in PromptStudio), it threw uncaught `TypeError` exceptions (`Cannot read properties of undefined/null` or `settings.customDictionary is not iterable`).

Additionally, when `settings.snippets` was empty, it returned the underlying `settings.customDictionary` array reference directly instead of a new array copy, leaking mutable state to callers.

### Solution
- Handle nullish, empty, or partial `settings` objects safely by validating array presence for both `customDictionary` and `snippets`.
- Safely extract non-empty string triggers from `snippets`.
- Always return a new array copy rather than mutating or returning internal state references.
- Added regression tests in `test/utils/snippets.test.js` covering nullish inputs, partial settings, empty dictionaries/snippets, and combined trigger extraction.

### Verification
- `node --test test/utils/snippets.test.js` (passes, 14/14 tests)
- `npm run build:renderer` (passes cleanly)
- `git diff --check` (clean)

### DIFF
--- a/src/utils/snippets.ts
+++ b/src/utils/snippets.ts
@@ -76,10 +76,16 @@ export function expandSnippets(text: string, snippets: Snippet[]): string {
  * Dictionary words plus snippet triggers — the hint list fed to the STT
  * prompt and cleanup-model dictionary suffix so triggers survive both.
  */
-export function getDictionaryHintWords(settings: {
-  customDictionary: string[];
-  snippets: Snippet[];
-}): string[] {
-  if (settings.snippets.length === 0) return settings.customDictionary;
-  return [...settings.customDictionary, ...settings.snippets.map((s) => s.trigger)];
+export function getDictionaryHintWords(settings?: {
+  customDictionary?: string[] | null;
+  snippets?: Snippet[] | null;
+} | null): string[] {
+  const dictionary = Array.isArray(settings?.customDictionary) ? settings.customDictionary : [];
+  const snippets = Array.isArray(settings?.snippets) ? settings.snippets : [];
+  if (snippets.length === 0) return [...dictionary];
+
+  const triggers = snippets
+    .map((s) => s?.trigger)
+    .filter((t): t is string => typeof t === "string" && t.length > 0);
+  return [...dictionary, ...triggers];
 }
--- a/test/utils/snippets.test.js
+++ b/test/utils/snippets.test.js
@@ -1,7 +1,7 @@
 const test = require("node:test");
 const assert = require("node:assert/strict");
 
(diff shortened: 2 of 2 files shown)


Chadpiha (approved):
Approved. Thanks for the contribution :)

 getDictionaryHintWords now safely handles nullish and partially hydrated settings, including missing dictionary and snippet collections, while preserving ordering and complete-settings behavior.
The new regression cases pass, along with 54 related tests and the full suite with no failures. Typecheck, lint, i18n checks, production build, and git diff --check also passed. No actionable findings.

PR #1674, opened 2026-08-16
feat: reliable cleanup on Flash Lite (dirty-output detector, re-pass, one-shot escalation)

andres03b:
## Problem

The fused Gemini pipeline (STT + cleanup in one call) on gemini-3.5-flash-lite frequently returns near-raw text on long Spanish dictations: fillers ("o sea", "eh") and immediate word repetitions ("las las") survive. Verified in DB: the fused pipeline DID run — it is model quality, not config.

## Changes

- **New `src/helpers/transcriptCleanliness.js`**: pure heuristic detector `looksUncleaned(text)` (filler density >= 2 or >= 1 immediate word repetition) plus `cleanupEscalationModel(model)` (Gemini Lite -> `gemini-3-flash-preview`, null otherwise). Unit-tested.
- **Fused gating (`audioManager.js`)**: a fused result that still looks dirty no longer claims `alreadyCleaned`, so the existing text-to-text cleanup route re-cleans it automatically with Lite. No new API plumbing.
- **One-shot escalation (`audioManager.js`)**: if the cleanup-route output still looks dirty and the model is a Lite variant, exactly one retry with `gemini-3-flash-preview` on the original text, inside try/catch (quota failure keeps the first pass). Gated on `route.kind === "cleanup"` — agent/voice-agent/translation untouched.
- **Hardened fused bridge directive (`geminiTranscription.js`)**: maps the custom prompt's `<transcript>`-tag references onto the attached audio (the Prompt Studio prompt was written for the two-step flow and confused the fused mode), names example fillers, and declares a near-unchanged transcript a failure.

## Verification

- `node --test` on transcriptCleanliness, geminiTranscription, dictationRouting, retryTranscriptionHandler: **78 pass, 0 fail** (TDD: red -> green).
- gentle-ai bounded review (reliability lens): **approved, 0 blockers**, 3 informative warnings (emphatic doubling false-positive costs one extra cheap call; `source` label does not distinguish the re-pass; wiring covered by live testing).
- Live dictation test pending before closing the slice.


andres03b:
Opened against the wrong repo by mistake (this belongs on my fork). Apologies for the noise.

PR #1684, opened 2026-08-17
docs(readme): fix broken star history chart

Dessalines39394:
The star history chart in the README is broken: the upstream service can no longer render it under GitHub's stargazer API restrictions, so the image fails to load. This switches the chart to star-history.dera.page, an alternative that uses a different data source requiring no API token and that serves both the chart image and the page from the same domain. The chart endpoint, query parameters, and hash-based link format are preserved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ React 19, TypeScript, Tailwind CSS v4, Electron 41, better-sqlite3, whisper.cpp,
 
 ## Star History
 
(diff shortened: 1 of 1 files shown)


Chadpiha:
Thanks for the contribution. The replacement endpoint currently works, but we don’t want to replace the README link with a newly launched, unofficial third-party service due to long-term reliability and trust concerns. We’ll remove or replace the broken chart independently if needed, so I’m closing this PR.

PR #1688, opened 2026-08-17
fix(debug-log): include app version in the System Info header

greatcoat:
**Problem:** Customer debug logs record platform, Node, and Electron versions but not the app's own version, so the first triage question — which release produced this log — is unanswerable.

**Fix:** Add `appVersion: app.getVersion()` as the first field of the System Info header that debugLogger writes when file logging initializes; regression test covers the header contents.

## Root cause
`debugLogger.initializeFileLogging()` logs a System Info header built only from `process`-derived fields; `app.getVersion()` was never included, and nothing else in the log carries the release version.

## Tests
- **RED:** new `test/helpers/debugLogger.test.js` (electron mocked via require-cache, per the existing `downloadUtils.test.js` pattern) fails on current main: `The input did not match the regular expression /appVersion/`.
- **GREEN:** passes with the fix; full suite 2091 tests, 0 fail (170 = known Electron-ABI better-sqlite3 local skips).

## Provenance
Contributing defect #4 from the CUS-113 RCA (Titan `wiki/areas/engineering/whisper-server-dll-not-found-rca.md`): triaging `debug-2026-08-17T12-23-22-137Z.log` required inferring the app version 1.8.x from Electron 41.10.0. Companion packaging fix: #1687.

### DIFF
--- a/src/helpers/debugLogger.js
+++ b/src/helpers/debugLogger.js
@@ -74,6 +74,7 @@ class DebugLogger {
 
       this.debug("Debug logging enabled", { logFile: this.logFile });
       this.info("System Info", {
+        appVersion: app.getVersion(),
         platform: process.platform,
         nodeVersion: process.version,
         electronVersion: process.versions.electron,
--- a/test/helpers/debugLogger.test.js
+++ b/test/helpers/debugLogger.test.js
@@ -1,0 +1,53 @@
+const test = require("node:test");
+const assert = require("node:assert/strict");
+const fs = require("fs");
+const os = require("os");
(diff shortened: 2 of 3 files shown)


Chadpiha (approved):
Thanks Josh! The fix correctly adds the application version to the debug log’s System Info header and keeps the current Electron test doubles aligned with the new API. I verified both startup and runtime debug-logging paths, with 14/14 targeted tests and the full 3,208-test suite completing with 0 failures. No regressions, cross-platform concerns, or API/dependency changes found.

PR #1694, opened 2026-08-17
feat: RFC - enable automatic updates by default

IdrisGit:
## Goal
The goal is fewer stale-version support reports and more users benefiting from the latest fixes and features.

## Proposal
this RFC proposes downloading updates automatically by default,  with a restart prompt once ready.

This helps reduce stale-version bug reports and helps users receive fixes and features without needing to notice or remember an update prompt. Users can opt out in Settings and keep the manual update flow.


greatcoat:
Closing — this shipped in v1.10.0 via #2120, along the lines you proposed. Automatic updates is now a Settings toggle (on by default for new installs, off for existing ones until they opt in), checks always run so the sidebar can still offer a manual update, and the update-available popup window is gone.

Thanks for the RFC — it's the direction we went.

PR #1695, opened 2026-08-17
feat(transcriptions): RFC - Full transcription history search and pagination

### LINKED ISSUE #1666: Transcription history is capped at the 50 most recent entries, and Ctrl/Cmd+K search only filters those 50 (follow-up to #307)

Setup: OpenWhispr 1.8.3, Windows 11, daily dictation since May — ~7,000 rows in
`transcriptions.db`. Code references below are against current `main`, not the shipped bundle.

The transcription history on the Home view only ever shows the 50 most recent entries. On a busy
dictation day that's a couple of hours of work, so older dictations effectively vanish from the
app — my first guess was that a retention setting was silently pruning them. Then I tried
Ctrl/Cmd+K, which is labelled "Search across notes and transcripts", and it doesn't reach the
older entries either.

What's actually going on (current `main`):

- The renderer pins the page size: `src/stores/transcriptionStore.ts` has
  `const DEFAULT_LIMIT = 50` / `currentLimit`, and the only caller, `ControlPanel.tsx`, calls
  `initializeTranscriptions(undefined, includeDiscarded)` — so nothing ever requests a different
  limit. `HistoryView.tsx` has no load-more path.
- The DB layer takes a limit but no offset: `DatabaseManager.getTranscriptions(limit = 50,
  { includeDiscarded })` in `src/helpers/database.js` runs `... ORDER BY timestamp DESC LIMIT ?`,
  and the `db-get-transcriptions` IPC handler in `src/helpers/ipcHandlers.js` passes the default
  through.
- Transcript search in the command palette is a client-side filter over what's already loaded: in
  `src/components/CommandSearch.tsx`, `filteredTranscripts` runs
  `transcriptions.filter((tr) => tr.text.toLowerCase().includes(query.toLowerCase()))` on the same
  50-row store, then `.slice(0, 5)`. Notes, by contrast, get a real DB-level search:
  `searchNotes()` over the `notes_fts` FTS5 table via `db-search-notes`. Transcripts have neither
  an FTS table nor a search handler.
- The data itself is fine: `transcriptRetentionDays` defaults to `0` (keep forever) and only
  prunes when `> 0`; `audioRetentionDays` (default 30) deletes `.webm` files and clears audio
  flags in `src/helpers/audioStorage.js`, never the text. All ~7,000 rows are intact in SQLite —
  the UI just can't reach the older ~6,950.

Relation to #307: that issue asked for two things — remove the 50-entry limit, and add a search
bar — and was closed as released in v1.6.0. The search bar did land (the Cmd+K palette), and it's
genuinely nice. But because transcript matching happens over the already-loaded 50 rows, the
original motivation of #307 ("look back at what I dictated days or weeks later") still isn't
reachable from the UI. I'm opening this separately rather than reviving a closed thread, but happy
for it to be folded back into #307 if you'd prefer.

Roughly by effort:

- Smallest: a "Load more" control that re-requests with a larger limit.
  `initializeTranscriptions(limit)` already takes a limit as its first argument and remembers it
  (`currentLimit`), and the limit param is plumbed renderer → IPC → SQL end-to-end, so this needs
  no new wiring. (A raised limit also doesn't suffer from the classic offset instability when new
  dictations prepend to a live list. If you'd rather paginate with a real offset,
  `getAgentConversationsWithPreview(limit = 50, offset = 0, ...)` in the same `database.js`
  already does `LIMIT ? OFFSET ?` — an in-repo pattern to copy.)
- Bigger, but what actually closes #307's intent: give transcripts what notes already have — an
  FTS5 table plus a `db-search-transcriptions` handler — so the Ctrl/Cmd+K palette searches the
  whole history instead of the current page.

- Checked the retention settings — data is kept forever by default, so this isn't retention;
  the rows are all in SQLite, just unreachable from the UI.
- Ctrl/Cmd+K search — filters only the 50 already-loaded entries (top-5 shown), so it can't
  substitute for either pagination or real search.
- Reading `transcriptions.db` directly works, but that's not really an answer for a UI app.

**Additional context**

Happy to test a build against a real ~7k-row history and report back render/query timings if list
size is a concern.

IdrisGit:
> [!NOTE]
> This is a feature proposal for maintainer review and feedback.

## Summary

This RFC proposes making complete local transcription history browsable and searchable without loading the full database into renderer memory.

## Current behavior

History loads up to 50 transcriptions. Command Search filters that loaded subset in memory, so older transcriptions cannot be found.

## Proposal

- Load History incrementally with indexed keyset pagination.
- Virtualize History rows to keep large histories responsive.
- Add an SQLite FTS5 index for complete transcription-history search.
- Rebuild the FTS index once during migration so existing transcripts are
  searchable after upgrade.
- Preserve existing deleted and discarded-history behavior.

## Expected outcome

Users can browse their full transcription history and search across their complete local transcription database. Command Search (ctrl+k) remains bounded to a small result set while its search scope is no longer limited to loaded rows.

## Implementation included in this RFC

- FTS5 external-content index and write-maintenance triggers.
- Atomic initial index rebuild for existing transcript rows.
- Cursor-based History IPC and store state.
- Full-history Command Search integration.
- Virtualized infinite-scroll History UI.
- SQLite-backed search coverage.

if approved fixes #1666


xlurie:
Disclosure: I filed #1666, which this PR says it would fix if approved — thanks for picking it up so quickly.

I replayed the PR's DDL and queries (the SQL only, not the PR's JS) against a copy of a real 87-day history: 7,055 rows, 3.10 M chars of `text`, ~14 MB DB, 97% of rows Cyrillic. Environment: `node:sqlite` (SQLite 3.51.2; the app bundles better-sqlite3 with 3.53.0, so these are not app timings). Results: full FTS `rebuild` under 200 ms; the file grew 2.9 MB, of which 1.9 MB is FTS pages and the rest freelist; `integrity-check` clean; the keyset walk covered all 142 pages in ~87 ms via the partial index, with all 7,055 ids present exactly once; insert/update/hard-delete trigger round-trips stayed consistent. Happy to rerun all of this against an actual build once the implementation settles.

Three observations:

1. `searchTranscriptions` orders by `timestamp DESC` where `searchNotes` uses `notes_fts.rank`, and the query plan shows `USE TEMP B-TREE FOR ORDER BY`, so search cost tracks match count: 0.3 ms at 25 matches, ~4 ms at 598 (`агент`), ~22 ms at 4,383. Still fast, but for `агент` the returned 20 all fall within the 6 newest calendar days of the 87-day history. Was recency ordering deliberate, and is reaching older matches planned?

2. My data has no two rows sharing a timestamp, so the walk never exercised the `id` tie-break the composite cursor exists for. A fixture with several same-second rows straddling a page boundary would cover that case.

3. Cyrillic under the default `unicode61` tokenizer (same as `notes_fts`): case folds correctly (`промпт`/`ПРОМПТ`/`Промпт` return identical sets), but `ё` is not folded to `е` — `ещё` and `еще` match near-disjoint sets here (90 vs 1,001 rows, 1 shared).

PR #1702, opened 2026-08-18
feat(models): add Gemini 3.5 and 3.1 Flash Lite models

### LINKED ISSUE #1698: Add gemini-3.5-flash-lite to the Gemini model picker (2.5 Flash-Lite is retired by Google)
### Summary

The Gemini model picker for Dictation Cleanup offers `gemini-2.5-flash-lite`, which Google has retired for new API keys, while the current replacement `gemini-3.5-flash-lite` is missing from the list. That leaves no working Flash-Lite option, and the models that do work add several seconds to every dictation.

### What happens

Selecting **Gemini 2.5 Flash-Lite** and pressing Test returns Google's own error:

```
Test failed: This model models/gemini-2.5-flash-lite is no longer available to new users.
Please update your code to use models/gemini-3.5-flash-lite for the latest features and improvements.
```

`gemini-3.5-flash-lite` is on Google's free tier and is the model Google explicitly points to, but it is not selectable in the app.

### Why it matters — measured

Dictating in Turkish, local Whisper `large-v3-turbo` on CUDA, cleanup via Gemini BYOK:

| Stage | Time |
|---|---|
| Local transcription (5.28 s clip, measured directly against the bundled whisper-server) | **0.26 – 0.55 s** |
| Network connect to `generativelanguage.googleapis.com` | 0.04 – 0.09 s |
| Remainder = cleanup generation | **~2.5 – 4.5 s** |

So cleanup accounts for roughly 85% of the wait, and the only selectable Gemini models are the heavier ones. Flash-Lite is exactly the tier that would fix this, which is presumably why #1674 targeted Flash Lite reliability in the first place.

### Why there is no workaround

Pointing the **Custom** provider at Google's OpenAI-compatible base (`https://generativelanguage.googleapis.com/v1beta/openai`) fails with *"Custom endpoint is invalid or not configured"*. Per `resolveConfiguredOpenAIBase` in `src/services/ai/openaiBase.ts`, this is intentional — `generativelanguage.googleapis.com` is on the `knownNonOpenAIUrls` blocklist. Additionally `ensureV1Suffix()` would rewrite that base to `.../openai/v1`, which is not a valid Google path. Both behaviours look correct on their own; together they mean the built-in picker is the only route, so the missing entry is a hard block.

Writing the model id into local storage directly does not help either: with the Gemini provider selected, the value is validated against the bundled list and reverts to `gemini-3.5-flash`.

### Request

Add `gemini-3.5-flash-lite` to the Gemini model picker, and consider retiring `gemini-2.5-flash-lite` from the list since Google no longer serves it to new keys.

Precedent: #837 added Gemini 3.5 Flash the same way.

### Environment

- OpenWhispr 1.8.3 (Windows installer)
- Windows 11 Home 26200
- Speech-to-text: local Whisper `large-v3-turbo`, CUDA build
- Cleanup: Google Gemini, BYOK, dictation language Turkish

### LINKED ISSUE #1699: Missing Google Gemini models
**Describe the bug**
In the selection of Google Gemini models, some interesting models with larger free limits are missing, like Gemini 3.1 Flash Lite or Gemini 3.5 Flash Lite

**To Reproduce**
Selecting models from any Google Gemini list in "Bring your own key" option.

**Expected behavior**
It would probably be best if a complete list of models was displayed, from which we could choose. If it is supposed to be curated by you in some way, there should be some general rule that, for example, immediately approves new models as they arrive. 

- 
- OS: Win 11
- Version 1.8.3

xAlcahest:
## Problem

Google retired `gemini-2.5-flash-lite` for new API keys: selecting it and pressing Test now returns Google's own error pointing at `models/gemini-3.5-flash-lite` as the replacement. That replacement was missing from the registry, so users with new keys had no working Flash-Lite option at all and cleanup ran on the heavier Gemini models, which per the measurements in #1698 accounts for most of the dictation wait.

## Solution

Add the current Flash-Lite tier to the Gemini section of `modelRegistryData.json`, which feeds every picker where Gemini appears (dictation cleanup, dictation agent, translation, note formatting, chat). Verified against Google's current lineup on [ai.google.dev/gemini-api/docs/models](https://ai.google.dev/gemini-api/docs/models): `gemini-3.5-flash-lite` is GA, multimodal, with thinking on by default ([model page](https://ai.google.dev/gemini-api/docs/models/gemini-3.5-flash-lite)), so it carries `supportsThinking` and the disable-thinking path can send `thinkingLevel: "minimal"` like the other Gemini 3 entries. `gemini-3.1-flash-lite` is GA and multimodal but absent from the [thinking support table](https://ai.google.dev/gemini-api/docs/thinking), so it gets no `supportsThinking` flag, since sending `thinkingConfig` to it would fail the request.

`gemini-2.5-flash-lite` stays in the list on purpose: Google still serves it to existing keys, and dropping it would break users who have it selected today.

Fixes #1698
Fixes #1699

## Changes

- src/models/modelRegistryData.json: add `gemini-3.5-flash-lite` and `gemini-3.1-flash-lite` entries with capability flags
- src/locales/*/translation.json: localized descriptions for the two models in all 10 languages
- test/helpers/geminiModelRegistry.test.js: pin the two entries and their flags, assert `gemini-2.5-flash-lite` stays present, and check every Gemini `descriptionKey` resolves in every locale


Chadpiha (approved):
Thanks! I verified the following:

- Both IDs are GA on Google's models page, and the supportsThinking split matches the thinking-support table: gemini-3.5-flash-lite accepts thinkingLevel: "minimal", gemini-3.1-flash-lite isn't listed (only the -image variant), so leaving it unflagged is right.
- The flag actually governs behaviour — gemini.ts and thinkingSuppression.ts gate solely on the registry supportsThinking, nothing keys off a gemini-3 ID prefix, so 3.1 Flash Lite won't receive thinkingConfig.
- Gemini models[0] is unchanged, so per-provider defaults (settingsStore, policyRules) are unaffected. Keeping gemini-2.5-flash-lite for existing keys is the correct call.
- All 10 locales updated in registry order; check-i18n and translationCoverage green.
- New test follows the gemmaMtpRegistry.test.js convention. CI green across builds/CodeQL.

I pushed one small follow-up commit onto your branch:

- Moved the descriptionKey check into test/locales/translationCoverage.test.js and generalised it to every descriptionKey in the registry (all providers, not just Gemini). It checks en only, since the existing en-parity test already covers the other locales — so the Gemini-only per-locale test was dropped as redundant.
- Synced the Gemini model list in CLAUDE.md with the registry while at it.

Issue #1706, opened 2026-08-18
Snippet expansion crashes dictation when snippets is nullish


hsusul:
## Describe the bug

After a dictation finishes, `useAudioRecording` always runs `expandSnippets(result.text, getSettings().snippets)`. `expandSnippets` reads `snippets.length` with no array guard.

`getDictionaryHintWords` already treats nullish/partial settings as an empty list (#1671 / #1672). The expansion path was not given the same guard. Cloud/settings apply can call `setSnippets(settings.snippets)` whenever `snippets !== undefined`, including `null`.

## Affected component

- `src/utils/snippets.ts` — `expandSnippets`
- Caller: `src/hooks/useAudioRecording.js`

## To Reproduce

On current `main`:

```js
expandSnippets("hello", [{ trigger: "x", replacement: "y" }])
// "hello"

expandSnippets("hello", null)
// TypeError: Cannot read properties of null (reading 'length')

expandSnippets("hello", undefined)
// TypeError: Cannot read properties of undefined (reading 'length')
```

## Expected behavior

Nullish or non-array `snippets` is treated as no snippets: return the transcript unchanged.

## Actual behavior

The renderer throws after transcription, so the dictation is not pasted.

## Environment

- OpenWhispr `main`
- Any platform. Hits when `getSettings().snippets` is null/undefined (cloud apply, partial settings).

## User impact

Dictation that should have pasted instead fails in the overlay with a TypeError.

## Root-cause hypothesis

`expandSnippets` assumed the settings store always holds an array. #1672 hardened the hint-word helper only.

## Proposed focused scope

- Guard `expandSnippets` like `getDictionaryHintWords`
- Add cases to `test/utils/snippets.test.js`
- No settings-store or SQLite changes

I intend to work on this.

PR #1707, opened 2026-08-18
fix(calendar): treat date-only upcoming events as local calendar days

### LINKED ISSUE #1705: Upcoming Meetings groups Google all-day events on the previous local day
## Describe the bug

Google all-day events store `start.date` as `YYYY-MM-DD` with no time or zone. `formatUpcomingDateGroup()` passes that string to `new Date(...)`, which the spec treats as UTC midnight. West of UTC that is still the previous local calendar day, so an all-day event for today is listed under yesterday's weekday heading (or a formatted date) instead of "Today".

Timed ISO timestamps and `Date` objects are fine. History grouping already uses `normalizeDbDate()` (#1272); upcoming grouping was left on the default parser. `normalizeDbDate` is not the right fix here: appending `Z` to a date-only string still means UTC midnight.

A missing `start_time` also throws (`Cannot read properties of null (reading 'getFullYear')`) and can blank the Upcoming Meetings list.

## Affected component

- `src/utils/dateFormatting.ts` — `formatUpcomingDateGroup`
- Caller: `src/components/UpcomingMeetings.tsx` (`event.start_time`)
- Writer: `src/helpers/googleCalendarManager.js` (`item.start?.dateTime || item.start?.date`)

## To Reproduce

On current `main`, in `America/Los_Angeles`, with local calendar date `2026-08-18`:

```js
formatUpcomingDateGroup("2026-08-18", t)
// "Monday, Aug 17"  — not Today

formatUpcomingDateGroup(new Date(2026, 7, 18, 12), t)
// "Today"

formatUpcomingDateGroup(null, t)
// throws
```

## Expected behavior

- A date-only `YYYY-MM-DD` is that local calendar day (Today / Tomorrow / weekday).
- Null or invalid values return an empty label instead of throwing.

## Actual behavior

Date-only strings shift back one local day west of UTC. Null throws.

## Environment

- OpenWhispr `main`
- Any timezone behind UTC (US/EU mornings). Google Calendar all-day events.

## User impact

All-day meetings (PTO, holidays, conferences) appear under the wrong day in Upcoming Meetings.

## Root-cause hypothesis

`new Date("YYYY-MM-DD")` is UTC midnight. Upcoming grouping never adopted a date-only local parse after #1272.

## Proposed focused scope

- Parse `YYYY-MM-DD` as a local calendar date
- Guard null/invalid input
- Extend `test/helpers/dateGroupFormatting.test.js`
- Do not change `normalizeDbDate` (history timestamps) or Microsoft Graph `T00:00:00Z` all-day handling

I intend to work on this.

hsusul:
## Summary
- Google all-day events store `start.date` as `YYYY-MM-DD`. `formatUpcomingDateGroup()` passed that to `new Date()`, which is UTC midnight — west of UTC that is still yesterday.
- Parse date-only strings as local calendar dates, and return an empty label for null/invalid input instead of throwing.

## Problem
On current `main` in `America/Los_Angeles` when today is 2026-08-18:

| Input | Before |
| --- | --- |
| `"2026-08-18"` (Google all-day) | `Monday, Aug 17` |
| `new Date(2026, 7, 18, 12)` | `Today` |
| `null` | throws `getFullYear` |

History grouping already uses `normalizeDbDate()` (#1272). That helper appends `Z` and is correct for SQLite timestamps, not for date-only all-day values.

## Root cause
ECMAScript date-only forms are UTC. Upcoming Meetings never got a local date-only parse.

## Fix
- `YYYY-MM-DD` → local `Date(y, m - 1, d)`
- Timed ISO / `Date` objects unchanged
- Null, non-strings, and invalid strings → `""`

## Scope
- `src/utils/dateFormatting.ts` and `test/helpers/dateGroupFormatting.test.js`
- No Google sync, Microsoft Graph `T00:00:00Z` all-day, or UI copy changes

## Tests

```bash
node --test test/helpers/dateGroupFormatting.test.js
```

Fixes #1705

### DIFF
--- a/src/utils/dateFormatting.ts
+++ b/src/utils/dateFormatting.ts
@@ -27,8 +27,32 @@ export function formatRelativeTime(
   return formatShortDate(dateStr);
 }
 
-export function formatUpcomingDateGroup(date: Date | string, t: (key: string) => string): string {
-  const d = typeof date === "string" ? new Date(date) : date;
+function toUpcomingGroupDate(date: Date | string | null | undefined): Date | null {
+  if (date instanceof Date) {
+    return Number.isNaN(date.getTime()) ? null : date;
+  }
+  if (typeof date !== "string") return null;
+  const trimmed = date.trim();
+  if (!trimmed) return null;
+
+  // Google all-day events are date-only (`YYYY-MM-DD`). `new Date("YYYY-MM-DD")`
+  // is UTC midnight, which is still the previous local day west of UTC.
+  const dateOnly = /^(\d{4})-(\d{2})-(\d{2})$/.exec(trimmed);
+  if (dateOnly) {
+    const parsed = new Date(Number(dateOnly[1]), Number(dateOnly[2]) - 1, Number(dateOnly[3]));
+    return Number.isNaN(parsed.getTime()) ? null : parsed;
+  }
+
+  const parsed = new Date(trimmed);
+  return Number.isNaN(parsed.getTime()) ? null : parsed;
+}
+
+export function formatUpcomingDateGroup(
+  date: Date | string | null | undefined,
+  t: (key: string) => string
+): string {
+  const d = toUpcomingGroupDate(date);
+  if (!d) return "";
   const now = new Date();
   const today = new Date(now.getFullYear(), now.getMonth(), now.getDate());
   const tomorrow = new Date(today);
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks for the thorough diagnosis on #1705 — the root cause (date-only `YYYY-MM-DD` parsing as UTC midnight) is exactly right, and the local-parse approach here was correct.

Unfortunately this PR was overtaken by the Coming up sidebar redesign that landed on main yesterday (a5e59ee4): `formatUpcomingDateGroup` and its tests were removed, and day grouping now lives in `groupEventsByDay` in `src/components/UpcomingMeetings.tsx` — which still has the same UTC-midnight bug via `new Date(event.start_time)`. Since the function this PR patches no longer exists, the branch conflicts with main and can't be rebased meaningfully.

Closing in favor of a fix targeting the new grouping code, which will carry the same date-only local-parse approach proposed here. The fix for #1706 (#1708) is merged — thanks again for both reports.

PR #1708, opened 2026-08-18
fix(snippets): do not crash snippet expansion when the list is nullish

### LINKED ISSUE #1706: Snippet expansion crashes dictation when snippets is nullish

hsusul:
## Summary
- Dictation always runs `expandSnippets(text, getSettings().snippets)` after transcription.
- A null/undefined snippets list threw on `.length` and dropped the paste. Guard it the same way `getDictionaryHintWords` already does.

## Problem
On current `main`:

```js
expandSnippets("hello", null)
// TypeError: Cannot read properties of null (reading 'length')
```

Cloud/settings apply can call `setSnippets(settings.snippets)` when `snippets !== undefined`, including `null`.

## Root cause
`expandSnippets` assumed the settings store always holds an array. #1672 hardened the STT hint helper only.

## Fix
Treat a non-array or empty list as “no snippets” and return the transcript unchanged.

## Scope
- `src/utils/snippets.ts` and `test/utils/snippets.test.js`
- No settings-store or SQLite changes

## Tests

```bash
node --test test/utils/snippets.test.js
```

Fixes #1706

### DIFF
--- a/src/utils/snippets.ts
+++ b/src/utils/snippets.ts
@@ -52,8 +52,8 @@ function buildMatcher(snippets: Snippet[]): SnippetMatcher | null {
  * matcher is memoized against the snippets array reference (the settings
  * store replaces the array on every change).
  */
-export function expandSnippets(text: string, snippets: Snippet[]): string {
-  if (!text || snippets.length === 0) return text;
+export function expandSnippets(text: string, snippets?: Snippet[] | null): string {
+  if (!text || !Array.isArray(snippets) || snippets.length === 0) return text;
   if (snippets !== cachedSnippets) {
     cachedSnippets = snippets;
     cachedMatcher = buildMatcher(snippets);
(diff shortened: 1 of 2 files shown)


gabrielste1n (approved):
Verified locally against current main: merge is clean, all 15 snippet tests pass, typecheck passes. The failing Build and Notarize check is pre-existing Prettier drift on main in files this PR doesn't touch. Guard matches the getDictionaryHintWords pattern from #1672. Thanks for the focused fix!

PR #1709, opened 2026-08-18
fix(reasoning): survive Anthropic responses without a text block

### LINKED ISSUE #1653: Cleanup failed: Cannot read properties of undefined (reading 'trim') — intermittent, possibly related to #1417
**Describe the bug**
Using the BYOK Anthropic provider (Cloud Providers → Anthropic) with a user-supplied API key, dictation cleanup intermittently fails with:

Cleanup failed
Your dictation was pasted without AI cleanup.
Cannot read properties of undefined (reading 'trim')

This happens roughly 1 in every 2–3 dictations — not consistently reproducible on every attempt. The raw dictation still gets pasted, just without the AI cleanup pass.

**To Reproduce**
1. Go to Settings → Cloud Providers → Anthropic (Sonnet 5.0 in my case)
2. Enter your own Anthropic API key and save
3. Dictate as normal (repeat a few times)
4. Roughly every 2nd–3rd attempt: see error "Cleanup failed... Cannot read properties of undefined (reading 'trim')"

**Expected behavior**
Dictation cleanup should run successfully and consistently. If a value the cleanup step relies on is sometimes missing/undefined, it should be handled gracefully instead of throwing an unhandled `undefined.trim()` error.

**Screenshots**
Image

- OS: Windows 11
- App: OpenWhispr desktop
- Version: 1.8.1

N/A

**Additional context**
- Provider: BYOK Anthropic (own API key)
- Possibly related to #1417, which fixed a `temperature` parameter rejection on Sonnet 5+/Opus 4.7+ in the same BYOK Anthropic → cleanup request path. This could be a regression/edge case from that fix, or an unrelated bug hitting the same code path — flagging the connection in case it's useful for triage.
- Frequency: ~1 in 2–3 dictations, not every time.

AdityaPainuli:
Closes #1653

Chased the reporter's debug log: the crash comes from the `process-anthropic-reasoning` handler reading `data.content[0].text.trim()` directly. Anthropic's `content` is a block array. When the response hits max_tokens before any text, or leads with a non-text block, `content[0].text` is undefined and the handler throws the exact "Cannot read properties of undefined (reading 'trim')" from the issue. The stack in the attached log (ipcHandlers.js:4279 on 1.8.1) is this line.

The intermittency fits too: the failing run in that log spent 11.9s in reasoning vs 1.3-3.7s for the successful runs around it, which is what a truncated long generation looks like.

## Changes

- `src/helpers/anthropicResponse.js` (new): joins all text blocks from the response; returns null when there are none. A second helper builds a real error message naming the stop_reason.
- `src/helpers/ipcHandlers.js`: handler uses it. No text now surfaces as "Anthropic returned no text before hitting the max_tokens limit" instead of a TypeError, and the raw dictation still pastes as before.

## Test

- `test/helpers/anthropicResponse.test.js`, 11 cases. One runs the pre-fix expression against the failing shapes and asserts it throws TypeError; the helper handles the same shapes.
- `npm test`: no new failures vs main (same 78 pre-existing local failures on both). `npm run lint` and prettier clean.


Chadpiha (approved):
Approved. I reviewed the Anthropic response handling end to end across live dictation cleanup, streaming cleanup, selection edits, and History re-transcription. The shared handler now safely extracts all text blocks, handles empty or non-text response content with a useful error, preserves raw-transcript fallback, and retains complete-output enforcement where required. The additional data in #1653 matches the fixed failure path and does not expose another gap. After merging the latest main, all focused tests pass (11/11), and I found no actionable regressions.

Thanks for the contribution @AdityaPainuli !

PR #1710, opened 2026-08-18
fix(dictation): start the mic without waiting for the cloud STT config

### LINKED ISSUE #1673: Dictation hotkey delays ~4–5s when logged out — get-stt-config auth blocks recording start
**Describe the bug**
Pressing the dictation hotkey while logged out (using local models) delays the recording start by ~4–5 seconds. Debug shows a stt-config/auth attempt before the MediaRecorder opens.

**To Reproduce**
1. Use OpenWhispr while signed out and with local STT selected.
2. Press the dictation hotkey.
3. Observe ~4–5s gap between the shortcut press and "Recording started"; logs show a get-stt-config/auth resolution and a "Not authenticated" error.

**Expected behavior**
Recording should start immediately (or within configured warm-hold window) regardless of cloud auth state.

**Screenshots**
N/A

- OS: Linux (KDE/Wayland)
- Version: regression introduced between 1.8.1 and 1.8.3

**Additional context**
- Relevant files:
  - src/hooks/useAudioRecording.js — calls await window.electronAPI.getSttConfig before starting
  - src/helpers/cloudConfigRequest.js — getAuthHeader() is awaited and can delay, producing "Not authenticated"
- Root cause: start path awaits getSttConfig which awaits auth/cookie resolution; when logged out that lookup blocks the start.
- Suggested fix: do not block mic start on getSttConfig. Start capture immediately and apply STT config asynchronously, or use a short timeout/cached fallback.

AdityaPainuli:
Closes #1673

Traced the stall in the start path. `performStartRecording` awaits `getSttConfig()` before the MediaRecorder opens. Logged out, that IPC sits ~4-5s in auth/cookie resolution and then fails with "Not authenticated". But the config it waited for can only change the start decision in one case: signed-in OpenWhispr Cloud, where `shouldUseStreaming` reads `sttConfig.dictation.mode`. Local STT returns false on the first line of that function, signed-out returns false before ever reaching the config. So the wait was pure latency for exactly the users hitting it.

## Changes

- `src/helpers/sttConfigPolicy.js` (new): `needsSttConfigBeforeStart(settings)`, true only for signed-in OpenWhispr Cloud mode.
- `src/hooks/useAudioRecording.js`: fires the config fetch in the background either way, so it still lands for later recordings, and only awaits it when the predicate says it matters. Fetch failures log a warning instead of stalling the start.

Behavior for signed-in cloud users is unchanged: they still wait, and streaming vs batch is still decided from the config.

Verified by code trace, not on a live Linux session; the gating logic itself is unit-tested.

## Test

- `test/helpers/sttConfigPolicy.test.js`, 5 cases (local, signed-out, BYOK, signed-in cloud, missing settings).
- `npm test`: no new failures vs main. `npm run lint` and prettier clean.


Chadpiha (approved):
Thanks for this @AdityaPainuli :) 

Approved. I traced recording startup and STT provider selection across local, signed-out, BYOK, self-hosted, and signed-in OpenWhispr Cloud flows. The configuration request is removed from the microphone-start critical path when it cannot influence routing, while managed cloud still waits where configuration can select batch versus streaming. This remains compatible with the current main BYOK routing behavior. After merging the latest main, all focused tests pass (5/5), and I found no actionable regressions.

PR #1711, opened 2026-08-18
fix(hotkeys): fall back when Hyprland rejects the bind under the Lua config provider

### LINKED ISSUE #1675: Dictation hotkey silently fails on Hyprland with Lua config provider (hyprctl keyword rejected)

AdityaPainuli:
Closes #1675

The nasty part of this bug: hyprctl exits 0 when the compositor rejects the command. Under the Lua config provider it answers "keyword can't work with non-legacy parsers. Use eval." with a clean exit code. `registerKeybinding()` ran execFileSync with `stdio: "pipe"`, never read the output, and returned true unconditionally. So the app believed the hotkey registered, the Settings UI reported OK, and the fallback chain in `hotkeyManager` never got a chance to run.

## Changes

- `src/helpers/hyprctlResponse.js` (new): a keyword response counts as success only when every non-empty line is "ok". Empty output is tolerated so a hyprctl build that prints nothing on success doesn't regress to a permanently failing bind.
- `src/helpers/hyprlandShortcut.js`: captures the bind response and throws on rejection. The existing catch logs it and returns false, which activates the fallbacks that were already there: alternate keys via `tryNativeFallbacks`, then globalShortcut/portal. Persisting the bind to the config file is also skipped on rejection, so a dead bind no longer gets written to disk.

I can't run Hyprland with the Lua provider here, so the compositor-side behavior rests on the issue's repro (`hyprctl binds` stays empty after a 0-exit bind) plus the documented "ok" response contract. The response validation itself is unit-tested.

Writing binds into `hyprland.lua` directly would be the fuller fix for Lua-provider setups; this PR keeps scope to "stop lying about success and let the fallback chain work".

## Test

- `test/helpers/hyprlandShortcutResponse.test.js`, 4 cases, including the exact Lua rejection string and a mixed ok+error response.
- `npm test`: no new failures vs main. `npm run lint` and prettier clean.

### DIFF
--- a/src/helpers/hyprctlResponse.js
+++ b/src/helpers/hyprctlResponse.js
@@ -1,0 +1,16 @@
+// hyprctl exits 0 even when the compositor rejects the command, reporting the
+// failure only in its response text — e.g. under the Lua config provider,
+// `hyprctl keyword bind` prints "keyword can't work with non-legacy parsers.
+// Use eval." and the bind never registers (#1675). A successful keyword
+// command answers "ok" (one per request); anything else is a rejection. Empty
+// output is tolerated so a hyprctl build that prints nothing on success does
+// not regress to a permanently failing bind.
+function isHyprctlResponseOk(output) {
+  const lines = String(output ?? "")
+    .split("\n")
+    .map((line) => line.trim())
+    .filter(Boolean);
+  return lines.every((line) => line.toLowerCase() === "ok");
+}
+
+module.exports = { isHyprctlResponseOk };
--- a/src/helpers/hyprlandShortcut.js
+++ b/src/helpers/hyprlandShortcut.js
@@ -3,6 +3,7 @@ const fs = require("fs");
 const path = require("path");
 const os = require("os");
 const debugLogger = require("./debugLogger");
+const { isHyprctlResponseOk } = require("./hyprctlResponse");
 
 const DBUS_SERVICE_NAME = "com.openwhispr.App";
 const DBUS_OBJECT_PATH = "/com/openwhispr/App";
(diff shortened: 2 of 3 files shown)


Chadpiha:
Thanks for the focused fix and root-cause analysis. This has been superseded by #1664, which has now merged and closed #1675.  
The merged implementation includes the same important validation for hyprctl error responses that exit successfully, along with native Lua-provider registration through hyprctl eval, Lua persistence, cleanup/update handling, and broader test coverage. Since the complete fix is now on main, I’m closing this PR rather than rebasing and merging the overlapping implementation. Thanks again for the contribution.

PR #1713, opened 2026-08-18
feat(onboarding): theme the flow for dark mode and fix review findings

clarkmaan:
Moves onboarding off hardcoded light values onto a token set that both themes bind, then closes the correctness issues a review of this branch turned up.

## Theming

- `--onboarding-*` gains a dark block; components bind tokens instead of carrying `dark:` variants, and portalled Radix panels resolve them via `body:has(.onboarding-canvas)`
- Dropdown rows reuse `.onboarding-list-row`'s rounded hover slab, replacing the square full-bleed band whose corners fought the sheet's 17px radius
- Both hover slabs bind `--onboarding-row-hover`; they were still painting `neutral-100`, i.e. near-white on a `#141414` surface
- Compact hero becomes a baked Bayer dither with a per-theme strip

## Fixes

- **Control panel could stay permanently invisible.** `ready-to-show` no longer shows it, but `did-finish-load` still cleared the 10s backstop, so a renderer that loaded and then failed left a hidden window with no fallback.
- **`setOnboardingActive(true)` had no unmount cleanup**, so an ErrorBoundary catch left every hotkey gated until restart.
- **The "Self-hosted" row selected BYOK** and dropped the choice; it now seeds the self-hosted field, policy permitting.
- **Enterprise onboarding never applied its provider** to the cleanup and note-formatting scopes. Transcription is deliberately untouched: `enterprise` is not one of `TRANSCRIPTION_POLICY_CATALOG.modes`, so enterprise governs the LLM scopes only.
- **A denied system-audio prompt aborted calendar OAuth** entirely.
- **The dictation demo inserted its transcript twice**, once from the demo event and once from the real auto-paste into the focused demo textarea.
- **Expanded onboarding wants 910px of work area and never scrolled**, clipping fixed-height steps out of reach on a 1366x768-class display.

## Housekeeping

- Remove the kitchen-sink gallery and the seven assets nothing references
- Progress dots derive from the live route instead of a fixed four
- Translate the remaining onboarding copy across all nine locales, matching each file's register; replace ten German strings left in the Russian file
- Dev panel gains theme and language switches for QA

## Verification

`npm run quality-check` clean, 2122 tests / 0 failures, `vite build` succeeds. The dark-mode hover slab and the dropdown panel were checked in the running app over CDP.


boseq:
Retargeted through https://github.com/boseq/openwhispr/pull/2 so these commits merge into the head branch of #1670 rather than `main`.

The source branch was copied at exact SHA `349054e1f2ba356af4bb9899075ed3926908217f`; commit history and authorship are unchanged.

Issue #858, opened 2026-05-27
KDE Wayland: Portal paste method sends Ctrl+V into Claude Code terminal (needs Ctrl+Shift+V)


gcgriffith:
Environment

OS: Kubuntu 25.10 (KDE Plasma, Wayland session)
OpenWhispr version: 1.7.2
Cursor IDE (3.5.33) with Claude Code terminal using Claude Code for VS Code 2.1.152

Description
On KDE Wayland, OpenWhispr correctly detects the cursor window class but uses the linux-fast-paste --portal (RemoteDesktop D-Bus) method which sends Ctrl+V. The Claude Code terminal embedded inside Cursor IDE interprets Ctrl+V as an image paste command, briefly showing "No image found in clipboard" and not pasting the transcribed text.
Pasting manually with Ctrl+Shift+V into the Claude Code terminal works correctly.
Debug log evidence
"detectedWindowClass": "cursor"
Attempting linux-fast-paste --portal (RemoteDesktop D-Bus)
Paste successful { "tool": "linux-fast-paste", "method": "portal" }
Expected behaviour
When the target window is a terminal emulator or contains a terminal (such as the Claude Code terminal inside Cursor IDE), OpenWhispr should use Ctrl+Shift+V instead of Ctrl+V.
What I tried

Installed xdotool and ydotool — both working
ydotoold daemon running (root-owned at /tmp/.ydotool_socket)
Set YDOTOOL_SOCKET=/tmp/.ydotool_socket in ~/.profile and ~/.config/environment.d/ydotool.conf
Forced Cursor to run as XWayland via --ozone-platform=x11 in desktop file
Deleted portal token (~/.cache/openwhispr/portal-paste-token) — portal re-acquires token and continues using same method
Voice paste works correctly into Konsole and Kate, but not into Claude Code terminal inside Cursor
A paste to Cursor's Agent also works... just not to the calude code terminal within the Cursor IDE 

Suggested fix
Add cursor to the terminal window class detection list so OpenWhispr uses Ctrl+Shift+V instead of Ctrl+V when pasting into Cursor IDE. Alternatively, add a user-configurable option to force Ctrl+Shift+V for specific window classes.

xAlcahest:
#827 will solve this.

PR #1715, opened 2026-08-18
fix: avoid preemptive xdotool probes on Wayland

IdrisGit:
Defer XWayland target detection until an X11 fallback is needed, avoiding xdotool crashes during normal Wayland portal pastes.


gabrielste1n:
This fixes the early xdotool call, but it also brings back an older paste bug.

Because Electron detection is disabled on Wayland, Cursor and VS Code may no longer be recognized correctly on KDE. Their integrated terminals can then receive Ctrl+V instead of Shift+Insert. The terminal treats that as “paste image,” so the dictated text is not inserted. This reintroduces #858, which was previously fixed by #873.

Could we keep the delayed xdotool behavior while preserving Cursor/VS Code detection? A regression test for KDE + Cursor expecting the portal to use --shift-insert would also help.

IdrisGit:
@gabrielste1n I have pushed another commit, this keeps the delayed `xdotool` behavior while restoring Electron detection on KDE Wayland.

KDE now obtains the active window class and `PID` through `kdotool`, with the existing `KWin D-Bus` path as the fallback. The `PID` is passed to the existing Electron detection, so Cursor and VS Code again use `Shift+Insert` rather than `Ctrl+V` when pasting through the portal.

`xdotool` remains deferred until an XWayland fallback is actually needed. Also added the requested regression test for KDE + Cursor, asserting the portal receives --shift-insert.

PR #1725, opened 2026-08-19
fix(windows): restore the captured target window before pasting

### LINKED ISSUE #859: Prevent Input focus from being repositioned.

gabrielste1n:
Fixes #859

## Summary

On Windows, dictation now pastes back into the window the user was recording into, even when another window steals the foreground during transcription. The fast-paste binary gains a `--restore-window <hwnd>` mode that re-activates the captured target before sending the paste keystroke, and the JS side threads the window handle that main already captures at record start into the paste path.

Credit to PR #1227 by @stantheman0128 for the `RestoreForegroundWindow` design (AttachThreadInput foreground-lock workaround, minimized-window restore, verify-then-settle); this PR adopts that C implementation against current main and wires it to the existing capture seam instead of adding a new one.

## Root cause

The paste chain had no notion of a target on Windows:

- `windows-fast-paste.exe` called `GetForegroundWindow()` immediately before `SendInput`, pasting into whatever was foreground at paste time.
- The `paste-text` handler restores focus by PID on macOS only (#668); the win32 branch merely blurred the overlay — a no-op, since the overlay is `focusable: false` and shown via `showInactive`.
- Meanwhile main already captures the foreground HWND at every dictation start (toggle, PTT, voice agent) via `selectionManager`'s `--detect-only` probe (the hex `TARGET %p` line) — but only used it for selection-edit verification, never for the paste.

So a console flash, an alt-tab, or Chromium yanking focus to its own UI during transcription sent the dictated text to the wrong window.

## Fix

- **`resources/windows-fast-paste.c`**: new `--restore-window <hwnd>` argument, parsed with `strtoull(..., 16)` to match the hex `TARGET %p` format the binary itself prints in `--detect-only` mode. `RestoreForegroundWindow()` (adopted from #1227) runs before terminal detection, so the ctrl+v / ctrl+shift+v decision is made for the restored window, not the thief. The 20ms settle sleep only applies when a window switch actually happened — in the common case where the target never lost the foreground, the paste goes out with no added latency. If the target is gone or can't be restored, it falls through to the current foreground — exact pre-fix behavior. `--capabilities` now advertises `focus-restore-v1`.
- **`src/helpers/selectionManager.js`**: `getWinTargetHwnd()` exposes the captured HWND, draining an in-flight probe first (same wait loop as `captureSelectedText`) so a paste racing the stop-press probe waits for the answer instead of getting `null`.
- **`src/helpers/ipcHandlers.js`**: the `paste-text` handler passes `targetWindow` on win32, mirroring the macOS `activateTargetPid` path.
- **`src/helpers/clipboard.js`**: `pasteWindows`/`pasteWithFastPaste` forward the handle as `["--restore-window", hwnd]`. The nircmd/PowerShell fallbacks are untouched.

No new binary invocation is added at record start: the HWND comes from the probe main already fires on every dictation start. That also means the rollout is **dormant-safe** — an older cached `windows-fast-paste.exe` ignores the unknown flag and behaves exactly as before the fix, with no premature-paste hazard. The behavior activates once CI (`build-windows-fast-paste.yml`) ships the rebuilt exe.

## Tests

- `test/helpers/clipboardRestore.test.js`: `pasteWithFastPaste` passes `["--restore-window", "1A2B3C"]` when a target was captured (hex id pins the `TARGET %p` ↔ base-16 parse coupling), and no args when nothing was captured.
- `test/helpers/selectionManager.test.js`: `getWinTargetHwnd()` returns the hex HWND parsed from a mocked win32 `--detect-only` probe; waits for an in-flight `captureTarget` probe before answering; returns `null` with no capture or a non-Windows target.
- Full suite, typecheck, and lint pass. The new tests fail without the corresponding code changes.

The C change cannot be compiled or exercised in this repo's test setup, so **end-to-end verification on a real Windows desktop is required** before this can be considered done: record from an input field (e.g. Brave, Windsurf), alt-tab away or let focus get stolen during transcription, and confirm the paste returns to the original field — including the minimized-window and closed-window cases.

## Limitations

- **Intra-window caret moves are not fixable by window-level restore.** If the target window stays foreground but moves its internal focus (e.g. Chromium shifting focus to its own toolbar/settings button, as in the original report's Brave case), the window is already foreground, the restore is a no-op, and Ctrl+V still lands wherever the app's caret is. Fixing that would require per-control focus tracking (UI Automation), out of scope here.
- `SW_RESTORE` will un-minimize a window the user deliberately minimized mid-transcription — consistent with nircmd's `win activate` semantics, and arguably correct since that's where they dictated.
- If focus was stolen before the stop press in toggle mode, the stop-press re-probe can capture the thief — the same trust model as the selection-edit capture already on main.
- Until the rebuilt exe ships (and for users with a stale cached copy), the flag is ignored and behavior is unchanged.


stantheman0128:
Thanks for the credit, and for wiring this to the existing `--detect-only` capture instead of adding a second probe.

The intra-window Chromium caret case is the right limitation to spell out. Window restore is a no-op when the top-level HWND never changed.

PR #1733, opened 2026-08-19
fix(network): treat the full IPv6 link-local range as private HTTP

### LINKED ISSUE #1731: Self-hosted HTTP allowlist misses most of IPv6 link-local (fe80::/10)
## Describe the bug

`isSecureHttpEndpoint()` allows HTTP only to private hosts so LAN LLM/STT servers work without TLS. IPv6 link-local is checked with `hostname.startsWith("fe80")`.

Link-local is **fe80::/10** (hextets `fe80`–`febf`), not the single hextet `fe80`. `urlAudioDownloader.js` already uses `/^fe[89ab]/` for SSRF. The self-hosted allowlist does not, so `http://[fe90::1]:8080` is rejected even though it is link-local.

`http://[fe80::1]:8080` still works (covered by the existing test).

## Affected component

- `src/utils/urlUtils.ts` — `isPrivateHost` / `isSecureHttpEndpoint`
- Callers: self-hosted transcription / OpenAI-compatible base URLs (`OpenAICompatiblePanel`, `transcriptionRoute`, etc.)
- Tests: `test/helpers/secureEndpoint.test.js`

## To Reproduce

On current `main`:

```js
isSecureHttpEndpoint("http://[fe80::1]:8080")  // true
isSecureHttpEndpoint("http://[fe90::1]:8080")  // false
isSecureHttpEndpoint("http://[fea0::1]:8080")  // false
isSecureHttpEndpoint("http://[febf::1]:8080")  // false
isSecureHttpEndpoint("http://[fe7f::1]:8080")  // false (not link-local)
```

## Expected behavior

Any fe80::/10 address is treated as private, matching `isPrivateIp` in `urlAudioDownloader.js`.

## Actual behavior

Only the `fe80` hextet is allowed; the rest of the link-local range requires HTTPS.

## Environment

- OpenWhispr `main`
- Linux/macOS IPv6 LAN; self-hosted Whisper/llama on a link-local address

## User impact

A self-hosted endpoint on `fe90::…` / `fea0::…` / `febf::…` cannot be saved/used over HTTP. The same machine on `fe80::1` works.

## Root-cause hypothesis

The private-host check copied the common `fe80` prefix instead of RFC 4291 fe80::/10. The downloader SSRF path was later widened; this allowlist was not.

## Proposed focused scope

- Treat IPv6 link-local as fe80::/10 in `isPrivateHost`
- Extend `test/helpers/secureEndpoint.test.js`
- Do not change download SSRF, enterprise SSRF, or IPv4 rules

I intend to work on this.

hsusul:
## Summary
- Self-hosted HTTP is allowed only on private hosts. IPv6 link-local was checked with `startsWith("fe80")`.
- Link-local is fe80::/10 (`fe80`–`febf`). LAN servers on `fe90::` / `fea0::` / `febf::` were rejected. Align with `isPrivateIp` in `urlAudioDownloader.js`.

## Problem
On current `main`:

| URL | Before |
| --- | --- |
| `http://[fe80::1]:8080` | allowed |
| `http://[fe90::1]:8080` | rejected |
| `http://[fea0::1]:8080` | rejected |
| `http://[febf::1]:8080` | rejected |
| `http://[fe7f::1]:8080` | rejected (not link-local) |

## Root cause
The allowlist used the common `fe80` prefix instead of RFC 4291 fe80::/10. The audio-download SSRF path already matches `/^fe[89ab]/`.

## Fix
Use the same fe80::/10 prefix test for IPv6 hosts in `isPrivateHost`. Unique-local `fc`/`fd` is unchanged.

## Scope
- `src/utils/urlUtils.ts` and `test/helpers/secureEndpoint.test.js`
- No download SSRF, enterprise SSRF, or IPv4 changes

## Tests

```bash
node --test test/helpers/secureEndpoint.test.js
```

Fixes #1731


Chadpiha (approved):
Approving — nice catch, @hsusul

You're right that `startsWith("fe80")` only covered a slice of the link-local range: fe80::/10 spans fe80–febf, and your `/^fe[89ab]/` matches it exactly (while correctly leaving out deprecated site-local fec0::/10). It also matches how `urlAudioDownloader.js` has classified link-local all along, so the two classifiers now agree on both the allow side and the block side — a nice consistency win.

Everything the wider range admits is non-routable reserved space, so the plain-HTTP gate stays safe. Clean, minimal, well-tested. Thank you — merging!

Chadpiha (approved):
Approving — nice catch, @hsusul. 🎉

You're right that `startsWith("fe80")` only covered a slice of the link-local range: fe80::/10 spans fe80–febf, and your `/^fe[89ab]/` matches it exactly (while correctly leaving out deprecated site-local fec0::/10). It also matches how `urlAudioDownloader.js` has classified link-local all along, so the two classifiers now agree on both the allow side and the block side — a nice consistency win.

Everything the wider range admits is non-routable reserved space, so the plain-HTTP gate stays safe. Clean, minimal, well-tested. Thank you — merging!

PR #1734, opened 2026-08-19
fix(ai): retry HTTP 408 request timeouts in the API retry strategy

### LINKED ISSUE #1732: API retry strategy never retries HTTP 408 Request Timeout
## Describe the bug

`createApiRetryStrategy()` retries network errors (no status), 429, and 5xx. HTTP **408 Request Timeout** is a transient 4xx (the server did not receive a complete request in time; RFC 9110 clients may retry) but is treated like 400/404 and never retried.

Cleanup, dictation-agent, and other `withRetry(..., createApiRetryStrategy())` callers therefore fail a flaky 408 on the first response instead of backing off.

## Affected component

- `src/utils/retry.ts` — `createApiRetryStrategy`
- Callers: `ReasoningService`, OpenAI/Gemini/Tinfoil inference providers
- Tests: `test/helpers/apiRetryStrategy.test.js`

## To Reproduce

On current `main`:

```js
const { shouldRetry } = createApiRetryStrategy();
shouldRetry(Object.assign(new Error("timed out"), { status: 408 }))
// false

shouldRetry(Object.assign(new Error("rate limited"), { status: 429 }))
// true

shouldRetry(Object.assign(new Error("unavailable"), { status: 503 }))
// true
```

## Expected behavior

408 retries like 429/5xx. Other 4xx (400, 401, 404, 409, 422) stay non-retryable.

## Actual behavior

408 is a hard failure.

## Environment

- OpenWhispr `main`
- Any cloud/BYOK reasoning call that surfaces `error.status === 408` (proxies, gateways, overloaded APIs)

## User impact

A single request-timeout from the API fails dictation cleanup / agent with no retry, even though the same helper already retries 429 and 503.

## Root-cause hypothesis

The 4xx comment assumed every client error is deterministic. 408 is a timeout, not a permanent rejection.

## Proposed focused scope

- Retry status 408 (and still 429 + 5xx)
- Add a unit test next to the existing 429 case
- No download retry, cloud-chunk, or i18n changes

I intend to work on this.

hsusul:
## Summary
- `createApiRetryStrategy()` retried network errors, 429, and 5xx, but not HTTP 408 Request Timeout.
- Cleanup / dictation-agent / other `withRetry` callers failed on the first 408 instead of backing off.

## Problem
On current `main`:

| Status | Retried? |
| --- | --- |
| no status (network) | yes |
| 400 / 404 | no |
| 408 | **no** |
| 429 | yes |
| 503 | yes |

RFC 9110 allows clients to retry 408; it is a timeout, not a deterministic rejection.

## Root cause
The 4xx branch treated every client error as permanent. 408 is transient, like 429.

## Fix
Retry 408 alongside 429 and 5xx. Leave other 4xx non-retryable.

## Scope
- `src/utils/retry.ts` and `test/helpers/apiRetryStrategy.test.js`
- No download retry, cloud-chunk upload, or i18n changes

## Tests

```bash
node --test test/helpers/apiRetryStrategy.test.js
```

Fixes #1732

### DIFF
--- a/src/services/ai/inferenceProviders/openai.ts
+++ b/src/services/ai/inferenceProviders/openai.ts
@@ -210,8 +210,10 @@ export const openaiProvider: InferenceProvider = {
       });
     }
 
+    const retryStrategy = createApiRetryStrategy();
     const response = await withRetry(async () => {
       let lastError: Error | null = null;
+      let lastRetryableError: Error | null = null;
 
       for (const { url: endpoint, type } of endpointCandidates) {
         const controller = new AbortController();
@@ -296,6 +298,9 @@ export const openaiProvider: InferenceProvider = {
             throw new Error("Request timed out after 30s");
           }
           lastError = error as Error;
+          if (retryStrategy.shouldRetry(lastError)) {
+            lastRetryableError = lastError;
+          }
           if (type === "responses") {
             logger.logReasoning("OPENAI_ENDPOINT_FALLBACK", {
               attemptedEndpoint: endpoint,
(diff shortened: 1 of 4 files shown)


Chadpiha (approved):
Thanks @hsusul for identifying this edge case and keeping the fix focused. I reviewed the retry flow end to end, including the OpenAI Responses-to-Chat fallback. With the follow-up preserving transient errors across fallback candidates, HTTP 408 now retries as intended while existing network, 429, 5xx, and permanent 4xx behavior remains intact.
The regression coverage, full test suite, and typecheck pass. Approved. Thanks again for the contribution!

Issue #1735, opened 2026-08-19
CI: two workflows both define a job named `tests`; only one runs the format gate on `main`


xlurie:
Two workflows each define a job named `tests`. Only one of them runs the format gate, and that one is also the only test gate on pushes to `main` — it has been failing since 2026-08-15, while the identically-named job next to it stays green. In the meantime the same trivial cleanup was authored three separate times. This is a report about the wiring, not about the unformatted files, which people keep fixing.

*(Links pinned to `e42a1ab`, current `main` at the time of writing.)*

### The collision

- [`.github/workflows/tests.yml`](https://github.com/OpenWhispr/openwhispr/blob/e42a1abc92eda3f33b27c60e6216ef2cdc51d171/.github/workflows/tests.yml) — `on: pull_request` only. Job `tests`: `npm ci --ignore-scripts`, `npm rebuild better-sqlite3`, `npm run typecheck`, `npm test`.
- [`.github/workflows/build-and-notarize.yml`](https://github.com/OpenWhispr/openwhispr/blob/e42a1abc92eda3f33b27c60e6216ef2cdc51d171/.github/workflows/build-and-notarize.yml) — `on: push` (`main`, `develop`) + `pull_request` (into `main`), both with `paths-ignore`, plus `workflow_dispatch`. Job `tests`: the same install/rebuild/test steps **plus** `npm run quality-check` (= `format:check && typecheck`, replacing the standalone typecheck step; `format:check` itself is `eslint . && cd src && eslint . && prettier --check "**/*.{js,jsx,ts,tsx,json,css,md}"`), `npm run i18n:check` and `npm audit`.

A comment on that job's `npm rebuild` step puts it plainly: *"This job is the only test gate on push to main."*

So the second is a superset of the first — but in a PR's check list they render as `Tests / tests` and `Build and Notarize / tests`. Same name, different scope. When the narrow one is green and the broad one is red, the red one reads like noise.

### What it has cost so far

Since `73ac9bb7` (2026-08-15T17:47Z) every `Build and Notarize` push run on `main` has failed, all of them on the `Run quality checks` step. The last green one was `b3a8368b`, 2026-08-15T03:02Z. During that window the same four-file Prettier cleanup was written three times:

- **#1703** (opened 2026-08-18T15:15, still open) — a standalone `prettier --write` on `SettingsPage.tsx`, `PasteToolsInfo.tsx`, `WorkspacesService.ts`, `snippets.ts`.
[...]

PR #1743, opened 2026-08-20
chore(format): fix prettier drift in windowConfig.js

AdityaPainuli:
Round two of #1712, same disease, new file.

`npm run format:check` fails on `main` again: `src/helpers/windowConfig.js` landed unformatted via #1718, merged after #1712 went in. Every open PR shows the red quality-check again because of it. I rebased my three fix PRs (#1709, #1710, #1711) onto current main yesterday expecting green and got this instead, so I traced it the same way.

This PR is `prettier --write` on that one file. +1/-5, wrapping only.

## Test

- `npm run quality-check` (the failing CI command): fails on clean main, passes with this diff.

Might be worth making the quality-check job a required status check, since drift keeps landing through PRs merged while it's red. Happy to leave that call to you.

### DIFF
--- a/src/helpers/windowConfig.js
+++ b/src/helpers/windowConfig.js
@@ -16,11 +16,7 @@ function resolveOverlayWindowType({ role, platform, linuxSession }) {
 
   // Sway asks wlroots whether an unmanaged XWayland surface wants focus.
   // "toolbar" opts in; "notification" keeps the existing text field focused.
-  if (
-    linuxSession.isSway &&
-    linuxSession.xwaylandAvailable &&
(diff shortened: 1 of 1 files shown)


Chadpiha:
Thank you, @AdityaPainuli, for catching and fixing this formatting failure.  
After checking against the latest main, src/helpers/windowConfig.js already contains the exact Prettier change from this PR, so merging it now would be a no-op. I’m closing this as superseded—not because the fix was incorrect.  
The underlying CI visibility and enforcement problem remains tracked in #1735 and #1745. Thanks again for the contribution and for helping keep CI healthy.

PR #1745, opened 2026-08-20
fix(ci): stop the two tests jobs from sharing a name and disagreeing

AdityaPainuli:
Refs #1735

I ran into this from the other side before the issue existed: three of my PRs (#1709, #1710, #1711) showed a green "tests" next to a red "tests", and it took a log dive to learn they're different workflows with different scopes. #1735 documents the full cost, including one formatting cleanup that got written three separate times.

## Changes

- `.github/workflows/build-and-notarize.yml`: job `tests` renamed to `quality`, the three `needs:` refs updated. A red gate now renders as "Build and Notarize / quality" instead of as the twin of the green unit-test job. The `main` ruleset has no `required_status_checks` referencing the old name, per the issue's own check; if some org-level config pins it, that needs a settings touch-up on merge.
- `.github/workflows/tests.yml`: the standalone PR job becomes globally unique `pr-quality`; Typecheck becomes `npm run quality-check` (format:check + typecheck), and translation plus dependency-audit checks are added. It remains unfiltered, so normal and documentation-only PRs both emit the authoritative gate.

After merge, `pr-quality` can safely become the required status check. `Build and Notarize / quality` remains the internal prerequisite for every platform build and the full gate on pushes to `main`.

## Test

- Both YAMLs parse and the job graph is intact: `quality` + build-linux/build-windows/build-macos all `needs: quality`, while `tests.yml` keeps its single `pr-quality` job.
- Node 24 local verification: `quality-check`, `i18n:check`, `npm audit` (0 vulnerabilities), and the DB-backed suite (3,217 passing, 0 failing).
- The PR check list supplies the final workflow-wiring evidence: distinct `Build and Notarize / quality` and `Tests / pr-quality` checks.


xlurie:
Author of #1735 here — verified this at head `9388e46`: it lands levers 1 and 3 from the issue (the name collision and the format-gate asymmetry). No stale `needs:` remains in the two workflows, both bases match current `main`, and the fix already shows on this PR's own Checks tab — `Build and Notarize / quality` and `Tests / tests` render as distinct runs. (The third `tests` job, in `release.yml`, is tag-only and never shares a check page with these two — out of scope here.)

The case it retires: on #1663 (head `852eff29`), `Build and Notarize / tests` failed while `Tests / tests` passed, started one second apart — same name, opposite verdicts, and the failure was `main`'s own prettier drift in `helpers/windowConfig.js`, not that PR's code. Post-merge a red gate carries its own name and is attributable at a glance. Both reds on this head are that identical drift — the one-file fix in #1743, which also explains the ordering: merged the other way around, `Tests / tests` goes red on every open PR until the drift lands.

Re-ran the ruleset check today: still no `required_status_checks` at the repository level pinning `tests` (org-level config isn't visible to me — the one thing worth a glance on merge). On `Closes #1735`: auto-close works for me; I'll take the required-check lever (2) separately once the new names have landed. One residual to expect: `quality` also runs `npm audit`, so it can still go red alone on a new advisory — outside both this diff and the collision.

Chadpiha (approved):
Thank you, @AdityaPainuli, for isolating the duplicate-name problem and closing the format-gate mismatch with a focused change.

I refreshed this branch onto current main and verified the merged head locally with Node 24: quality-check, i18n:check, npm audit (0 vulnerabilities), and the DB-backed suite (3,217 passing, 0 failing). The job graph also retains all three platform dependencies on quality.

I changed Closes #1735 to Refs #1735 so the issue stays open until the required-check ruleset is configured and verified.

PR #1747, opened 2026-08-20
feat: add RTL text support for transcripts, notes, and chats

### LINKED ISSUE #1402: RTL support for Arabic, Persian, and Urdu
Issue:
While the underlying OpenWhispr model can accurately transcribe languages such as Arabic, Persian, and Urdu, the user interface completely lacks the concept of text directionality. RTL texts are displayed left-aligned within a Left-to-Right (LTR) layout, leading to broken formatting and misplaced punctuation.

Current State in the Code:
• The SUPPORTED_UI_LANGUAGES array in src/i18n.ts does not include any RTL languages.
• The dir="rtl" attribute is not used anywhere in the project, meaning the application does not apply any directional rules.

Proposed Solution:
Establish RTL support based on the existing i18n system through the following steps:
1. Directionality Logic: Add a function that maps the language code to its direction (e.g., 'ar' returns 'rtl') instead of manual hardcoding.
2. Root Direction: Apply the dir attribute at the App Root level so the entire structure inherits it automatically when the language changes.
3. CSS Logical Properties: Use margin-inline-start and text-align: start instead of left and right to ensure the layout flips automatically.
4. Transcript Direction: Enable transcripts to be displayed in the correct direction even if the app UI is in English, as the dictation language may differ from the UI language.
5. Adding Languages: Include Arabic (and later Persian and Urdu) in the supported languages.

Rejected Alternatives:
• Auto-detection Only: Unreliable for mixed content; it is better as a complement to explicit settings.
• Separate CSS files for RTL: Increases maintenance difficulty; Logical Properties are a better and lighter solution.
• Keeping the Current State: Makes reading the output very difficult for users of these languages.

Additional Note:
I need to reach an agreement on point four specifically: should the text direction follow the UI language, the selected dictation language, or be detected per transcript? I am ready to work on a PR as soon as the methodology is approved.

IdrisGit:
## What
Users writing in RTL languages (Arabic, Hebrew, Persian, Urdu) saw broken alignment in transcripts, notes, and chats.

## How
Scoped to **user-generated content only** (UI stays LTR). Containers now use `dir="auto"` so the browser picks direction per element from the text itself, and mirror-sensitive utilities were switched to **tailwind v4** logical variants (`ps-/pe-`, `ms-/me-`, `border-s-`, `rounded-bs/be/ts/te`).

- `languageRegistry.json`: `"dir": "rtl"` for `ar`, `he`, `fa`, `ur`
- `languageSupport.ts`: add `getLanguageDirection()` helper
- `input`/`textarea`: default `dir="auto"`
- `dir="auto"` on chat bubbles/input, transcript history + raw text, live preview, note title/body, markdown, meeting segments, conversation/note list titles, toasts

fixes #1402

### DIFF
--- a/src/components/TranscriptionPreviewOverlay.tsx
+++ b/src/components/TranscriptionPreviewOverlay.tsx
@@ -322,7 +322,10 @@ export default function TranscriptionPreviewOverlay() {
                     : "border-border/25 bg-background/30",
               ].join(" ")}
             >
-              <p className="select-text text-[13px] leading-[1.52] text-foreground whitespace-pre-wrap break-words [text-wrap:pretty]">
+              <p
+                className="select-text text-[13px] leading-[1.52] text-foreground whitespace-pre-wrap break-words [text-wrap:pretty]"
+                dir="auto"
+              >
                 {activeText}
               </p>
             </div>
--- a/src/components/chat/ChatInput.tsx
+++ b/src/components/chat/ChatInput.tsx
@@ -114,7 +114,7 @@ export function ChatInput({
     <div className="shrink-0 px-3 pb-3 pt-1">
       <div
         className={cn(
-          "flex items-center gap-2 min-h-11 pl-4 pr-1.5 rounded-full",
+          "flex items-center gap-2 min-h-11 ps-4 pe-1.5 rounded-full",
           GLASS_SURFACE,
           "border border-black/10 dark:border-white/14",
           "transition-all duration-200",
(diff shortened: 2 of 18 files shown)


greatcoat:
Closing — RTL shipped in v1.10.0 via #1869, covering the same ground: document direction follows the active locale, layouts use logical properties throughout, and `dir="auto"` is applied to transcripts, notes, chat and live preview. Arabic also landed as a full UI language.

Thanks for getting this moving and for scoping it properly to user-generated content.

PR #1753, opened 2026-08-20
fix(calendar): fail OAuth loopback immediately on state mismatch

### LINKED ISSUE #1750: Calendar OAuth hangs for 120s after a callback with a mismatched state
## Describe the bug

Google/Microsoft calendar connect uses `runOAuthLoopbackFlow()` with an ephemeral `127.0.0.1` server.

If the provider (or a second tab) returns with an authorization `code` but a `state` that does not match the in-flight PKCE state, the handler writes `400 Invalid request` and **returns without `cleanup()` or `reject()`**. The promise stays pending until `OAuth_TIMEOUT_MS` (120s). The neighboring `error=` query path already cleans up and rejects immediately.

A request with no `code` (browser `favicon.ico`, bare `/`) must keep waiting — rejecting those would abort a real in-flight login.

## Affected component

- `src/helpers/oauthLoopbackFlow.js` — `runOAuthLoopbackFlow`
- Callers: `googleCalendarOAuth.js`, `microsoftCalendarOAuth.js`

## To Reproduce

On current `main` (electron mocked, `shell.openExternal` no-op):

```js
const flow = runOAuthLoopbackFlow({
  errorParam: "gcal_error",
  buildAuthUrl: (redirectUri) => redirectUri,
  handleCallback: async () => ({}),
});
// wait until listen(), then:
await fetch(`${redirectUri}/?code=stolen&state=wrong`);
// flow is still pending after 400ms; it only rejects at 120s with "OAuth flow timed out"
```

Observed on `4348e424`: HTTP 400, body contains `Invalid request`, `Promise.race` vs 400ms timeout → `still-pending`.

## Expected behavior

A callback that includes `code` but the wrong `state` should close the server and reject the flow immediately (same as the `error=` branch). Requests without a `code` should still only get 400 and leave the flow running.

## Actual behavior

The browser shows "Invalid request" and the app keeps "connecting" until the 120s timeout.

## Environment

- OpenWhispr `main`
- Calendar connect (Google or Microsoft)

## User impact

A CSRF-mismatched or stale-tab callback leaves calendar connect spinning for two minutes instead of failing fast.

## Root-cause hypothesis

The `!code || returnedState !== state` branch sends 400 but never calls `cleanup()` / `reject()`. Timeout is the only exit.

## Proposed focused scope

- If `code` is present and `state` does not match, `cleanup()` and reject
- Leave no-`code` requests as 400-and-wait (favicon / empty GET)
- Do not change token exchange, PKCE, or hosted callback HTML
- Add a unit test with a mocked `electron.shell`

I intend to work on this.

hsusul:
## Summary

- Calendar connect (Google/Microsoft) uses an ephemeral `127.0.0.1` PKCE loopback server.
- A callback that includes `code` but the wrong `state` already returned HTTP 400; the promise kept waiting until the 120s timeout.
- Reject those attempts immediately, matching the existing `error=` branch. Requests with no `code` (favicon / bare GET) still only get 400 and keep waiting.

Fixes #1750

## Problem

After a stale-tab or CSRF-mismatched redirect, the browser shows "Invalid request" while the app stays on "connecting" for two minutes.

## Root cause

`runOAuthLoopbackFlow` treated `!code || returnedState !== state` as a 400 and returned without `cleanup()` or `reject()`. Only `error=` and the timeout closed the server.

## Fix

If a `code` is present and `state` does not match, call `cleanup()` and reject with `OAuth state mismatch`. Leave no-code requests as 400-and-wait so a favicon fetch cannot abort a real login.

## Behavior before vs after

| Callback | Before | After |
| --- | --- | --- |
| `?code=…&state=wrong` | 400, hang 120s | 400, reject immediately |
| `/favicon.ico` (no code) | 400, keep waiting | unchanged |
| `?error=access_denied` | reject immediately | unchanged |
| matching `code` + `state` | success | unchanged |

## Scope & non-goals

- Only `oauthLoopbackFlow.js` and a new unit test.
- Does not change token exchange, PKCE generation, or hosted callback HTML.

## Tests

- `test/helpers/oauthLoopbackFlow.test.js` (mocked `electron.shell`)

## Validation

- `ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/oauthLoopbackFlow.test.js` → PASS (3 tests)
- `npx prettier --check src/helpers/oauthLoopbackFlow.js test/helpers/oauthLoopbackFlow.test.js` → PASS

### DIFF
--- a/src/helpers/oauthLoopbackFlow.js
+++ b/src/helpers/oauthLoopbackFlow.js
@@ -57,8 +57,17 @@ function runOAuthLoopbackFlow({ buildAuthUrl, handleCallback, errorParam }) {
     const codeVerifier = crypto.randomBytes(32).toString("base64url").slice(0, 43);
     const codeChallenge = crypto.createHash("sha256").update(codeVerifier).digest("base64url");
     const state = crypto.randomBytes(32).toString("hex");
+    let callbackClaimed = false;
 
     const server = http.createServer(async (req, res) => {
+      // Accepted requests can outlive server.close(), so only the first
+      // terminal callback may settle the flow or exchange a code.
+      if (callbackClaimed) {
+        res.writeHead(400, { "Content-Type": "text/html" });
+        res.end("<html><body><h3>Invalid request.</h3></body></html>");
+        return;
+      }
+
       try {
         const url = new URL(req.url, `http://127.0.0.1`);
         const returnedState = url.searchParams.get("state");
@@ -66,6 +75,7 @@ function runOAuthLoopbackFlow({ buildAuthUrl, handleCallback, errorParam }) {
         const error = url.searchParams.get("error");
 
         if (error) {
+          callbackClaimed = true;
           redirect(res, { [errorParam]: error });
           cleanup();
           reject(new Error(`OAuth error: ${error}`));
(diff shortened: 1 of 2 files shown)


Chadpiha (approved):
Approved ✅

I traced the Google and Microsoft calendar OAuth journeys through the shared
loopback helper. The original state-mismatch fix and follow-up lifecycle
hardening now ensure:

- Code-bearing state mismatches reject immediately.
- Harmless no-code requests leave the flow running.
- The first terminal callback exclusively owns the flow.
- Overlapping malformed, mismatched, or duplicate callbacks cannot settle the
  flow or start another token exchange.
- Catch and timeout paths claim the flow before closing/rejecting, preventing
  accepted late callbacks from creating a ghost connection.
- Existing provider-error and successful authorization behavior remains intact.

The new regressions exercise real localhost connections, including requests
accepted before `server.close()`. Removing either catch/timeout ownership
assignment causes its corresponding test to fail.

Thanks for the contribution :)

PR #1757, opened 2026-08-21
fix(auth): show Apple sign-in on Windows and Linux

### LINKED ISSUE #1748: Windows/Linux: "Continue with Apple" is hard-gated to macOS, so Apple-created accounts have no way to sign in
**Problem.** The Apple sign-in button renders only on macOS. On Windows and Linux it is not disabled or hidden — it is never put in the DOM. A user who created their account with Sign in with Apple therefore has no route into the desktop app on those platforms except discovering the password-reset flow themselves.

**Fix.** Remove the `isMacOS &&` wrapper around the Apple button so it renders on every platform, matching Google and Microsoft. The backend, the OAuth route and the Windows deep-link plumbing already support it; this is UI-only.

## Evidence

`openwhispr-desktop/src/components/AuthenticationStep.tsx` @ `4d1a6eaf`

```
:94   const isMacOS = getCachedPlatform() === "darwin";
:593  {isMacOS && (            ← wraps the entire Apple <Button>, closes :618
:620  <Button … handleSocialSignIn("google")>      ← no wrapper
:645  <Button … handleSocialSignIn("microsoft")>   ← no wrapper
```

Google and Microsoft have byte-identical `disabled` props and no platform condition. Apple is the only gated provider.

## The server has never had this restriction

`openwhispr-api/api/desktop-signin/[provider].ts:5` — `ALLOWED_PROVIDERS = new Set(["google", "microsoft", "apple"])`, with no platform check anywhere in the handler.

Verified live against production:

```
$ curl -sI "https://auth.openwhispr.com/api/desktop-signin/apple?callbackURL=…"
HTTP 302 → appleid.apple.com/auth/authorize
   client_id     = com.gizmolabs.openwhispr.signin
   scope         = email name
   response_mode = form_post
```

`src/lib/auth.ts:203-217` builds one shared URL with no per-provider branching, and Windows deep-link registration is complete (`main.js:210/212`, `:231`, `:1737-1746`; `electron-builder.json:5-8`). Token storage has no platform branch either.

## Why it is like this

Deliberate scoping when the feature shipped, not a technical limit. Commit `ffc16fcc` (2026-04-30, Gabriel Stein), *feat(auth): Sign in with Apple on macOS (#691)*:

> "Apple is already wired into better-auth on the API side; **the desktop OAuth deep-link flow is provider-agnostic**, so this is a UI-only change. … Render an Apple button above Google/Microsoft, **gated to darwin**"

## Repro

1. Create an account with Sign in with Apple (any platform).
2. Install the desktop app on Windows or Linux and open the sign-in screen.
3. Only Google, Microsoft and email/password are offered. No Apple button, and no message explaining why.

## Impact

Every Windows and Linux user whose account was created through Apple. No UI string anywhere tells them Apple is macOS-only, so it reads as broken rather than unsupported. For scale, a code comment in `openwhispr-api/lib/teams-launch.ts:27-29` records 1,056 production users on `privaterelay.appleid.com` — all Apple-created identities. That is an upper bound on exposure, not a count of affected users, but it is not small.

## Notes for whoever picks this up

* Check the Apple button's `disabled` condition still makes sense off-macOS — it gates on `oauthProtocolRegistered`, which is what deep-link registration sets.
* Worth a look at the silent-failure path while in here: `src/lib/auth.ts:215-216` fires `openExternalLink` and returns `{}` without awaiting an outcome, so any failure after the browser opens can never surface. Separate issue if it is real.

---

Found while investigating [CUS-121](https://linear.app/openwhispr/issue/CUS-121/apple-app-store-subscription-not-recognised-on-windows-app). The customer hit exactly this, set a password as a workaround, and told us the Apple button was missing — he was right.

zhongwater123:
## Summary

- show Apple alongside the other OAuth providers on every desktop platform
- remove the now-unused renderer platform check
- verify through the current `AuthenticationStep` renderer harness that Apple renders with a Linux platform mock and dispatches `signInWithSocial("apple")`

## Maintenance update

- synchronized the branch with `main@834a0771fea146bd3e0b693abd9193cbd288705c`
- preserved the latest email discovery, onboarding resume-draft, SSO, bidirectional-text, portal, and Nucleo icon changes
- kept the regression on the project's shared renderer test infrastructure
- mutation boundary remains explicit: restoring the macOS-only gate fails at `Apple sign-in should render on Linux`

## Testing

Run on Windows with Node.js 24.19.0:

- `node --import tsx --test test/components/authenticationStep.test.js` — 1 passed
- ESLint for `AuthenticationStep.tsx` and its renderer test — passed
- Prettier check for both PR files — passed
- `cd src && tsc --noEmit` — passed
- Node 24 renderer production build — passed

The full suite was not rerun during this maintenance update. The focused behavior test, typecheck, lint/format checks, and production renderer build are current against the synchronized branch; official PR CI provides the repository-wide signal.

Closes #1748


boseq (approved):
Tested this in the real app on Linux and Windows, comparing 1.10.2 with the same install running this branch's JS.

- Ubuntu 26.04 (GNOME, x64 deb): 1.10.2 shows Google, Microsoft and SSO. This branch adds Apple. Clicking it opened the browser on Apple's "sign in to OpenWhispr" page.
- Windows 11: same result on the sign-in screen, Apple now shows between Google and Microsoft. I couldn't get a click through in the VM, so the browser step is only verified on Linux.

Outside the app: the new test fails on main and passes here. A renderer harness with the protocol unregistered shows Apple disabled with the same hint as Google and Microsoft.

Two small things in the test, neither blocking:

- `AuthenticationStep` no longer imports `utils/platform`, so `harness.platform` has no effect and the darwin/win32/linux loop runs the same code three times. A fixed `"linux"` mock is enough, since it's what makes a restored macOS gate fail (without it the real `getPlatform()` falls back to `"darwin"` under Node).
- Nothing covers the disabled state when the protocol isn't registered. It could go in the same test: stub `getOAuthProtocolRegistered` to resolve `false` and check the Apple tile's `disabled` and `title`.

Also, the description says a restored gate fails at "should render on Linux". With the loop, it fails first at "should render on win32".

I didn't do a full Apple sign-in, since I don't have a test Apple account.

boseq (approved):
Pushed 99df8d3 with the two test changes from my earlier review: a fixed Linux platform mock instead of the loop, plus a check that Apple is disabled with the protocol hint when `openwhispr://` isn't registered. It fails on main at "Apple sign-in should render on Linux", and removing the protocol gate from the tiles fails it too. I updated the description to match. CI is green.

Issue #943, opened 2026-06-12
This tool needs an auto update, option box, in the settings.


bhowiebkr:
I use this tool basically updates multiple times every day, and I want to turn off the auto update card that pops up in the top right, it's really annoying. This tool gets way too many updates. Like, literally every time I launch a tool, it wants to update. I go for lunch and the tool wants to update. I go pick up my kids from school and the tool wants to update. I go to the bathroom and take a dump and the tool wants to update. 

It's too much for me. Maybe, a silent update feature? or, just push, one update a day or something? Maybe just use the, Windows push notification thing, instead of having a continuous, widget, in the top right, of windows all the time when there an update?.

gabrielste1n:
@bhowiebkr will add a disable button in the settings :) coming in next release!

gabrielste1n:
you can finally turn off this notification in v1.7.3 via preferences :)

PR #1760, opened 2026-08-21
feat(updater): add setting to disable auto-install on quit

### LINKED ISSUE #1460: Setting to disable silent install-on-quit after an update has downloaded (`autoInstallOnAppQuit`)

xlurie:
## Summary

`autoInstallOnAppQuit` is hardcoded `true` (`src/updater.js:68`), so a downloaded update installs silently on quit — a version change the user never approved. This adds one toggle in Settings → System → Updates: **Install updates automatically**, default **ON** — zero behavior change for every existing install. When OFF, a downloaded update waits for an explicit **Install & Restart**. Fixes #1460. Scope is deliberately the toggle only: the update check and the manual Download button are untouched, and a ready-to-install banner can follow up separately (#1694 explores adjacent prompt UX). It completes the update preferences alongside the notification toggle (#943) and the check toggle (#1662).

This revives @xAlcahest's #1529, which was closed unmerged in his queue cleanup without ever receiving review: the implementation is his (credited via Co-authored-by). I rebased it onto current `main` (it applies cleanly), re-verified the mechanism against the vendored electron-updater 6.8.9, and re-ran all the gates below. Happy to close this in favor of a reopened #1529 if he prefers.

## How

- `autoUpdater.autoInstallOnAppQuit` is driven by a persisted `UPDATE_AUTO_INSTALL` env setting (only the literal string `"false"` disables — anything else, including unset, stays ON). Env persistence rather than the renderer settings store because the flag must be correct in the main process at boot: the userData `.env` is loaded before `UpdateManager` is constructed, while a renderer store hydrates only when the settings UI mounts.
- Electron-free policy helper `src/helpers/updateInstallPolicy.js` (the `updateCheckPolicy.js` pattern from #1662); IPC pair + preload bridge + Settings toggle following the auto-start toggle idiom; strings in all ten locales.
- Runtime setter: electron-updater 6.8.9 latches quit-handler registration at download time (`BaseUpdater.executeDownload` → `addQuitHandler()`, which returns without registering when the flag is off) and re-checks the flag inside the hook at quit time. Plain assignment therefore covers disabling; enabling after a download that completed while gated re-registers via `addQuitHandler()` (a protected API — guarded with a `typeof` check).

## Platform notes

- **Windows/Linux:** the flag is honored whether flipped before or after a download (quit-time re-check + the setter's re-register).
- **macOS** (source read of `MacUpdater`, not runtime-tested): there is no `addQuitHandler`; with the flag ON, Squirrel staging starts at download-done, and as far as the source shows a staged update installs on relaunch even if the toggle is flipped off afterwards. With the flag OFF at download time, staging waits for **Install & Restart** (`quitAndInstall` handles the `!autoInstallOnAppQuit` case). The toggle fully governs any download that starts after it is flipped, so it stays visible on all platforms. Worth noting: the `!autoInstallOnAppQuit` branches in `MacUpdater` were unreachable while the flag was hardcoded — this PR activates them for opted-out users.

## Tests

The three test files from #1529, re-run on current `main` (22 tests): policy-helper unit tests (env parsing, including the `"true"`/unset cases); install-gate behavior on a `Module._load` mock harness in the style of `test/helpers/updater.test.js` (gate at setup, runtime flips, re-register fires only for enable-after-gated-download, tolerates a missing `addQuitHandler`); an env round-trip through a tmpdir-userData `EnvironmentManager` including a simulated restart.

`npm run typecheck`, `npm run lint`, `npm run quality-check`, `npm run i18n:check`, and `npm run build:renderer` all pass; the full `npm test` delta vs clean `main` on the same machine is exactly the 22 new passing tests (pre-existing local failures identical on both refs).


xlurie:
@xAlcahest — correcting one sentence in this description. I offered to close this in favour of a reopened #1529. That was not available when I wrote it: #1529 closed on 2026-08-19 together with your other open PRs when their head repository went away (the API reports none for it now, and the fork 404s for me), and I did not check that before offering. Whether a Reopen route still exists is something you can see and I cannot.

The commits are still in this repo: `refs/pull/1529/head` resolves to `adf53864`, both of yours intact (`git fetch https://github.com/OpenWhispr/openwhispr refs/pull/1529/head && git checkout -b <branch> FETCH_HEAD`). If you would rather carry this yourself, open a PR from them and I close #1760 — no reason needed. If you would rather have your name on the commits here instead of in a trailer, one word and I replace my squash with your two commits rebased onto `main`. If neither, nothing to do: the implementation is yours and stays credited via Co-authored-by whichever way this lands. Apart from the rebase, my diff differs from `adf53864` by small edits in three files.

For the choice: rebased onto current `main`, merges clean at `a9cf27b4`, the gates in the description re-run on this rebase; no review yet. This needs no answer.

greatcoat:
Closing — superseded by #2120 in v1.10.0, which adds the toggle you asked for under Settings → System → Updates as part of a wider updater rework: the update-available popup window is removed entirely and the sidebar carries the update instead.

Thanks to you and @xAlcahest for the groundwork here, and for the careful note on why the flag has to be correct in main at boot.

PR #1768, opened 2026-08-21
fix(notes): format future timestamps as dates

### LINKED ISSUE #1767: fix(notes): do not label future timestamps as now
### Component
`src/utils/dateFormatting.ts` — `formatRelativeTime`

### Reproduction
With the current `upstream/main` implementation:

```js
formatRelativeTime(new Date(Date.now() + 2 * 60 * 60 * 1000).toISOString(), key => key)
// "notes.list.timeNow"
```

### Actual behavior
Any timestamp in the future produces the `notes.list.timeNow` label because its negative minute difference is less than one. A note timestamp two hours ahead therefore appears to have been updated now.

### Expected behavior
Future timestamps should not use an elapsed-time label such as “now”, “minutes ago”, or “hours ago”. They should fall back to the existing short-date presentation, just as timestamps older than seven days do.

### User impact
Clock skew, delayed sync metadata, or a future-dated imported record can make note recency misleading in the notes UI.

### Root cause
The minute threshold checks only `minutes < 1`; it does not distinguish negative (future) elapsed time from a timestamp less than one minute old.

### Scope
I intend to add deterministic coverage and make the smallest formatting-only fix. Existing past-time labels and invalid-date behavior will remain unchanged.

### Environment
Node.js 24.9.0; reproducible with the pure TypeScript utility without Electron, credentials, network, or downloaded models.

hsusul:
## Summary

Prevent future note timestamps from being displayed as “now”.

## Problem

`formatRelativeTime` treated every negative elapsed-minute value as less than one minute, so a timestamp two hours in the future returned the `notes.list.timeNow` label.

## Root cause

The elapsed-time thresholds did not distinguish future timestamps from timestamps that are less than one minute old.

## Solution

Return the existing short-date representation for future timestamps. Past relative-time labels and invalid-date handling remain unchanged.

## Tests

- Added deterministic coverage with a mocked clock for a timestamp two hours in the future.
- Existing date-grouping tests continue to pass.

## Validation

- `node --test test/utils/dateFormatting.test.js test/utils/dateGrouping.test.js` — PASS (5 tests).
- `npm run lint` — PASS.
- `npm run format:check` — PASS.
- `npm run typecheck` — PASS.
- `npm run i18n:check` — PASS.
- `npm run build:renderer` — PASS.

## Compatibility and risk

This changes only future timestamps; all existing past-time thresholds and the short-date formatter are preserved. No API shape or locale keys changed.

## Documentation

No documentation change is needed for this internal display correction.

Fixes #1767. Duplicate searches found no existing issue or pull request for future relative timestamps.

### DIFF
--- a/src/utils/dateFormatting.ts
+++ b/src/utils/dateFormatting.ts
@@ -30,6 +30,7 @@ export function formatRelativeTime(
   const date = normalizeDbDate(dateStr);
   if (Number.isNaN(date.getTime())) return "";
   const diff = Date.now() - date.getTime();
+  if (diff < 0) return formatShortDate(dateStr);
   const minutes = Math.floor(diff / 60000);
   const hours = Math.floor(diff / 3600000);
   const days = Math.floor(diff / 86400000);
--- a/test/utils/dateFormatting.test.js
+++ b/test/utils/dateFormatting.test.js
@@ -1,0 +1,14 @@
+const test = require("node:test");
+const assert = require("node:assert/strict");
+
+const load = () => import("../../src/utils/dateFormatting.ts");
+
+test("future timestamps use a calendar date instead of the now label", async (t) => {
+  t.mock.timers.enable({ apis: ["Date"], now: Date.parse("2026-08-21T12:00:00.000Z") });
+  const { formatRelativeTime } = await load();
+
+  assert.notEqual(
+    formatRelativeTime("2026-08-21T14:00:00.000Z", (key) => key),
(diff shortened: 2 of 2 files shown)


Chadpiha (approved):
Thanks @hsusul !

This is a focused fix for future timestamps: it falls back to the localized short date instead of producing a misleading negative relative time. The strengthened regression test asserts the exact output and fails if the production guard is removed, while existing current and past timestamp behavior remains covered. Good to merge.

PR #1769, opened 2026-08-21
fix(notes): keep note content visible above the floating chat panel

IdrisGit:
## Summary
Opening the floating chat panel covered the bottom of the note/transcript, forcing a manual scroll to see the latest content. This reserves scroll space for the panel and auto-reveals the content bottom on open/close.

## Changes
- Reserve panel space via an inherited `--floating-inset` CSS variable written by a callback ref + ResizeObserver — no React re renders on panel resize (NoteEditor)
- On open, force-pin the active view's scroller to its content bottom; on panel resize mid-session, re-pin only if already near the bottom so reading position isn't yanked (NoteEditor)
- Transcript scroller uses `pb-[var(--floating-inset,96px)]`: closed keeps the original 96px bottom-bar clearance, open reserves only panel height + gap (MeetingTranscriptChat); editor views use the same var with a 0px fallback (RichTextEditor)
- Cleanup removes the var instead of zeroing it so fallbacks apply on close; the browser clamp returns the scroll to the natural bottom
- Fold the chat auto-scroll ResizeObserver into the existing layout effect in `useStickToBottom` (re-pins when streamed content height settles late)

## Testing
I have tested it locally and it's working fine
image

### DIFF
--- a/src/components/notes/EmbeddedChat.tsx
+++ b/src/components/notes/EmbeddedChat.tsx
@@ -22,6 +22,8 @@ interface EmbeddedChatProps {
   activeConversationId?: number | null;
   onSwitchConversation?: (id: number) => void;
   onNewChat?: () => void;
+  /** Floating panel ref; NoteEditor reserves scroll space with it. */
+  floatingPanelRef?: React.Ref<HTMLDivElement>;
 }
 
 function EmptyState() {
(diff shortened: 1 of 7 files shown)


Chadpiha (approved):
Approved, thanks @IdrisGit . My follow-up fixes address the three layout issues:
- Pinned note/transcript content now follows viewport resizing.
- Panel growth no longer pulls readers who scrolled away from the bottom.
- Maximum-height chat preserves visible note content.

PR #1772, opened 2026-08-21
fix(chat): persist conversation migration links

### LINKED ISSUE #1770: Conversation migration does not persist cloud acknowledgement
## Problem

The renderer's one-time assistant conversation migration creates cloud conversations but only updates the Zustand store with the returned cloud ID. It never persists that ID through the existing `markConversationSynced` IPC bridge.

## Reproduction

On a clean `upstream/main` checkout, use a fake `window.electronAPI` with one local conversation whose `cloud_id` is null and a successful cloud create response. Call `startConversationMigration()` twice. Both calls select the same local row as unsynced and issue a cloud create; no `markConversationSynced` call occurs.

## Expected behavior

After a successful migration, the local SQLite row should be linked and marked synced, matching the normal `SyncService` conversation path. A later launch should not migrate the same conversation again.

## Impact and scope

Users with existing assistant history can re-upload the same history on every migration/restart, or receive duplicate/conflicting cloud records depending on server idempotency. The focused fix should persist the acknowledgement and preserve the existing cloud create behavior.

I intend to submit the focused fix and deterministic regression test.

hsusul:
Fixes #1770

## Summary

- persist the cloud acknowledgement during assistant conversation migration
- avoid overwriting newer in-memory conversation state with a stale migration snapshot
- delete a cloud orphan when the local row disappears during migration
- add a deterministic fake-IPC migration regression test

## Problem

`startConversationMigration` created cloud conversations but only updated the Zustand store with the returned cloud ID. It did not call the existing `markConversationSynced` SQLite bridge, so the local row remained unsynced and the migration repeated on later launches.

## Fix

Persist the acknowledgement through `markConversationSynced`, preserve the current renderer row when applying the returned ID, and retire the cloud row if the local row was deleted while the request was in flight.

## Validation

- `node --test test/stores/chatStoreMigration.test.js` — passed
- `git diff --check` — passed
- `npm test` — pre-existing environment/loader failures on clean upstream baseline: 2,894 tests, 1,769 passed, 1,117 failed, 7 skipped; failures were unrelated module-export and renderer/native test failures

No generated files, dependencies, or unrelated paths were changed.


Chadpiha (approved):
Approved ✅

I traced the conversation sync flow end to end, including an in-flight cloud create receiving newer local messages, competing cloud links, retries, orphan cleanup, and subsequent no-op syncs.
Create acknowledgements are now transactionally checked against the original snapshot. Newer local state remains pending and retries with the complete message history, while losing or orphaned cloud creates are retired. The unused duplicate migration path was also removed, leaving one authoritative sync flow.
The new race regressions, existing conversation pull/tombstone coverage, full test suite, typecheck, and all CI/platform builds pass. Good to merge, thank you!

PR #1776, opened 2026-08-22
fix(windows): block untrusted file navigation

hsusul:
## Summary

The control-panel navigation guard treated every `file://` URL as trusted. In packaged builds, `DevServerManager.getAppUrl(true)` returns `null`, so the old handler also dereferenced that value while installing the guard.

This adds a small navigation policy helper that keeps the development app route and DevTools available, rejects arbitrary file URLs, and safely handles the packaged-app null URL. The regression tests cover both the packaged file-navigation boundary and development behavior.

## Validation

- `NODE_OPTIONS='--import=tsx' node --test test/helpers/windowManagerNavigation.test.js`
- `npm run typecheck`
- `npm run i18n:check`
- `npm run build:renderer`
- `git diff --check`
- `npm run format:check` reaches an existing upstream Prettier failure in `src/components/notes/floatingChatLayout.ts`; ESLint passes.

### DIFF
--- a/src/helpers/navigationGuard.js
+++ b/src/helpers/navigationGuard.js
@@ -1,3 +1,5 @@
+// will-navigate also fires for renderer-initiated location.reload() with the
+// window's own URL (appUrl), so that URL must stay allowed.
 function isAllowedAppNavigation(url, appUrl) {
   if (url.startsWith("devtools://")) return true;
   if (!appUrl) return false;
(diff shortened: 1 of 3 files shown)


Chadpiha (approved):
Thanks @hsusul, this was a good find!

You were right on both counts: `getAppUrl(true)` returns null in packaged builds, so the old handler threw on every `will-navigate` and the guard effectively never ran in production — and blanket `file://` trust was too broad.

I pushed a commit on top that adjusts one behavior: renderer-initiated `location.reload()` fires `will-navigate` with the window's own `file://` URL (sign-out, onboarding restart, and the error-boundary reload all rely on it), so the guard now derives the packaged index.html URL via `getAppFilePath()` and allows exactly that URL (query/hash ignored) while blocking every other `file://` target. Blocked non-web URLs are also dropped instead of handed to `shell.openExternal`. Tests cover the reload, foreign-file, and scheme-spoof cases, and the full suite is green. Approving — thanks for catching the fail-open!

Issue #1774, opened 2026-08-22
[BUG] Windows: Parakeet transcription fails when Windows username contains non-ASCII characters


o101010:
**Describe the bug**
Local Parakeet transcription fails on Windows when the user's Windows username/path contains non-ASCII characters.

OpenWhispr passes the Parakeet model paths to the bundled `sherpa-onnx-ws-win32-x64.exe`. On Windows, sherpa-onnx incorrectly decodes non-ASCII characters in these paths.

For example, my Windows username contains `ö`:

C:\Users\Hölla\...

The path is passed to sherpa-onnx, but the character is corrupted and the resulting path cannot be found.
`'C:\\Program Files\\OpenWhispr\\resources\\bin\\sherpa-onnx-ws-win32-x64.exe' '--tokens=C:\\Users\\H÷lla\\.cache\\openwhispr\\parak'

I reproduced the underlying issue independently with the latest sherpa-onnx v1.13.6 Windows x64 release using `sherpa-onnx-offline-websocket-server.exe`.

An ASCII-only path works correctly, while a path containing `ö` or `ë` fails:

`sherpa-onnx` interprets the latter as: `C:\Temp\H÷lla-test` and reports that the file does not exist.

I allready open an issue on [sherpa-onnx github](https://github.com/k2-fsa/sherpa-onnx/issues/3885)

I think there are two possible approaches for OpenWhispr:

1. Sanitize/relocate the local Parakeet model path to an ASCII-only directory before passing it to sherpa-onnx, avoiding the Windows Unicode path issue.
2. Wait for sherpa-onnx to properly fix Unicode path handling on Windows, then update the bundled sherpa-onnx binary in OpenWhispr.

The second option would be preferable as the long-term solution, but **the first would prevent other unexpected behaviours on unknown bugs in Onnx.**

**To Reproduce**

1. Install local LLM Parakeet TDT 0.6B
2. Press your shortcut to execute dictation
3. You have an error message with Onnx failed to initialized

**Expected behavior**
Voice dictation should start normally and the audio should be transcribed using the local Parakeet TDT 0.6B model.

**Screenshots**

<img width="344" height="327" alt="Image" src="https://github.com/user-attachments/assets/07f39b78-787a-4b5e-929b-31d6d8699c82" />

**Desktop (please complete the following information):**

- OS: Windows 11 25h2
- Browser : N/A
- Version : 1.8.3

**Additional context**
[...]

PR #1777, opened 2026-08-22
fix(cli): enforce request body byte limit

hsusul:
## Summary

The CLI bridge advertises a 1 MiB request body limit, but the guard compared JavaScript character count instead of UTF-8 byte count. Multibyte payloads could therefore exceed the intended limit before parsing and dispatch.

This changes the check to `Buffer.byteLength(raw, "utf8")` and adds a regression test proving a sub-1 MiB character-count payload whose encoded bytes exceed 1 MiB is rejected with HTTP 400 before the database mutation runs.

## Validation

- `NODE_OPTIONS='--import=tsx' node --test test/helpers/cliBridgeRequestBody.test.js`
- `npm run typecheck`
- `npm run i18n:check`
- `npm run build:renderer`
- `git diff --check`
- `npm run format:check` reaches an existing upstream Prettier failure in `src/components/notes/floatingChatLayout.ts`; ESLint passes.

### DIFF
--- a/src/helpers/cliBridge.js
+++ b/src/helpers/cliBridge.js
@@ -30,9 +30,14 @@ async function findAvailablePort() {
 function readJsonBody(req) {
   return new Promise((resolve, reject) => {
     let raw = "";
+    // Count wire bytes per chunk: string concatenation decodes each Buffer
+    // chunk, so measuring the decoded string would rescan the whole body on
+    // every chunk and miscount sequences split across chunk boundaries.
+    let receivedBytes = 0;
     req.on("data", (chunk) => {
       raw += chunk;
-      if (Buffer.byteLength(raw, "utf8") > MAX_REQUEST_BODY_BYTES) {
+      receivedBytes += Buffer.byteLength(chunk);
+      if (receivedBytes > MAX_REQUEST_BODY_BYTES) {
         reject(new Error("Request body too large"));
         req.destroy();
       }
--- a/src/helpers/cliBridge.js
+++ b/src/helpers/cliBridge.js
@@ -32,7 +32,7 @@ function readJsonBody(req) {
     let raw = "";
     req.on("data", (chunk) => {
       raw += chunk;
-      if (raw.length > MAX_REQUEST_BODY_BYTES) {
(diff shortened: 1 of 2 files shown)


Chadpiha (approved):
Approved — the fix is correct and verified end-to-end. `raw.length` counted UTF-16 code units, so multibyte payloads could reach ~4x the intended 1 MiB cap before rejection; the byte-based check closes that. Traced the rejection path: `readJsonBody` rejects → `_handleRequest` maps it to `400 validation_error` before any route handler or DB write runs, and the regression test pins exactly that (char count < 1 MiB, bytes > 1 MiB, no `saveNote` call).

I've pushed a small follow-up commit onto this branch: the limit is now tracked incrementally (`receivedBytes += Buffer.byteLength(chunk)`) instead of re-measuring the accumulated string on every chunk — one pass per chunk, and exact even when a multibyte sequence is split across chunk boundaries. Tests, Prettier, and ESLint all pass with it.

One note for the record: the branch name references #1774, but that issue (Windows Parakeet failure with non-ASCII usernames) is unrelated and remains open — this PR intentionally does not close it.

Chadpiha (approved):
Approved — the fix is correct and verified end-to-end. `raw.length` counted UTF-16 code units, so multibyte payloads could reach ~4x the intended 1 MiB cap before rejection; the byte-based check closes that. Traced the rejection path: `readJsonBody` rejects → `_handleRequest` maps it to `400 validation_error` before any route handler or DB write runs, and the regression test pins exactly that (char count < 1 MiB, bytes > 1 MiB, no `saveNote` call).

I've pushed a small follow-up commit onto this branch: the limit is now tracked incrementally (`receivedBytes += Buffer.byteLength(chunk)`) instead of re-measuring the accumulated string on every chunk — one pass per chunk, and exact even when a multibyte sequence is split across chunk boundaries. Tests, Prettier, and ESLint all pass with it.

One note for the record: the branch name references #1774, but that issue (Windows Parakeet failure with non-ASCII usernames) is unrelated and remains open — this PR intentionally does not close it.

PR #1780, opened 2026-08-23
fix(diarization): guard nullish inputs and format non-finite timestamps in speakerMerge

### LINKED ISSUE #1779: splitIntoSentences and formatTimestamp in speakerMerge produce invalid output on nullish or non-finite inputs
### Describe the bug

In `src/helpers/speakerMerge.js`:

1. `splitIntoSentences(text)` passes `text` directly to `new Intl.Segmenter(undefined, { granularity: "sentence" }).segment(text)`. When `text` is `null` or `undefined`, `Intl.Segmenter` coerces the argument to string (`"null"` or `"undefined"`), causing `splitIntoSentences(null)` to return `["null"]`, and `splitIntoSentences(undefined)` to return `["undefined"]`. In environments without `Intl.Segmenter`, it throws `TypeError: text.split is not a function`.
2. `formatTimestamp(seconds)` does not sanitize non-finite or negative numbers. When `seconds` is `undefined`, `NaN`, or `Infinity`, it formats timestamps as `"NaN:NaN"`. When `seconds` is negative (e.g. `-5`), `Math.floor` produces negative components (`"-1:-5"`).
3. `formatSpeakerTranscript(mergedSegments)` calls `mergedSegments.map(...)` directly without verifying that `mergedSegments` is an array, throwing a `TypeError` when called with `null` or `undefined`.

### To Reproduce

1. Call `splitIntoSentences(null)` -> returns `["null"]`.
2. Call `splitIntoSentences(undefined)` -> returns `["undefined"]`.
3. Call `formatTimestamp(undefined)` -> returns `"NaN:NaN"`.
4. Call `formatTimestamp(-5)` -> returns `"-1:-5"`.
5. Call `formatSpeakerTranscript(null)` -> throws `TypeError: Cannot read properties of null (reading 'map')`.

### Expected Behavior

- `splitIntoSentences` should return `[]` for nullish, non-string, or whitespace-only inputs.
- `formatTimestamp` should treat negative, nullish, or non-finite values as 0 seconds (`"0:00"`).
- `formatSpeakerTranscript` should safely return `""` for nullish or empty arrays.

hsusul:
Fixes #1779

### Problem
In `src/helpers/speakerMerge.js`:
1. `splitIntoSentences(text)` passed `text` directly to `Intl.Segmenter`, which coerces `null` or `undefined` into string values (`"null"` / `"undefined"`), causing `splitIntoSentences(null)` to return `["null"]`. In environments without `Intl.Segmenter`, calling `splitIntoSentences(null)` threw a `TypeError`.
2. `formatTimestamp(seconds)` did not sanitize non-number, non-finite, or negative inputs, returning `"NaN:NaN"` on undefined/NaN/Infinity or negative minute/second components like `"-1:-5"` on negative numbers.
3. `formatSpeakerTranscript(mergedSegments)` threw `TypeError: Cannot read properties of null (reading 'map')` when `mergedSegments` was nullish.
4. `mergeSpeakersWithText(segments, text)` produced segments with `text: null` when text was null.

### Solution
- Added explicit string and empty-string guards at the top of `splitIntoSentences`.
- Sanitized `seconds` in `formatTimestamp` so that non-finite, sub-zero, or nullish inputs safely default to 0 seconds (`"0:00"`).
- Added array guard in `formatSpeakerTranscript` to safely return `""` for nullish or empty inputs.
- Sanitized input text in `mergeSpeakersWithText` to default to `""`.
- Added comprehensive regression tests in `test/helpers/speakerMerge.test.js`.

### Verification
- `node --test test/helpers/speakerMerge.test.js` (passes, 17/17 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/speakerMerge.js
+++ b/src/helpers/speakerMerge.js
@@ -1,4 +1,7 @@
 function splitIntoSentences(text) {
+  if (typeof text !== "string" || text.trim().length === 0) {
+    return [];
+  }
   if (typeof Intl !== "undefined" && Intl.Segmenter) {
     try {
       const segmenter = new Intl.Segmenter(undefined, { granularity: "sentence" });
@@ -16,9 +19,13 @@ function splitIntoSentences(text) {
 }
 
 function formatTimestamp(seconds) {
-  const hrs = Math.floor(seconds / 3600);
-  const mins = Math.floor((seconds % 3600) / 60);
-  const secs = Math.floor(seconds % 60);
+  const totalSeconds =
+    typeof seconds === "number" && Number.isFinite(seconds) && seconds > 0
+      ? Math.floor(seconds)
+      : 0;
+  const hrs = Math.floor(totalSeconds / 3600);
+  const mins = Math.floor((totalSeconds % 3600) / 60);
+  const secs = totalSeconds % 60;
   if (hrs > 0) {
     return `${hrs}:${String(mins).padStart(2, "0")}:${String(secs).padStart(2, "0")}`;
   }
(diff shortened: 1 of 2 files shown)


Chadpiha:
Thanks @hsusul! Closing with #1779 — this is a case where the validation you're after already exists, one layer up at the real boundary.

The `merge-speaker-text` IPC handler validates everything before these helpers run: it rejects non-array segments, non-string text, and non-finite durations, caps sizes, and re-sanitizes each segment (`speaker` → sliced string or `"unknown"`, `start`/`end` → finite number or 0). `formatTimestamp`'s only other caller feeds it internal constants that are `Number.isFinite`-guarded at the call site. So the helpers can only ever see sanitized input, and duplicating the checks one layer deeper is the pattern our conventions steer away from (see the note on #1840).

Thanks for the attention to the diarization path! 🙏

PR #1782, opened 2026-08-23
fix(dictionary): split bulk import text on carriage return line breaks

### LINKED ISSUE #1781: parseDictionaryImportText fails to split on carriage return line breaks
### Describe the bug

In `src/helpers/dictionaryImport.js`, `parseDictionaryImportText(text)` splits bulk import text using `text.split(/[,\n]/)`.

When text contains carriage return line endings without line feeds (`\r`), such as from clipboard paste on older formats or raw carriage returns (`"Alice\rBob\rCarol"`), `/[,\n]/` fails to split the text. As a result, the entire text containing embedded `\r` characters is treated as a single dictionary entry (e.g. `"Alice\rBob\rCarol"`), corrupting the imported custom dictionary words.

### To Reproduce

1. Call `parseDictionaryImportText("Alice\rBob\rCarol")`.
2. Observe output: `["Alice\rBob\rCarol"]`.

### Expected Behavior

`parseDictionaryImportText` should split on `\r`, `\n`, `\r\n`, and commas, cleanly returning `["Alice", "Bob", "Carol"]`.

hsusul:
Fixes #1781

### Problem
In `src/helpers/dictionaryImport.js`, `parseDictionaryImportText` used `.split(/[,\n]/)` to parse bulk-imported dictionary words.

When pasted text used carriage return line endings without line feeds (`\r`), such as from clipboard sources or older format pastes (e.g. `"Alice\rBob\rCarol"`), `/[,\n]/` failed to split the words. As a result, the entire text was kept as a single corrupted entry containing embedded `\r` characters.

### Solution
- Updated the regex split pattern to `/[,\r\n]/` so that carriage return line breaks (`\r`), line feeds (`\n`), CRLF sequences (`\r\n`), and commas all split words cleanly.
- Added regression tests in `test/helpers/dictionaryImportFormat.test.js` covering `\r`, `\r\n`, and mixed delimiters.

### Verification
- `node --test test/helpers/dictionaryImportFormat.test.js` (passes, 4/4 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/dictionaryImport.js
+++ b/src/helpers/dictionaryImport.js
@@ -7,7 +7,7 @@
  */
 export function parseDictionaryImportText(text) {
   return String(text ?? "")
-    .split(/[,\n]/)
+    .split(/[,\r\n]/)
     .map((w) => w.trim())
     .filter(Boolean);
 }
--- a/test/helpers/dictionaryImportFormat.test.js
+++ b/test/helpers/dictionaryImportFormat.test.js
@@ -21,3 +21,15 @@ test("import accepts mixed commas and new lines and skips blanks", async () => {
     "Carol",
   ]);
 });
+
+test("import accepts carriage return and CRLF line endings", async () => {
+  const { parseDictionaryImportText } = await load();
+  assert.deepEqual(parseDictionaryImportText("Alice\rBob\rCarol"), ["Alice", "Bob", "Carol"]);
+  assert.deepEqual(parseDictionaryImportText("Alice\r\nBob\r\nCarol"), ["Alice", "Bob", "Carol"]);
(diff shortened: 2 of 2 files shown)


Chadpiha:
Thanks @hsusul! Closing with #1781 — good news on this one: Windows `\r\n` paste already imports cleanly today. The split on `\n` leaves `"word\r"`, and the existing `.trim()` on each entry strips the `\r` (your CRLF test assertions pass against main as-is). And bare-CR separators can't reach the parser in practice: the only caller reads from a textarea, whose value normalizes CR/CRLF to LF per the HTML spec before the parser ever sees it.

If we ever add file-based dictionary import, this one-character hardening would become genuinely useful — happy to revisit then. Thanks! 🙏

PR #1794, opened 2026-08-23
fix(meeting): handle nullish inputs in partitionPendingMicFinals and isWithinRetractWindow

### LINKED ISSUE #1793: partitionPendingMicFinals and isWithinRetractWindow throw TypeError on nullish inputs
### Describe the bug

In `src/helpers/meetingMicHoldback.js`:

1. `partitionPendingMicFinals({ pending, now, force = false, isDuplicate })` iterates over `pending` directly with `for (const entry of pending)`. When `pending` is `null` or `undefined`, it throws `TypeError: pending is not iterable`.
2. When `isDuplicate` is omitted or not a function, calling `isDuplicate(entry)` throws `TypeError: isDuplicate is not a function`.
3. `isWithinRetractWindow({ candidate, systemTimestamp, windowMs })` accesses `candidate.timestamp` directly without checking whether `candidate` is `null` or `undefined`, throwing `TypeError: Cannot read properties of null/undefined (reading 'timestamp')`.

### To Reproduce

1. Call `partitionPendingMicFinals({ pending: null, now: 1000 })` -> throws `TypeError: pending is not iterable`.
2. Call `partitionPendingMicFinals({ pending: [{ releaseAt: 0 }], now: 1000 })` -> throws `TypeError: isDuplicate is not a function`.
3. Call `isWithinRetractWindow({ candidate: null, systemTimestamp: 1000, windowMs: 500 })` -> throws `TypeError`.

### Expected Behavior

- `partitionPendingMicFinals` should default nullish `pending` to `[]` and safely check `typeof isDuplicate === "function"`.
- `isWithinRetractWindow` should return `false` when `candidate` is nullish or timestamps are non-finite numbers.

hsusul:
Fixes #1793

### Problem
In `src/helpers/meetingMicHoldback.js`:
1. `partitionPendingMicFinals` iterated over `pending` without checking for nullish values, throwing a `TypeError: pending is not iterable`.
2. When `isDuplicate` was omitted or not a function, calling `isDuplicate(entry)` threw a `TypeError`.
3. `isWithinRetractWindow` accessed `candidate.timestamp` directly without checking whether `candidate` was nullish or whether timestamps and windows were valid finite numbers, throwing a `TypeError`.

### Solution
- Defaulted `pending` to `[]` in `partitionPendingMicFinals` and added a `typeof isDuplicate === "function"` check.
- Added nullish and `Number.isFinite` checks in `isWithinRetractWindow`.
- Added unit tests in `test/helpers/meetingMicHoldback.test.js`.

### Verification
- `node --test test/helpers/meetingMicHoldback.test.js` (passes, 25/25 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/meetingMicHoldback.js
+++ b/src/helpers/meetingMicHoldback.js
@@ -15,13 +15,13 @@ const partitionPendingMicFinals = ({ pending, now, force = false, isDuplicate })
   const duplicates = [];
   const releases = [];
 
-  for (const entry of pending) {
+  for (const entry of pending || []) {
     if (!force && entry.releaseAt > now) {
       deferred.push(entry);
       continue;
     }
 
-    if (isDuplicate(entry)) {
+    if (typeof isDuplicate === "function" && isDuplicate(entry)) {
       duplicates.push(entry);
       continue;
     }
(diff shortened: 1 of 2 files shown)


Chadpiha:
Thanks @hsusul — closing with #1793, same reasoning as the #1840 note.

`partitionPendingMicFinals`'s only caller passes the `meetingPendingMicFinals` closure array, which is initialized to `[]` and only ever reassigned to other arrays — and the `length === 0` guard immediately above the call would itself throw first if it were ever nullish. `isWithinRetractWindow`'s only caller filters `candidate.timestamp == null` and `systemTimestamp != null` before calling, and `committedAt` is always a `Date.now()` value (the legacy-null case is already pinned by an existing test). I did verify all 24 existing tests pass against your version, so no pinned behavior was disturbed — the change is just behavior-neutral for every input the pipeline can produce.

Appreciate the care with this module — it's one of the trickier ones! 🙏

PR #1796, opened 2026-08-23
fix(catalog): handle non-array summaries and nullish fields in normalizeBedrockCatalog

### LINKED ISSUE #1795: normalizeBedrockCatalog throws TypeError on non-array summaries or non-string model fields
### Describe the bug

In `src/helpers/bedrockCatalog.js`:

1. `profileIdByModelId(inferenceProfileSummaries)` and `normalizeBedrockCatalog(modelSummaries, inferenceProfileSummaries)` iterate directly over `inferenceProfileSummaries || []` and `modelSummaries || []`. When passed non-array objects or primitives, it throws `TypeError: (intermediate value) is not iterable`.
2. When sorting models via `models.sort((a, b) => a.vendor.localeCompare(b.vendor) || a.label.localeCompare(b.label))`, if a model summary contains non-string `providerName`, `modelName`, or `modelId`, calling `.localeCompare()` throws `TypeError: a.vendor.localeCompare is not a function`.

### To Reproduce

1. Call `normalizeBedrockCatalog({}, {})` -> throws `TypeError`.
2. Call `normalizeBedrockCatalog(null, undefined)` with non-string fields -> throws `TypeError`.

### Expected Behavior

- `normalizeBedrockCatalog` and `profileIdByModelId` should safely accept non-array inputs, defaulting to empty arrays.
- `vendor` and `label` should safely be coerced to strings to ensure deterministic sorting without throwing.

hsusul:
Fixes #1795

### Problem
In `src/helpers/bedrockCatalog.js`:
1. `profileIdByModelId` and `normalizeBedrockCatalog` iterated over input summaries assuming they were always arrays, throwing a `TypeError` if non-array objects or primitives were passed.
2. In `models.sort((a, b) => a.vendor.localeCompare(b.vendor) || a.label.localeCompare(b.label))`, missing or non-string `providerName` or `modelName` could lead to `TypeError: a.vendor.localeCompare is not a function`.

### Solution
- Added `Array.isArray()` guards in `profileIdByModelId` and `normalizeBedrockCatalog`.
- Guaranteed string values for `vendor` and `label` before sorting.
- Added test coverage in `test/helpers/bedrockCatalog.test.js`.

### Verification
- `node --test test/helpers/bedrockCatalog.test.js` (passes, 5/5 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/bedrockCatalog.js
+++ b/src/helpers/bedrockCatalog.js
@@ -4,10 +4,13 @@
 
 function profileIdByModelId(inferenceProfileSummaries) {
   const map = new Map();
-  for (const profile of inferenceProfileSummaries || []) {
+  const profiles = Array.isArray(inferenceProfileSummaries) ? inferenceProfileSummaries : [];
+  for (const profile of profiles) {
     if (!profile?.inferenceProfileId) continue;
-    for (const model of profile.models || []) {
-      const modelId = model?.modelArn?.split("/").pop();
+    const models = Array.isArray(profile.models) ? profile.models : [];
+    for (const model of models) {
+      const modelId =
+        typeof model?.modelArn === "string" ? model.modelArn.split("/").pop() : undefined;
       if (modelId && !map.has(modelId)) map.set(modelId, profile.inferenceProfileId);
     }
   }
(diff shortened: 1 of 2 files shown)


Chadpiha:
Thanks @hsusul — of this batch, this is the one closest to a real boundary, so I want to explain the close (with #1795) properly.

The data path is: `EnterpriseProviderConfig` → `bedrock-list-models` IPC handler → official AWS SDK. The SDK's deserializer either yields a real array (or `undefined`, which the existing `|| []` covers) or throws — on non-2xx, malformed JSON, network, and auth failures alike — and every throw lands in the handler's try/catch, which maps it through `mapEnterpriseError` to a visible "Could not load models" state in the UI. The `profileSummaries` argument is a locally built literal array in the handler itself. The genuinely-optional `modelName`/`providerName` fields are already handled by the existing `||` fallbacks.

The key trade-off: the one input this PR changes (AWS returning a 200 that violates its own schema) currently produces that *surfaced* error — after the PR it would produce an empty model picker reported as success, which hides the failure from the user. If we ever harden this boundary further, we'd do it in the IPC handler with an explicit error rather than coercion in the pure helper.

Thanks for thinking about API resilience — that instinct is exactly right, it just belongs one layer up here. 🙏

PR #1800, opened 2026-08-23
fix(uploads): normalize provider names and clamp verbose segment timestamps

### LINKED ISSUE #1799: timestampRequestFields misses mixed-case providers and mapVerboseSegments allows negative/inverted timestamps
### Describe the bug

In `src/helpers/uploadTimestamps.js`:

1. `timestampRequestFields(provider, model)` compares `provider` and `model` using strict case-sensitive comparisons against `"openai"`, `"groq"`, `"mistral"`. When passed capitalized provider names (e.g. `"OpenAI"`, `"Groq"`, `"Mistral"`) or untrimmed strings, it returns `null` instead of the appropriate verbose format fields.
2. In `mapVerboseSegments(responseData)`, negative start timestamps (`start < 0`) are not clamped to `0`, and when an upstream provider returns an inverted segment where `end < start`, the invalid end time is preserved instead of being clamped to at least `start`.

### To Reproduce

1. Call `timestampRequestFields("OpenAI", "whisper-1")` -> returns `null` (expected `{ response_format: "verbose_json" }`).
2. Call `mapVerboseSegments({ segments: [{ start: -2, end: 3, text: "test" }] })` -> returns `[{ start: -2, end: 3, text: "test" }]` (negative start timestamp).
3. Call `mapVerboseSegments({ segments: [{ start: 5, end: 3, text: "test" }] })` -> returns `[{ start: 5, end: 3, text: "test" }]` (inverted timestamp).

### Expected Behavior

- `timestampRequestFields` should normalize provider and model strings (case-insensitively and trimmed).
- `mapVerboseSegments` should clamp `start` to `Math.max(0, seg.start)` and ensure `end >= start`.

hsusul:
Fixes #1799

### Problem
In `src/helpers/uploadTimestamps.js`:
1. `timestampRequestFields` used case-sensitive comparisons, failing when provider names had leading/trailing whitespace or mixed casing (e.g. `"OpenAI"`, `"Groq"`, `"Mistral"`).
2. `mapVerboseSegments` did not bound negative start timestamps (`start < 0`) and allowed inverted segment ranges where `end < start`.

### Solution
- Normalized `provider` and `model` with `trim().toLowerCase()` in `timestampRequestFields`.
- Clamped `start` to `Math.max(0, seg.start)` and guaranteed `end = Math.max(start, rawEnd)` in `mapVerboseSegments`.
- Added unit tests in `test/helpers/uploadTimestamps.test.js`.

### Verification
- `node --test test/helpers/uploadTimestamps.test.js` (passes, 6/6 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/uploadTimestamps.js
+++ b/src/helpers/uploadTimestamps.js
@@ -5,16 +5,13 @@
 // 400 on unknown fields) must keep the plain-text request untouched — a
 // missing timestamp is a degraded export, never a failed upload.
 function timestampRequestFields(provider, model) {
-  const normProvider = typeof provider === "string" ? provider.trim().toLowerCase() : "";
-  const normModel = typeof model === "string" ? model.trim().toLowerCase() : "";
-
-  if (normProvider === "openai" && normModel === "whisper-1") {
+  if (provider === "openai" && model === "whisper-1") {
     return { response_format: "verbose_json" };
   }
-  if (normProvider === "groq") {
+  if (provider === "groq") {
     return { response_format: "verbose_json" };
   }
-  if (normProvider === "mistral") {
+  if (provider === "mistral") {
     return { timestamp_granularities: "segment" };
   }
   return null;
--- a/src/helpers/uploadTimestamps.js
+++ b/src/helpers/uploadTimestamps.js
@@ -5,13 +5,16 @@
 // 400 on unknown fields) must keep the plain-text request untouched — a
 // missing timestamp is a degraded export, never a failed upload.
 function timestampRequestFields(provider, model) {
-  if (provider === "openai" && model === "whisper-1") {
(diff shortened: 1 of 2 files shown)


Chadpiha (approved):
Thanks @hsusul, the timestamp clamps are a good catch !

`mapVerboseSegments` parses verbose_json straight off the provider's HTTP response, and that's exactly the kind of boundary we want hardened: negative or inverted segment timestamps now clamp instead of flowing into note anchors, and your test locks the semantics in.

I pushed a commit on top that scopes the PR to those clamps. The provider/model normalization came out because `resolveTranscriptionRoute` already constructs `route.provider` as a fixed lowercase literal and canonicalizes the model through `resolveByokModel`, so the values reaching `timestampRequestFields` are guaranteed — and the surrounding handler compares them strictly, which we want to keep consistent. Thanks again for the contribution!

PR #1802, opened 2026-08-23
fix(reasoning): make stripThinkingTags case-insensitive and clean stray closing tags

### LINKED ISSUE #1801: stripThinkingTags fails on mixed-case thinking tags and leaves stray closing tags
### Describe the bug

In `src/helpers/stripThinking.js`:

1. `stripThinkingTags(text)` uses lowercase-only regex patterns (e.g. `/<think>(?:(?!<think>)[\s\S])*?<\/think>/` and `/<think>[\s\S]*$/`). When reasoning models or API proxies emit titlecased/uppercase tags (such as `<Think>...</Think>` or `<THINK>...</THINK>`), the thinking block is not stripped.
2. If an orphan or stray closing tag `</think>` exists (e.g. when streaming starts mid-response or after the opening tag was sliced), `stripThinkingTags` leaves the orphan closing tag in the output text.

### To Reproduce

1. Call `stripThinkingTags("<Think>reasoning</Think>Meeting Notes")` -> returns `"<Think>reasoning</Think>Meeting Notes"` (expected `"Meeting Notes"`).
2. Call `stripThinkingTags("<THINK>reasoning</THINK>Meeting Notes")` -> returns `"<THINK>reasoning</THINK>Meeting Notes"` (expected `"Meeting Notes"`).
3. Call `stripThinkingTags("</think>Meeting Notes")` -> returns `"</think>Meeting Notes"` (expected `"Meeting Notes"`).

### Expected Behavior

- `stripThinkingTags` should use case-insensitive regex matching for thinking tags.
- Any stray closing `</think>` tags should be cleaned up from the output.

hsusul:
Fixes #1801

### Problem
In `src/helpers/stripThinking.js`:
1. `stripThinkingTags` only matched lowercase `<think>` and `</think>` tags, failing to strip thinking blocks when models or API proxies output mixed-case or uppercase tags (e.g. `<Think>...</Think>`, `<THINK>...</THINK>`).
2. Orphan closing tags (`</think>`) emitted when streaming started after the opening tag were preserved in the output.

### Solution
- Added case-insensitive matching (`i` flag) to `<think>...</think>` and trailing unterminated `<think>` removal regexes.
- Added a final pass to remove any leftover stray `</think>` tags (`replace(/<\/think>/gi, "")`).
- Added unit tests in `test/helpers/stripThinking.test.js`.

### Verification
- `node --test test/helpers/stripThinking.test.js` (passes, 10/10 tests)
- `git diff --check` (clean)


gabrielste1n (approved):
Traced both call sites end to end — `ReasoningService.ts:414` (raw chat-completions JSON from Corti/Groq/user-configured LAN endpoints) and `localReasoningBridge.js:52` (local llama-server output) — and both receive uncontrolled model-emitted text, so this is exactly the boundary where we validate. Nice catch: beyond leaking tags, the old case-sensitive trailing rule could treat `<think>a</THINK>b` as unterminated and strip the entire answer; the case-insensitive pairing fixes that worse failure too. Lowercase behavior is unchanged and all tests pass. Filed #1861 to mirror this in the streaming filter. Thank you @hsusul!

PR #1804, opened 2026-08-24
fix(gpu): refresh banner after settings close

### LINKED ISSUE #1778: [BUG] GPU acceleration pop‑up fails to update when already enabled in settings
**Describe the bug**
When I installed OpenWhispr, one of the first things I did was enable GPU acceleration in the settings menu. However, after closing the settings menu, a banner at the top of the screen still said that GPU acceleration should be enabled. When I clicked the button, it took me to the AI model section of the settings because GPU acceleration had already been enabled. 

**To Reproduce**

1. Install OpenWhispr as a fresh install for all users (Admin) on Windows 10
2. Configure it following the default flow
3. Go to 'Settings' and then 'Speech to text'
4. Turn on GPU acceleration
5. Close the 'Settings' menu
6. See that the enable GPU acceleration banner is there at the top of the screen in the app
7. Click the enable button in the banner
8. In the opened 'Settings' menu, see that GPU acceleration is already enabled under 'Speech to text'
9. Close the 'Settings menu again
10. See that the banner is still there.

**Expected behavior**
I expected that OpenWhispr would not show me the Enable GPU Acceleration banner because I had already enabled it in the settings, or at the very least, clicking the banner button to enable GPU acceleration should have made the app check and realize that acceleration had already been enabled, and then automatically dismiss the banner. 

**Screenshots**

Image

Image

- OS: Windows 10 22H2
- Browser: N/A
- Version: OpenWhispr 1.8.3

zhongwater123:
## Summary

- refresh the home-screen GPU offer after Settings closes
- keep the existing CUDA, Vulkan, and intelligence detection behavior in a focused hook
- add renderer lifecycle coverage for install success followed by closing Settings

Fixes #1778

## Root cause

ControlPanel cached the GPU offer, but its detection effect did not observe the Settings modal state. A successful GPU pack installation only updated local state inside TranscriptionModelPicker, and the main process did not publish a completion event. Closing Settings therefore left the home banner on its previous result until another eligible setting changed or the app restarted.

The fix pauses the home probe while Settings is open and re-runs the existing probe when Settings closes. It does not change GPU installation, persistence, provider selection, or translations.

## Validation

- Node 24.19.0: focused GPU lifecycle and policy tests (11/11 passed)
- Regression mutation check: removing the Settings-close dependency makes the new test fail with one probe instead of two
- TypeScript typecheck
- Full repository ESLint
- Prettier check for changed files
- Renderer production build on Windows

## Full-suite note

The full Windows test run reaches an existing failure in test/helpers/assemblyAiStreaming.test.js because its Electron mock leaves electron.app undefined for debugLogger. The same failure reproduces on an unchanged main worktree at fa11ddc, so this PR does not modify that unrelated area. The real GPU-pack download flow was not repeated; the regression test exercises the renderer lifecycle with mocked Electron status responses.

### DIFF
--- a/src/components/ControlPanel.tsx
+++ b/src/components/ControlPanel.tsx
@@ -59,7 +59,7 @@ import WindowControls from "./WindowControls";
 
 import { getCachedPlatform } from "../utils/platform";
 import { isAccessibilitySkipped } from "../utils/permissions";
-import { eligibleGpuOffers, type GpuOffers } from "../utils/gpuBannerPolicy";
+import { useGpuBannerAvailability } from "../hooks/useGpuBannerAvailability";
 import {
   setActiveNoteId,
   setActiveFolderId,
@@ -156,10 +156,6 @@ export default function ControlPanel({ initialSettingsSection }: ControlPanelPro
     folderId: number;
     event: any;
   } | null>(null);
-  const [gpuAccelAvailable, setGpuAccelAvailable] = useState<GpuOffers>({
-    transcription: false,
-    intelligence: null,
-  });
   const [gpuBannerDismissed, setGpuBannerDismissed] = useState(
     () => localStorage.getItem("gpuBannerDismissedUnified") === "true"
   );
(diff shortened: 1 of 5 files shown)


gabrielste1n (approved):
Great find and a clean fix — thanks for the thorough root-cause analysis on the issue and the mutation-checked regression test. I pushed one small follow-up commit hoisting the duplicated installHookDom test helper into the shared rendererTestHarness. Merging!

PR #1806, opened 2026-08-24
fix(audio): validate settings and finite duration in shouldSaveDiscardedRecording

### LINKED ISSUE #1805: shouldSaveDiscardedRecording permits Infinity duration and non-object settings
### Describe the bug

In `src/helpers/discardedRecording.js`:

1. `shouldSaveDiscardedRecording(settings, durationSeconds, policyState = null)` evaluates `durationSeconds >= MIN_DISCARDED_DURATION_SECONDS`. When `durationSeconds` is `Infinity` or non-finite, it returns `true` instead of `false`.
2. When `settings` is a non-object primitive (e.g. `true` or `"invalid"`), `settings.dataRetentionEnabled` and `settings.saveDiscardedTranscriptions` evaluate improperly or can cause unexpected policy rule evaluations.

### To Reproduce

1. Call `shouldSaveDiscardedRecording(settings, Infinity)` -> returns `true` (expected `false`).
2. Call `shouldSaveDiscardedRecording("invalid", 3)` -> returns `false` but doesn't validate object type upfront.

### Expected Behavior

- `shouldSaveDiscardedRecording` should return `false` if `!settings || typeof settings !== "object"`.
- `shouldSaveDiscardedRecording` should return `false` if `!Number.isFinite(durationSeconds)` or `durationSeconds < MIN_DISCARDED_DURATION_SECONDS`.

hsusul:
Fixes #1805

### Problem
In `src/helpers/discardedRecording.js`:
1. `shouldSaveDiscardedRecording` checked `durationSeconds >= MIN_DISCARDED_DURATION_SECONDS`, which allowed `Infinity` to improperly evaluate to `true`.
2. When `settings` was passed as a non-object primitive, it wasn't validated upfront.

### Solution
- Added `if (!settings || typeof settings !== "object") return false;` upfront.
- Required `Number.isFinite(durationSeconds) && durationSeconds >= MIN_DISCARDED_DURATION_SECONDS`.
- Added unit tests in `test/helpers/discardedRecording.test.js`.

### Verification
- `node --test test/helpers/discardedRecording.test.js` (passes, 5/5 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/discardedRecording.js
+++ b/src/helpers/discardedRecording.js
@@ -8,7 +8,10 @@ import {
 export const MIN_DISCARDED_DURATION_SECONDS = 1;
 
 export function shouldSaveDiscardedRecording(settings, durationSeconds, policyState = null) {
-  if (!settings) return false;
+  if (!settings || typeof settings !== "object") return false;
+  if (!Number.isFinite(durationSeconds) || durationSeconds < MIN_DISCARDED_DURATION_SECONDS) {
+    return false;
+  }
   const dataRetentionEnabled = policyState
     ? effectiveLocalHistoryEnabled(policyState, settings.dataRetentionEnabled)
     : settings.dataRetentionEnabled;
@@ -18,6 +21,5 @@ export function shouldSaveDiscardedRecording(settings, durationSeconds, policySt
   if (!settings.saveDiscardedTranscriptions) return false;
   if (!dataRetentionEnabled) return false;
   if (!(audioRetentionDays > 0)) return false;
-  if (!(durationSeconds >= MIN_DISCARDED_DURATION_SECONDS)) return false;
   return true;
 }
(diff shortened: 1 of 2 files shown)


Chadpiha:
Thanks for this one, @hsusul — closing along with #1805 after tracing where both inputs come from.

`durationSeconds` is pure wall-clock arithmetic: `(Date.now() - recordingStartTime) / 1000`, or `null` when there's no start time (which already returns false today). It never comes from media metadata, so the well-known "WebM reports Infinity duration" situation can't feed this helper — the value is always null or finite. And `settings` is always a Zustand store snapshot, so it's always an object (property access on primitives doesn't throw in JS either, so the non-object path already returned false).

The direction of the guard matters here too: this feature exists to *preserve* discarded audio (#964/#907 — the min-duration check only filters accidental sub-second Escape taps). If an unknown duration ever did appear, silently declining to save would lose exactly the audio the feature protects, so failing open is the safer default for this one.

Really appreciate the attention to the audio pipeline — thank you!

PR #1808, opened 2026-08-24
fix(notifications): normalize source in getNotificationTimeoutMs and sanitize timer duration

### LINKED ISSUE #1807: getNotificationTimeoutMs misses mixed-case source and NotificationDismissTimer doesn't validate duration
### Describe the bug

In `src/helpers/notificationTimer.js`:

1. `getNotificationTimeoutMs(source)` compares `source === "calendar"` strictly without case or whitespace normalization. When `source` is passed as `"Calendar"`, `"CALENDAR"`, or has surrounding whitespace, it improperly falls back to `DETECTION_TIMEOUT_MS` (30s) instead of `CALENDAR_TIMEOUT_MS` (2 min).
2. In `NotificationDismissTimer.prototype.start(durationMs)`, non-finite or non-positive `durationMs` values (such as `null`, `undefined`, `NaN`, `-5`) are passed directly into `setTimeout`, resulting in immediate or unpredictable timeout firing.
3. In `NotificationDismissTimer.prototype._arm`, when the timeout expires, calling `this._onTimeout()` throws `TypeError: this._onTimeout is not a function` if `this._onTimeout` is nullish or not a function.

### To Reproduce

1. Call `getNotificationTimeoutMs("Calendar")` -> returns `30000` (expected `120000`).
2. Call `timer.start(NaN)` -> `_deadline` is `NaN`.
3. Call `timer.start()` with no callback -> throws `TypeError`.

### Expected Behavior

- `getNotificationTimeoutMs` should normalize `source` case-insensitively with trimming.
- `NotificationDismissTimer.prototype.start` should sanitize `durationMs` to a positive finite number (defaulting to `DETECTION_TIMEOUT_MS`).
- `_onTimeout` should only be invoked if `typeof this._onTimeout === "function"`.

hsusul:
Fixes #1807

### Problem
In `src/helpers/notificationTimer.js`:
1. `getNotificationTimeoutMs(source)` used strict lowercase comparison against `"calendar"`, falling back to 30s detection timeout when `source` was capitalized or padded with whitespace (e.g. `"Calendar"`).
2. `NotificationDismissTimer.prototype.start(durationMs)` did not sanitize `durationMs`, allowing non-finite or non-positive numbers to disrupt timer scheduling.
3. `NotificationDismissTimer` threw a `TypeError` when the timer expired if `this._onTimeout` was missing or not a function.

### Solution
- Normalized `source` with `trim().toLowerCase()`.
- Sanitized `durationMs` to ensure a positive finite number before passing to `_arm`.
- Guarded callback invocation with `typeof this._onTimeout === "function"`.
- Added unit tests in `test/helpers/notificationTimer.test.js`.

### Verification
- `node --test test/helpers/notificationTimer.test.js` (passes, 10/10 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/notificationTimer.js
+++ b/src/helpers/notificationTimer.js
@@ -9,7 +9,8 @@ const CALENDAR_TIMEOUT_MS = 2 * 60 * 1000;
 const MIN_RESUME_MS = 5 * 1000;
 
 function getNotificationTimeoutMs(source) {
-  return source === "calendar" ? CALENDAR_TIMEOUT_MS : DETECTION_TIMEOUT_MS;
+  const normSource = typeof source === "string" ? source.trim().toLowerCase() : "";
+  return normSource === "calendar" ? CALENDAR_TIMEOUT_MS : DETECTION_TIMEOUT_MS;
 }
 
 // Auto-dismiss countdown for the meeting notification overlay, pausable while
@@ -24,7 +25,9 @@ class NotificationDismissTimer {
 
   start(durationMs) {
     this.cancel();
-    this._arm(durationMs);
+    const validDuration =
+      Number.isFinite(durationMs) && durationMs > 0 ? durationMs : DETECTION_TIMEOUT_MS;
+    this._arm(validDuration);
   }
 
   pause() {
@@ -55,7 +58,9 @@ class NotificationDismissTimer {
     this._timer = setTimeout(() => {
       this._timer = null;
       this._deadline = null;
-      this._onTimeout();
+      if (typeof this._onTimeout === "function") {
+        this._onTimeout();
+      }
     }, durationMs);
   }
 }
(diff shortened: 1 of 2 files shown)


Chadpiha:
Thanks @hsusul — closing along with #1807 (see the #1840 note for the general convention).

`source` is an internal two-value enum: the only producers are literal `"audio"` and `"calendar"` strings inside `meetingDetectionEngine.js`, and it never crosses IPC or persistence on the way to this helper. Worth noting the engine itself strict-compares `source === "calendar"` in two other places (pref gating and the Join button) — so normalizing in just this one reader would imply a tolerance the rest of the pipeline doesn't have. Similarly, the timer's only production `start()` call receives one of two constants, and the constructor always gets an inline callback — the fallbacks would convert a loud future bug into a notification that silently never dismisses.

Thanks as always for the tidy tests! 🙏

PR #1816, opened 2026-08-25
fix(notes): handle nullish and non-iterable placeholders in isRegenerableNoteTitle

### LINKED ISSUE #1815: isRegenerableNoteTitle throws TypeError when placeholders is null or non-iterable
### Describe the bug

In `src/helpers/regenerableNoteTitle.js`:

`isRegenerableNoteTitle(title, placeholders = [], calendarEventName = null)` has a default parameter `placeholders = []`, which only triggers when `placeholders` is `undefined`. When callers pass `null` or a non-iterable object/primitive for `placeholders`, iterating with `for (const p of placeholders)` throws `TypeError: placeholders is not iterable`.

### To Reproduce

1. Call `isRegenerableNoteTitle("My Note", null)` -> throws `TypeError: placeholders is not iterable`.
2. Call `isRegenerableNoteTitle("My Note", {})` -> throws `TypeError: placeholders is not iterable`.

### Expected Behavior

`isRegenerableNoteTitle` should safely accept `null`, `undefined`, or non-iterable values for `placeholders`, defaulting to an empty collection and returning `false` for non-matching titles without throwing.

hsusul:
Fixes #1815

### Problem
In `src/helpers/regenerableNoteTitle.js`:
`isRegenerableNoteTitle` defaulted `placeholders = []`, which only triggers when `placeholders === undefined`. If passed `null` or a non-iterable value, iterating with `for (const p of placeholders)` threw `TypeError: placeholders is not iterable`.

### Solution
- Fall back to `[]` if `placeholders` is `null` or not iterable before iterating.
- Added unit tests in `test/helpers/regenerableNoteTitle.test.js`.

### Verification
- `node --test test/helpers/regenerableNoteTitle.test.js` (passes, 6/6 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/regenerableNoteTitle.js
+++ b/src/helpers/regenerableNoteTitle.js
@@ -9,7 +9,12 @@ export function isRegenerableNoteTitle(title, placeholders = [], calendarEventNa
   if (trimmed === "") return true;
 
   const set = new Set(BUILTIN_PLACEHOLDERS);
-  for (const p of placeholders) {
+  const placeholderList =
+    Array.isArray(placeholders) ||
+    (placeholders && typeof placeholders[Symbol.iterator] === "function")
+      ? placeholders
+      : [];
+  for (const p of placeholderList) {
     if (typeof p === "string" && p.trim()) set.add(p.trim().toLowerCase());
   }
   if (set.has(trimmed.toLowerCase())) return true;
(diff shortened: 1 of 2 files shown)


Chadpiha:
Thanks @hsusul! Closing together with #1815.

This helper has a single caller (`PersonalNotesView.tsx`), which always passes a literal inline array of three `t(...)` results — and react-i18next's `t()` always returns a string — so `placeholders` is a string array by construction on every call. The one parameter that genuinely crosses a trust boundary here is `title` (it comes from the DB and can be null), and the helper already guards that with `typeof title === "string"`. Our convention is to draw the line exactly there: validate what crosses a boundary, trust what internal code guarantees.

The tests themselves were well done and I ran them green — this is purely a scope-of-validation call, not a quality one. Thanks for the contribution, and keep 'em coming!

PR #1818, opened 2026-08-25
fix(dock): handle nullish and missing parameters in resolveDockVisibility

### LINKED ISSUE #1817: resolveDockVisibility throws TypeError when called without arguments or with nullish parameter
### Describe the bug

In `src/helpers/dockPolicy.js`:

`resolveDockVisibility({ platform, controlPanelVisible })` directly destructures its first parameter without defaulting to an empty object. When called with `null`, `undefined`, or with no arguments, it throws:

`TypeError: Cannot destructure property 'platform' of 'undefined' / 'null'`.

### To Reproduce

1. Call `resolveDockVisibility()` -> throws `TypeError`.
2. Call `resolveDockVisibility(null)` -> throws `TypeError`.
3. Call `resolveDockVisibility(undefined)` -> throws `TypeError`.

### Expected Behavior

`resolveDockVisibility` should safely handle `null`, `undefined`, or missing arguments, returning `null` off-darwin or when `platform` is not darwin, without throwing an error.

hsusul:
Fixes #1817

### Problem
In `src/helpers/dockPolicy.js`:
`resolveDockVisibility({ platform, controlPanelVisible })` threw `TypeError: Cannot destructure property 'platform' of 'undefined' / 'null'` when called without arguments or when passed `null`/`undefined`.

### Solution
- Defaulted parameters with `params || {}` before destructuring.
- Added unit tests in `test/helpers/dockPolicy.test.js`.

### Verification
- `node --test test/helpers/dockPolicy.test.js` (passes, 4/4 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/dockPolicy.js
+++ b/src/helpers/dockPolicy.js
@@ -10,7 +10,8 @@
 
 // Whether the Dock icon should be visible right now.
 // Returns null off macOS, where there is no Dock to act on.
-export function resolveDockVisibility({ platform, controlPanelVisible }) {
+export function resolveDockVisibility(params = {}) {
+  const { platform, controlPanelVisible } = params || {};
   if (platform !== "darwin") return null;
   return !!controlPanelVisible;
 }
--- a/test/helpers/dockPolicy.test.js
+++ b/test/helpers/dockPolicy.test.js
@@ -27,3 +27,12 @@ test("there is no Dock to act on outside macOS", async () => {
     assert.equal(resolveDockVisibility({ platform, controlPanelVisible: true }), null);
   }
 });
+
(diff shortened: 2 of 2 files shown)


Chadpiha:
Thanks @hsusul! Closing with #1817 — same reasoning as the note on #1840.

`dockPolicy.js` is a deliberately pure policy module with exactly one caller (`dockManager.js`), which always passes `{ platform: process.platform, controlPanelVisible: this._controlPanelVisible }` — a guaranteed string and a `!!`-coerced boolean. Nullish/missing params can't occur, and if a future refactor ever did pass them, the TypeError at the call site would be the more useful signal than a silent `null`.

Appreciate you reading through the dock policy code — thanks!

PR #1821, opened 2026-08-25
feat(calendar): add provider-neutral availability tool

boseq:
## Summary

- add a read-only get_calendar_availability tool over the shared local calendar cache
- combine selected Google, Microsoft, and Apple calendars into privacy-safe busy intervals and available slots
- normalize provider free/busy and self-declined semantics, including all-day events, overlap boundaries, buffers, and minimum slot duration
- add IPC/preload wiring, prompt guidance, UI status/icon support, types, translations, and focused coverage

## Scope

- no new OAuth scopes, token flows, or provider connections
- no raw event titles, attendees, meeting links, account identifiers, or provider names are returned to the model
- results are explicitly labeled as local-calendar-cache coverage for the next seven local calendar days

## Validation

- 45 focused calendar availability/provider/database/tool tests
- TypeScript typecheck
- full ESLint
- i18n consistency check
- Swift typecheck
- renderer production build
- Prettier and git diff checks

## Manual testing

Validated locally with Google, Microsoft, and Apple events in a 9 AM–6 PM range, including minimum-duration filtering and cross-provider busy interval union.


Chadpiha (approved):
Approving — solid work. I did a full pass on this: read every changed file, traced the flow end-to-end (prompt context → tool call → renderer validation → IPC → service → cache query → interval math → localized facts back to the model), and ran the suite, typecheck, i18n check, and a renderer production build locally.

Things I specifically verified:

- **Privacy boundary holds.** Only times, durations, and flags ever reach the model; the projection fails closed on malformed payloads, and unknown IPC errors collapse to a generic message with an allowlist for the correctable ones.
- **Cache coverage vs. the 7-day horizon** checks out for all three providers (Google 9d lookahead + 1d token TTL, Microsoft 14d/7d TTL + lookback, Apple 8d snapshot + 2h lookback), including ongoing multi-day events since all three provider queries are overlap/end-time based.
- **Microsoft all-day date-only storage** — I confirmed against Graph's documented behavior that all-day events stay date-pure at midnight in the preferred zone, and the format change can't regress reminders or the upcoming UI because those queries exclude `is_all_day` rows.
- **Migration safety** — table ordering is right, fresh DBs no-op, and the forced REST resync populates the new columns; until then `unknown` conservatively blocks, which is the right failure direction.

I pushed two small commits onto the branch:

1. `fix(calendar): load shared availability helper as ESM` — the constants-sharing commit imported a CJS (`module.exports`) helper into the renderer graph, which hangs the Vite-SSR chat-streaming tests (that's why the CI `tests` jobs were stuck). Converted the helper to ESM per the `meetingJoinUrl.js` convention, and split `computeCalendarAvailability` out so the service validates each request once. 
2. `fix(calendar): accept null optional args in availability tool` — models often send explicit `null` for optional params; the renderer now treats null as absent like the service's `??` defaults already did (with a regression test), and the seven-day wording in the schema is now derived from `MAX_AVAILABILITY_HORIZON_DAYS`.

Two behavioral notes for the record, both fine by me: unanswered invitations block availability on all providers (only an explicit decline or free/transparent unblocks — matches provider free/busy semantics), and broad multi-day questions intentionally get a clarifying question about daily hours before per-day calls.

PR #1612, opened 2026-08-13
fix(macos): make auto-paste layout-aware for non-QWERTY keyboards

IdrisGit:
## Summary

Fixes macOS auto-paste and synthetic selection copy on non-QWERTY keyboard layouts.

The native helper previously posted US-QWERTY physical key codes for Cmd+C and Cmd+V. On layouts such as standard Dvorak, the same physical positions produce different characters, so paste could trigger Cmd+. instead.

The helper now resolves the key code for the logical `c` or `v` shortcut using the active ASCII-capable input source and `UCKeyTranslate`, with the Command modifier state applied during resolution.

## Behavior

- Supports standard Dvorak, Dvorak - QWERTY Cmd, Colemak, AZERTY, QWERTZ, and other installed layouts.
- Does not switch or otherwise mutate the user’s input source.
- Retains the existing US-QWERTY key codes as a fallback if macOS cannot resolve the current layout.
- Keeps the existing clipboard preservation, focus recovery, and accessibility-permission behavior unchanged.

> [!IMPORTANT]
> I need help with testing this fix, as I use Linux and currently don't have a macOS system lying around for testing
> @vsaw if you can help with testing that would be nice, I can connect with you if you want to understand how to test this changes locally
> if good this PR will fixes #1478 and fixes  #643

PR #1822, opened 2026-08-25
fix(macos): resolve paste/copy key codes for non-QWERTY layouts

io41:
## Problem

`resources/macos-fast-paste.swift` hardcodes the virtual key codes it posts: `0x09` for `v` and `0x08` for `c`. Those values are positions on a QWERTY layout, not characters. On Dvorak, Colemak, and similar layouts they land on different keys, so the synthesized `⌘V` and `⌘C` are not paste and copy.

Both paths are affected:

- Auto-paste after dictation silently does nothing (or triggers whatever shortcut happens to live on that key).
- `--copy` selection capture returns `COPY_OK` because the frontmost app resolved fine, but the clipboard was never populated — so the caller believes it captured a selection that it did not.

On a Dvorak layout, `v` is at key code 47 and `c` is at key code 34.

## Fix

Look the key code up in the active keyboard layout with `UCKeyTranslate` instead of hardcoding it. `keyCodeForCharacter(_:)` walks key codes 0–127 through the current `TISCopyCurrentKeyboardLayoutInputSource()` layout and returns the one that produces the requested character.

If the lookup fails, it falls back to the QWERTY key code **for the same mode** — `0x08` in copy mode, `0x09` in paste mode. Keeping the fallback mode-aligned matters: falling back to a single constant would let a failed lookup in copy mode post `⌘V`, overwriting the user's selection instead of copying it.

No behaviour change on QWERTY: the lookup returns the same values that were previously hardcoded.

## Notes

- Adds `import Carbon` for the Text Input Services and `UCKeyTranslate` APIs.
- The `--copy` stdout contract (`COPY_OK <pid> <name>`), exit codes, and the ordering of the `AXIsProcessTrusted()` check and frontmost-app resolution are all unchanged.
- Self-contained in the one file — no build-script or caller changes.

## Testing

- `swiftc` compiles clean.
- On a non-QWERTY (Dvorak) layout, the lookup resolves `v` to key code 47 and `c` to key code 34, confirming the resolution path runs rather than silently falling back.
- Verified end to end in a local packaged build on macOS 15 (arm64): dictation auto-paste inserts text, and selection capture populates the clipboard so in-place selection edits work.

### DIFF
--- a/resources/macos-fast-paste.swift
+++ b/resources/macos-fast-paste.swift
@@ -1,14 +1,54 @@
 import Cocoa
+import Carbon
 
 if !AXIsProcessTrusted() {
     exit(2)
+}
+
+// Look up which key code produces a given character in the current keyboard
+// layout. Hardcoded key codes (0x09 for 'v', 0x08 for 'c') are only correct on
+// QWERTY, so Cmd+V / Cmd+C land on the wrong key on Dvorak, Colemak, etc.
+func keyCodeForCharacter(_ target: UniChar) -> CGKeyCode? {
+    guard let sourceRef = TISCopyCurrentKeyboardLayoutInputSource()?.takeRetainedValue(),
+          let layoutDataRef = TISGetInputSourceProperty(sourceRef, kTISPropertyUnicodeKeyLayoutData) else {
+        return nil
+    }
+    let layoutData = unsafeBitCast(layoutDataRef, to: CFData.self)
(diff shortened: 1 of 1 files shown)


Chadpiha:
Thanks for the fix and the end-to-end Dvorak testing! We’re consolidating this work in #1612, which covers both auto-paste and selection copy, accounts for Command-specific layouts, and adds regression coverage.
Closing this as superseded by #1612. Really appreciate your contribution and testing @io41!

PR #1825, opened 2026-08-25
Calendar availability tool: let the assistant answer "when am I free?" across all connected calendars

gabrielste1n:
## What

Adds `get_calendar_availability`, a new agent tool available in every chat surface (voice assistant panel, main chat, note chat, folder chat), that computes **free time slots and busy blocks across all connected calendars** — Google, Microsoft, and Apple — for an arbitrary date range.

This unlocks:
- *"What times am I free on Thursday?"*
- *"Write me an email and suggest some times that work for me Wed–Thursday next week."*

## How

**The model now has a clock.** The agent system prompt carries the current date, time, weekday, and IANA timezone — previously nothing injected any of these, so "next Wednesday" was unresolvable by any tool.

**Local-first availability.** Free/busy is computed from the shared `calendar_events` SQLite table (already deduped across providers), not live provider APIs — no new OAuth scopes, works offline, uniform across all three providers:
- New `DatabaseManager.getEventsInRange(fromIso, toIso)` with overlap semantics, built on `dedupedEventsQuery` so Apple mirrors of Google/Microsoft events never double-book. Unlike `getUpcomingEvents`, all-day rows are included (an all-day OOO matters for availability).
- New cross-provider IPC channel `calendar-get-events-in-range` (deliberately not the historical `gcal-` prefix).

**Availability semantics** (`src/utils/availability.ts`, pure and unit-tested):
- Free slots within a per-day window (default 09:00–18:00 local, model-overridable via `day_start`/`day_end`), minimum 30 minutes.
- Confirmed and tentative events hold time; each busy block carries its status so the model can caveat holds.
- Meetings the user **declined** count as free (from stored attendee RSVP data).
- All-day events are reported per day but never subtracted — without the providers' free/busy field there's no way to tell a blocking OOO from a transparent birthday, so the model surfaces them as caveats.

**31-day horizon.** All three providers now sync 31 days ahead (was 7 for Google/Apple, 14 for Microsoft) so any "next week" question is answerable. Google's `syncToken` and Graph's `deltaLink` permanently encode the window they were created with, so a one-time `user_version 2` migration clears stored tokens to force a full re-sync over the new range. Days outside the synced horizon (including past days, which are never synced) are reported as `unknown` so the model says so instead of guessing.

No API-side changes: tools are client-declared and client-executed, and the streaming endpoint forwards them as-is.

## Tests

- `test/utils/availability.test.js` — interval math: overlap merging, min-slot filtering, declined-event exclusion, all-day separation, midnight-spanning clipping; inputs built from local components so results are timezone-independent.
- `test/services/availabilityTool.test.js` — defaults, range swap/cap, ±1-day fetch pad, unknown-day marking, IPC failure.
- `test/services/agentSystemPrompt.test.js` — date/timezone line and tool instruction wiring.
- `test/helpers/calendarDatabase.test.js` — `getEventsInRange` boundaries, all-day inclusion, cancelled exclusion, Apple-mirror dedupe, and one-time-only token reset.
- Full suite green (2782 pass), `typecheck`, `lint`, `i18n:check` clean (tool name/status keys in all 10 locales), and the Swift calendar listener compiles with the widened lookahead.


gabrielste1n:
Superseded by #1821, which merged an independent implementation of the same tool with provider-normalized free/busy semantics (availability_status / self_response_status), buffer and min-slot controls, and a privacy-safe output shape. Closing this rather than rebasing a duplicate. The one piece worth porting — extending the 7-day availability horizon to ~31 days so requests like "Wed–Thu next week" are answerable — will come as a small follow-up on top of main.

PR #1826, opened 2026-08-25
feat(dictation): replace processing rainbow beam with brand Signal glow

gabrielste1n:
## What

The dictation pill's processing/thinking state (the glowing border while transcription + cleanup run) previously used the BorderBeam `colorful` variant — nine hues orbiting a 40px circle at low opacity. At that size it read as faint chromatic noise, especially in light theme, and the rainbow palette didn't match the brand.

This replaces it with a **Signal glow** in brand blues:

- **Gradient ring** — cyan → blue → indigo band with a white-hot comet head, orbiting at the same 1.6s cadence as the Beam
- **Comet bloom** — soft blur trailing the head
- **Breathing ambient halo** — a wide blue radial bloom that pulses, legible from across the screen

## How

- The glow layers anchor on a new `voice-pill-beam-anchor` wrapper **outside** the BorderBeam, whose compact (`sm`) preset sets `overflow: hidden` on its wrapper and would crop the halo. The processing footprint is a fixed 40×40 circle, so the layers rotate freely.
- Animations are gated on `data-active` so the always-visible idle pill costs nothing; a 220ms opacity fade covers enter/exit. `prefers-reduced-motion` disables the motion entirely.
- Agent Mode is unchanged: it keeps its ocean Beam for listening/thinking. The Beam props are now fixed to the agent values since plain dictation no longer activates it, and the obsolete `plain-dictation-processing-glow` light-theme tuning is removed.

## Testing

- `test/helpers/voicePillStructure.test.js` updated to pin the new structure: anchor root, Signal glow active on plain thinking, Beam inactive for plain dictation, agent beam behavior unchanged. All 50 pill tests pass.
- Full suite passes (`npm test`), `npm run typecheck` and eslint clean.
- Verified visually in light + dark via an offscreen render of the shipped CSS.


gabrielste1n:
Two follow-up commits on this branch:

- **130c0246** — the study page's wide halo can't ship as-is: the pill docks 4px from the native window's right/bottom edges, and the transparent always-on-top window must stay minimal because it captures clicks across its whole rect. The halo + comet bloom are now a tighter, brighter corona that fits that budget (verified against simulated window bounds in both themes — no clip line).
- **6b93c8c4** — fixes the long-standing stutter at the start of the recording expansion: the native window grow used to fire in the same commit that starts the pill's 300ms width transition, so the anchored setBounds landed mid-animation. The window now snaps during the entrance's 420ms static hold instead, while nothing animates and the resize mask fully covers it.

PR #1828, opened 2026-08-25
fix(updater): handle null and non-object prefs in appUpdatesEnabled

### LINKED ISSUE #1827: appUpdatesEnabled throws TypeError when called with null
### Describe the bug

In `src/helpers/updateCheckPolicy.js`:

`appUpdatesEnabled({ notificationsEnabled, notifyUpdates } = {})` has default parameter `= {}`, which only triggers when the argument is `undefined`. When called with `null` or a non-object value, it throws:

`TypeError: Cannot destructure property 'notificationsEnabled' of 'null' as it is null.`

### To Reproduce

1. Call `appUpdatesEnabled(null)` -> throws `TypeError`.

### Expected Behavior

`appUpdatesEnabled` should safely handle `null` and non-object values, defaulting to `true` (check-by-default behavior) without throwing an error.

hsusul:
Fixes #1827

### Problem
In `src/helpers/updateCheckPolicy.js`:
`appUpdatesEnabled({ notificationsEnabled, notifyUpdates } = {})` threw `TypeError: Cannot destructure property 'notificationsEnabled' of 'null' as it is null` when passed `null`.

### Solution
- Defaulted parameters safely to handle `null` and non-object values before destructuring.
- Added unit tests in `test/helpers/updateCheckPolicy.test.js`.

### Verification
- `node --test test/helpers/updateCheckPolicy.test.js` (passes, 3/3 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/updateCheckPolicy.js
+++ b/src/helpers/updateCheckPolicy.js
@@ -3,7 +3,9 @@
 // firewalled machines never surface a connection error dialog (#1605).
 // Only an explicit false disables — missing prefs (renderer sync not arrived
 // yet) keep check-by-default behavior.
-function appUpdatesEnabled({ notificationsEnabled, notifyUpdates } = {}) {
+function appUpdatesEnabled(prefs = {}) {
+  const { notificationsEnabled, notifyUpdates } =
+    prefs && typeof prefs === "object" ? prefs : {};
   return notificationsEnabled !== false && notifyUpdates !== false;
 }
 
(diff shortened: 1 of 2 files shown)


Chadpiha:
Thank you @hsusul — tidy PR and a thorough verification list, as always.

Closing this along with #1827: both call sites (`updater.js`) pass `windowManager?.notificationPrefs`, which is initialized in the constructor to a complete object and only ever key-mutated behind an existing `!prefs || typeof prefs !== "object"` guard at the IPC boundary (`sync-notification-preferences`) — so `null` can't reach the helper. The `undefined` case (missing windowManager via `?.`) is already covered by the `= {}` default and pinned by an existing test.

There's also a behavioral reason we'd keep the throw: this helper enforces the #1605 contract ("only an explicit false disables update checks"). If pref state were ever corrupted to null, silently returning `true` would quietly re-enable update checks — for that invariant we'd rather crash loudly than fail open.

Appreciate you poking at the updater gating — thanks again!

PR #1830, opened 2026-08-25
fix(transcription): handle nullish params and normalize cloudTranscriptionMode in resolveStreamingFallbackTarget

### LINKED ISSUE #1829: resolveStreamingFallbackTarget throws TypeError when called without arguments and ignores mixed-case cloudTranscriptionMode
### Describe the bug

In `src/helpers/transcriptionFallback.js`:

1. `resolveStreamingFallbackTarget({ useLocalWhisper, cloudTranscriptionMode, isSignedIn })` does not default its parameter, throwing:
   `TypeError: Cannot destructure property 'useLocalWhisper' of 'undefined' / 'null'` when called without arguments or when passed `null`/`undefined`.
2. `cloudTranscriptionMode === "openwhispr"` uses strict case-sensitive matching without whitespace trimming. If passed `"OpenWhispr"` or `" openwhispr "`, it treats the mode as non-cloud and defaults to `"byok"`.

### To Reproduce

1. Call `resolveStreamingFallbackTarget()` -> throws `TypeError`.
2. Call `resolveStreamingFallbackTarget(null)` -> throws `TypeError`.
3. Call `resolveStreamingFallbackTarget({ useLocalWhisper: false, cloudTranscriptionMode: "OpenWhispr", isSignedIn: true })` -> returns `"byok"` (expected `"cloud"`).

### Expected Behavior

`resolveStreamingFallbackTarget` should safely handle `null`, `undefined`, or missing parameters, and normalize `cloudTranscriptionMode` case-insensitively.

hsusul:
Fixes #1829

### Problem
In `src/helpers/transcriptionFallback.js`:
1. `resolveStreamingFallbackTarget` threw `TypeError: Cannot destructure property 'useLocalWhisper' of 'undefined' / 'null'` when called without arguments or passed `null`/`undefined`.
2. `cloudTranscriptionMode === "openwhispr"` did not normalize casing or whitespace, treating `"OpenWhispr"` as non-cloud and returning `"byok"`.

### Solution
- Defaulted parameters with `params || {}` before destructuring.
- Normalized `cloudTranscriptionMode` with `trim().toLowerCase()`.
- Added unit tests in `test/helpers/transcriptionFallback.test.js`.

### Verification
- `node --test test/helpers/transcriptionFallback.test.js` (passes, 4/4 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/transcriptionFallback.js
+++ b/src/helpers/transcriptionFallback.js
@@ -1,11 +1,10 @@
 // Where a streaming session's batch fallback goes. "skip" keeps a signed-out
 // cloud user's audio from being diverted to a leftover BYOK provider.
-export function resolveStreamingFallbackTarget({
-  useLocalWhisper,
-  cloudTranscriptionMode,
-  isSignedIn,
-}) {
-  const isCloudMode = !useLocalWhisper && cloudTranscriptionMode === "openwhispr";
+export function resolveStreamingFallbackTarget(params = {}) {
+  const { useLocalWhisper, cloudTranscriptionMode, isSignedIn } = params || {};
+  const normMode =
+    typeof cloudTranscriptionMode === "string" ? cloudTranscriptionMode.trim().toLowerCase() : "";
+  const isCloudMode = !useLocalWhisper && normMode === "openwhispr";
   if (isCloudMode) return isSignedIn ? "cloud" : "skip";
   return "byok";
 }
--- a/test/helpers/transcriptionFallback.test.js
+++ b/test/helpers/transcriptionFallback.test.js
@@ -38,3 +38,26 @@ test("BYOK mode falls back to the user's own provider", async () => {
     "byok"
   );
 });
+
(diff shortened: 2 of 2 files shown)


Chadpiha:
Thanks @hsusul — closing with #1829 (same convention as the #1840 note).

The only caller passes `getSettings()`, which is a Zustand state snapshot and can't be nullish. And `cloudTranscriptionMode` is written as a lowercase literal by every write site in the app's history, so old localStorage can't hold a mixed-case value either. The deciding factor on the normalization half: this value is compared case-sensitively in ~15 other places (audioManager, dictationStreamingRouting, the settings store's own `deriveTranscriptionMode`, …). Normalizing in just this one reader would make the fallback route disagree with the streaming session router for the same stored value — if mixed case ever became real, the fix would belong at the write/store boundary so all readers stay consistent.

Thanks for the continued energy! 🙏

PR #1832, opened 2026-08-25
fix(onboarding): normalize demo and input kinds in isOnboardingInputAllowed

### LINKED ISSUE #1831: isOnboardingInputAllowed fails on mixed-case or untrimmed demo and input kinds
### Describe the bug

In `src/helpers/onboardingInputPolicy.js`:

`isOnboardingInputAllowed(onboardingActive, demoKind, inputKind)` performs strict case-sensitive checks against `ONBOARDING_DEMO_KINDS` and `demoKind === inputKind`. When `demoKind` or `inputKind` is passed with mixed casing (e.g. `"Dictation"`, `"Assistant"`) or surrounding whitespace, it returns `false` during active onboarding demos instead of allowing the corresponding input.

### To Reproduce

1. Call `isOnboardingInputAllowed(true, "Dictation", "dictation")` -> returns `false` (expected `true`).
2. Call `isOnboardingInputAllowed(true, " Assistant ", "assistant")` -> returns `false` (expected `true`).

### Expected Behavior

`isOnboardingInputAllowed` should normalize `demoKind` and `inputKind` case-insensitively with whitespace trimming.

hsusul:
Fixes #1831

### Problem
In `src/helpers/onboardingInputPolicy.js`:
`isOnboardingInputAllowed` used strict case-sensitive checks against `ONBOARDING_DEMO_KINDS`, returning `false` during onboarding demos if `demoKind` or `inputKind` had mixed casing or whitespace (e.g. `"Dictation"`, `"Assistant"`).

### Solution
- Normalized `demoKind` and `inputKind` with `trim().toLowerCase()`.
- Added unit tests in `test/helpers/onboardingInputPolicy.test.js`.

### Verification
- `node --test test/helpers/onboardingInputPolicy.test.js` (passes, 5/5 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/onboardingInputPolicy.js
+++ b/src/helpers/onboardingInputPolicy.js
@@ -2,7 +2,9 @@ const ONBOARDING_DEMO_KINDS = new Set(["dictation", "assistant"]);
 
 function isOnboardingInputAllowed(onboardingActive, demoKind, inputKind) {
   if (!onboardingActive) return true;
-  return ONBOARDING_DEMO_KINDS.has(demoKind) && demoKind === inputKind;
+  const normDemo = typeof demoKind === "string" ? demoKind.trim().toLowerCase() : "";
+  const normInput = typeof inputKind === "string" ? inputKind.trim().toLowerCase() : "";
+  return ONBOARDING_DEMO_KINDS.has(normDemo) && normDemo === normInput;
 }
 
 module.exports = { ONBOARDING_DEMO_KINDS, isOnboardingInputAllowed };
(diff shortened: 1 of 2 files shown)


Chadpiha:
Thanks for this, @hsusul! Closing along with #1831 — here's the reasoning.

`demoKind` can only ever be stored after passing two strict `ONBOARDING_DEMO_KINDS.has()` checks: once at the IPC boundary (`onboarding-demo-begin` in `ipcHandlers.js`) and again in `windowManager.beginOnboardingDemo`. Every `inputKind` call site is a hardcoded lowercase literal in `windowManager.js`, and the TS type is the exact union `"dictation" | "assistant"`. So a mixed-case or padded kind can't reach this function in the app.

The strictness here is also deliberate: this is a fail-closed gate (there's even a test on main named "unknown demo kinds fail closed while onboarding is active"). Normalizing at the policy layer would make it looser than the two strict gates it sits behind, so a typo'd kind would be tolerated in one layer and rejected in two others — we'd rather it fail loudly and consistently.

Thanks for looking so closely at the onboarding flow — it's brand new (#1670), so extra eyes on it are very welcome!

PR #1834, opened 2026-08-25
fix(speaker-count): validate primitive type in normalizeStoredSpeakerCount

### LINKED ISSUE #1833: normalizeStoredSpeakerCount accepts boolean true and array values as valid speaker counts
### Describe the bug

In `src/helpers/speakerCount.js`:

`normalizeStoredSpeakerCount(value)` converts its input using `const count = Number(value)` and checks `Number.isInteger(count) && count >= 1`.

In JavaScript:
- `Number(true)` is `1`, so `normalizeStoredSpeakerCount(true)` returns `1`.
- `Number(["2"])` is `2`, so `normalizeStoredSpeakerCount(["2"])` returns `2`.

When non-numeric types (such as booleans or arrays) arrive via cloud sync or loosely structured stored data, `normalizeStoredSpeakerCount` erroneously returns an integer instead of degrading to `null`.

### To Reproduce

1. Call `normalizeStoredSpeakerCount(true)` -> returns `1` (expected `null`).
2. Call `normalizeStoredSpeakerCount(["2"])` -> returns `2` (expected `null`).

### Expected Behavior

`normalizeStoredSpeakerCount` should only accept numbers or numeric strings (`typeof value === "number" || typeof value === "string"`), returning `null` for booleans, arrays, and other non-numeric types.

hsusul:
Fixes #1833

### Problem
In `src/helpers/speakerCount.js`:
`normalizeStoredSpeakerCount` converted inputs directly with `Number(value)`, causing boolean `true` to evaluate to `1` and array `["2"]` to evaluate to `2` instead of rejecting non-numeric types as `null`.

### Solution
- Restricted `normalizeStoredSpeakerCount` to only numbers and strings (`typeof value === "number" || typeof value === "string"`) before numeric conversion.
- Added unit tests in `test/helpers/speakerCount.test.js`.

### Verification
- `node --test test/helpers/speakerCount.test.js` (passes, 3/3 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/speakerCount.js
+++ b/src/helpers/speakerCount.js
@@ -4,6 +4,7 @@ const { MAX_SPEAKER_COUNT } = require("../constants/speakerDetection.json");
 // 1 is meaningless. Cloud sync persists whatever the server sends, so invalid
 // values arrive as data rather than as programmer error and must degrade to null.
 function normalizeStoredSpeakerCount(value) {
+  if (typeof value !== "number" && typeof value !== "string") return null;
   const count = Number(value);
   if (!Number.isInteger(count) || count < 1) return null;
   return Math.min(count, MAX_SPEAKER_COUNT);
--- a/test/helpers/speakerCount.test.js
+++ b/test/helpers/speakerCount.test.js
@@ -7,6 +7,7 @@ const { MAX_SPEAKER_COUNT } = require("../../src/constants/speakerDetection.json
 test("keeps a usable stored count", () => {
   assert.equal(normalizeStoredSpeakerCount(1), 1);
   assert.equal(normalizeStoredSpeakerCount(3), 3);
+  assert.equal(normalizeStoredSpeakerCount("3"), 3);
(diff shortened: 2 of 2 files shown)


Chadpiha:
Thanks @hsusul — nice clean diff and the tests are in exactly our house style.

Closing this one together with #1833 after tracing where the stored value can actually come from: the cloud path is validated server-side with `z.number().int().nullish()` into a Postgres `INTEGER` column, the local path is a SQLite `INTEGER` read via better-sqlite3 (which can only yield numbers or null, and rejects binding booleans/arrays), and the renderer-side twin is typed `number | null | undefined`. So a boolean or array can't arrive at this helper in practice, and per our convention we keep type validation at those boundaries rather than re-checking downstream.

One thing worth knowing for the future: this helper has an intentional twin — `normalizeSpeakerCount` in `src/services/uploadNotes.ts` — and `test/services/uploadNotes.test.js` exists specifically to hold the two to identical outputs. Any behavior change to one needs the twin and the parity test updated in the same PR.

Really appreciate you digging into this area — thank you!

PR #1838, opened 2026-08-25
fix(chat): handle nullish messages and malformed elements in conversationChunker

### LINKED ISSUE #1837: chunkConversation throws TypeError when messages is nullish and formatChunkText handles missing fields improperly
### Describe the bug

In `src/helpers/conversationChunker.js`:

1. `chunkConversation(title, messages)` calls `messages.filter(...)` without validating that `messages` is an array. When called with `null`, `undefined`, or non-array inputs, it throws:
   `TypeError: Cannot read properties of null (reading 'filter')`
2. In `formatChunkText(title, messages)`, if `title` is nullish or a non-string, `${title}\n` outputs `"null\n..."` or `"undefined\n..."`. If any message object has nullish `role` or `content`, `${m.role}: ${m.content}` stringifies to `"undefined: undefined"`.

### To Reproduce

1. Call `chunkConversation("Title", null)` -> throws `TypeError`.
2. Call `chunkConversation("Title", undefined)` -> throws `TypeError`.
3. Call `chunkConversation(null, [{ role: "user", content: "Hello" }])` -> formatted text begins with `"null\n"`.

### Expected Behavior

- `chunkConversation` should safely handle nullish or non-array `messages`, returning `[]`.
- `formatChunkText` should safely sanitize `title` and message properties.

hsusul:
Fixes #1837

### Problem
In `src/helpers/conversationChunker.js`:
1. `chunkConversation` called `messages.filter(...)` without ensuring `messages` was an array, throwing `TypeError: Cannot read properties of null (reading 'filter')` when called with `null` or `undefined`.
2. `formatChunkText` emitted `"null\n..."` if `title` was nullish and `"undefined: undefined"` if message items were missing role/content properties.

### Solution
- Array-coerced and filtered nullish/non-object message items in `chunkConversation`.
- Safely sanitized `title` and formatted messages with safe defaults in `formatChunkText`.
- Added complete test suite in `test/helpers/conversationChunker.test.js`.

### Verification
- `node --test test/helpers/conversationChunker.test.js` (passes, 4/4 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/conversationChunker.js
+++ b/src/helpers/conversationChunker.js
@@ -2,7 +2,10 @@ const CHUNK_SIZE = 5;
 const CHUNK_OVERLAP = 2;
 
 function chunkConversation(title, messages) {
-  const relevant = messages.filter((m) => m.role !== "system");
+  const messageList = Array.isArray(messages) ? messages : [];
+  const relevant = messageList.filter(
+    (m) => m && typeof m === "object" && m.role !== "system"
+  );
   if (relevant.length === 0) return [];
 
   if (relevant.length <= CHUNK_SIZE) {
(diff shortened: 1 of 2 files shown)


Chadpiha:
Thanks @hsusul! Closing along with #1837, same reasoning as the note over on #1840: `messages` always comes from better-sqlite3's `.all()` (an array of rows, never null), and the `agent_messages`/`agent_conversations` tables have carried `NOT NULL` + `CHECK(role IN (...))` constraints since they were created — so no database, however old, can hold the shapes this guards against. The caller additionally wraps the call in a try/catch that logs and continues.

One small heads-up for next time: the diff also tweaks the chunk text for existing inputs (title trimming, header omission for empty titles, defaulting missing roles) — changes like that affect what gets embedded, so they're best split into their own PR where they can be weighed on their own merits.

Appreciate the thorough test-writing as always! 🙏

Issue #1435, opened 2026-08-02
Parakeet streaming: opening of longer dictations dropped from paste while live preview shows full text (occurs with cleanup disabled)


GuanHukd:
## Summary

With text cleanup **enabled**, the **beginning of a longer dictation is silently dropped** from the pasted output. The live transcription preview shows the complete text, and disabling cleanup pastes the complete text — so the audio, the transcription, and the transcript assembly are all correct. The loss occurs somewhere in the cleanup stage.

The amount lost **varies between runs of the same passage**, which suggests a race rather than a fixed truncation.

## Environment

- OpenWhispr **1.8.1**
- Windows 11 Pro 25H2 (build 26200)
- AMD Ryzen AI 9 365 w/ Radeon 880M (no NVIDIA GPU), 16 GB RAM
- STT: **Local → Parakeet Unified EN 0.6B** (int8, CPU) — streaming path via `sherpa-onnx-ws`
- Cleanup: OpenAI `GPT-4.1 Mini` (non-reasoning), BYOK

## Reproduction

Dictate a passage of roughly 30 seconds / five sentences with natural pauses between them.

**Spoken:**
> Right, so we need the MHRA notification sorted before the standstill period ends. Second thing, Parfetts want the twin-pack listing confirmed by Friday. Third, chase the Bao'an factory about the duty stamps. Fourth, Booker's rate of sale on the PWR Wall is up, so we should push more facings. And finally, book the InterTabac stand before prices go up.

**Cleanup DISABLED — complete (all five points present):**
> Right, so we need the MHRA notification sorted before the standstill period ends. Second thing, Parfits want the Twin Pack list in confirmed by Friday. Third, chase the bunk battery above the duty stands. Fourth, booker's rate of sale on the PWR wall is up, so we should push more facings. And finally, book the into the backstand before prices go up

**Cleanup ENABLED, run 1 — first sentence missing entirely:**
> Second thing, Parfetts on the twin pack listed confirmed by Friday. Third, chase the Borough Factory about the duty stamps for Booker's rate of sale on the PWR wall is up, so we should push more facings. And finally, book the into backstand before the prices go up.

**Cleanup ENABLED, run 2 — first sentence and most of the second missing:**
> Listed confirmed by Friday third. Chase the bone factory back. The duty stamps for Booker's rate of sale on the PWR wall is up, so we should push more faces. And finally, book them into the backstand before prices go up.

[...]

Issue #871, opened 2026-05-30
Audio capture intermittently produces empty (~110 byte) webm files — transcription fails with 'FFmpeg: End of file' before whisper-server runs


wlaitinen:
## Summary

OpenWhispr v1.7.2 on Windows 11 **intermittently records empty `.webm` files**
(~110 bytes, header only, no audio data). When that happens, transcription
fails with an FFmpeg "End of file" error before `whisper-server` is ever
invoked. The same recording session produces a mix of valid recordings
(100 KB – 2+ MB) and broken empty recordings (~110 bytes), with no obvious
user-visible difference at record time.

## Symptom (UI error)

```
Transcription failed: Local Whisper failed: Error invoking remote method
'transcribe-local-whisper': Error: FFmpeg exited with code 3753488571:
  avdevice 60. 3.100 / 60. 3.100
  libavfilter 9. 12.100 / 9. 12.100
  libswscale 7. 5.100 / 7. 5.100
  libswresample 4. 12.100 / 4. 12.100
  libpostproc 57. 3.100 / 57. 3.100
[matroska,webm @ ...] 0x00 at pos 36 (0x24) invalid as first byte of an EBML number
[in#0 @ ...] Error opening input: End of file
Error opening input file C:\Users\WILLIA~1\AppData\Local\Temp\whisper-input-1780124484064.webm
Error opening input files: End of file
```

## Evidence on disk

Listing `%APPDATA%\open-whispr\audio\` after several short dictations in one
session:

```
   227,306 bytes  08:02:00  OpenWhispr-2026-05-30-07-02-00-17.webm   ← valid
       110 bytes  08:01:24  OpenWhispr-2026-05-30-07-01-24-16.webm   ← BROKEN: header only
   108,487 bytes  08:00:54  OpenWhispr-2026-05-30-07-00-54-15.webm   ← valid
   110,420 bytes  07:55:11  OpenWhispr-2026-05-30-06-55-11-14.webm   ← valid
```

The 110-byte file is a webm container with no audio frames — just enough
EBML header for FFmpeg to start parsing, then EOF (which matches the
"position 36 invalid as first byte of an EBML number" in the error). It
indicates audio capture either never began or ended before any audio bytes
were written, yet the recording was still "completed" and handed downstream
to FFmpeg unconditionally.

## Environment

- **OpenWhispr**: 1.7.2 (fresh uninstall + reinstall)
- **OS**: Windows 11
[...]

PR #1840, opened 2026-08-25
fix(audio): guard evaluateFinishedRecording against null and non-object parameters

### LINKED ISSUE #1839: evaluateFinishedRecording throws TypeError when called with null
### Describe the bug

In `src/helpers/recordingValidation.js`:

`evaluateFinishedRecording({ blobSize, receivedAudioData } = {})` defaults its parameter using `= {}`, which only triggers when the argument is `undefined`. When called with `null` or a non-object primitive, it throws:

`TypeError: Cannot destructure property 'blobSize' of 'null' as it is null.`

### To Reproduce

1. Call `evaluateFinishedRecording(null)` -> throws `TypeError`.

### Expected Behavior

`evaluateFinishedRecording` should safely handle `null` and non-object inputs, returning `{ usable: false, reason: "no-audio-data" }` without throwing an error.

hsusul:
Fixes #1839

### Problem
In `src/helpers/recordingValidation.js`:
`evaluateFinishedRecording({ blobSize, receivedAudioData } = {})` threw `TypeError: Cannot destructure property 'blobSize' of 'null' as it is null` when passed `null`.

### Solution
- Handled `null` and non-object parameters safely using `params && typeof params === "object" ? params : {}`.
- Added unit tests in `test/helpers/recordingValidation.test.js`.

### Verification
- `node --test test/helpers/recordingValidation.test.js` (passes, 10/10 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/recordingValidation.js
+++ b/src/helpers/recordingValidation.js
@@ -3,7 +3,9 @@ import { isEmptyRecording } from "./recordingGuard.js";
 // Decide whether a finished MediaRecorder session carries real audio before we
 // hand it to the transcription backend. A fast tap can flush only the container
 // header, or deliver no chunks at all, which crashes FFmpeg. See issue #871.
-export function evaluateFinishedRecording({ blobSize, receivedAudioData } = {}) {
+export function evaluateFinishedRecording(params = {}) {
+  const { blobSize, receivedAudioData } =
+    params && typeof params === "object" ? params : {};
   if (!receivedAudioData) {
     return { usable: false, reason: "no-audio-data" };
   }
--- a/test/helpers/recordingValidation.test.js
+++ b/test/helpers/recordingValidation.test.js
@@ -43,10 +43,14 @@ test("rejects 255 bytes (just under the threshold)", async () => {
   });
 });
 
-test("treats missing / undefined args as unusable (defensive, does not throw)", async () => {
+test("treats missing / undefined / null args as unusable (defensive, does not throw)", async () => {
   const { evaluateFinishedRecording } = await load();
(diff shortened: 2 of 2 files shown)


Chadpiha:
Thanks for the careful work here, @hsusul — the tests are well written and the verification checklist (typecheck, lint, i18n, build) is genuinely appreciated.

I'm going to close this one, along with #1839, for a reason that's about our conventions rather than your code: `evaluateFinishedRecording` has exactly one caller (`audioManager.js` in `finalizeBatchRecording`), and it always passes a freshly built object literal with a Blob that's guaranteed non-null a few lines earlier — so `null` can't reach the helper through any real path. Our house rule (per CLAUDE.md) is to validate at real boundaries (IPC, external APIs, persisted data) and trust internal callers past that point.

There's also a subtle reason we prefer the current behavior: if a future refactor ever *did* pass `null` here, we'd want the loud crash at the bug site — a swallowed error here would surface as a silent empty transcription, which is a failure mode we've worked hard to eliminate (see #871, #1435).

Thanks again, and please keep the contributions coming! 🙏

Chadpiha:
One general note, since you've clearly got an eye for edge cases: our convention (CLAUDE.md) is to validate exactly once, at real trust boundaries — IPC payloads arriving from the renderer, external API responses not already mediated by a typed SDK, persisted data an older app version could have written in a different shape, user-editable files — and to trust internal callers past that point, so that impossible states fail loudly instead of being silently coerced. Hardening PRs at those boundaries are very welcome! The strongest ones start from an observable app behavior (a crash log, a repro via real user actions) rather than a direct helper call. Would love to see more contributions in that direction — thanks for the energy you're putting into OpenWhispr.

PR #1847, opened 2026-08-26
fix(calendar): support pre-parsed conference_data and guard primitives in getMeetingJoinUrl

### LINKED ISSUE #1846: getMeetingJoinUrl fails on pre-parsed conference_data objects and extractMeetingUrl fails on string/iterable inputs
### Describe the bug

In `src/helpers/meetingJoinUrl.js`:

1. `getMeetingJoinUrl(event)` assumes `event.conference_data` is always a JSON string. When `conference_data` is provided as an already-parsed object (e.g. from structured IPC payloads or memory caches), `JSON.parse(event.conference_data)` executes on `"[object Object]"` and throws a SyntaxError, erroneously returning `null`.
2. `extractMeetingUrl(candidates)` checks `!Array.isArray(candidates)`, returning `null` if `candidates` is passed as a single string or a `Set` of candidate texts.

### To Reproduce

1. Call `getMeetingJoinUrl({ conference_data: { entryPoints: [{ entryPointType: "video", uri: "https://zoom.us/j/123" }] } })` -> returns `null` (expected `"https://zoom.us/j/123"`).
2. Call `extractMeetingUrl("Join here: https://zoom.us/j/123456789")` -> returns `null` (expected `"https://zoom.us/j/123456789"`).

### Expected Behavior

- `getMeetingJoinUrl` should handle both serialized JSON strings and pre-parsed objects in `conference_data`.
- `extractMeetingUrl` should accept single strings, arrays, and iterables.

hsusul:
Fixes #1846

### Problem
In `src/helpers/meetingJoinUrl.js`:
1. `getMeetingJoinUrl` assumed `event.conference_data` was always a serialized JSON string. If passed an already-parsed object, `JSON.parse` failed with a SyntaxError on `"[object Object]"` and caught to `null`.
2. Primitive non-object `event` inputs were not guarded before property access.

### Solution
- Supported both JSON-serialized strings and parsed objects for `conference_data`.
- Validated that `event` is an object and guarded entry points array handling.
- Added unit tests in `test/helpers/meetingJoinUrl.test.js`.

### Verification
- `node --test test/helpers/meetingJoinUrl.test.js` (passes, 14/14 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/meetingJoinUrl.js
+++ b/src/helpers/meetingJoinUrl.js
@@ -37,14 +37,19 @@ function isMeetingUrl(value) {
 }
 
 export function getMeetingJoinUrl(event) {
-  if (!event) return null;
-  const hangoutLink = event.hangout_link?.trim();
+  if (!event || typeof event !== "object") return null;
+  const hangoutLink =
+    typeof event.hangout_link === "string" ? event.hangout_link.trim() : "";
   if (hangoutLink) return hangoutLink;
   if (!event.conference_data) return null;
   try {
-    const data = JSON.parse(event.conference_data);
-    const uri = data?.entryPoints
-      ?.find((ep) => ep.entryPointType === "video" && ep.uri?.trim())
+    const data =
+      typeof event.conference_data === "string"
+        ? JSON.parse(event.conference_data)
+        : event.conference_data;
+    const entryPoints = Array.isArray(data?.entryPoints) ? data.entryPoints : [];
+    const uri = entryPoints
+      .find((ep) => ep?.entryPointType === "video" && typeof ep?.uri === "string" && ep.uri.trim())
       ?.uri?.trim();
     return uri || null;
   } catch {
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks @hsusul! Closing: `conference_data` has exactly one non-null producer — `JSON.stringify(item.conferenceData)` in `googleCalendarManager.js`, unchanged since the feature first landed — written into a SQLite TEXT column, and every reader hands back raw rows (renderer events are structured-clone copies of those rows over IPC). The pre-parsed-object shape doesn't exist anywhere in the codebase. Two smaller notes from verifying this: primitives already return `null` on current main (three of the four new assertions pass unchanged against main), and a non-string `hangout_link` throwing loudly is the behavior we prefer for an impossible state. The real boundaries (Google/Graph API responses) are normalized once in the calendar managers, and we trust internal callers past that point. Closing #1846 along with this.

PR #1849, opened 2026-08-26
fix(settings): guard getLlmRequestTimeoutSeconds against null and non-object parameters

### LINKED ISSUE #1848: getLlmRequestTimeoutSeconds throws TypeError when called with null
### Describe the bug

In `src/helpers/llmRequestTimeout.js`:

`getLlmRequestTimeoutSeconds({ streaming = false } = {})` uses default destructuring parameter syntax `= {}`, which only applies when the parameter is `undefined`. When called with `null` or a non-object primitive, it throws:

`TypeError: Cannot destructure property 'streaming' of 'null' as it is null.`

### To Reproduce

1. Call `getLlmRequestTimeoutSeconds(null)` -> throws `TypeError`.

### Expected Behavior

`getLlmRequestTimeoutSeconds` should safely handle `null` and non-object inputs, returning the non-streaming default timeout (30 seconds) without throwing an error.

hsusul:
Fixes #1848

### Problem
In `src/helpers/llmRequestTimeout.js`:
`getLlmRequestTimeoutSeconds({ streaming = false } = {})` threw `TypeError: Cannot destructure property 'streaming' of 'null' as it is null` when passed `null`.

### Solution
- Guarded destructuring with `params && typeof params === "object" ? params : {}`.
- Added unit tests in `test/helpers/llmRequestTimeout.test.js`.

### Verification
- `node --test test/helpers/llmRequestTimeout.test.js` (passes, 3/3 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/llmRequestTimeout.js
+++ b/src/helpers/llmRequestTimeout.js
@@ -1,6 +1,7 @@
 const LLM_REQUEST_TIMEOUT_SECONDS = 30;
 const LLM_STREAMING_TIMEOUT_SECONDS = 60;
 
-export function getLlmRequestTimeoutSeconds({ streaming = false } = {}) {
+export function getLlmRequestTimeoutSeconds(params = {}) {
+  const { streaming = false } = params && typeof params === "object" ? params : {};
   return streaming ? LLM_STREAMING_TIMEOUT_SECONDS : LLM_REQUEST_TIMEOUT_SECONDS;
 }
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks @hsusul! Closing: all five call sites of `getLlmRequestTimeoutSeconds` pass either no argument (`ReasoningService.ts:324`, `gemini.ts:88`, `openai.ts:220`, `tinfoil.ts:43`) or the inline literal `{ streaming: true }` (`ReasoningService.ts:624`) — the parameter is a compile-time constant at every site, so `null` or a primitive can never arrive. Coercing those to `{}` would mask a future caller bug as a silently-wrong 30s timeout on a streaming request rather than a loud TypeError, which is the opposite of what we want. Closing #1848 along with this.

PR #1851, opened 2026-08-26
fix(transcription): handle nullish inputs safely in cloudChunkPolicy helpers

### LINKED ISSUE #1850: cloudChunkPolicy helper functions throw TypeErrors when passed nullish error or results arguments
### Describe the bug

In `src/helpers/cloudChunkPolicy.js`:

1. `isTransientChunkError(err)` accesses `err.code` and `err.statusCode` without null checking, throwing `TypeError: Cannot read properties of null (reading 'code')` when passed `null` or `undefined`.
2. `isNetworkLevelFailure(err)` accesses `err.statusCode` without null checking, throwing `TypeError` on nullish inputs.
3. `summarizeChunkResults(results)` uses `for (const result of results)`, throwing `TypeError: results is not iterable` when passed `null`.
4. `assembleChunkTranscript(results, ...)` accesses `results.length`, throwing `TypeError: Cannot read properties of null (reading 'length')` when passed `null`.

### To Reproduce

1. Call `isTransientChunkError(null)` -> throws `TypeError`.
2. Call `isNetworkLevelFailure(null)` -> throws `TypeError`.
3. Call `summarizeChunkResults(null)` -> throws `TypeError`.
4. Call `assembleChunkTranscript(null, 10, 100)` -> throws `TypeError`.

### Expected Behavior

These functions should defensively handle nullish and non-object/non-iterable inputs without throwing runtime errors.

hsusul:
Fixes #1850

### Problem
In `src/helpers/cloudChunkPolicy.js`:
1. `isTransientChunkError` and `isNetworkLevelFailure` threw `TypeError` when called with `null` or `undefined`.
2. `summarizeChunkResults` threw `TypeError: results is not iterable` on `null`.
3. `assembleChunkTranscript` threw `TypeError: Cannot read properties of null (reading 'length')` on `null`.

### Solution
- Added defensive null/type guards to `isTransientChunkError`, `isNetworkLevelFailure`, `summarizeChunkResults`, and `assembleChunkTranscript`.
- Added unit tests in `test/helpers/cloudChunkPolicy.test.js`.

### Verification
- `node --test test/helpers/cloudChunkPolicy.test.js` (passes, 30/30 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/cloudChunkPolicy.js
+++ b/src/helpers/cloudChunkPolicy.js
@@ -53,6 +53,7 @@ const FATAL_CHUNK_CODES = new Set(["AUTH_EXPIRED", "LIMIT_REACHED"]);
 const NON_RETRYABLE_CHUNK_CODES = new Set([...FATAL_CHUNK_CODES, "NO_SPEECH_DETECTED"]);
 
 function isTransientChunkError(err) {
+  if (!err || typeof err !== "object") return true;
   if (NON_RETRYABLE_CHUNK_CODES.has(err.code)) return false;
   return !err.statusCode || err.statusCode >= 500;
 }
@@ -62,7 +63,7 @@ function isTransientChunkError(err) {
 // before the next attempt. Errors carrying a statusCode or a business code
 // prove the connection works.
 function isNetworkLevelFailure(err, { timedOut = false } = {}) {
-  return timedOut || (!err.statusCode && !err.code);
+  return timedOut || (!err?.statusCode && !err?.code);
 }
 
 // Fatal TLS alerts and HTTP/2/QUIC protocol errors poison every stream on the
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks @hsusul! Closing: all four of these helpers are called only from `chunkedCloudTranscribe` in `ipcHandlers.js`. The error arguments come exclusively from `catch` blocks over producers that construct real `Error` objects (and the `isTransientChunkError(failure)` call at `:462` is additionally dominated by `if (!failure) break;` at `:446`), while `results` is the locally built `new Array(totalChunks).fill(null)` from `:393` and is never reassigned. So nullish inputs can't reach these functions on any real path, and the guards would turn an impossible state into a silent empty transcript / zeroed summary instead of a loud failure. Closing #1850 along with this.

PR #1853, opened 2026-08-26
fix(audio): default now to current timestamp and validate finite numbers in isMicWarm

### LINKED ISSUE #1852: isMicWarm returns false when now parameter is omitted and does not validate non-finite inputs
### Describe the bug

In `src/helpers/micWarmState.js`:

`export const isMicWarm = (lastWarmedAt, now, ttlMs = MIC_WARM_TTL_MS) => ...`

When `isMicWarm(lastWarmedAt)` is called without explicitly passing the `now` timestamp argument:
`now` is `undefined`, causing `now - lastWarmedAt` to evaluate to `NaN`. `elapsed >= 0 && elapsed < ttlMs` evaluates to `false`, even if the microphone was warmed milliseconds ago.

### To Reproduce

1. Call `isMicWarm(Date.now())` -> returns `false` (expected `true`).

### Expected Behavior

`isMicWarm` should default `now` to `Date.now()` and validate finite numbers for `now` and `ttlMs`.

hsusul:
Fixes #1852

### Problem
In `src/helpers/micWarmState.js`:
When `isMicWarm(lastWarmedAt)` was called without providing the `now` parameter, `now` was `undefined`, producing `NaN` for elapsed calculation and always returning `false` regardless of when the microphone was warmed.

### Solution
- Defaulted `now` to `Date.now()`.
- Validated that timestamps and TTL values are finite numbers.
- Added unit tests in `test/helpers/micWarmState.test.js`.

### Verification
- `node --test test/helpers/micWarmState.test.js` (passes, 11/11 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/micWarmState.js
+++ b/src/helpers/micWarmState.js
@@ -9,8 +9,19 @@ export const WARMUP_ACQUIRE_TIMEOUT_MS = 15000;
 
 // True only when the mic was warmed within the TTL. A negative elapsed (clock
 // jumped backwards) returns false so we err on the side of re-warming.
-export const isMicWarm = (lastWarmedAt, now, ttlMs = MIC_WARM_TTL_MS) => {
-  if (typeof lastWarmedAt !== "number" || lastWarmedAt <= 0) return false;
-  const elapsed = now - lastWarmedAt;
-  return elapsed >= 0 && elapsed < ttlMs;
+export const isMicWarm = (lastWarmedAt, now = Date.now(), ttlMs = MIC_WARM_TTL_MS) => {
+  if (typeof lastWarmedAt !== "number" || !Number.isFinite(lastWarmedAt) || lastWarmedAt <= 0) {
+    return false;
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: `isMicWarm` has exactly two production call sites, both in `audioManager.js` (`:1070`, `:4229`), and both pass `Date.now()` explicitly — the omitted-`now` shape never occurs. `lastWarmedAt` is always `this._micWarmedAt`, which is only ever assigned `0` or `Date.now()` internally and never crosses an IPC/DB/user-file boundary, and non-finite values already return `false` on main via the `elapsed >= 0` comparison, so the new guards don't change behavior even hypothetically. Per our convention (validate once at real trust boundaries, trust internal callers past that point), we'd rather a future caller bug fail loudly than be coerced to a sentinel. Closing #1852 along with this.

PR #1859, opened 2026-08-26
Restore Voice Assistant auto-paste into focused inputs

boseq:
## Summary

- restore Voice Assistant response delivery into a focused writable input when Auto-Paste is enabled
- capture an opaque editable-target session before processing and revalidate it before pasting the completed response
- when no writable input is selected, copy the completed response to the clipboard and keep it visible in the floating Assistant panel
- if a captured target changes or paste fails, fall back to the same safe clipboard delivery instead of losing the response
- show the existing Copy button in its Copied state for six seconds after an automatic clipboard fallback
- keep selection replacement intact and add protected-field checks for macOS, Windows, and Linux
- update tests, documentation, and all supported locales

## Testing

- npm run lint
- npm run typecheck
- npm run build:renderer
- npm test — 3,043 tests; 2,865 passed, 177 skipped, 1 todo, 0 failed
- npm run i18n:check
- npm run compile:text-monitor (macOS arm64)


Chadpiha (approved):
Approved — reviewed end-to-end: traced the full delivery chain (capture → caret session → suppressed streaming → revalidate → paste / fail-closed fallback) on all three platforms, plus the streaming/cancel changes against the ControlPanel chat. Merged tree runs 2950 tests / 0 failures locally; typecheck, lint, and i18n:check clean.

What I specifically verified:

Revalidation is real on every platform. macOS compares the frontmost pid before trusting the app-local AX focus, Windows compares the live foreground HWND against the session target, Linux compares the active window — switching apps mid-answer falls back to panel + clipboard instead of pasting into the wrong window. Session kinds (caret vs selection) can't be cross-used, and the live-selection checks in all three probes close the clipboard-capture blind spot as advertised.
Cancelled and empty responses never paste and never touch the clipboard, and the explicit-cancel vs unmount disambiguation is pinned by a real live-root unmount test. Nice work on the stale-binary guard in isFocusedEditable (settling on the first monitor line instead of hanging).
No regressions in the existing paths: all changed shared seams (_bankPanelAgentCommand options, captureSelectedText options, useCopyFeedback API) were checked against their other consumers; the pasted:false → paste_failed change turns the previously silent accessibility-skipped "success" into a truthful error, which reads as a fix.

I pushed one commit to the branch: 58643ae1 closes a gap on the Linux Wayland (AT-SPI) path — atspi-pid targets carry no window class or exe name, so the terminal-signature check saw an empty signature and a terminal's empty prompt could verify as a writable caret (pasting a multi-line answer into a shell executes it). The fix resolves the pid's executable name before probing (same ps helper macOS uses) and refuses ATSPI_ROLE_TERMINAL inside both probe sources. Note the .c change gets its first compile in build-linux-text-monitor.yml, which only triggers on push to main — worth a manual workflow_dispatch before relying on the rebuilt release binary.

Two deliberate behavior changes I'm signing off on but want on the record: (1) explicit Esc/Stop cancels no longer persist partial replies to the saved conversation — this applies to the ControlPanel chat too, and was previously racy either way; (2) with auto-paste on and no writable caret, every completed standalone answer overwrites the user's clipboard, including purely informational queries, with the 6-second Copied state as the only signal. Both match the PR description; flagging so they're a decision, not an accident.

PR #1866, opened 2026-08-26
fix(windows): use endpoint-loopback for meeting audio (#1265)

### LINKED ISSUE #1265: No sound recorded within MS Teams meeting - Affected 1.7.4 and above - Windows
**Describe the bug**
The sound is no longer grabbed during Teams meetings with version 1.7.4 and above (up to 1.7.6).

**What was tested**
- all versions between 1.7.2 and 1.7.6 (both setup and portable version)
- Bluetooth headsets and built-in speaker/mic (also with disabled exclusive mode for applications)
- Other sound sources (mainly web browsers, voice is fetched from youtube without any issues)
- All parakeet, nemotron models and Whisper base 

**To Reproduce**
- just not recorded sound from the Teams meeting

**Desktop**

- OS: Windows 11 25H2 26200.8655
- Browser: Edge 150 / Vivaldi latest
- MS Teams 26183.1903.4892.4448

**Additional context**
Could be related to one of the fixes for 1.7.4 (copied from release notes):

"Native Windows meeting audio. Meetings now capture sound from every app on every output device instead of only your default speakers, via a native Windows helper. Older Windows automatically falls back to the previous method. (https://github.com/OpenWhispr/openwhispr/pull/960)"

amarpreet2209:
## What

Fixes #1265. MS Teams uses WASAPI exclusive mode for call audio, which bypasses the shared-mode audio engine. The existing process-loopback helper captures in shared mode, so it receives only silence during Teams meetings and the transcription is empty.

## Solution

Add endpoint-loopback capture mode to the native helper. This captures audio from the default render endpoint after Windows mixes all application audio together, hearing all apps regardless of their audio mode.

## Validation

- 13 new tests pass
- 20 existing meetingMicGate tests pass
- Manual verification with MS Teams device test

### DIFF
--- a/resources/windows-system-audio-helper.c
+++ b/resources/windows-system-audio-helper.c
@@ -1,16 +1,24 @@
 /**
  * Windows System Audio Helper
  *
- * Captures system audio for meeting transcription via WASAPI process
- * loopback (VAD\Process_Loopback, Windows 10 2004+). Runs in EXCLUDE mode
- * against OpenWhispr's own process tree, so it hears every application on
- * every render endpoint — independent of the default output device — while
- * never re-capturing OpenWhispr's own sounds.
+ * Captures system audio for meeting transcription via WASAPI.
+ *
+ * Two capture modes:
+ *   endpoint-loopback (default):
+ *     Captures from the default render endpoint after mixing. Hears all
+ *     applications including those using WASAPI exclusive mode (MS Teams,
+ *     Zoom, etc.). Captures from the default output device only.
+ *
+ *   process-loopback (Windows 10 2004+):
+ *     Captures via VAD\Process_Loopback in EXCLUDE mode against
+ *     OpenWhispr's own process tree. Hears every application on every
+ *     render endpoint, but cannot capture exclusive-mode streams.
  *
  * Commands:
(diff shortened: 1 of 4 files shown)


stantheman0128:
Support. Teams WASAPI exclusive mode bypassing shared-mode process-loopback is a real empty-transcript path (same cabinet as #1453, different handle). Capturing the default render endpoint after the mix is the right height: one helper mode instead of per-app shared-mode patches.

I did not rerun Teams live here. The new `windowsEndpointLoopback` tests plus the existing meetingMicGate suite are the right shape. Happy to see this land.

boseq:
Thanks for digging into this — you were right that endpoint loopback hears Teams when process loopback doesn't.

The exclusive-mode part doesn't hold up though: Teams' call audio is an ordinary shared-mode session, and endpoint loopback can't capture exclusive streams either. Windows just keeps that stream out of process loopback entirely — silence even when we aimed it straight at the Teams process.

We didn't take endpoint loopback as the default since it only hears the default output device, which is what #960 fixed. Instead we kept the native capture and made the existing silent-capture fallback actually fire; it was disarming after the first sound from any app. That's #2233, where you're credited for finding the working path. Closing this one — thanks again.

PR #1867, opened 2026-08-26
Add an opposite-edge dictation pill for Agent mode

boseq:
## Summary

- keep the existing Agent modal and shared voice-pill styling intact
- show a separate focusless dictation surface on the opposite edge of the same display
- reuse the production pill entrance timing, processing glow, and real microphone waveform in the companion
- host Live Transcript in the separate companion window when preview mode is enabled
- carry the recording input identity across IPC so Agent and translation hotkeys cannot activate the dictation companion
- keep plain dictation from animating the existing Agent footer pill
- preserve ordinary dictation routing, tap-to-toggle, and push-to-talk behavior while Agent is open

## Validation

- `npm test` (2,977 passed, 184 skipped, 0 failed)
- `npm run quality-check`
- `npm run i18n:check`
- `npm run build:renderer`


Chadpiha (approved):
Approving ✅ — with some additional commits pushed to the branch
Reviewed this end-to-end rather than hunk-by-hunk: traced the companion window's full lifecycle through the main process, walked eleven user journeys in the renderer, and re-verified every finding from the previous round against the actual code rather than the commit messages.

The fix round holds up. All 13 findings from the last review are genuinely fixed, and each one is pinned by a test that fails without its fix (I checked by reverting). The two that aren't automatically pinned — the macOS click-through reset and the App-level close-intent wiring — are the ones we'd agreed to cover manually, so that's as designed.

The feature does what the PR says it does. The companion appears only for plain dictation while the panel is open, on the opposite edge, focusless and content-protected, reusing the production entrance timing, processing glow, and real waveform data. Assistant and translation recordings never activate it, and plain dictation never animates the panel footer. I also traced the panel-closed path specifically to confirm no regression for everyone not using Agent mode: the only deltas are one extra lifecycle IPC message that the main process treats identically to idle, and a timer that early-returns before doing any work while the panel is closed.

Two things I found and fixed on the branch
I've pushed three commits rather than leaving them as review comments, since one was a merge conflict you'd have had to resolve anyway. Please sanity-check them — happy to back any of it out if you disagree with the approach.

9b7ccfd1 — the handback gap was still open in the common case. The close-intent ownership fix closed it only when the footer is in a pill-visible phase. With a settled response on screen the footer sits in the actions phase, where assistantActionsSuppressPill holds the main pill at opacity 0 until the content fade completes — while the companion has already hidden at close intent. Pressing Esc mid-recording left a hot mic with no visible surface anywhere for roughly 200–260ms. I moved the decision into a pure helper (shouldSuppressPillForAssistantActions) next to resolveCompanionPillInteractive, matching how the interactivity fix was structured, so the rule is now "final actions suppress the pill, except for activity that was just handed back at close intent." Also hoisted voicePillIsProcessing, which was already being computed inline for voiceActivity.

I checked this against the errorDownplay path, where the companion deliberately stays alive: no new duplicate-pill window, because dictationErrorSuppressesPill is a separate OR term that still suppresses independently.

c45b4662 — a hidden companion still counted as available. _isAgentDictationPillAvailable checked created + not destroyed + ready, but not visible. Since showAgentDictationPill early-returns during onboarding and nothing re-shows it on exit, a panel left open across an onboarding transition could pass the fail-closed gate and start a recording with nothing on screen — the exact class this round was meant to eliminate. Adding isVisible() closes it structurally rather than per-path, and the existing blocked-press re-kick makes it self-heal on the next press. The other half: _hideNormalAppSurfaces hides every sibling popup surface but not the companion, so re-entering onboarding left it floating over the flow. It's in that list now.

One note on the second commit: adding isVisible() broke two fakes in windowManagerMeetingNotification.test.js that model a live companion. I taught the fakes isVisible: () => true rather than adding a defensive ?. — the window is always a real BrowserWindow in production, and there's already one moveTop?.() in there that exists only to satisfy test fakes.

63cc0d1d — merged main to clear the conflict. It was just the dual import in useAudioRecording.js: your resolveLifecycleInputKind against main's createAssistantResponseDelivery from #1859. Both are used, so both stay. I checked how the companion composes with what's landed on main since your base — [#1859](https://github.com/OpenWhispr/openwhispr/issues/1859)'s auto-paste resolves the user's real app rather than the panel's follow-up input on macOS, and #1710's non-blocking start only shortens the preparing window. [#1826](https://github.com/OpenWhispr/openwhispr/issues/1826) and [#1863](https://github.com/OpenWhispr/openwhispr/issues/1863) turned out to be ancestors of your base, so the companion inherits the current pill polish automatically.

Verification
3075 passed / 0 failed / 185 skipped, with typecheck, format:check, and i18n:check all clean on the merged tree.

PR #1872, opened 2026-08-27
fix(calendar): handle nullish and missing options safely in CalendarSyncInterval

### LINKED ISSUE #1871: CalendarSyncInterval constructor throws TypeError when called without options and _getInterval fails on non-numeric inputs
### Describe the bug

In `src/helpers/calendarSyncInterval.js`:

`constructor(syncFn, { intervalMs, maxIntervalMs, logScope })` directly destructures its second argument without defaulting it. When called as `new CalendarSyncInterval(syncFn)` or `new CalendarSyncInterval(syncFn, null)`, it throws:

`TypeError: Cannot destructure property 'intervalMs' of 'undefined' as it is undefined.`

Furthermore, when `intervalMs` or `maxIntervalMs` are omitted or non-numeric, `_getInterval()` computes `NaN`, resulting in invalid timer scheduling.

### To Reproduce

1. Call `new CalendarSyncInterval(() => Promise.resolve())` -> throws `TypeError`.
2. Call `new CalendarSyncInterval(() => Promise.resolve(), null)` -> throws `TypeError`.

### Expected Behavior

`CalendarSyncInterval` should safely default options and ensure `_getInterval()` produces valid positive finite intervals.

hsusul:
Fixes #1871

### Problem
In `src/helpers/calendarSyncInterval.js`:
`constructor(syncFn, { intervalMs, maxIntervalMs, logScope })` threw `TypeError: Cannot destructure property 'intervalMs' of 'undefined'` when called without an options object or with `null`.

### Solution
- Defaulted constructor options safely with `options && typeof options === "object" ? options : {}`.
- Validated `intervalMs` and `maxIntervalMs` to always be positive finite integers with safe defaults.
- Added unit tests in `test/helpers/calendarSyncInterval.test.js`.

### Verification
- `node --test test/helpers/calendarSyncInterval.test.js` (passes, 5/5 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/calendarSyncInterval.js
+++ b/src/helpers/calendarSyncInterval.js
@@ -2,15 +2,30 @@ const debugLogger = require("./debugLogger");
 
 const FOCUS_SYNC_THROTTLE_MS = 30 * 1000;
 
+const DEFAULT_INTERVAL_MS = 60 * 1000;
+const DEFAULT_MAX_INTERVAL_MS = 5 * 60 * 1000;
+
 // Interval runner for calendar REST providers: exponential backoff on
 // consecutive failures (intervalMs × 2 per failure, capped at maxIntervalMs)
 // and a 30s throttle on focus-triggered syncs.
 class CalendarSyncInterval {
-  constructor(syncFn, { intervalMs, maxIntervalMs, logScope }) {
-    this.syncFn = syncFn;
-    this.intervalMs = intervalMs;
-    this.maxIntervalMs = maxIntervalMs;
-    this.logScope = logScope;
+  constructor(syncFn, options = {}) {
+    const {
+      intervalMs = DEFAULT_INTERVAL_MS,
+      maxIntervalMs = DEFAULT_MAX_INTERVAL_MS,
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: CalendarSyncInterval has exactly 2 production call sites (googleCalendarManager.js:27, microsoftCalendarManager.js:62), both of which pass an inline arrow function and a hardcoded options object literal with positive finite integer literals (`{ intervalMs: 2 * 60 * 1000, maxIntervalMs: 30 * 60 * 1000, logScope: "gcal"/"mcal" }`) — the options argument can never be undefined, null, or non-numeric, and appleCalendarManager.js only imports the FOCUS_SYNC_THROTTLE_MS constant. Per our convention (validate once at real trust boundaries, trust internal callers past that point), these guards can never fire, and coercing a bad syncFn to a no-op would hide a future caller bug behind a runner that silently never syncs — we'd rather that fail loudly. Closing #1871 along with this.

PR #1874, opened 2026-08-27
fix(sync): support Date and numeric timestamps and guard null notes in cloudSyncGuards

### LINKED ISSUE #1873: normalizeTimestamp throws TypeError on Date/number inputs and buildNoteUpdatePayload throws on nullish notes
### Describe the bug

In `src/helpers/cloudSyncGuards.js`:

1. `normalizeTimestamp(value)` directly calls `value.replace(" ", "T")...`. When passed numeric epoch timestamps, `Date` instances, or non-string values, it throws:
   `TypeError: value.replace is not a function`
2. `buildNoteUpdatePayload(note, cloudFolderId)` and `buildNoteCreatePayload(note, cloudFolderId)` access properties on `note` without null guarding, throwing:
   `TypeError: Cannot read properties of null (reading 'title')`

### To Reproduce

1. Call `normalizeTimestamp(new Date())` -> throws `TypeError: value.replace is not a function`.
2. Call `buildNoteUpdatePayload(null)` -> throws `TypeError: Cannot read properties of null`.

### Expected Behavior

- `normalizeTimestamp` should support string ISO/SQLite formats, numeric timestamps, and Date instances, degrading invalid non-strings to `""`.
- `buildNoteUpdatePayload` should guard against nullish and non-object note parameters.

hsusul:
Fixes #1873

### Problem
In `src/helpers/cloudSyncGuards.js`:
1. `normalizeTimestamp` threw `TypeError: value.replace is not a function` when called with `Date` instances, epoch millisecond numbers, or booleans.
2. `buildNoteUpdatePayload` and `buildNoteCreatePayload` threw `TypeError: Cannot read properties of null (reading 'title')` when passed nullish or non-object `note` inputs.

### Solution
- Added support for `Date` instances and numeric timestamps in `normalizeTimestamp`, safely degrading other non-strings to `""`.
- Added null safety for `note` in `buildNoteUpdatePayload`.
- Added unit tests in `test/helpers/cloudSyncGuards.test.js`.

### Verification
- `node --test test/helpers/cloudSyncGuards.test.js` (passes, 14/14 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/cloudSyncGuards.js
+++ b/src/helpers/cloudSyncGuards.js
@@ -7,6 +7,13 @@
 // sub-second cloud value at the same instant ('Z' > '.').
 export function normalizeTimestamp(value) {
   if (!value) return "";
+  if (value instanceof Date && !Number.isNaN(value.getTime())) {
+    return value.toISOString();
+  }
+  if (typeof value === "number" && Number.isFinite(value)) {
+    return new Date(value).toISOString();
+  }
+  if (typeof value !== "string") return "";
   const iso = value.replace(" ", "T").replace(/Z$/, "");
   return (/\.\d+$/.test(iso) ? iso : `${iso}.000`) + "Z";
 }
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: cloudSyncGuards has exactly one production consumer, SyncService.ts. Every value reaching normalizeTimestamp — directly (SyncService.ts:2380, 2405, 2406, 2573, 2586, 2606) or via isCloudEntryNewer (SyncService.ts:1568, 1637, 1937, 1986, 2138) — is a string or nullish by construction: cloud rows are JSON API responses typed `updated_at: string`, local rows are SQLite DATETIME (TEXT) columns written by `datetime('now')`, and the `since` cursors come from localStorage; nullish values were already handled by the existing `if (!value) return ""`. Likewise buildNoteUpdatePayload/buildNoteCreatePayload only ever receive non-null DB note rows: pushNote null-checks its fetch (SyncService.ts:746-747) and the batch paths iterate rows returned by getPendingNotes (SyncService.ts:1707-1736). A Date, number, boolean, or null note can never flow in. Per our convention (validate once at real trust boundaries, trust internal callers past that point), these guards would also mask a future caller bug by coercing to "" / {} — silently breaking last-writer-wins ordering instead of failing loudly. Closing #1873 along with this.

PR #1878, opened 2026-08-27
fix(linux): treat COSMIC Terminal as a terminal for paste

### LINKED ISSUE #1876: Linux auto-paste uses Ctrl+V in COSMIC Terminal
## Describe the bug

Linux auto-paste sends Ctrl+Shift+V in terminal emulators and Ctrl+V elsewhere. The window-class list in `clipboard.js` (kept in sync with `resources/linux-fast-paste.c`) does not include COSMIC Terminal.

COSMIC Terminal’s app id is `com.system76.CosmicTerm`; the process is `cosmic-term`. Neither contains the existing `"terminal"` substring (`CosmicTerm` has no `terminal` spelling). Paste uses Ctrl+V, which terminals typically bind to a control character instead of clipboard paste.

## Affected component

- `src/helpers/clipboard.js` — `LINUX_TERMINAL_CLASSES` / `isLinuxTerminalWindowClass`
- `resources/linux-fast-paste.c` — `terminal_classes` (must stay in sync)
- Tests: `test/helpers/linuxTerminalClasses.test.js`

## To Reproduce

On current `main`:

```js
clipboard.isLinuxTerminalWindowClass("org.gnome.Ptyxis")     // true (after #1659)
clipboard.isLinuxTerminalWindowClass("com.system76.CosmicTerm") // false
clipboard.isLinuxTerminalWindowClass("cosmic-term")          // false
```

## Expected behavior

COSMIC Terminal is treated as a terminal so auto-paste uses Ctrl+Shift+V.

## Actual behavior

Dictation pastes with Ctrl+V in COSMIC Terminal.

## Environment

- OpenWhispr `main`
- Linux COSMIC Desktop (Pop!_OS) with COSMIC Terminal focused

## User impact

Pasted transcriptions do not appear (or insert a literal ^V) when dictating into COSMIC Terminal.

## Root-cause hypothesis

Same gap as Ptyxis/GNOME Console (#1656 / #1659): a newer terminal’s app id was never added to the class list.

## Proposed focused scope

- Add `cosmicterm` / `cosmic-term` to both class lists
- Extend `linuxTerminalClasses.test.js`
- Do not change paste backends, Wayland portal order, or other app classes

I intend to work on this.

hsusul:
## Summary

- Linux auto-paste uses Ctrl+Shift+V in terminals and Ctrl+V elsewhere.
- COSMIC Terminal’s app id is `com.system76.CosmicTerm` (process `cosmic-term`). Neither matches the existing `"terminal"` substring (`CosmicTerm` ≠ `terminal`).
- Add those signatures to `clipboard.js` and `linux-fast-paste.c` (lists must stay in sync).

Fixes #1876

## Problem

Dictating into COSMIC Terminal pastes with Ctrl+V, which typically inserts a control character instead of clipboard text.

## Root cause

Same gap as Ptyxis/GNOME Console (#1659): a newer terminal’s window class was never added to `LINUX_TERMINAL_CLASSES`.

## Fix

- Add `cosmicterm` (matches `com.system76.CosmicTerm`) and `cosmic-term` (process name)

## Behavior before vs after

| Window class | Before | After |
| --- | --- | --- |
| `com.system76.CosmicTerm` | not a terminal | terminal (Ctrl+Shift+V) |
| `cosmic-term` | not a terminal | terminal |
| `org.gnome.Ptyxis` | terminal | unchanged |
| `firefox` | not a terminal | unchanged |

## Scope & non-goals

- Only the two class lists and the existing terminal-class test
- Does not change paste backends, portal order, or other app classes

## Tests

- `test/helpers/linuxTerminalClasses.test.js`

## Validation

- `ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/linuxTerminalClasses.test.js` → PASS (4 tests)
- `npx prettier --check src/helpers/clipboard.js test/helpers/linuxTerminalClasses.test.js` → PASS

### DIFF
--- a/resources/linux-fast-paste.c
+++ b/resources/linux-fast-paste.c
@@ -353,7 +353,7 @@ static const char *terminal_classes[] = {
     "terminator", "xterm", "urxvt", "rxvt", "tilix", "terminology",
     "wezterm", "foot", "st", "yakuake", "ghostty", "guake", "tilda",
     "hyper", "tabby", "sakura", "warp", "termius", "waveterm",
-    "ptyxis", "kgx", "org.gnome.console", NULL
+    "ptyxis", "kgx", "org.gnome.console", "cosmicterm", "cosmic-term", NULL
 };
 
 static int is_terminal(const char *wm_class) {
--- a/src/helpers/clipboard.js
+++ b/src/helpers/clipboard.js
@@ -60,6 +60,8 @@ const LINUX_TERMINAL_CLASSES = [
   "ptyxis",
   "kgx",
   "org.gnome.console",
+  "cosmicterm",
+  "cosmic-term",
 ];
 
 // macOS reports localized app names rather than window classes, and iTerm2 has
--- a/test/helpers/linuxTerminalClasses.test.js
+++ b/test/helpers/linuxTerminalClasses.test.js
@@ -38,6 +38,12 @@ test("Ptyxis and GNOME Console window classes are terminals", () => {
   assert.equal(clipboard.isLinuxTerminalWindowClass("kgx"), true);
 });
 
+test("COSMIC Terminal window classes are terminals", () => {
+  const clipboard = new ClipboardManager();
+  assert.equal(clipboard.isLinuxTerminalWindowClass("com.system76.CosmicTerm"), true);
+  assert.equal(clipboard.isLinuxTerminalWindowClass("cosmic-term"), true);
+});
+
 test("existing terminals still match", () => {
   const clipboard = new ClipboardManager();
   assert.equal(clipboard.isLinuxTerminalWindowClass("kitty"), true);


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: the repro in #1876 doesn't hold — on current main, `isLinuxTerminalWindowClass("com.system76.CosmicTerm")` already returns `true`, because matching is case-insensitive substring and the existing `"st"` entry (suckless terminal) matches the "st" in "sy**st**em76"; the C binary's `strcasestr` (resources/linux-fast-paste.c:359) behaves identically. Every production reporter that can identify COSMIC Terminal emits that app id — `xdotool getwindowclassname` (src/helpers/clipboard.js:1470, src/helpers/selectionManager.js:573), KWin resourceClass (selectionManager.js:589), `hyprctl activewindow` (selectionManager.js:603), and X11 XGetClassHint in the binary (linux-fast-paste.c:774) — cosmic-term's desktop file pins `StartupWMClass=com.system76.CosmicTerm`, so main already sends Ctrl+Shift+V on all of them. In the environment the issue names (COSMIC Wayland), the list isn't consulted at all: COSMIC is none of GNOME/KDE/Hyprland/wlroots (src/helpers/linuxSession.js), so `windowSignals` is empty and paste routes through Shift+Insert (clipboard.js:1562). The only string this PR newly matches is the bare comm name `cosmic-term`, which reaches a matcher solely through the /proc/<pid>/comm fallback for the Konsole X11 WM_CLASS-blindness quirk (clipboard.js:1497) and the AT-SPI pid path (selectionManager.js:499), which requires cosmic-term to expose an accessibility tree — neither scenario is claimed or evidenced here. Per our convention (validate once at real trust boundaries, trust internal callers past that point), we extend these lists only for identifier values a production reporter actually emits and fails to match — and every reporter here already matches. Closing #1876 along with this.

PR #1881, opened 2026-08-27
fix(dictionary): support Set and iterable snapshots in agentNameDictionaryChanges

### LINKED ISSUE #1880: agentNameDictionaryChanges ignores Set/iterable dictionaries and findStoredWord throws on non-string word
### Describe the bug

In `src/helpers/agentNameDictionary.js`:

1. `findStoredWord(words, word)` executes `word.toLowerCase()`. When `word` is not a string, it throws:
   `TypeError: word.toLowerCase is not a function`
2. `agentNameDictionaryChanges(dictionary, newName, oldName)` checks `Array.isArray(dictionary) ? dictionary : []`. When `dictionary` is passed as a `Set` or other iterable collection of words, it defaults to `[]` and suggests adding the word even if it exists in the set.

### To Reproduce

1. Call `agentNameDictionaryChanges(new Set(["OpenWhispr"]), "OpenWhispr")` -> returns `{ add: ["OpenWhispr"], remove: [] }` (expected `{ add: [], remove: [] }`).

### Expected Behavior

- `agentNameDictionaryChanges` should accept arrays, Sets, and iterables.
- `findStoredWord` should safely guard non-string lookups.

hsusul:
Fixes #1880

### Problem
In `src/helpers/agentNameDictionary.js`:
- `agentNameDictionaryChanges` strictly checked `Array.isArray(dictionary)`, ignoring non-array iterables (such as `Set`), which resulted in calculating redundant add deltas.
- `findStoredWord` called `word.toLowerCase()` unconditionally, throwing `TypeError` if `word` was not a string.

### Solution
- Supported Sets and generic iterables via `Array.from`.
- Added defensive type check for `word` in `findStoredWord`.
- Added unit tests in `test/helpers/agentNameDictionary.test.js`.

### Verification
- `node --test test/helpers/agentNameDictionary.test.js` (passes, 15/15 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/agentNameDictionary.js
+++ b/src/helpers/agentNameDictionary.js
@@ -1,5 +1,6 @@
 function findStoredWord(words, word) {
-  const needle = word.toLowerCase();
+  if (typeof word !== "string" || !Array.isArray(words)) return undefined;
+  const needle = word.trim().toLowerCase();
   return words.find((w) => typeof w === "string" && w.trim().toLowerCase() === needle);
 }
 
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: agentNameDictionaryChanges has exactly one production call site (src/utils/agentName.ts:13, via setAgentName and ensureAgentNameInDictionary), which always passes getSettings().customDictionary — store state that is a plain array on every assignment path: initialization goes through readStringArray (Array.isArray-guarded JSON from localStorage, settingsStore.ts:185), and every writer (setCustomDictionary settingsStore.ts:1784, updateCustomDictionary settingsStore.ts:1810/1826, applyCustomDictionaryFromExternal settingsStore.ts:1841, startup reconcile settingsStore.ts:3159) sets an array sourced from SQLite's getDictionary() `rows.map(...)` (database.js:1427) or an array literal. A Set never exists in that flow, and the existing `Array.isArray(dictionary) ? dictionary : []` already handles any non-array without throwing. findStoredWord is module-private and only ever called with trimmedNew/trimmedOld (agentNameDictionary.js:28-29), which the `typeof === "string"` guards two lines above guarantee are non-empty strings — the claimed TypeError cannot occur. Per our convention (validate once at real trust boundaries, trust internal callers past that point), these guards and the widened Iterable contract are dead code on an internal-only path, and we'd rather a future caller bug fail loudly than be coerced to undefined. Closing #1880 along with this.

PR #1883, opened 2026-08-27
fix(auth): validate tenant string type and normalize environment in cortiAuth

### LINKED ISSUE #1882: assertValidTarget in cortiAuth permits null/undefined tenant due to regex type coercion and getCortiToken throws on null credentials
### Describe the bug

In `src/helpers/cortiAuth.js`:

1. `assertValidTarget(environment, tenant)` tests tenant with `TENANT_PATTERN.test(tenant)` where `const TENANT_PATTERN = /^[a-zA-Z0-9_-]+$/`. Because `RegExp.prototype.test()` coerces arguments to strings, `TENANT_PATTERN.test(null)` coerces `null` to `"null"` (which matches the alphanumeric pattern and returns `true`), bypassing tenant validation when `tenant` is `null` or `undefined`.
2. Valid environment names with mixed case (e.g. `"EU"`, `"US"`) or leading/trailing whitespace are rejected because `CORTI_ENVIRONMENTS.has(environment)` is case-sensitive.
3. `getCortiToken(credentials, fetchImpl)` destructures `credentials` without defaulting, throwing `TypeError: Cannot destructure property 'environment' of 'null' as it is null` when passed `null`.

### To Reproduce

1. Call `getCortiToken({ environment: "eu", tenant: null, clientId: "id", clientSecret: "s" })` -> `assertValidTarget` allows `tenant: null` to pass without throwing `Invalid Corti tenant name`.
2. Call `getCortiToken(null)` -> throws `TypeError: Cannot destructure property 'environment' of 'null'`.

### Expected Behavior

- `assertValidTarget` should require `typeof tenant === "string"` and normalize `environment` (trimming and lowercasing).
- `getCortiToken` should guard against nullish `credentials`.

hsusul:
Fixes #1882

### Problem
In `src/helpers/cortiAuth.js`:
1. `assertValidTarget` evaluated `TENANT_PATTERN.test(tenant)` without checking `typeof tenant === "string"`. In JavaScript, `RegExp.prototype.test(null)` coerces `null` to `"null"` (and `undefined` to `"undefined"`), matching the alphanumeric tenant regex and permitting invalid/nullish tenant names.
2. `assertValidTarget` was sensitive to casing/whitespace for valid Corti environment targets (e.g. `"EU"`).
3. `getCortiToken` threw `TypeError: Cannot destructure property 'environment' of 'null'` when called with nullish credentials.

### Solution
- Checked that `typeof tenant === "string"` and `typeof environment === "string"`, and normalized environment strings with `.trim().toLowerCase()`.
- Guarded credentials parameter in `getCortiToken`.
- Added unit tests in `test/helpers/cortiAuth.test.js`.

### Verification
- `node --test test/helpers/cortiAuth.test.js` (passes, 5/5 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/cortiAuth.js
+++ b/src/helpers/cortiAuth.js
@@ -12,16 +12,21 @@ const TOKEN_FETCH_TIMEOUT_MS = 12_000;
 const tokenCache = new Map();
 
 function assertValidTarget(environment, tenant) {
-  if (!CORTI_ENVIRONMENTS.has(environment)) {
+  if (typeof environment !== "string" || !CORTI_ENVIRONMENTS.has(environment.trim().toLowerCase())) {
     throw new Error(`Invalid Corti environment: ${environment}`);
   }
-  if (!TENANT_PATTERN.test(tenant)) {
+  if (typeof tenant !== "string" || !TENANT_PATTERN.test(tenant.trim())) {
     throw new Error("Invalid Corti tenant name");
   }
 }
 
-async function getCortiToken({ environment, tenant, clientId, clientSecret }, fetchImpl) {
+async function getCortiToken(credentials = {}, fetchImpl) {
+  const { environment, tenant, clientId, clientSecret } =
+    credentials && typeof credentials === "object" ? credentials : {};
   assertValidTarget(environment, tenant);
+
+  const normalizedEnv = environment.trim().toLowerCase();
+  const normalizedTenant = tenant.trim();
 
   // The secret is part of the key (digested, never stored raw) so edited
   // credentials always re-authenticate; a key without it makes a connection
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: getCortiToken has exactly 3 production call sites (src/helpers/cortiTranscription.js:37 — reached from ipcHandlers.js:4212, 5897, and 9194 — plus ipcHandlers.js:1174 via _mintStoredCortiToken and ipcHandlers.js:1250 in test-provider-connection), and assertValidTarget's only caller is getCortiToken itself. Every path coerces tenant to a nonempty string first ((tenant || "").trim() || "base" in audioManager.js:404, transcriptionRoute.ts:273, ipcHandlers.js:1170/1253), and environment comes solely from the two-option lowercase select in TranscriptionModelPicker.tsx (values "us"/"eu", "us" fallback at every consumer), so a nullish tenant, mixed-case environment, or null credentials object never reaches this function — and even the synthetic null-coercion case only admits "null"/"undefined", which already sit inside the charset TENANT_PATTERN enforces, so nothing unsafe can reach the URL. Per our convention (validate once at real trust boundaries, trust internal callers past that point), the added guards defend unreachable states, and the trim/lowercase normalization actually loosens validation — main fails loudly on " EU ", which we prefer to silently coercing it into a working request. Closing #1882 along with this.

PR #1892, opened 2026-08-27
fix(chat): normalize provider casing and guard settings in resolveChatRoute

### LINKED ISSUE #1890: resolveChatRoute throws TypeError on nullish settings and misroutes mixed-case cloud providers to local
### Describe the bug

In `src/helpers/chatRouting.js`:

1. `resolveChatRoute({ provider, lanUrl, customApiKey, isEnterpriseProvider = false })` directly destructures its parameter without defaulting it, throwing `TypeError: Cannot destructure property 'provider' of 'undefined' as it is undefined` when called without arguments or with `null`.
2. `CLOUD_CHAT_PROVIDERS.has(provider)` performs a case-sensitive check on exact lowercase names. If `provider` has mixed casing or surrounding whitespace (e.g. `"OpenAI"`, `"Anthropic"`, `" groq "`), it returns `false` and erroneously routes chat requests to `kind: "local"`.
3. Non-string `lanUrl` or `customApiKey` inputs throw `TypeError` when calling `?.trim()`.

### To Reproduce

1. Call `resolveChatRoute({ provider: "OpenAI" })` -> returns `{ kind: "local", baseUrl: "", apiKey: "" }` (expected `kind: "provider"`).
2. Call `resolveChatRoute(null)` -> throws `TypeError: Cannot destructure property 'provider' of 'null'`.

### Expected Behavior

- `resolveChatRoute` should safely handle nullish settings.
- `provider` casing and whitespace should be normalized so recognized cloud providers route to `kind: "provider"`.

hsusul:
Fixes #1890

### Problem
In `src/helpers/chatRouting.js`:
- `resolveChatRoute` threw `TypeError: Cannot destructure property 'provider' of 'undefined'` when called without settings or with `null`.
- `CLOUD_CHAT_PROVIDERS.has(provider)` failed for valid provider names with mixed casing or leading/trailing whitespace (e.g. `"OpenAI"`, `" Anthropic "`), causing valid cloud providers to route as `kind: "local"`.
- Non-string `lanUrl` and `customApiKey` threw `TypeError` when evaluating `?.trim()`.

### Solution
- Safely defaulted `settings` in `resolveChatRoute`.
- Normalized `provider` with `typeof provider === "string" ? provider.trim().toLowerCase() : ""`.
- Added string type checks for `lanUrl` and `customApiKey`.
- Added unit tests in `test/helpers/chatRouting.test.js`.

### Verification
- `node --test test/helpers/chatRouting.test.js` (passes, 8/8 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/chatRouting.js
+++ b/src/helpers/chatRouting.js
@@ -11,15 +11,18 @@ const CLOUD_CHAT_PROVIDERS = new Set([
 
 // Resolve Chat from Chat-owned settings only. In particular, this must never
 // consult Dictation Cleanup's mode or endpoint.
-export function resolveChatRoute({ provider, lanUrl, customApiKey, isEnterpriseProvider = false }) {
+export function resolveChatRoute(settings = {}) {
+  const { provider, lanUrl, customApiKey, isEnterpriseProvider = false } =
+    settings && typeof settings === "object" ? settings : {};
+
   // An explicit self-hosted URL is the caller's declared route — it wins even
   // over a stale enterprise provider id left in settings.
-  const baseUrl = lanUrl?.trim() || "";
+  const baseUrl = typeof lanUrl === "string" ? lanUrl.trim() : "";
   if (baseUrl) {
     return {
       kind: "self-hosted",
       baseUrl,
-      apiKey: customApiKey?.trim() || "",
+      apiKey: typeof customApiKey === "string" ? customApiKey.trim() : "",
     };
   }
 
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: resolveChatRoute has exactly 2 production call sites (src/services/ReasoningService.ts:545 and :745), both of which pass an inline object literal — so nullish settings can never reach the destructure. The provider/lanUrl/customApiKey fields at both sites come from selectResolvedLLMConfig(settings, "chatIntelligence") via the single caller useChatStreaming.ts:424-438: provider is always a canonical lowercase registry id or "" (settingsStore.ts:2584 guarantees a string; every writer of chatAgentProvider uses modelRegistryData.json ids, and the managed-enterprise path validates provider to exactly "bedrock"/"azure" at its real trust boundary in enterpriseManagedConfig.mjs before it's ever stored), lanUrl is a non-empty string or undefined (gated by !!chatConfig.remoteUrl at useChatStreaming.ts:263), and customApiKey is a string or undefined — all already safe with the existing ?.trim(). Per our convention (validate once at real trust boundaries, trust internal callers past that point), these guards only mask a future caller bug by coercing it to a silent kind:"local" route — and the casing normalization would be inconsistent anyway, since every downstream provider lookup (useChatStreaming.ts:265-276, ReasoningService.ts:576-581) stays case-sensitive. Note the quality CI job also fails prettier on chatRouting.js. Closing #1890 along with this.

PR #1894, opened 2026-08-27
fix(calendar): guard nullish events and safe reset iteration in CalendarReminderScheduler

### LINKED ISSUE #1893: notificationKey and isReminderEligible in calendarReminderScheduler throw TypeError on nullish events
### Describe the bug

In `src/helpers/calendarReminderScheduler.js`:

1. `notificationKey(event)` directly accesses `event.provider` and `event.id`. When `event` is `null` or non-object, it throws:
   `TypeError: Cannot read properties of null (reading 'provider')`
2. `isReminderEligible(event)` directly accesses `event.attendees_count`. When `event` is `null` or non-object, it throws:
   `TypeError: Cannot read properties of null (reading 'attendees_count')`
3. `reset(provider)` iterates directly over `this.notifiedMeetings` with a `for...of` loop while mutating the Set with `this.notifiedMeetings.delete(key)`.

### To Reproduce

1. Call `notificationKey(null)` -> throws `TypeError: Cannot read properties of null (reading 'provider')`.
2. Call `isReminderEligible(null)` -> throws `TypeError: Cannot read properties of null (reading 'attendees_count')`.

### Expected Behavior

- `notificationKey` and `isReminderEligible` should safely guard against nullish and non-object events.
- `reset(provider)` should safely iterate over a snapshot of the Set keys.

hsusul:
Fixes #1893

### Problem
In `src/helpers/calendarReminderScheduler.js`:
1. `notificationKey(event)` and `isReminderEligible(event)` directly accessed properties on `event` without guarding against null or non-object values, throwing `TypeError: Cannot read properties of null`.
2. `reset(provider)` mutated `this.notifiedMeetings` with `.delete(key)` while iterating over it via `for...of`.

### Solution
- Added defensive guards for nullish and malformed events in `notificationKey` and `isReminderEligible`.
- Guarded database manager method return values with `Array.isArray`.
- Created an array snapshot with `Array.from(this.notifiedMeetings)` during deletion in `reset()`.
- Added unit tests in `test/helpers/calendarReminderScheduler.test.js`.

### Verification
- `node --test test/helpers/calendarReminderScheduler.test.js` (passes, 14/14 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/calendarReminderScheduler.js
+++ b/src/helpers/calendarReminderScheduler.js
@@ -4,13 +4,18 @@ const { getMeetingJoinUrl } = require("./meetingJoinUrl");
 const MEETING_REMINDER_LEAD_MS = 60 * 1000;
 
 function notificationKey(event) {
-  return `${event.provider || "google"}:${event.id}`;
+  if (!event || typeof event !== "object") return ":";
+  return `${event.provider || "google"}:${event.id || ""}`;
 }
 
 // A solo event with no invitees and no meeting link is a time block (focus
 // time, a personal reminder), not a meeting — every provider syncs those.
 function isReminderEligible(event) {
-  return event.attendees_count > 0 || getMeetingJoinUrl(event);
+  if (!event || typeof event !== "object") return false;
+  return (
+    (typeof event.attendees_count === "number" && event.attendees_count > 0) ||
+    Boolean(getMeetingJoinUrl(event))
+  );
 }
 
 // Provider-agnostic meeting reminder scheduling. Reads only the shared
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: `notificationKey` and `isReminderEligible` are module-private (only the `CalendarReminderScheduler` class is exported), and every event they receive is a `calendar_events` row from `DatabaseManager.getActiveEvents()`/`getUpcomingEvents()` (database.js:3818, 3870) — `.all().map(stripDedupeColumn)` always yields an array of plain row objects with `id TEXT PRIMARY KEY`, `provider TEXT NOT NULL DEFAULT 'google'`, and numeric `attendees_count`, never a nullish element or a non-array. The scheduler's single production instantiation (main.js:480) receives the real DatabaseManager, so the `Array.isArray(this.databaseManager?.…)` guards are dead code — and as written they run each SQLite query twice. The `reset()` change addresses a non-issue: deleting entries from a Set during `for...of` iteration is well-defined in JavaScript, and `reset` is only ever called with the hardcoded literals "google"/"microsoft"/"apple" (googleCalendarManager.js:66,101,326; microsoftCalendarManager.js:101,128,363; appleCalendarManager.js:348). Per our convention (validate once at real trust boundaries, trust internal callers past that point), these guards would only convert a future caller bug into silent sentinel keys (":" ) and silently dropped reminders instead of a loud TypeError. Closing #1893 along with this.

PR #1896, opened 2026-08-27
fix(notes): preserve transcript scroll during resize

boseq:
## Summary

- stop transcript and floating-chat resize observers from overriding explicit upward scroll intent
- share the follow-bottom state machine across live transcript/chat scrollers
- keep automatic live-follow enabled while the reader is at the bottom, and restore it only after returning to the true bottom
- add regression coverage for the resize-plus-upward-scroll race

## Root cause

The virtualized meeting transcript remeasures row heights while the Notes window or pane is resized. Each `totalSize` change reran a layout effect that wrote `scrollTop = scrollHeight` while the reader was still inside the near-bottom threshold. Repeated measurements could therefore reset small upward scroll deltas before they escaped the threshold, leaving the transcript apparently pinned.

The optional floating chat layout had a second resize-driven bottom correction with the same race.

## Test plan

- `node --import tsx --test test/helpers/scrollFollowState.test.js test/components/floatingChatLayout.test.js test/components/meetingTranscriptChat.test.js` — 11 passed
- focused renderer/test ESLint — 0 errors
- Prettier check for all changed files — passed
- `npm test` — 3,077 passed, 0 failed, 185 skipped, 1 existing todo

## Manual QA

1. Open a note with a long live transcript.
2. Resize the Notes/Transcript area continuously.
3. Scroll upward and stop in the middle while transcript updates continue.
4. Confirm resize measurements and new partials do not move the reader.
5. Scroll to the true bottom and confirm automatic live-follow resumes.

### DIFF
--- a/src/components/chat/ChatMessages.tsx
+++ b/src/components/chat/ChatMessages.tsx
@@ -12,12 +12,23 @@ interface ChatMessagesProps {
 export function ChatMessages({ messages, emptyState, onOpenNote }: ChatMessagesProps) {
   // Follow the stream only while the user is at the bottom; scrolling up to
   // re-read must not be yanked back down by the next token.
-  const { scrollRef, handleScroll } = useStickToBottom<HTMLDivElement>(messages);
+  const {
+    scrollRef,
+    handleScroll,
+    handleWheel,
+    handleTouchStart,
+    handleTouchMove,
+    handleTouchEnd,
+  } = useStickToBottom<HTMLDivElement>(messages);
 
   return (
     <div
       ref={scrollRef}
       onScroll={handleScroll}
+      onWheel={handleWheel}
+      onTouchStart={handleTouchStart}
+      onTouchMove={handleTouchMove}
+      onTouchEnd={handleTouchEnd}
       className={cn("flex-1 overflow-y-auto agent-chat-scroll", "px-3 py-2")}
     >
       {messages.length === 0 ? (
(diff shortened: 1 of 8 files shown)


Chadpiha (approved):
Approving — thanks for the clear root-cause writeup; it matched what I found when tracing this end-to-end.

What I verified:

The core mechanism is sound: upward wheel/touch intent detaches follow before any resize pin can land, the rejoin state machine behaves correctly against a moving bottom and fractional HiDPI metrics, and the new floatingChatLayout test genuinely fails without the fix.
The MeetingTranscriptChat restructure (wrapper div + ResizeObserver instead of the totalSize layout effect) is inert to the virtualizer, and the partial-only mount race from your second commit is fixed with no observer leaks.
All three useStickToBottom consumers attach the full handler set; ChatMessages streaming and the dictation LiveTranscriptPanel entrance/reset choreography still behave correctly; dead code is fully removed.
Full suite green, typecheck/lint/prettier clean, and zero textual or semantic conflicts with current main.

I pushed two small commits onto the branch:

stopFollowing in floatingChatLayout.ts now requires the gesture target to be inside the active scroller (same containment check the hook's isReaderGesture has). Without it, wheeling up over the recording header pill or consent strip — inside the content root but outside the scroller — detached follow with nothing ever scrolling, so readerMovedAway never latched and follow stayed wedged off. Regression test included (verified red before the guard).
A test pinning the readerMovedAway rejoin gate: a controller that rejoins whenever it's at the bottom passed every existing test, leaving the gate — the piece that prevents a pre-detach pin echo from re-following — unguarded.

PR #1901, opened 2026-08-27
fix(auth): guard nullish parameters and non-string tokens in accountScopeBinding

### LINKED ISSUE #1900: accountScopeBinding helpers throw TypeError on nullish parameters and non-string tokens
### Describe the bug

In `src/helpers/accountScopeBinding.js`:

1. `hashToken(token)` directly calls `crypto.createHash("sha256").update(token, "utf8")`. When `token` is `null`, `undefined`, or a non-string type, Node's crypto library throws:
   `TypeError [ERR_INVALID_ARG_TYPE]: The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received null`
2. `evaluateScopeRequest` and `resolveBootAccountScope` destructure their object parameter without defaulting, throwing `TypeError: Cannot destructure property of 'undefined'` or `'null'` when called without arguments or with `null`.

### To Reproduce

1. Call `binding.resolveBootAccountScope(null)` -> throws `TypeError`.
2. Call `binding.evaluateScopeRequest(null)` -> throws `TypeError`.

### Expected Behavior

- `hashToken` should return `""` for non-string tokens.
- `evaluateScopeRequest` and `resolveBootAccountScope` should safely handle missing or nullish options.

hsusul:
Fixes #1900

### Problem
In `src/helpers/accountScopeBinding.js`:
- `hashToken(token)` called `crypto.createHash().update(token)` without checking `typeof token === "string"`, throwing `TypeError [ERR_INVALID_ARG_TYPE]` on nullish or invalid values.
- `evaluateScopeRequest` and `resolveBootAccountScope` threw `TypeError` when called with `null` or without arguments.

### Solution
- Checked that `token` is a non-empty string in `hashToken`.
- Safely defaulted object destructuring in `evaluateScopeRequest` and `resolveBootAccountScope`.
- Added unit tests in `test/helpers/accountScopeBinding.test.js`.

### Verification
- `node --test test/helpers/accountScopeBinding.test.js` (passes, 5/5 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/accountScopeBinding.js
+++ b/src/helpers/accountScopeBinding.js
@@ -15,6 +15,7 @@ const BINDING_VERSION = 1;
 const bindingFile = () => path.join(app.getPath("userData"), "account-scope-binding.json");
 
 function hashToken(token) {
+  if (typeof token !== "string" || token.length === 0) return "";
   return crypto.createHash("sha256").update(token, "utf8").digest("hex");
 }
 
@@ -22,7 +23,9 @@ function hashToken(token) {
 // arriving while a token exists is a "validated signed-out" transition and
 // must carry the current generation, so a stray early null call cannot
 // discard a valid binding.
-function evaluateScopeRequest({ accountId, expectedGeneration, token, generation }) {
+function evaluateScopeRequest(request = {}) {
+  const { accountId, expectedGeneration, token, generation } =
+    request && typeof request === "object" ? request : {};
   if (accountId !== null && (typeof accountId !== "string" || accountId.trim().length === 0)) {
     return { ok: false, code: "INVALID_ACCOUNT" };
   }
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: evaluateScopeRequest has exactly one production call site (src/helpers/ipcHandlers.js:1965) and resolveBootAccountScope exactly one (main.js:439), both invoked with inline object literals — the argument can never be null/undefined, and the renderer-controlled fields crossing the IPC boundary (accountId, expectedGeneration) are already validated by the existing function bodies. hashToken has exactly two call sites, both internal to the module: accountScopeBinding.js:48 runs only after the pre-existing non-empty-string token check, and accountScopeBinding.js:68 (persist) is reached only via ipcHandlers.js:1982 after evaluateScopeRequest has guaranteed a truthy token, which tokenStore only ever yields as string|null — so a nullish or non-string token never occurs. Per our convention (validate once at real trust boundaries, trust internal callers past that point), these guards protect paths that don't exist, and the "" sentinel in hashToken would silently persist an unmatchable binding instead of failing loudly if a caller bug ever appeared. Closing #1900 along with this.

PR #1903, opened 2026-08-27
fix(dev): support HTTPS and sanitize options in DevServerManager

### LINKED ISSUE #1902: DevServerManager.waitForDevServer fails on HTTPS dev server URLs and lacks unit test coverage
### Describe the bug

In `src/helpers/devServerManager.js`:

`DevServerManager.waitForDevServer` unconditionally requires `http` and hardcodes port fallback to `80`:
```javascript
const http = require("http");
const urlObj = new URL(url);

const result = await new Promise((resolve) => {
  const req = http.get(
    {
      hostname: urlObj.hostname,
      port: urlObj.port || 80,
      path: urlObj.pathname,
      timeout: 2000,
    },
    ...
```

When `url` uses the `https:` protocol (or Vite is configured with an HTTPS dev certificate), `waitForDevServer` connects over cleartext HTTP to port 80 instead of TLS on port 443, resulting in connection failure. Additionally, `maxAttempts` and `delay` are not validated against non-positive numbers.

### To Reproduce

1. Call `DevServerManager.waitForDevServer("https://localhost:5183/")` -> attempts cleartext `http.get` on port 80.

### Expected Behavior

- `waitForDevServer` should select the appropriate client module (`https` or `http`) and default the port appropriately (443 for HTTPS, 80 for HTTP).
- `maxAttempts` and `delay` should be sanitized.

hsusul:
Fixes #1902

### Problem
In `src/helpers/devServerManager.js`:
- `DevServerManager.waitForDevServer` unconditionally used cleartext `http.get` and defaulted the port to `80`, failing when checking HTTPS URLs (or dev servers configured with HTTPS certificates).
- Handled invalid/negative `maxAttempts` and `delay` values without sanitization.

### Solution
- Selected the appropriate client module (`https` or `http`) based on the URL protocol and defaulted the port accordingly (443 for HTTPS, 80 for HTTP).
- Sanitized `maxAttempts` and `delay` parameters.
- Exported `parseDevServerPort` for testability.
- Added comprehensive unit tests in `test/helpers/devServerManager.test.js`.

### Verification
- `node --test test/helpers/devServerManager.test.js` (passes, 4/4 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/devServerManager.js
+++ b/src/helpers/devServerManager.js
@@ -18,17 +18,22 @@ const DEV_SERVER_URL = `http://localhost:${DEV_SERVER_PORT}/`;
 
 class DevServerManager {
   static async waitForDevServer(url = DEV_SERVER_URL, maxAttempts = 30, delay = 1000) {
-    for (let i = 0; i < maxAttempts; i++) {
+    const attempts = Number.isInteger(maxAttempts) && maxAttempts > 0 ? maxAttempts : 30;
+    const intervalDelay = Number.isInteger(delay) && delay >= 0 ? delay : 1000;
+
+    for (let i = 0; i < attempts; i++) {
       try {
-        const http = require("http");
         const urlObj = new URL(url);
+        const isHttps = urlObj.protocol === "https:";
+        const client = isHttps ? require("https") : require("http");
+        const defaultPort = isHttps ? 443 : 80;
 
         const result = await new Promise((resolve) => {
-          const req = http.get(
+          const req = client.get(
             {
               hostname: urlObj.hostname,
-              port: urlObj.port || 80,
-              path: urlObj.pathname,
+              port: urlObj.port || defaultPort,
+              path: urlObj.pathname + urlObj.search,
               timeout: 2000,
             },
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: waitForDevServer has exactly 5 production call sites (src/helpers/windowManager.js:149, 514, 1810, 2009, 2169), all of which call it with zero arguments, so the URL is always the hardcoded `http://localhost:${DEV_SERVER_PORT}/` constant (explicit validated port, http-only) and maxAttempts/delay are always the integer defaults 30/1000 — the https branch, port-80 fallback, and sanitization guards can never execute. Our Vite dev server config (src/vite.config.mjs) has no `https` option, and the PR doesn't make DEV_SERVER_URL https-capable either, so the new code path is unreachable even after the fix. Per our convention (validate once at real trust boundaries, trust internal callers past that point), we prefer a future caller bug to fail loudly rather than be coerced to sentinel defaults. Closing #1902 along with this.

PR #1918, opened 2026-08-28
fix(linux): restore meeting prompts for later calls on Linux

### LINKED ISSUE #1917: Meeting prompts stop appearing after dismissing the first notification on Linux
**Describe the bug**
On Linux, particularly Hyprland, meeting prompts may stop appearing after the first notification is dismissed or externally closed.  

After ending the first meeting, joining a later meeting may not trigger another prompt unless OpenWhispr is restarted.

**To Reproduce**

1. Join a meeting and wait for the meeting notification.
2. Dismiss it or close it using Hyprland.
3. End the meeting.
4. Later, join another meeting.
5. No notification appears.

**Expected behavior**
Later meetings should receive their own prompt. Closing a notification through Hyprland should not block future prompts, and OpenWhispr’s own microphone activity should not keep the first meeting marked as active.

- OS: Linux
- Version: v1.9.1

IdrisGit:
### Summary
Fixes meeting notifications becoming stuck after the first detected call on Linux.  

Users will now receive prompts for later meetings after the previous meeting ends. Closing a notification through Hyprland also releases it correctly instead of blocking future prompts.

### Behavior
- Ignores OpenWhispr’s own microphone activity when deciding whether a meeting is still active.
- Allows a later calling application to trigger a new prompt.
- Avoids repeated prompts during the same ongoing call.
- Preserves the five-minute cooldown when the user explicitly clicks Dismiss.
- Preserves the 30-second automatic expiry without applying a cooldown.

fixes #1917

### DIFF
--- a/src/helpers/audioActivityDetector.js
+++ b/src/helpers/audioActivityDetector.js
@@ -52,6 +52,35 @@ class AudioActivityDetector extends EventEmitter {
     this._externalMicActive = false;
     this._lastEmittedExternalMicReliable = false;
     this._lastEmittedExternalMicActive = false;
+    this._externalCapturePids = new Set();
+    this._promptedCapturePids = new Set();
+  }
+
+  _markPrompted() {
+    this.hasPrompted = true;
+    this._promptedCapturePids = new Set(this._externalCapturePids);
+  }
+
+  _rearmPromptForSourceChange() {
+    if (!this.hasPrompted || !this._pidScopedCapability || !this._externalCapturePids.size) return;
+    if (!this._promptedCapturePids.size) {
+      this._promptedCapturePids = new Set(this._externalCapturePids);
+      return;
+    }
+
+    const previousSourceGone = [...this._promptedCapturePids].every(
+      (pid) => !this._externalCapturePids.has(pid)
+    );
+    if (!previousSourceGone) return;
+
+    this.hasPrompted = false;
+    this._promptedCapturePids.clear();
+    debugLogger.info("Re-armed meeting prompt for a changed capture source", {}, "meeting");
+  }
+
+  _isExternalLinuxSourceOutput(sourceOutput, excludedProcessIds) {
+    const processId = Number(sourceOutput?.properties?.["application.process.id"]);
+    return Number.isInteger(processId) && processId > 0 && !excludedProcessIds.has(processId);
   }
 
   getExternalMicState() {
@@ -151,6 +180,7 @@ class AudioActivityDetector extends EventEmitter {
 
   resetPrompt() {
     this.hasPrompted = false;
+    this._promptedCapturePids = new Set();
     this._clearSustainedTimer();
     this.audioActiveStart = null;
     debugLogger.info("Audio detection prompt reset (no cooldown)", {}, "meeting");
@@ -160,6 +190,7 @@ class AudioActivityDetector extends EventEmitter {
     this.consecutiveChecks = 0;
     this.audioActiveStart = null;
     this.hasPrompted = false;
+    this._promptedCapturePids = new Set();
     this._clearResetTimer();
   }
 
@@ -172,6 +203,7 @@ class AudioActivityDetector extends EventEmitter {
     this._activeMicPids.clear();
     this._activeSources = 0;
     this._lastKnownMicState = false;
+    this._externalCapturePids = new Set();
     this._clearCooldownReevalTimer();
     this._linuxOwnershipRequest++;
     this._linuxReconcileQueued = false;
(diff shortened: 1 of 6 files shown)


gabrielste1n (approved):
Verified end to end in an isolated worktree, tracing every consumer of the changed signals against the merge base. The core fix is right: leaving `audio:sustained-audio` in `activeDetections` after a card vanished without a response suppressed every future prompt for the rest of the app session, and releasing just that one detection (rather than clearing all of them, which would strand `_notificationQueue`) is the correct shape. The re-entrancy is also correctly guarded — `handleNotificationResponse` clears `activeDetections` before it dismisses, and the replacement branch in `showMeetingNotification` nulls `_pendingNotificationData` before `close()`, so neither path can double-report.

I pushed one commit fixing two things I found that reached beyond the intended improvement.

**1. The polling fallback claimed mic-ownership reliability from a snapshot it stops refreshing.** `_checkLinux` began calling `_setPidScopedCapability(true)`, but `_check()` short-circuits on `_userRecording` for the whole of a recording, so nothing refreshes `_activeMicPids` once a meeting starts. `_activateAutoEnd` passes that `reliable` straight into `beginSession`, where `ownership = eligible && reliable && externalMicActive`. Measured, Linux polling mode with an app holding the mic when recording starts:

| | after first poll | during recording |
|---|---|---|
| merge base | `reliable: false` | unchanged → fallback mode, 60s-silence auto-end works |
| this PR | `reliable: true, externalMicActive: true` | **frozen** → ownership mode |

In ownership mode `evaluate()` returns early while `externalMicActive` is true, only `handleExternalMicState` can leave that mode, and it never fires because polling is gated. `handleMeetingProcessExit` also early-returns unless the mode is `fallback`, so not even the meeting app exiting stops the recording. Auto-end was silently dead for the session.

It is reachable on any machine with a working `pactl`: `fallbackToPolling` deliberately surrenders reliability first — "Announce the reliability loss before the snapshot is reset" — and then `_startPolling()` → `_check()` → `_checkLinux()` set it straight back. Any `pactl subscribe` exit (a PulseAudio/PipeWire restart) lands there.

Reliability is now scoped to a live listener, which is the only thing that pushes every transition into the snapshot. Polling keeps its PID data for the re-arm, so your fix still works in that mode.

**2. The re-arm had no dwell time, so a mid-call capture swap re-prompted.** `_rearmPromptForSourceChange` cleared `hasPrompted` the moment the prompted PID set went disjoint, with no requirement that the mic ever went idle. Reconciles are throttled to 1s, so a teardown-and-recreate inside one window is observed atomically as "old gone, new present" — I measured two prompts for one continuous call. That contradicts the PR's own "avoids repeated prompts during the same ongoing call", and it's the hazard `OWNERSHIP_CONFIRM_MS` already exists for ("a device flap emits MIC_STOP and MIC_START in one reconcile pass, an app rebuilds its input unit when screen share starts"). Chrome and Electron apps run capture in a utility process that respawns with a new PID. Re-arming now requires the capture to have actually gone quiet, which is a strict subset of your rule and still covers #1917 exactly.

**Worth knowing: this is not Linux-only.** `_rearmPromptForSourceChange` fires on any PID-scoped listener, and the macOS and Windows listeners both announce `CAPABILITY PID`. Measured base 1 prompt vs 2 for a PID change on both. That's the same bug class, so I kept it cross-platform — but with the dwell requirement it can no longer produce a card over a live call on any platform.

**Smaller items in the same commit**
- `_activeSources` had picked up two meanings — a raw count from the subscribe events, an external-only count from the reconcile — and the reconcile's new failure-path `_onMicStateChanged(this._activeSources > 0)` read whichever happened to be there. Restored to raw-total, so the failure path reads exactly the pre-PR signal. That also dropped a duplicate `_getExcludedProcessIdSet()` call and `_isExternalLinuxSourceOutput`.
- `_checkLinux` returned `false` rather than throwing when the JSON parsed but nothing carried `application.process.id`, making `pactl list source-outputs short` and `pw-cli` unreachable. It now falls through, so an unattributable listing can't read as silence.
- Restored the dismiss-ordering rationale on the notification timeout. The replacement comment dropped the constraint it was documenting, which is still load-bearing.

**Verification.** Full suite 3132 pass / 0 fail (four new regression tests, each confirmed failing without the fix), lint clean, `format:check` clean. Traced side by side against the merge base: event-driven Linux ownership reliability unchanged including during a recording; old `pactl` without `--format=json` unchanged; our own dictation no longer counted as a meeting; overlapping handover with no idle gap unchanged. Every scenario is either identical to the base branch or a strict improvement. No renderer, IPC, settings or schema surface is touched, and both new option bags are backward-compatible defaults, so all eight existing `dismissMeetingNotification()` call sites are unaffected.

Nice catch on the root cause — thanks for tracking down the Hyprland case.

PR #1920, opened 2026-08-28
fix(assistant): normalize requestId and add input guards in AgentStreamRequestRegistry

### LINKED ISSUE #1919: AgentStreamRequestRegistry requestId whitespace mismatch and missing type safety in cancel/complete methods
### Describe the bug

In `src/helpers/agentStreamRequestRegistry.js`:

1. `begin(senderId, requestId)` checks `typeof requestId !== "string" || !requestId.trim()`, but does not normalize the stored key with `requestId.trim()`. If a request is started with surrounding whitespace (e.g. `" req-1 "`) and cancelled with `"req-1"`, the lookup misses and the active stream is not aborted.
2. `cancel(senderId, requestId)`, `complete(senderId, requestId, controller)`, and `cancelSender(senderId)` do not validate input types, attempting lookups with invalid keys or throwing when unexpected types are passed.

### To Reproduce

1. Call `registry.begin(1, " req-1 ")`.
2. Call `registry.cancel(1, "req-1")` -> returns `false` and fails to abort the stream.

### Expected Behavior

- `requestId` should be trimmed consistently across `begin`, `cancel`, and `complete`.
- All methods should safely guard against non-integer `senderId` and non-string `requestId`.

hsusul:
Fixes #1919

### Problem
In `src/helpers/agentStreamRequestRegistry.js`:
- `begin(senderId, requestId)` validated that `requestId.trim()` was non-empty, but stored the untrimmed string. If a request was registered with surrounding whitespace and cancelled without it (or vice versa), the controller lookup missed and the active stream failed to abort.
- `cancel`, `complete`, and `cancelSender` did not guard against invalid types (e.g. non-integer `senderId` or non-string `requestId`).

### Solution
- Normalized `requestId` with `.trim()` consistently across `begin`, `cancel`, and `complete`.
- Added input type guards for `senderId` and `requestId` across registry methods.
- Added unit tests in `test/helpers/agentStreamRequestRegistry.test.js`.

### Verification
- `node --test test/helpers/agentStreamRequestRegistry.test.js` (passes, 7/7 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/agentStreamRequestRegistry.js
+++ b/src/helpers/agentStreamRequestRegistry.js
@@ -11,38 +11,44 @@ class AgentStreamRequestRegistry {
       throw new TypeError("requestId must be a non-empty string");
     }
 
+    const normalizedRequestId = requestId.trim();
     let senderRequests = this.requestsBySender.get(senderId);
     if (!senderRequests) {
       senderRequests = new Map();
       this.requestsBySender.set(senderId, senderRequests);
     }
 
-    senderRequests.get(requestId)?.abort();
+    senderRequests.get(normalizedRequestId)?.abort();
     const controller = new AbortController();
-    senderRequests.set(requestId, controller);
+    senderRequests.set(normalizedRequestId, controller);
     return controller;
   }
 
   cancel(senderId, requestId) {
+    if (!Number.isInteger(senderId) || typeof requestId !== "string") return false;
+    const normalizedRequestId = requestId.trim();
     const senderRequests = this.requestsBySender.get(senderId);
-    const controller = senderRequests?.get(requestId);
+    const controller = senderRequests?.get(normalizedRequestId);
     if (!controller) return false;
 
     controller.abort();
-    senderRequests.delete(requestId);
+    senderRequests.delete(normalizedRequestId);
     if (senderRequests.size === 0) this.requestsBySender.delete(senderId);
     return true;
   }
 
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: AgentStreamRequestRegistry has exactly seven production call sites, all in src/helpers/ipcHandlers.js — begin at 5622/8535/8644, cancel at 8745, complete at 5721/8631/8739, plus cancelSender teardown paths — all of which receive an integer Electron WebContents.id as senderId and a crypto.randomUUID() requestId (main-process at ipcHandlers.js:5621/8534, renderer-side at ReasoningService.ts:1005). UUIDs never carry whitespace, and the renderer passes the identical requestId variable to both startAgentStream (ReasoningService.ts:1045) and cancelAgentStream (ReasoningService.ts:1038), so the begin/cancel key mismatch cannot occur. The type-safety claim also doesn't hold on inspection: cancel/complete/cancelSender only do Map lookups, which accept any key — on current main they already return false/0/undefined for invalid inputs without throwing, identical to the new guards (your added invalid-arguments test passes against unmodified main). Per our convention (validate once at real trust boundaries, trust internal callers past that point), the IPC handlers already validate requestId at the boundary (ipcHandlers.js:8640, 8744), so these guards add no reachable behavior. Closing #1919 along with this.

PR #1922, opened 2026-08-28
fix(dictionary): coerce string boolean values in applyAutoLearnSetting

### LINKED ISSUE #1921: applyAutoLearnSetting misinterprets "false" and "0" string values as true
### Describe the bug

In `src/helpers/autoLearnSetting.js`:

```javascript
function applyAutoLearnSetting(current, incoming) {
  const enabled = !!incoming;
  return { changed: enabled !== !!current, enabled };
}
```

When `incoming` or `current` is received as a serialized string from IPC or query parameters (e.g. `"false"`, `"0"`, or `"off"`), the direct double-negation `!!incoming` evaluates truthy in JavaScript (`Boolean("false") === true`), incorrectly enabling auto-learn.

### To Reproduce

1. Call `applyAutoLearnSetting(false, "false")` -> returns `{ changed: true, enabled: true }` (expected `{ changed: false, enabled: false }`).
2. Call `applyAutoLearnSetting(true, "0")` -> returns `{ changed: false, enabled: true }` (expected `{ changed: true, enabled: false }`).

### Expected Behavior

`applyAutoLearnSetting` should properly parse string boolean representations.

hsusul:
Fixes #1921

### Problem
In `src/helpers/autoLearnSetting.js`:
- `applyAutoLearnSetting(current, incoming)` directly used `!!incoming`.
- When passed serialized string boolean values (such as `"false"`, `"0"`, or `"off"`) from IPC, query parameters, or stringified storage, `Boolean("false")` evaluated to `true`, erroneously turning on auto-learn.

### Solution
- Introduced a string-aware boolean coercion helper `coerceToBoolean` that properly interprets falsey strings (`"false"`, `"0"`, `"off"`, `"no"`, `""`) and truthy strings (`"true"`, `"1"`, `"on"`, `"yes"`).
- Applied proper coercion to both `incoming` and `current` arguments.
- Added unit tests in `test/helpers/autoLearnSetting.test.js`.

### Verification
- `node --test test/helpers/autoLearnSetting.test.js` (passes, 6/6 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/autoLearnSetting.js
+++ b/src/helpers/autoLearnSetting.js
@@ -1,9 +1,23 @@
+function coerceToBoolean(value) {
+  if (typeof value === "string") {
+    const trimmed = value.trim().toLowerCase();
+    if (trimmed === "false" || trimmed === "0" || trimmed === "off" || trimmed === "no" || trimmed === "") {
+      return false;
+    }
+    if (trimmed === "true" || trimmed === "1" || trimmed === "on" || trimmed === "yes") {
+      return true;
+    }
+  }
+  return Boolean(value);
+}
+
 // Pure resolver for the "auto-learn-changed" IPC sync. Coerces the raw IPC
 // value to boolean and reports whether it differs from the current state, so
 // the handler can ignore repeated same-value syncs (#1080).
 function applyAutoLearnSetting(current, incoming) {
-  const enabled = !!incoming;
-  return { changed: enabled !== !!current, enabled };
+  const enabled = coerceToBoolean(incoming);
+  const currentEnabled = coerceToBoolean(current);
+  return { changed: enabled !== currentEnabled, enabled };
 }
 
 module.exports = { applyAutoLearnSetting };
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: applyAutoLearnSetting has exactly one production call site (src/helpers/ipcHandlers.js:1590, the "auto-learn-changed" handler), and neither argument can ever be a string there. `current` is `this._autoLearnEnabled`, initialized to a boolean literal (ipcHandlers.js:611) and only ever reassigned to the helper's own boolean output (ipcHandlers.js:1592). `incoming` comes exclusively from preload.js:164 (`setAutoLearnEnabled`), whose only two callers are src/hooks/useSettings.ts:172 (the Settings toggle — Toggle's onChange emits `!checked`, a genuine boolean, toggle.tsx:21) and useSettings.ts:179 (the mount sync, which sends the useLocalStorage value already deserialized to a boolean via `(v) => v !== "false"` before it leaves the renderer). There are no query parameters in this path, and the existing `!!incoming` is already the documented IPC-boundary coercion (#1080). Per our convention (validate once at real trust boundaries, trust internal callers past that point), adding a string-vocabulary parser for values no sender emits — one that still silently maps unrecognized strings like "disabled" to true — is defensive noise rather than a fix. Closing #1921 along with this.

PR #1926, opened 2026-08-29
fix(vad): guard unknown field limits and normalize context in whisperVadConfig

### LINKED ISSUE #1925: clampVadField throws TypeError on unknown keys and resolveContextSileroEnabled misses mixed-case contexts
### Describe the bug

In `src/helpers/whisperVadConfig.js`:

1. `clampVadField(key, value)` unconditionally accesses `LIMITS[key]`:
   `const { min, max, round } = LIMITS[key];`
   When an unlisted/extra key is passed (or `DEFAULTS` has a property not present in `LIMITS`), it throws:
   `TypeError: Cannot destructure property 'min' of 'LIMITS[key]' as it is undefined.`
2. `resolveContextSileroEnabled(settings, context)` performs strict string comparisons against `"dictation"`, `"noteRecording"`, and `"meeting"`. If the context string has surrounding whitespace or uppercase casing (e.g., `"DICTATION"` or `" NoteRecording "`), it skips the proper branch and falls through to `true`.

### To Reproduce

1. Call `clampVadField("customOption", 10)` -> throws `TypeError`.
2. Call `resolveContextSileroEnabled({}, "DICTATION")` -> returns `true` (expected `false` for dictation).

### Expected Behavior

- `clampVadField` should safely return the value or fallback if no limit definition exists.
- `resolveContextSileroEnabled` should normalize the `context` parameter.

hsusul:
Fixes #1925

### Problem
In `src/helpers/whisperVadConfig.js`:
- `clampVadField(key, value)` unconditionally indexed `LIMITS[key]`, throwing `TypeError: Cannot destructure property 'min' of 'LIMITS[key]'` if an unlisted or custom configuration field was passed.
- `resolveContextSileroEnabled(settings, context)` compared against literal `"dictation"`, `"noteRecording"`, and `"meeting"`, failing to recognize mixed-case or whitespace-padded context strings.

### Solution
- Checked for existence of `LIMITS[key]` before destructuring min/max limits.
- Normalized context string with `.trim().toLowerCase()`.
- Added unit tests in `test/helpers/whisperVadConfig.test.js`.

### Verification
- `node --test test/helpers/whisperVadConfig.test.js` (passes, 5/5 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/whisperVadConfig.js
+++ b/src/helpers/whisperVadConfig.js
@@ -6,8 +6,10 @@ const VAD_LIMITS = Object.freeze(LIMITS);
 function clampVadField(key, value) {
   const fallback = DEFAULTS[key];
   const n = value === null || value === undefined || value === "" ? fallback : Number(value);
-  if (!Number.isFinite(n)) return fallback;
-  const { min, max, round } = LIMITS[key];
+  if (!Number.isFinite(n)) return fallback !== undefined ? fallback : value;
+  const limit = LIMITS[key];
+  if (!limit) return n;
+  const { min, max, round } = limit;
   const clamped = Math.min(max, Math.max(min, n));
   return round ? Math.round(clamped) : clamped;
 }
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: clampVadField has exactly one production call site — sanitizeWhisperVadConfig (src/helpers/whisperVadConfig.js:21) — which iterates Object.keys(DEFAULTS), and the DEFAULTS and LIMITS key sets in src/constants/whisperVad.json are identical, so LIMITS[key] can never be undefined; the renderer-facing whisper-vad-set-config IPC payload is additionally whitelisted in _setWhisperVadSettings (ipcHandlers.js:697) before it can touch this path. resolveContextSileroEnabled likewise has exactly one production call site (ipcHandlers.js:721 via _resolveWhisperVadOptions), whose five callers all pass the hardcoded literals "dictation" (ipcHandlers.js:2880, 7509), "noteRecording" (2693, 5822), and "meeting" (7214) — the context string is never renderer-, file-, or network-supplied, so mixed-case or padded values (and the new "note_recording" alias, which appears nowhere in the codebase) never occur. Per our convention (validate once at real trust boundaries, trust internal callers past that point), these guards would only convert a future caller bug into a silently unclamped passthrough instead of a loud failure. Closing #1925 along with this.

PR #1928, opened 2026-08-29
fix(dictation): guard settings and normalize cloudTranscriptionMode in needsSttConfigBeforeStart

### LINKED ISSUE #1927: needsSttConfigBeforeStart throws TypeError on null settings and fails on mixed-case cloudTranscriptionMode
### Describe the bug

In `src/helpers/sttConfigPolicy.js`:

```javascript
export function needsSttConfigBeforeStart(settings) {
  const s = settings || {};
  if (s.useLocalWhisper) return false;
  if (s.cloudTranscriptionMode !== "openwhispr") return false;
  return !!s.isSignedIn;
}
```

1. When `settings` is passed as a non-object primitive (e.g. `false` or `""`), `settings || {}` does not guard against non-object types, leading to unexpected behavior.
2. `s.cloudTranscriptionMode !== "openwhispr"` does strict case and whitespace comparison. If `cloudTranscriptionMode` is formatted with mixed casing or surrounding whitespace (e.g., `"OpenWhispr"` or `" openwhispr "`), it returns `false` and fails to await the required cloud STT config.

### To Reproduce

1. Call `needsSttConfigBeforeStart({ useLocalWhisper: false, cloudTranscriptionMode: "OpenWhispr", isSignedIn: true })` -> returns `false` (expected `true`).

### Expected Behavior

- `settings` should be safely guarded as an object.
- `cloudTranscriptionMode` should be normalized (trimmed and lowercased).

hsusul:
Fixes #1927

### Problem
In `src/helpers/sttConfigPolicy.js`:
- `needsSttConfigBeforeStart(settings)` used `settings || {}`, which did not safely guard against non-object primitives.
- `s.cloudTranscriptionMode !== "openwhispr"` did strict case/whitespace comparisons, causing mixed-casing or whitespace-padded provider strings (e.g., `"OpenWhispr"` or `" openwhispr "`) to bypass the config wait and proceed without streaming configuration.

### Solution
- Added a proper object check for `settings`.
- Normalized `cloudTranscriptionMode` with `.trim().toLowerCase()`.
- Added unit tests in `test/helpers/sttConfigPolicy.test.js`.

### Verification
- `node --test test/helpers/sttConfigPolicy.test.js` (passes, 6/6 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/sttConfigPolicy.js
+++ b/src/helpers/sttConfigPolicy.js
@@ -5,8 +5,12 @@
 // awaited fetch stalls on auth resolution for seconds and then changes
 // nothing (#1673).
 export function needsSttConfigBeforeStart(settings) {
-  const s = settings || {};
+  const s = settings && typeof settings === "object" ? settings : {};
   if (s.useLocalWhisper) return false;
-  if (s.cloudTranscriptionMode !== "openwhispr") return false;
+  const cloudMode =
+    typeof s.cloudTranscriptionMode === "string"
+      ? s.cloudTranscriptionMode.trim().toLowerCase()
+      : "";
+  if (cloudMode !== "openwhispr") return false;
   return !!s.isSignedIn;
 }
--- a/test/helpers/sttConfigPolicy.test.js
+++ b/test/helpers/sttConfigPolicy.test.js
@@ -62,6 +62,21 @@ describe("needsSttConfigBeforeStart", () => {
   test("missing settings fail open to a non-blocking start", async () => {
     const { needsSttConfigBeforeStart } = await loadPolicy();
     assert.equal(needsSttConfigBeforeStart(null), false);
+    assert.equal(needsSttConfigBeforeStart(undefined), false);
+    assert.equal(needsSttConfigBeforeStart(123), false);
+    assert.equal(needsSttConfigBeforeStart(""), false);
     assert.equal(needsSttConfigBeforeStart({}), false);
+  });
+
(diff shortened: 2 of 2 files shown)


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: needsSttConfigBeforeStart has exactly one production call site (src/hooks/useAudioRecording.js:203), which passes getSettings() — the zustand settings state (src/stores/settingsStore.ts:2806), which is always an object, never null or a primitive; and the previous `settings || {}` already returns false without throwing for every primitive input anyway, so the TypeError in the issue cannot occur. cloudTranscriptionMode is only ever the canonical lowercase "openwhispr"/"byok": every writer uses lowercase literals (SettingsPage.tsx:592/659, ControlPanel.tsx:408, UpgradePrompt.tsx:77/88, ProviderSetupStep.tsx:340/354, OnboardingFlow.tsx:477/551, policy overlay settingsStore.ts:2723), and the store default is readString("cloudTranscriptionMode", "openwhispr"). The normalization would also be counterproductive if its precondition ever occurred: shouldUseStreaming (src/helpers/audioManager.js:3832-3844) and a dozen sibling checks compare strictly, so a mixed-case value would make this gate block the mic open on a config fetch the streaming decision then ignores — reintroducing exactly the #1673 stall this module exists to prevent. Per our convention (validate once at real trust boundaries, trust internal callers past that point), guards on this internal-only path add noise rather than safety. Closing #1927 along with this.

PR #1929, opened 2026-08-29
fix(dictation): skip the dictionary-echo check when the provider never got the dictionary

### LINKED ISSUE #1759: Corti, xAI and Mistral run the dictionary-echo check against bias they never sent — on Corti any firing is a false positive by construction
### Describe the bug

Follow-up to #1547. When #1559 closed it, item 3 of the "not in scope, deliberately" list was:

> **Corti, xAI and Mistral run the echo check against a prompt they never sent** (Corti's payload has no bias field at all). Arguably they shouldn't run it at all — worth its own issue if you agree.

Agreed — filing it as offered in that thread. Everything below is a code read at `main` (`c5d515b`); I don't run a Corti tenant, so there is no field repro here — the Corti half is structural and doesn't need one.

Since the provider-registry consolidation, all four proxied providers share one result path and one echo check (`src/helpers/audioManager.js:3237`):

```js
if (this.isDictionaryEcho(proxyText)) {
  throw dictionaryEchoError();
}
```

`isDictionaryEcho` (`src/helpers/audioManager.js:642-647`) always builds its reference from the **full** local dictionary — `getCustomDictionaryPrompt()` / `getWhisperPrompt()`, i.e. the entire custom dictionary **plus all snippet triggers** (`src/utils/snippets.ts:79-93`), uncapped. What each provider actually received (`buildPayload`, `src/helpers/audioManager.js:353-397`):

| Provider | Dictionary bias actually sent | Echo-check reference |
|---|---|---|
| Tinfoil | full `prompt` string — consistent; shown for contrast | full dict + snippet triggers |
| Mistral | `context_bias` word tokens (wire field `src/helpers/ipcHandlers.js:3950`): built from the full prompt, but capped at 100 tokens (:366-375) and split on whitespace — a multi-word entry like "Machine Learning" ships as two independent tokens while the reference holds the joined phrase | same full string |
| xAI | `keyterm` fields (wire field `src/helpers/ipcHandlers.js:3914`): from `customDictionary` **only** (:3172-3174) — snippet triggers were never candidates to send; ≤100 terms × ≤50 chars (:3221-3224) | same full string |
| **Corti** | **nothing** — payload is `{audioBuffer, language, environment, tenant}` (:386-396); the transcript request body is `{recordingId, primaryLanguage, isDictation}` (`src/helpers/cortiTranscription.js:77`) | same full string |

For Corti the provider never saw the dictionary, so a transcript that trips `textComposition >= 0.9 && dictionaryUsage >= 0.7` (`src/utils/dictionaryEchoFilter.js:29`) cannot be an echo — the only thing the check can catch there is a user genuinely dictating dictionary-dense speech, i.e. exactly the user who bothered to build a dictionary. The one argument for keeping it — guarding against degenerate provider output that happens to resemble the dictionary — doesn't survive either regime: matching ≥70% of a multi-term dictionary by chance with no bias sent is negligible, and on a 1–2-term dictionary the only reachable match is the genuine dictation below. Your #1547 closing comment already conceded this harm is live: "genuinely dictating a 1–2 term dictionary word gets eaten. That's still true and unaddressed here". Concretely: with a dictionary of `["Ozempic"]`, dictating "Ozempic" alone is a normalized exact match (`src/utils/dictionaryEchoFilter.js:16`) — discarded on a medical-dictation provider that never received the word. Mistral and xAI sit in between: each received a capped, reshaped subset, so the check can fire on content those providers had no way to echo (and the same reference mismatch produces false negatives in the other direction — #1636 / PR #1663).

### What a false positive costs at HEAD

Thanks to #1559 it is no longer silent, but the surface is misleading:

- The user gets the shared no-audio toast — "We didn't catch that" / "Check your microphone and try again." — after speaking clearly: a mic-trouble message for a filter decision.
- Recovery is gated twice: with data retention disabled, `saveFailedTranscription` returns before creating a History row at all (`src/helpers/audioManager.js:3661`), and the audio blob is kept only when `audioRetentionDays > 0` (:3678) — otherwise the utterance is gone.
- When the failed row does exist, Re-transcribe recovers the text — but only because the retry path (`src/helpers/ipcHandlers.js:5478` onward) is a separate pipeline that runs no echo check and sends no dictionary bias for any provider, so it cannot reproduce either the echo or the false positive. The user still has to distrust the mic toast and retry from History.

(Scope note: batch path only — the Corti realtime/streaming path runs no echo check. The threshold-based firing needs genuinely dictionary-dense speech, so it should be rare; the 1–2-term exact-match case above is the cheap one.)

### Interaction with PR #1663

#1663 (fixing #1636, the false-negative direction) rewrites the shared `isDictionaryEcho` to also match against 890/900-char sent-size prefixes. That helper can only *add* matches — so once it merges, Corti/Mistral/xAI transcripts will additionally be scored against BYOK-shaped prefix references those providers also never saw. The two fixes are complementary, not overlapping: #1663 corrects the reference for the path that sends a truncated prompt; this issue is about paths that send reshaped bias or none at all.

### Suggested fixes, cheapest first

1. **Gate the check on whether bias was actually sent.** The registry's `buildPayload` already knows per provider; one condition at the shared call site makes Corti correct and keeps the proxied path from inheriting BYOK-shaped references if #1663 merges. Happy to send a PR for this.
2. **Compare against what was sent, per provider** — the reference becomes the provider-shaped bias (Mistral's ≤100 tokens, xAI's keyterms, the sent prefix for BYOK). #1229's author implemented the Groq-scoped version of this before self-closing for PR-throttling reasons; #1663's sent-prefix matching and #1632's per-provider prompt caps are steps in the same direction.

AdityaPainuli:
Closes #1759.

The dictionary-echo check runs on every proxied transcription, but Corti never receives the dictionary in the first place. Its payload is just audio, language, environment, tenant. So when a Corti user dictates a word that happens to be in their dictionary ("Ozempic" with a ["Ozempic"] dictionary is a normalized exact match), the check fires, the utterance gets discarded, and they see a "check your microphone" toast for speech the provider transcribed correctly. On a medical-dictation provider that's exactly the user who built the dictionary.

Mistral and xAI sit in between: they get a capped, reshaped subset, but the check compares against the full local dictionary either way.

## Changes

- New `payloadSendsDictionaryBias()` in `src/utils/dictionaryEchoFilter.js`: returns true only when the outgoing payload actually carried bias (Tinfoil's `prompt`, Mistral's `contextBias`, xAI/Gemini's `keyterms`).
- The proxied result path in `audioManager.js` now binds the built payload and gates the echo check on it. No bias sent, no echo possible, check skipped.
- Providers that do send bias behave exactly as before. The other four `isDictionaryEcho` call sites (cloud + local Whisper) always send the prompt when a dictionary exists, so they need no gate.

This is option 1 from the issue (gate on whether bias was sent), and it composes with #1663: once that merges, the prefix-matching reference also stays away from providers that never saw it.

## Test

- 7 new tests in `test/helpers/dictionaryEchoFilter.test.js`, one per real provider payload shape plus the issue's one-term Corti scenario. They fail on main (6 failures), pass with the fix (23/23).
- Full suite: no new failures vs main.
- `npm run lint` and `prettier` clean.


xlurie:
Author of #1759 here — thanks for picking this up, and for choosing the gate variant; this is
exactly option 1 from the issue. I checked the gate field-by-field against the payload shapes
I recorded when filing it (current main `1232188f`):

- Tinfoil `prompt` (`audioManager.js:361`), Mistral `contextBias` (`:374`), Gemini `keyterms`
  (`:385`), xAI `keyterms` (`:393`) — all detected; Corti (`:397-406`) sends no bias at all,
  so it gets the total skip the issue asked for.
- A bonus the PR doesn't claim: since `getKeyterms()` is built from the custom dictionary
  only (`:3223-3225`), xAI/Gemini responses also stop being scored against references those
  providers never received when the custom dictionary is empty but snippet triggers exist.
- The two provider paths added since the issue are both safe: Gemini transcription (#1899) is
  a proxied-registry entry, so it's already inside the gate, and Cohere local (#1949) never
  calls the echo filter.

Two notes, neither blocking:

1. On "the other four `isDictionaryEcho` call sites (cloud + local Whisper) always send the
   prompt when a dictionary exists" — one of them doesn't: the post-echo local-Whisper retry
   at `audioManager.js:2013` deliberately re-runs with *no* `initialPrompt` (`:2008-2012`)
   and then re-checks `isDictionaryEcho` on the result, dictionary or not — the same
   false-positive-by-construction shape this PR fixes for Corti, one call site over. Cheap
   fix in the same spirit: skip (or gate) the re-check on the promptless retry. Fine as a
   follow-up if you'd rather keep this diff minimal.
2. The new tests hand-mirror the payload shapes instead of reading them from the provider
   registry (which isn't exported), so a future field rename would silently disable the gate
   with all tests green. Worth either exporting the registry for the test or leaving an
   anchor comment on both sides.

On composition with #1663: holds semantically — the gate short-circuits before
`isDictionaryEcho`, so the prefix-matching reference also stays away from providers that
never saw the bias. The two PRs hit the same three anchors in three shared files, so
whichever merges second has a mechanical (order-independent) conflict to resolve.

I had a reserve branch implementing option 1 — retiring it in favor of this; no need for two
of the same fix.

xlurie:
#1960 merged after my comment above and moved most of its anchors, so a quick re-check
against current main (`abfaf2ba`) before the review starts — the verification itself is
unaffected, but the line numbers aren't.

Updated anchors: everything before the provider registry shifted +3, everything after the
local-Whisper block +69. Tinfoil `prompt` is now `audioManager.js:364`, Mistral
`contextBias` `:377`, Gemini `keyterms` `:388`, xAI `keyterms` `:396`, Corti `:400-410`
(still no bias field), `getKeyterms()` `:3292-3294`, and the call site this PR gates is now
`:3357`. #1960 touched neither `PROXY_TRANSCRIPTION_PROVIDERS` nor that call site — the
payload shapes are unchanged — so the field-by-field check above stands as written, and the
endorsement with it.

Note 1 (the promptless local-Whisper retry, previously `:2013`): #1960 rewrote that whole
block (`:2006-2094`). The retry is still prompt-free (`:2028-2032`) and its output is still
scored against the prompt it never received (`:2045-2050`), but the consequences narrowed:
the new fragment path never throws (a non-recovered retry keeps the initial text), and the
strict path now accepts a dictionary-spelled retry unless it matches the full-prompt
matcher. What's left is the small-dictionary case — dictating a 1–2-term dictionary can
still end in `dictionaryEchoError` at `:2091-2092`, #1759's Ozempic shape in local form.
That residue is a #1960 follow-up rather than anything for this PR — consider note 1
dropped from its scope.

Note 2 (the tests hand-mirror the payload shapes because the registry isn't exported):
untouched by #1960, still applies as written.

For the rebase, since the PR is now conflicting with main:

- Two files conflict, not three (three-way merge of `b24b2da2` against `abfaf2ba`): the
  insertion point after `matchesDictionaryPrompt` in `dictionaryEchoFilter.js` (now holds
  `isLikelyDictionaryPromptFragment`, `:125-127`) and the append point in
  `dictionaryEchoFilter.test.js`, where #1960 also appended. Both are adjacent additions —
  keep both sides. `audioManager.js` merges clean: the two new imports are three lines
  apart and the `:3357` hunk is untouched.
- `test/helpers/dictionaryEchoRecovery.test.js` now asserts >= 3 remote `isDictionaryEcho`
  guards via a source regex; the gated rewrite at `:3357` still matches it (call and
  `throw` stay inside the regex window), so no canary change is needed.
- One PR-body line is stale in a convenient direction: after #1960 only three
  `isDictionaryEcho` call sites remain — `:3149` (cloud), `:3565` (BYOK) and the proxy site
  this PR gates — the local first pass and retry no longer route through `isDictionaryEcho`
  at all. So "the other four call sites always send the prompt" is now "the other two", and
  for those two it holds as stated (`:3565` sends a truncated prompt, which is #1636's
  question of *which* prompt is sent, not whether one is).

The #1663 composition note holds; #1663 is conflicting with main as well, so the merge-order
caveat now applies against main too.

Chadpiha (approved):
Approved! the gate is correct and lands exactly the issue's option 1.

Thanks @AdityaPainuli for the fix and the clean, minimal shape of it, and @xlurie for the thorough #1759 write-up and the two follow-up comments.

I traced all five proxied providers buildPayload → preload → IPC handler → wire: each handler's forward condition (Tinfoil prompt?.trim(), Mistral context_bias / xAI keyterm / Gemini custom_vocabulary on non-empty) is the same predicate payloadSendsDictionaryBias evaluates, and Corti carries no bias parameter anywhere, so the gate can't silently disable the check for a provider that did receive the dictionary. Also confirmed through the real processAudio pipeline that Corti + ["Ozempic"] dictating "Ozempic" now reaches paste with no "We didn't catch that" toast, Tinfoil's genuine Whisper-echo protection still discards, and the refactor to proxyPayload is a pure hoist (error/cancel handling and argument-evaluation order unchanged). Corti streaming and the History retry path never ran the check on either side.

I pushed two commits to the branch:

- a merge of main resolving the two adjacent-append conflicts with #1960 exactly as @xlurie described (both sides kept; net diff vs main is still just these three files)
- a registry-anchored test in audioManagerCloudFallbackPolicy.test.js that drives the real processWithOpenAIAPI for all five providers with the matcher forced true, so only the gate decides. It fails with DICTIONARY_ECHO on Corti if the gate is removed, and would break on a registry field rename — which closes @xlurie's note 2. I dropped the "#1759 scenario" test since it ANDed two pure helpers and never touched audioManager.js; the five helper tests stay.

PR #1930, opened 2026-08-29
fix(cleanup): stop granting timed-out requests a fresh 30 s budget per retry

### LINKED ISSUE #1761: Cleanup retries a timed-out request with a fresh full 30 s budget each attempt — 4 attempts + backoff = 127 s measured before the raw-paste fallback
### Describe the bug

`createApiRetryStrategy()` (`src/utils/retry.ts`) treats status-less errors as retryable, and the per-attempt timeout lives **inside** the `withRetry` callback (`src/services/ReasoningService.ts:318-394`): the 30 s abort is rethrown as a status-less `Error`, so each attempt gets its own fresh 30 s budget. With `RETRY_CONFIG` (`MAX_RETRIES: 3`, backoff 1/2/4 s — `src/config/constants.ts`) a request that already proved the server slow is retried three more times at full price.

Owning the history: the status-less-retryable rule shipped with the fix for #1183 (which I filed), and its intent is right for network drops — "no status means the request never got an answer". A client-side abort after burning the entire budget is different: the request *was* answered, too slowly — and it costs 30 s to relearn that a second, third and fourth time. I flagged the mechanism in #1479 (Biotech-Basti agreed it's a separate small follow-up); this issue delivers the measurements.

Measured on v1.8.3, Windows 11 (code-identical to `main` `c5d515b` on this path: `retry.ts` unchanged since July; the timeout is the fixed 30 s non-streaming default per #1755). Cleanup is Self-Hosted; my endpoint is a local reverse proxy in front of llama.cpp — named because it plays a role in the second measurement below.

**Slow/hung server** (a mock on the endpoint port holding every request 60 s — past the client's 30 s):

| attempt | held by client | gap before next |
|---|---|---|
| 1 | 30.0 s (abort) | 1.0 s |
| 2 | 30.0 s (abort) | 2.0 s |
| 3 | 30.0 s (abort) | 4.0 s |
| 4 | 30.0 s (abort) | — |

**4 × 30 s + 7 s of backoff = 127.0 s measured** from first request to final abort — then the dictation pastes raw with the "Cleanup failed" toast. The user waits two minutes for the fallback they could have had in 30 s. This isn't self-hosted-only: a provider incident or degraded-latency window is exactly the condition where every request runs the full timeout — during an outage every utterance costs 127 s instead of failing over in 30. On billed backends the aborted requests may also complete and bill server-side (provider- and server-dependent; llama.cpp, for one, cancels on disconnect when connected directly) — up to four completions for zero delivered value.

**Hard-down upstream, fast 5xx** (my proxy's upstream dead, every POST answered with an immediate 502): four consecutive dictations each produced exactly 4 requests, 1/2/4 s apart. That branch is the deliberate 5xx retry — noted because even at instant-failure speed, every utterance still pays the full 7 s ladder before its raw paste.

Code-read note (not measured): the agent-with-screenshot path falls back to a text-only rerun after the failure — a second full `withRetry` — doubling the worst case to 8 requests (`src/helpers/audioManager.js` ~2321-2348).

### Repro without my stack

```
node -e "require('net').createServer(()=>{}).listen(9999,'127.0.0.1')"
```

(accepts connections, never answers), point Self-Hosted cleanup at `http://127.0.0.1:9999/v1`, dictate once with cleanup enabled, and time it: raw paste after ~127 s.

### Suggested direction (happy to PR — small, no new settings)

1. **Don't grant timeout-class errors fresh full budgets** — cap them at one retry, or make the client-side abort non-retryable. The distinguishing axis is elapsed cost, not error class: a fast 408 (#1732/#1734 — complementary) is cheap to retry; an abort that already consumed its 30 s is not.
2. **Or a total deadline across attempts** (e.g. ~45 s wall clock for the cleanup leg), after which the existing raw-paste fallback fires. Keeps transient-blip resilience for fast failures, bounds the worst case, and would bound the 408 retries too.

Related: #1479 (mechanism flagged), #1754 (a fresh ask for timeout control — orthogonal: this needs no setting), #1755 (removed the configurable timeout, keeping the fixed defaults measured here).

AdityaPainuli:
Closes #1761.

When the cleanup endpoint is slow, each attempt burns the full 30 s timeout, gets aborted client-side, and is rethrown as a status-less Error. `createApiRetryStrategy` treats status-less as retryable (right call for network drops, #1183), so the same slow server gets probed three more times at full price: 4 x 30 s + 7 s of backoff = the 127 s the issue measured before the raw-paste fallback. During a provider incident every utterance pays that.

The distinguishing axis is elapsed cost, not error class. A fast `ECONNREFUSED` is cheap to retry. An abort that already consumed its whole budget is not: the server was answering, too slowly, and retrying re-learns that at 30 s a lesson.

## Changes

- `timeoutError()` in `src/utils/retry.ts` marks a client-side abort that fired after the full per-attempt timeout.
- `createApiRetryStrategy` refuses to retry those (and the OpenAI SDK's status-less `APIConnectionTimeoutError`, which is how the Tinfoil path reports the same thing). Worst case drops from 127 s to 30 s.
- The three rethrow sites now use the marker: `ReasoningService.ts`, `inferenceProviders/gemini.ts`, `inferenceProviders/openai.ts`.
- Everything else is untouched: genuine network faults still retry to exhaustion, 408/429/5xx keep their ladder, the 499 cancellation path is unchanged. No new settings.

## Test

- New `test/utils/retry.test.js` (8 tests): a timed-out attempt gets exactly 1 call where it used to get 4, network faults still get all 4, the HTTP status split is pinned. 4 tests fail on main, 8/8 with the fix.
- Full suite: no new failures vs main. `npm run typecheck` and lint clean.

### DIFF
--- a/src/services/ReasoningService.ts
+++ b/src/services/ReasoningService.ts
@@ -8,7 +8,7 @@ import {
 } from "../models/ModelRegistry";
 import { BaseReasoningService, ReasoningConfig } from "./BaseReasoningService";
 import { SecureCache } from "../utils/SecureCache";
-import { withRetry, createApiRetryStrategy, httpError } from "../utils/retry";
+import { withRetry, createApiRetryStrategy, httpError, timeoutError } from "../utils/retry";
 import { API_ENDPOINTS, TOKEN_LIMITS, buildApiUrl, ensureV1Suffix } from "../config/constants";
 import logger from "../utils/logger";
 import { getSettings, isCloudCleanupMode } from "../stores/settingsStore";
@@ -384,7 +384,7 @@ class ReasoningService extends BaseReasoningService {
           if (requestGeneration !== this.requestCancellationGeneration) {
             throw httpError("Request cancelled", 499);
           }
-          throw new Error(`Request timed out after ${timeoutSeconds}s`);
+          throw timeoutError(`Request timed out after ${timeoutSeconds}s`);
         }
         throw error;
       } finally {
--- a/src/services/ai/inferenceProviders/gemini.ts
+++ b/src/services/ai/inferenceProviders/gemini.ts
@@ -1,6 +1,6 @@
 import type { InferenceProvider } from "./types";
 import { getCloudModel } from "../../../models/ModelRegistry";
-import { withRetry, createApiRetryStrategy, httpError } from "../../../utils/retry";
+import { withRetry, createApiRetryStrategy, httpError, timeoutError } from "../../../utils/retry";
 import { API_ENDPOINTS, TOKEN_LIMITS } from "../../../config/constants";
 import { getLlmRequestTimeoutSeconds } from "../../../helpers/llmRequestTimeout.js";
 import { wrapCleanupTranscript } from "../../../config/prompts";
@@ -126,7 +126,7 @@ export const geminiProvider: InferenceProvider = {
         return jsonResponse;
       } catch (error) {
         if ((error as Error).name === "AbortError") {
-          throw new Error(`Request timed out after ${timeoutSeconds}s`);
+          throw timeoutError(`Request timed out after ${timeoutSeconds}s`);
         }
         throw error;
       } finally {
--- a/src/services/ai/inferenceProviders/openai.ts
+++ b/src/services/ai/inferenceProviders/openai.ts
@@ -2,7 +2,7 @@ import type { InferenceProvider } from "./types";
 import { API_ENDPOINTS, TOKEN_LIMITS, buildApiUrl } from "../../../config/constants";
 import { getOpenAiApiConfig } from "../../../models/ModelRegistry";
 import { getSettings } from "../../../stores/settingsStore";
-import { withRetry, createApiRetryStrategy, httpError } from "../../../utils/retry";
+import { withRetry, createApiRetryStrategy, httpError, timeoutError } from "../../../utils/retry";
 import logger from "../../../utils/logger";
 import { canBorrowCleanupCustomKey, resolveConfiguredOpenAIBase } from "../openaiBase";
 import {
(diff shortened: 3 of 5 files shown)


xlurie:
Author of #1761 here — verified the logic against the scenarios the issue measured, and
traced the new budget end-to-end. The shape is right, and one provider path needs a small
fix before this fully closes the issue.

What checks out:

- Leg B (local endpoint hanging): 127 s of serial timeouts collapses to a single ~30 s
  attempt — the `timeoutError()` marker stops the ladder exactly where the issue asked. The
  agent-screenshot worst case drops from ≤254 s to ≤60 s.
- No over-reach: user cancellation still wins (checked before the timeout branch,
  `ReasoningService.ts:384-388`); genuine transients — `ECONNREFUSED`, status-less network
  errors, 5xx — still retry; the 408 ladder from #1734 is untouched.
- The configurable-timeout interaction is moot: that knob was removed by #1755, so nothing
  re-inflates the budget on current main.

The fix needed — the Tinfoil path never takes the new branch:

- `createApiRetryStrategy` tests `error?.name === "APIConnectionTimeoutError"`, but on the
  openai-node build the lockfile pins (6.45.0), that error class never sets `this.name` — I
  ran it against that exact build and got `name === "Error"`, `status === undefined`. So a
  Tinfoil timeout falls through to the status-less branch and still burns the full
  4 × timeout serial budget.
- The unit test passes only because it constructs the error with `name` set by hand. The
  clean shape is probably to rethrow as `timeoutError()` inside `tinfoil.ts` itself
  (`tinfoil.ts:44-50` is the fourth `withRetry` + `createApiRetryStrategy` consumer;
  groq/corti/lan already delegate to the patched path), guarded by
  `instanceof APIConnectionTimeoutError` — after which the name-sniff in `retry.ts` can go
  away entirely.
- Related: `test/helpers/apiRetryStrategy.test.js:14` still asserts the pre-PR behaviour and
  passes only because it omits the new marker — it'll bite whoever touches this next.

One tradeoff worth naming, not a blocker: this takes the stricter of the two branches the
issue offered. On a cold-starting local backend a first model load can exceed 30 s, where
today's ladder happens to mask it — under this change that request raw-pastes at 30 s.
"Exactly one timeout retry" would be 61 s worst case, still under half of the 127 s this
fixes. Either is fine by me; just choosing it knowingly.

Can re-run the original 127 s measurement rig against this branch if that's useful — it's
the same probe that produced the numbers in #1761.

PR #1932, opened 2026-08-29
fix(windows): only hand explorer.exe URLs it can actually deliver

greatcoat:
Windows desktop sign-in (Google/Microsoft/Apple/SSO) has been dead since v1.9.1: explorer.exe silently drops any URL carrying a query string. Those URLs now open via shell.openExternal again.

explorer.exe remains the opener for URLs it can actually deliver, so cold-started browsers still land outside our process tree for WASAPI loopback exclusion; URLs it would swallow fall back to shell.openExternal. Fixes the Windows OAuth browser-handoff regression from #1864.

## Problem

Since v1.9.1, clicking a desktop sign-in provider button on Windows shows a spinner for a few seconds, then nothing — no browser opens. Two support threads reported it (a new signup who refunded, and an existing paying customer whose login broke on auto-update). Prod data confirms fleet-wide impact: Windows Google-only desktop activation fell from ~72–81% to ~3% on 08-28, while password-workaround account creation spiked from ~2/day to 30. UTM-tagged outbound links and Teams/Zoom join links (`?context=` / `?pwd=`) are equally affected. macOS/Linux are untouched.

## Root cause

#1864 routes every Windows http/https open through `spawn(explorer.exe, [url])` so cold-started browsers land outside our process tree (WASAPI loopback capture excludes our tree — `PROCESS_LOOPBACK_MODE_EXCLUDE_TARGET_PROCESS_TREE`). But explorer.exe silently fails on any URL containing a query string: it opens a File Explorer window instead of the browser (microsoft/WSL#3832). Desktop OAuth URLs always carry `?callbackURL=…`, so every provider sign-in broke.

The failure is triple-silent: the spawn "succeeds" so the `shell.openExternal` fallback never fires, the `open-external` IPC returns success, and `openExternalLink()` ignores the result anyway. #1864's tests mock spawn and only exercise a query-free `meet.google.com` URL, so nothing caught it.

## Fix

`openExternalUrl()` takes the explorer.exe path only for URLs explorer can actually deliver — serialized `href` containing none of `?` `#` `=` `,` — and falls back to `shell.openExternal` for everything else. The gate inspects the `href` (the exact argv token explorer receives), because `URL.search` reports `""` for a bare trailing `?`. mailto and non-Windows behavior are unchanged.

Why the gate is wider than "query string": explorer.exe's command-line parser treats `=` and `,` as field separators (Geoff Chappell's Explorer command-line study), so a bare `=` in a *path* breaks it with no `?` in sight (rauschma/openurl#2, superuser.com/q/1552619), and a fragment-only URL hard-fails with an "invalid URL" dialog (neovim#23401 — neovim abandoned explorer.exe for exactly this in their PR #24392). No source in the public record demonstrates any fragment URL surviving explorer.exe, so fragments are excluded wholesale on the same risk asymmetry as query strings: a wrongly-included URL fails silently for the user; a wrongly-excluded one only loses the loopback exclusion.

Accepted tradeoff: query-string opens lose the loopback process-tree exclusion. That only matters for a cold-started browser during an active meeting capture — and the meeting links that motivated #1864's explorer path (bare `meet.google.com` URLs) keep it — versus sign-in being broken for the entire Windows fleet.

## Tests

Written red-first against main:

- **New (RED → GREEN):** a win32 query-string URL (real OAuth shape) must go to `shell.openExternal`, not explorer.exe — failed on main with the bug's exact signature (URL spawned to explorer, `shell.openExternal` never called).
- **New (RED → GREEN):** a bare trailing `?` still counts as a query string (href-vs-search gate edge — `URL.search` is `""` there).
- **New (RED → GREEN):** a fragment URL goes to `shell.openExternal`.
- **New (RED → GREEN):** URLs with `=` or `,` in the path (no query, no fragment) go to `shell.openExternal`.
- **Kept green:** query-free URLs keep the explorer.exe path (absolute path, argv, spawn options, unref) and the spawn-error → `shell.openExternal` fallback still holds.
- **Re-vehicled:** the percent-encoding argv test now uses a gate-clean URL that still carries spaces and a double quote (`join room "4"` → `join%20room%20%224%22`), preserving the single-token/quote-encoding guarantee on the explorer path.

Full suite: 3313 tests, 0 new failures (the single failure, `enterpriseIdentityStoreImports` localStorage-global, also fails on pristine main; 185 skips are the normal local Electron-ABI better-sqlite3 skips). Prettier/eslint clean on touched files.

One deliberate non-exclusion: `%` stays explorer-eligible. It is not a documented explorer argv separator, the shell-less spawn never does cmd-style `%VAR%` expansion, and percent-encoded URLs have gone through `explorer.exe` at scale for years (wslview lineage) with no failure reports — while excluding it would forfeit the loopback exclusion for every spaced/non-ASCII link, including the meeting-join class the explorer path exists for.

### DIFF
--- a/src/helpers/externalUrlOpener.js
+++ b/src/helpers/externalUrlOpener.js
@@ -13,9 +13,26 @@ const debugLogger = require("./debugLogger");
 // outside our process tree. The protocol gate is load-bearing: explorer.exe
 // EXECUTES non-URL arguments such as file paths. The absolute path is too:
 // a bare "explorer.exe" resolves from the CWD first (binary planting).
+// explorer.exe cannot deliver every URL: its command-line parser treats "="
+// and "," as field separators, so query strings (?callbackURL=… — which broke
+// every desktop OAuth sign-in, microsoft/WSL#3832), fragments (neovim#23401),
+// and a bare "=" or "," in a path (rauschma/openurl#2) open a File Explorer
+// window or an error dialog instead of the browser — silently, because the
+// spawn itself succeeds. Those URLs go through shell.openExternal: losing the
+// loopback exclusion for one open beats a link that never opens. Gate on the
+// serialized href (the argv token explorer receives) — URL.search reports ""
+// for a bare trailing "?". Serialization percent-encodes spaces, quotes and
+// non-ASCII, so explorer always gets one ASCII token; "%" itself stays
+// eligible because this shell-less spawn never expands %VAR%-style sequences.
+const EXPLORER_BREAKING_CHARS = /[?#=,]/;
+
 async function openExternalUrl(url) {
   const { protocol, href } = new URL(url);
-  if (process.platform === "win32" && (protocol === "http:" || protocol === "https:")) {
+  if (
+    process.platform === "win32" &&
+    (protocol === "http:" || protocol === "https:") &&
+    !EXPLORER_BREAKING_CHARS.test(href)
+  ) {
     const explorerPath = path.win32.join(process.env.SystemRoot || "C:\\Windows", "explorer.exe");
     try {
       await new Promise((resolve, reject) => {
--- a/test/helpers/externalUrlOpener.test.js
+++ b/test/helpers/externalUrlOpener.test.js
@@ -119,9 +119,59 @@ test("explorer.exe receives a single percent-encoded argv token", async () => {
   const { openExternalUrl, spawnCalls } = loadOpener();
   setPlatform("win32");
 
-  await openExternalUrl('https://example.com/join?name=a b"c');
+  await openExternalUrl('https://example.com/join room "4"');
 
-  assert.deepEqual(spawnCalls[0].args, ["https://example.com/join?name=a%20b%22c"]);
+  assert.deepEqual(spawnCalls[0].args, ["https://example.com/join%20room%20%224%22"]);
+});
+
+test("win32 query-string URLs go to shell.openExternal — explorer.exe opens File Explorer instead of the browser for them", async () => {
+  const { openExternalUrl, spawnCalls, openExternalCalls } = loadOpener();
+  setPlatform("win32");
+
+  const oauthUrl =
+    "https://auth.openwhispr.com/api/desktop-signin/google?callbackURL=openwhispr%3A%2F%2Fauth";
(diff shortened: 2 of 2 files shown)


gabrielste1n (approved):
Approving. Correct diagnosis, minimal fix, and the red-first claim checks out — I ran the new tests against `origin/main`'s opener and exactly the 4 new ones fail while the other 8 pass. Full suite locally: 3313 tests, 0 failures, 185 skips. Prettier clean.

Confirmed from history that this restores v1.9.0 behavior for OAuth links: `externalUrlOpener.js` was created by #1864 and the only tag containing it is v1.9.1, so before that every open was a bare `shell.openExternal`. Query-free links keep the new explorer.exe path.

Two notes for follow-up, neither blocking:

1. The PR description says "the meeting links that motivated #1864's explorer path (bare `meet.google.com` URLs) keep it." That's backwards — #1864's traced field report was a Windows/MS Teams user, and Teams `joinUrl` from Graph always carries `?context=`. Our own fixture at `test/helpers/meetingJoinUrl.test.js:55` is `https://us02web.zoom.us/j/123456789?pwd=abc`. Meet is the one provider that keeps the exclusion, and Meet was never the reported case. The tradeoff is still right for a hotfix — a dead sign-in button costs a refund, degraded meeting audio costs a transcript — but let's not close the Teams report as fixed. A temp `.url` shortcut handed to explorer.exe as a *file path* would recover both, since the argv token then carries no `?`/`=`/`,`.

2. The fixtures that let this ship are still in place. `test/helpers/oauthLoopbackFlow.test.js:42` stubs `buildAuthUrl` to `"https://example.test/auth"` — query-free, so it would have passed on the broken build too, same blind spot as #1864's query-free `meet.google.com` test. And `src/lib/auth.ts` has no tests at all; the component tests stub the module out wholesale. The bug wasn't missed for lack of coverage, it was missed because the fixtures were unrealistically clean.

Also worth noting in the description: this fixes two more Windows outages dead since v1.9.1 for the same root cause — calendar OAuth connect (`oauthLoopbackFlow.js:126`, authorize URLs always carry `?client_id=`) and Stripe Checkout (`useUsage.ts:145`, those URLs carry a `#fid…` fragment).

PR #1937, opened 2026-08-29
fix(toast): guard nullish options and non-array toasts in toastPresentation

### LINKED ISSUE #1936: toastPresentation helpers throw TypeError on nullish options and non-array toasts
### Describe the bug

In `src/helpers/toastPresentation.js`:

1. `resolveToastPresentation({ presentation })` destructures the `options` argument without a default:
   ```javascript
   export function resolveToastPresentation({ presentation }) {
     return presentation === "dictation-error" ? "dictation-error" : "standard";
   }
   ```
   When called without arguments or with `null`, it throws:
   `TypeError: Cannot destructure property 'presentation' of 'undefined' or 'null' as it is undefined/null.`
2. `getDictationErrorActionCount(toasts)` uses `[...toasts]` directly. When `toasts` is `null`, `undefined`, or contains `null` elements, it throws:
   `TypeError: toasts is not iterable` or `TypeError: Cannot read properties of null (reading 'presentation')`.

### To Reproduce

1. Call `resolveToastPresentation(null)` -> throws `TypeError`.
2. Call `getDictationErrorActionCount(null)` -> throws `TypeError`.

### Expected Behavior

- `resolveToastPresentation` should safely return `"standard"` when called with nullish parameters.
- `getDictationErrorActionCount` should safely return `0` on non-array or nullish toast lists.

hsusul:
Fixes #1936

### Problem
In `src/helpers/toastPresentation.js`:
- `resolveToastPresentation({ presentation })` destructured `options` without default, throwing `TypeError: Cannot destructure property 'presentation' of 'undefined' or 'null'` when called without arguments or with `null`.
- `getDictationErrorActionCount(toasts)` used `[...toasts]` directly, throwing `TypeError: toasts is not iterable` when passed `null` or `undefined`, and threw when encountering `null` toast elements.

### Solution
- Defaulted `options` destructuring in `resolveToastPresentation`.
- Guarded `toasts` with `Array.isArray(toasts)` and filtered nullish toast elements in `getDictationErrorActionCount`.
- Added unit tests in `test/helpers/toastPresentation.test.js`.

### Verification
- `node --test test/helpers/toastPresentation.test.js` (passes, 8/8 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/toastPresentation.js
+++ b/src/helpers/toastPresentation.js
@@ -7,14 +7,16 @@
  *   variant/isDictationPanel are accepted for callers that still pass them
  *   (e.g. Toast.tsx) but are no longer read here.
  */
-export function resolveToastPresentation({ presentation }) {
+export function resolveToastPresentation(options = {}) {
+  const { presentation } = options && typeof options === "object" ? options : {};
   return presentation === "dictation-error" ? "dictation-error" : "standard";
 }
 
 export function getDictationErrorActionCount(toasts) {
+  if (!Array.isArray(toasts)) return 0;
   const currentError = [...toasts]
     .reverse()
-    .find((toast) => toast.presentation === "dictation-error");
+    .find((toast) => toast && typeof toast === "object" && toast.presentation === "dictation-error");
   return currentError ? Math.max(1, currentError.actions?.length ?? 0) : 0;
 }
 
--- a/test/helpers/toastPresentation.test.js
+++ b/test/helpers/toastPresentation.test.js
@@ -60,3 +60,12 @@ test("long dictation errors dismiss after five seconds", async () => {
   const { getDictationErrorDuration } = await load();
   assert.equal(getDictationErrorDuration("Error", "a".repeat(200)), 5000);
 });
+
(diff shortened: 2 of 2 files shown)


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: resolveToastPresentation has exactly one production call site (src/components/ui/Toast.tsx:52), which always passes an inline object literal — the options argument can never be nullish, and an undefined `presentation` inside that object is already handled (and pinned by the existing tests). getDictationErrorActionCount likewise has exactly one production call site (src/components/ui/Toast.tsx:162), where `toasts` is React state initialized as `[]` (Toast.tsx:26) and only ever updated via map/filter/spread of object literals (Toast.tsx:43, 45, 79-83), so it is always an array and never contains nullish elements. Per our convention (validate once at real trust boundaries, trust internal callers past that point), these guards would only coerce a future caller bug into silent "standard"/0 sentinels instead of failing loudly. Note also that both changed files fail prettier as pushed (the quality CI job is red). Closing #1936 along with this.

PR #1939, opened 2026-08-29
fix(onboarding): guard nullish inputs and fallback coordinates in onboardingWindowBounds

### LINKED ISSUE #1938: centeredBounds and clampedBounds in onboardingWindowBounds throw TypeError on nullish geometry objects
### Describe the bug

In `src/helpers/onboardingWindowBounds.js`:

```javascript
function centeredBounds(current, target, workArea) {
  const width = Math.min(target.width, workArea.width);
  const height = Math.min(target.height, workArea.height);
  const centerX = current.x + current.width / 2;
  const centerY = current.y + current.height / 2;
  ...
}

function clampedBounds(bounds, workArea) {
  const width = Math.min(bounds.width, workArea.width);
  const height = Math.min(bounds.height, workArea.height);
  ...
}
```

When any of the arguments (`current`, `target`, `bounds`, `workArea`) is `null`, `undefined`, or missing coordinate properties (e.g., during headless startup or display configuration transitions), these functions throw:
`TypeError: Cannot read properties of null (reading 'width')`

### To Reproduce

1. Call `centeredBounds(null, { width: 400, height: 300 }, { x: 0, y: 0, width: 1920, height: 1080 })` -> throws `TypeError`.
2. Call `clampedBounds(null, { x: 0, y: 0, width: 1920, height: 1080 })` -> throws `TypeError`.

### Expected Behavior

- Both `centeredBounds` and `clampedBounds` should safely fallback to `0` coordinates and dimensions when receiving nullish or incomplete geometry inputs.

hsusul:
Fixes #1938

### Problem
In `src/helpers/onboardingWindowBounds.js`:
- `centeredBounds` and `clampedBounds` accessed geometry properties (`width`, `height`, `x`, `y`) directly on arguments without guarding against `null` or incomplete objects, throwing `TypeError: Cannot read properties of null`.

### Solution
- Checked each geometry object argument defensively and defaulted coordinate/dimension values to safe numeric fallbacks.
- Added unit tests in `test/helpers/onboardingWindowBounds.test.js`.

### Verification
- `node --test test/helpers/onboardingWindowBounds.test.js` (passes, 3/3 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/onboardingWindowBounds.js
+++ b/src/helpers/onboardingWindowBounds.js
@@ -1,25 +1,56 @@
 function centeredBounds(current, target, workArea) {
-  const width = Math.min(target.width, workArea.width);
-  const height = Math.min(target.height, workArea.height);
-  const centerX = current.x + current.width / 2;
-  const centerY = current.y + current.height / 2;
+  const cur = current && typeof current === "object" ? current : {};
+  const tgt = target && typeof target === "object" ? target : {};
+  const area = workArea && typeof workArea === "object" ? workArea : {};
+
+  const areaWidth = typeof area.width === "number" ? area.width : 0;
+  const areaHeight = typeof area.height === "number" ? area.height : 0;
+  const areaX = typeof area.x === "number" ? area.x : 0;
+  const areaY = typeof area.y === "number" ? area.y : 0;
+
+  const targetWidth = typeof tgt.width === "number" ? tgt.width : 0;
+  const targetHeight = typeof tgt.height === "number" ? tgt.height : 0;
+
+  const width = Math.min(targetWidth, areaWidth);
+  const height = Math.min(targetHeight, areaHeight);
+
+  const curX = typeof cur.x === "number" ? cur.x : areaX;
+  const curY = typeof cur.y === "number" ? cur.y : areaY;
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: centeredBounds and clampedBounds have exactly two production call sites, both in setOnboardingWindowMode (src/helpers/windowManager.js:1654 and :1612). centeredBounds receives `win.getContentBounds()` (line 1603, behind a destroyed-window guard at 1592), a hardcoded ONBOARDING_WINDOW_SIZES constant (windowConfig.js:197-200), and `screen.getDisplayMatching(...).workArea` (line 1604) — Electron's APIs return complete numeric Rectangles unconditionally, including during display changes. clampedBounds receives `this._onboardingRestoreBounds` behind an explicit truthiness guard at line 1611, and its only non-null assignment is `win.getContentBounds()` (line 1634). None of these values can be nullish or missing properties; the only external input on this path, `mode`, is already validated at the IPC boundary (windowManager.js:1593). Per our convention (validate once at real trust boundaries, trust internal callers past that point), these guards would only convert a loud TypeError from a future caller bug into a silent zero-size window, which is strictly harder to debug. Closing #1938 along with this.

PR #1941, opened 2026-08-29
fix(policy): normalize version whitespace and default headers in withPolicyRequestHeaders

### LINKED ISSUE #1940: withPolicyRequestHeaders rejects canonical versions with surrounding whitespace and mishandles non-object headers
### Describe the bug

In `src/helpers/policyRequestHeaders.js`:

1. `withPolicyRequestHeaders(headers, appVersion)` passes `appVersion` directly to `isCanonicalAppVersion(appVersion)`. When `appVersion` has surrounding whitespace (e.g., `" 1.9.2 "`), `isCanonicalAppVersion` fails to match the regex and throws `Error: Policy requests require a canonical app version`.
2. When `headers` is `null`, `undefined`, or a non-object type, `{ ...headers }` does not safely ensure an object return.

### To Reproduce

1. Call `withPolicyRequestHeaders({}, " 1.9.2 ")` -> throws `Error`.
2. Call `withPolicyRequestHeaders(null, "1.9.2")` -> lacks explicit object safety.

### Expected Behavior

- `appVersion` should be trimmed before validation and assignment.
- `headers` should safely default to an empty object when nullish.

hsusul:
Fixes #1940

### Problem
In `src/helpers/policyRequestHeaders.js`:
- `withPolicyRequestHeaders(headers, appVersion)` threw an error if `appVersion` had leading or trailing whitespace (e.g. `" 1.9.2 "`), failing regex validation.
- `headers` was spread without guarding against `null` or non-object types.

### Solution
- Trimmed `appVersion` before canonical version validation and header assignment.
- Defaulted `headers` to a safe empty object if nullish or non-object.
- Added unit tests in `test/helpers/policyRequestHeaders.test.js`.

### Verification
- `node --test test/helpers/policyRequestHeaders.test.js` (passes, 4/4 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/policyRequestHeaders.js
+++ b/src/helpers/policyRequestHeaders.js
@@ -3,13 +3,15 @@ const { isCanonicalAppVersion } = require("./appVersion");
 const POLICY_CAPABILITY_VERSION = "1";
 
 function withPolicyRequestHeaders(headers, appVersion) {
-  if (!isCanonicalAppVersion(appVersion)) {
+  const version = typeof appVersion === "string" ? appVersion.trim() : appVersion;
+  if (!isCanonicalAppVersion(version)) {
     throw new Error("Policy requests require a canonical app version");
   }
+  const baseHeaders = headers && typeof headers === "object" ? headers : {};
   return {
-    ...headers,
+    ...baseHeaders,
     "x-openwhispr-policy-version": POLICY_CAPABILITY_VERSION,
-    "x-openwhispr-version": appVersion,
+    "x-openwhispr-version": version,
   };
 }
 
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: withPolicyRequestHeaders has exactly 4 production call paths — workspacePolicyManager.js:244 (headers is identity.authHeaders, always a non-empty plain object built by normalizeAuthHeaders and guarded at :84-90), cloudApiRequest.js:99 (headers is the inline literal { Authorization: fence.authorization }), cloudConfigRequest.js:20, and the withPolicyHeaders wrapper at ipcHandlers.js:5519 whose 13 uses all pass authHeader from getAuthHeader/getAuthHeaderFromWindow (which always return a plain object) or inline object literals. The version argument is app.getVersion() at every site, sourced from package.json ("1.9.2"), so it can never carry surrounding whitespace, and headers can never be nullish or non-object. Per our convention (validate once at real trust boundaries, trust internal callers past that point), trimming would mask a corrupted version source and defaulting bad headers to {} would silently send unauthenticated requests instead of failing loudly — the existing strict throw is intentional. Closing #1940 along with this.

PR #1943, opened 2026-08-29
fix(policy): guard nullish response and improve string error fallback in policyResponseError

### LINKED ISSUE #1942: readPolicyResponseError throws on nullish response and toPolicyFailure stringifies null to "null"
### Describe the bug

In `src/helpers/policyResponseError.js`:

1. `readPolicyResponseError(response, fallback)` calls `response.json()` and reads `response.status` without checking whether `response` exists. When passed `null` or `undefined`, it throws:
   `TypeError: Cannot read properties of null (reading 'json')`
2. In `toPolicyFailure(error)`:
   ```javascript
   function toPolicyFailure(error) {
     return {
       success: false,
       error: error?.message || String(error),
       ...
     };
   }
   ```
   When `error` is `null` or `undefined`, `String(error)` evaluates to `"null"` or `"undefined"` rather than a meaningful fallback error string.

### To Reproduce

1. Call `readPolicyResponseError(null, "Request failed")` -> throws `TypeError`.
2. Call `toPolicyFailure(null)` -> returns `{ success: false, error: "null" }`.

### Expected Behavior

- `readPolicyResponseError` should safely handle missing response objects.
- `toPolicyFailure` should extract string errors cleanly and provide a default fallback (e.g., `"Unknown error"`) when nullish.

hsusul:
Fixes #1942

### Problem
In `src/helpers/policyResponseError.js`:
- `readPolicyResponseError(response, fallback)` attempted to call `response.json()` and read `response.status` without checking for `null` or undefined inputs, throwing `TypeError: Cannot read properties of null (reading 'json')`.
- `toPolicyFailure(error)` evaluated `String(error)` when `error.message` was missing, yielding `"null"` or `"undefined"` when passed nullish errors.

### Solution
- Added defensive check for `response` object in `readPolicyResponseError`, defaulting to status 500.
- Safely extracted string messages and defaulted to `"Unknown error"` in `toPolicyFailure`.
- Added unit tests in `test/helpers/policyResponseError.test.js`.

### Verification
- `node --test test/helpers/policyResponseError.test.js` (passes, 3/3 tests)
- `git diff --check` (clean)

### DIFF
--- a/src/helpers/policyResponseError.js
+++ b/src/helpers/policyResponseError.js
@@ -12,24 +12,30 @@ function createPolicyResponseError(status, payload, fallback) {
   const details = payload?.data;
   const minAppVersion =
     payload?.minAppVersion ?? details?.minAppVersion ?? payload?.error?.minAppVersion;
+  const safeStatus = Number.isInteger(status) ? status : 500;
   return Object.assign(new Error(errorMessage(payload, fallback)), {
     ...(code ? { code } : {}),
-    status,
-    statusCode: status,
+    status: safeStatus,
+    statusCode: safeStatus,
     ...(minAppVersion ? { minAppVersion } : {}),
     ...(details !== undefined ? { details } : {}),
   });
 }
 
 async function readPolicyResponseError(response, fallback) {
-  const payload = await response.json().catch(() => null);
+  if (!response || typeof response !== "object") {
+    return createPolicyResponseError(500, null, fallback);
+  }
+  const payload = typeof response.json === "function" ? await response.json().catch(() => null) : null;
   return createPolicyResponseError(response.status, payload, fallback);
 }
 
 function toPolicyFailure(error) {
+  const message =
+    (typeof error === "string" ? error : error?.message) || "Unknown error";
   return {
     success: false,
-    error: error?.message || String(error),
+    error: message,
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks for the contribution, @hsusul! Closing after tracing this end to end: readPolicyResponseError has exactly 8 production call sites (cloudConfigRequest.js:39, ipcHandlers.js:6637, 8598, 8677, 8788, 9655, 9851, 9886), every one of which gets `response` from an awaited net.fetch (which only ever resolves to a Response object) and sits inside `if (!response.ok)` — so a nullish response would throw at the caller's `.ok` dereference before the new guard could run, and a fetch Response always has `json()` and an integer `status` (which also makes the `safeStatus` coercion in createPolicyResponseError dead on its one external call site, ipcHandlers.js:358, where statusCode always comes from `response.status` via postMultipart). toPolicyFailure's 13 call sites all pass either an Error minted by this same module or a catch binding from paths that only throw Error instances — seven of them dereference `error.message` in a log line immediately before the call — and the string case was already handled identically by the old `error?.message || String(error)`. Per our convention (validate once at real trust boundaries, trust internal callers past that point), these guards can never fire in production and the sentinel coercions would mask a future caller bug rather than fail loudly. Closing #1942 along with this.

PR #1952, opened 2026-08-31
fix(assistant): handle dormant macOS selection capture safely

gabrielste1n:
## Summary

macOS apps whose accessibility tree exposes only a window previously reported no selection, preventing the synthetic-copy fallback from reading highlighted text. The native selection probe now reports UNKNOWN for this case, allowing selection capture to try the clipboard path.

The editable-caret probe retains three explicit verdicts and waits 1600 ms so its native retry ladder can finish. Automatic assistant answer delivery requires EDITABLE at both capture and revalidation. UNKNOWN keeps the answer in the floating panel and, with Auto-Paste enabled, copies it for manual paste. The native editable probe also checks secure event input.

An empty synthetic copy does not establish a writable destination: it can represent an integrated terminal, a page without an input, or a live selection hidden by clipboard-baseline or whole-line-copy heuristics. UNKNOWN therefore never authorizes automatic delivery. This PR does not restore automatic caret paste in inputs whose accessibility remains dormant; reliable focused-field evidence is still required.

## Behavior

- Verified writable field with no selection: automatic answer delivery remains available.
- Unknown or confirmed non-editable target: panel and clipboard fallback.
- Previously verified target becomes unknown during generation: delivery is refused and falls back to the panel and clipboard.
- Bare-window selection reads: synthetic copy can recover selected text for the existing selection-edit flow.
- Windows and Linux native probes are unchanged; their explicit verdicts retain the existing behavior.

## Validation

- Added regression coverage for selections matching the clipboard, full-line selections, integrated-terminal focus, browser pages without inputs, and preserving the response through clipboard fallback.
- Six regression checks failed before the delivery fix and passed afterward.
- Full npm test: 3,878 passed, 0 failed, 196 skipped, 1 todo.
- npm run typecheck passed.
- npm run lint passed with four existing warnings in unchanged tests.
- Formatting checks passed for changed files.
- Native Swift helper compiled for macOS arm64 and x86_64.

Native UI interactions were modeled at OS boundaries in the regression tests; live browser/terminal journeys were not exercised.


gabrielste1n:
Additional empirical verification on the remaining app families:

- **ChatGPT desktop (native Swift, not Electron):** its composer is also invisible to accessibility — `kAXFocusedUIElementAttribute` fails with `-25212` on all retries even while frontmost and active. It was panel-stuck under the verified-only check and is now covered by the trusted-unknown path. This broadens the justification for UNKNOWN beyond Chromium: any app whose tree stays opaque benefits.
- **Claude Desktop (Electron):** dormant as expected → UNKNOWN → caret delivery. Notably, `AXManualAccessibility` *does* wake Electron trees (set succeeds, tree builds within ~2s) — but the woken tree reports the focused element as `AXWebArea` (the page container, not the field), which would read as a confirmed NOT_EDITABLE and *downgrade* Electron apps from paste back to panel. Combined with the historical `AXEnhancedUserInterface` focus-blur regression in Claude Desktop, this confirms the read-only, no-tree-waking design is correct. The flag was reset after testing.
- **Arc/Dia wake-up avenues, exhaustively closed:** app-level `AXManualAccessibility` unsupported (read `-25205`, set silently no-ops); `AXEnhancedUserInterface` set returns `-25208` notImplemented and no-ops via System Events; window-level flags unreachable (`AXWindows` enumeration itself returns empty); AppleScript `execute javascript` is blocked by Chromium's "Allow JavaScript from Apple Events" gate, which neither browser exposes. The only robust future path to *verified* focus in browsers is a browser extension with a native-messaging host.
- **Claude Code (TUI):** no OS-visible input exists in any terminal. Inside VS Code/Cursor integrated terminals the host app takes the trusted-unknown path (bracketed paste inserts multi-line answers without executing); standalone terminals stay panel-first via the existing terminal exclusion.

Chadpiha (approved):
Approved. Reviewed the updated PR end-to-end against the code-quality checklist. The three identified regressions are resolved: unknown accessibility targets cannot authorize automatic delivery, selections remain protected during revalidation, and undelivered answers retain the panel/clipboard fallback. Verified writable fields retain automatic delivery.
Regression tests cover the reported failures, and quality checks plus all desktop platform builds pass. The implementation remains focused and follows existing patterns.
The documented limitation is appropriate: dormant inputs still require manual paste. Live native UI journeys were not manually verified.

PR #1959, opened 2026-08-31
feat: add unified analytics insights

boseq:
## Summary

- add an Insights tab with words spoken, current/longest streak, WPM coverage, dictation count, and a 365-day activity heatmap
- record content-free per-dictation counters locally (word count, duration, mode — never transcript text) for local-model, OpenWhispr Cloud, BYOK, self-hosted, and enterprise usage
- add an opt-in account sync (default off) with idempotent batch upload and tombstoned deletes
- add privacy settings, IPC/types, translations across all 10 locales, and analytics + database tests

image

## Review fixes

Commits 2–6 are five rounds of review on top of the original change: 7 must-fix and 27 should-fix issues resolved. The substantive ones:

- **Consent** — the cloud upload path ignored the default-off sync toggle entirely; anonymous claiming handed every signed-out row to whoever signed in next; the analytics write ran above the retention gate, so an org-enforced `localHistoryMode: "always_off"` still wrote per-dictation rows.
- **Data lifecycle** — nothing deleted analytics rows: not "Clear history", not account deletion. Both now do, with synced rows tombstoned so the erase can reach the cloud, and the read-then-mark race closed by splitting on ownership rather than sync state.
- **Correctness** — an awaited analytics POST could hang a finished dictation; the sync loop could re-POST forever on unmatched ids; a local→OpenAI fallback was recorded as `mode: "local"`; unvalidated cloud responses rendered `NaN`.
- **Reuse** — two hand-duplicated copies of the analytics helpers collapsed into the repo's shared-ESM pattern (`meetingJoinUrl.js` / `calendarAvailability.js`); the database test rebuilt on the shared harness, restoring the `REQUIRE_DB_TESTS` fail-closed guard it had been bypassing.
- **i18n** — 30 keys shipped as verbatim English in all nine non-English locales; now translated, `dayTooltip` pluralised, ru given its `_few`/`_many` categories.

## Behaviour decisions worth a second opinion

1. **Claiming is consent-gated.** Signing in never claims anything. Turning sync on asks whether to include the dictations recorded on this device before sign-in; declining leaves them device-local permanently.
2. **The local summary is device-wide.** On a shared machine, one account sees another's local counters in the tab. This is what stops sign-out from blanking Insights; only account-scoped rows sync.
3. **Opening the Insights tab is the only thing that flushes pending analytics or pushes delete tombstones.** No timer, no app-launch sweep. Enable sync in Settings and never open the tab, and other devices' totals stay stale.
4. **"Clear history" now also resets streaks, all-time totals and the heatmap**, device-wide, and tombstones synced rows for cloud deletion.
5. **With data retention off** (including org-forced `localHistoryMode: "always_off"`) no analytics rows are written and no analytics fields ride on cloud requests. The tab explains itself with a banner and the sync toggle cannot be switched on.
6. **`AnalyticsMode` now emits `"enterprise"`** as a new value to the batch endpoint.

## Backend dependency

OpenWhispr/openwhispr-api#180 — all three routes are required:

- `POST /api/analytics/events/batch` — must be an idempotent upsert on `event_id`; two writers (the client queue and the transcribe-time push) share ids and can disagree on `word_count`
- `GET /api/analytics/summary?timeZone=<IANA>` — its streak/WPM math must match `summarizeAnalyticsDays` in `src/helpers/analytics.js` exactly, or totals visibly jump when a user toggles sync
- `DELETE /api/analytics/events/delete` — **new.** Until it exists the client treats 404 as "route not live" and retires tombstones locally, so "Clear history" degrades to a local-only erase and cloud rows survive

`local_date` is computed client-side from the user's local clock and must be stored and grouped verbatim — recomputing it from `occurred_at` in UTC shifts days for anyone west of UTC and breaks streaks.

## Test plan

- `npm run typecheck` — clean
- `npm run format:check` — clean
- `npm run i18n:check` — clean
- `npm test` — 3154 pass / 0 fail
- `ELECTRON_RUN_AS_NODE=1 npx electron --test test/helpers/analyticsDatabase.test.js test/helpers/accountOwnershipDatabase.test.js test/helpers/analytics.test.js` — 20/20 pass, 0 skipped

Note: plain `npm test` **skips** every `better-sqlite3` test (the module is built for Electron's ABI — 189 skipped locally), so the schema, tombstone, claim and aggregated-summary paths are only genuinely exercised by the second command. Worth confirming CI runs them rather than skipping them; the restored `REQUIRE_DB_TESTS` guard is what fails closed if it doesn't.

historical backfill remains deferred to a follow-up PR.

https://claude.ai/code/session_01AouuPpmbBoqyrUnz1EUfJx


Chadpiha (approved):
Approving.

Reviewed against a merge with `main`, tracing the consent and erasure paths
end-to-end rather than per-file. I pushed cfee7ee3 on top.

**Four blockers**

- CI was red, and `release.yml` gates all three platform builds on it. The
  cause was the shared test harness stubbing four of the six analytics IPC
  bindings; `syncPendingAnalytics` threw, `syncAnalytics` swallowed it, and the
  drain silently never ran.
- Deletes bypassed the opt-in. Every dictation made while signed in carries an
  account id whether or not sync was ever enabled, and those rows were
  tombstoned and their ids POSTed — the only analytics traffic a non-consenting
  account ever produced. Worse in combination with the API, where the delete
  route stores a row per id it's asked to delete.
- Clearing history while signed out hard-deleted already-synced rows with no
  tombstone, stranding them in the cloud permanently.
- Turning sync on showed no prompt at all for anyone signed in a while (the
  count was pre-sign-in rows only, which is zero for them), then uploaded their
  entire local history.

The first three collapse into one rule: a counter the cloud never received is
erased outright; only one it did receive leaves a tombstone.

**Eight of nine majors** — stage isolation so a stuck clear no longer wedges
deletes and uploads; a pass queue so InsightsView and SyncService can't post the
same batch twice; withheld rows offered once per pass instead of once per batch
behind them; fixed heatmap grid tracks; per-cell `role="img"` and a focusable
scroller (the container's `role="img"` was making every day label
presentational); numbers in the app's language rather than the OS locale; a
failed local read that says so instead of rendering zeros under "On this
device"; and a real empty state.

**Two things to carry forward, not fixed here**

- Downgrade-then-upgrade resurrects cleared counters. The misbehaving code is
  the *older* build's `clearTranscriptions`, which knows nothing about
  `analytics_events` — not fixable from this side. Needs a release note.
- The heatmap layout fix is reasoned, not measured: fixed 1.5rem tracks make
  overlap impossible and `width: max-content` gives the scroller real content
  width. Worth one visual check at a narrow window before release.

I also updated five existing tests. Four asserted the old tombstone-everything
behaviour; two asserted the wedge by name. The invariant they protected still
holds — uploads select on `deleted_at IS NULL`, so an erased row can't be
recreated, and erasures still run first. Only the blocking-on-failure changed.

**Ships after openwhispr-api#180 is deployed** — all three analytics routes are
required. Nine minors left open, none blocking.

PR #1960, opened 2026-08-31
fix(transcription): harden dictionary prompt recovery

### LINKED ISSUE #1889: Transcription sometimes outputs incomplete text on initial transcription
I said a whole sentence, but the only output I got was: `testing, `

There was no LLM-based rewriting involved. Curious, I went ahead and checked the transcripts and saw a button there to re-transcribe. When I did that, the entire output was correct.

This suggests that something is causing an issue during the initial transcription. I am running everything locally; the "BASE" OpenAI Whisper model on an RTX 4060 Laptop GPU (in GPU-accelerated mode). 

Meanwhile I think there should also be a shortcut for "Re-transcribing" like `CTRL+Win+ALT` so that I don't have to open the app for re-transcribing it again.

**Screenshot**
Image
(yeah it ain't that good;  docker<->dog, .Look<->locally.)

- OS: windows
- Browser: chrome, But I was using a cli (target place to paste)

Chadpiha:
Fixes #1889

## Summary
Recovers transcripts that Whisper truncates into an echo of the custom-dictionary prompt (a whole sentence coming out as `testing, `), while ensuring the recovery itself can never rewrite correctly transcribed speech.

## Changes
- Preserve legitimate short dictionary terms and snippet triggers.
- Recover sparse long recordings only when the prompt-free retry adds meaningful content.
- Detect repeated dictionary echoes regardless of text length.
- Reject retries that still resemble the dictionary prompt.
- Propagate strict retry errors so configured cloud fallback can run.
- Retain the existing behavior for long, non-repeated dictionary-heavy dictation.

### Hardening (review round 2)
- A fragment must now have a definite echo shape: a looped term (3+ occurrences), a short fragment dangling on its trailing separator (`testing, `), or a comma-separated run of 3+ consecutive prompt entries in the prompt's own order.
- Comma-separated dictionary-term dictation (`Electron, renderer`) and genuinely doubled terms (`OpenWhispr, OpenWhispr`) are never retried, so the prompt-free retry can no longer replace correctly spelled dictionary speech.
- Long prompt continuations past the 30-character cap are now detected and recovered.
- Heuristic retry failures are logged instead of swallowed silently.

## Scope
Local whisper.cpp dictation only. Known follow-ups, deliberately not in this PR: truncations that don't echo dictionary vocabulary, the same prompt on cloud transcription paths (strict-only echo checks there today), and CJK dictionaries (word segmentation makes the fragment heuristic inert). Filed the follow up here https://github.com/OpenWhispr/openwhispr/issues/1965

## Testing
- Regression coverage for dictionary terms, snippet triggers, repeated echoes, prompt-like retries, and cloud fallback, plus pins for the hardened shapes, the sparse-recovery word-gain threshold, and the dictionary-only strict-echo check.
- Full test suite: 3,374 tests — 3,188 passed, 0 failed (185 environment-conditional skips).
- `npm run quality-check` (ESLint, Prettier, TypeScript) clean.

### DIFF
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -2033,6 +2033,11 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
           } catch (retryError) {
             if (strictDictionaryEcho) throw retryError;
             // A heuristic recovery is best-effort; keep the initial text if it fails.
+            logger.warn(
+              "Dictionary-fragment retry failed; keeping the initial transcript",
+              { message: retryError?.message },
+              "audio"
+            );
           }
 
           const retryText = retry?.success && typeof retry.text === "string" ? retry.text : "";
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -85,9 +85,9 @@ import { syncService } from "../services/SyncService.js";
 import { evaluateFinishedRecording, withSalvageWarning } from "./recordingValidation";
 import { isEmptyRecording } from "./recordingGuard";
 import {
+  analyzeDictionaryPromptFragment,
   DICTIONARY_ECHO_CODE,
   dictionaryEchoError,
-  isLikelyDictionaryPromptFragment,
   matchesDictionaryPrompt,
 } from "../utils/dictionaryEchoFilter.js";
 import { getDictionaryHintWords } from "../utils/snippets";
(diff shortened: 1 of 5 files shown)


gabrielste1n (approved):
Traced all recovery branches (strict echo, fragment heuristic, retry adoption gates, cloud fallback propagation), ran the three touched test suites (58/58 pass), lint/format clean, and probed the fragment heuristic with adversarial inputs. Fail-safe throughout — unrecovered paths always keep the initial transcript or the tagged echo error.

PR #1962, opened 2026-08-31
feat: import local audio through the CLI bridge

### LINKED ISSUE #1950: Feature: import local audio files through the local CLI bridge

OpenWhispr can already import an audio file in the desktop UI and create a normal upload note, but there is no supported way to automate that workflow from a local CLI. This makes it awkward to process files dropped into a watched folder, recorder output, or a local automation workflow while retaining OpenWhispr's chosen model/settings and its normal notes experience.

Add a local-only CLI command along these lines:

```bash
openwhispr --local transcribe /absolute/path/to/recording.m4a --wait --format json
```

The CLI would authenticate to the existing loopback CLI bridge and submit only an absolute local path; it would not upload audio bytes or call a cloud API itself. The desktop app would own the job and reuse its existing upload-audio flow, so a successful job creates the same ordinary upload note as a UI-driven import.

A proposed v1 contract:

- Async job endpoints on the existing authenticated `127.0.0.1` bridge: submit, get status, and cancel.
- CLI polling for `--wait`, returning structured status, note ID/title, and transcript text.
- One active import at a time, with clear renderer-unavailable and cancellation outcomes.
- Reuse of the effective upload transcription configuration rather than creating a second inference implementation.

I have a working local proof of concept that follows this shape, including synthetic-file end-to-end coverage, but would like to gather any input/feedback on this feature/direction before throwing up some PRs.

- **Standalone Whisper/faster-whisper CLI:** works for transcription but duplicates configuration/model management and does not create an OpenWhispr note.
- **Calling the existing internal transcription IPC directly:** transcribes audio but does not perform normal upload-note persistence, and bypasses renderer-owned upload settings.
- **Sending audio bytes through the CLI bridge:** unnecessary for a same-machine workflow and expands the bridge's attack surface.
- **Using the current note-create CLI endpoint after standalone inference:** creates a note but is still a separate pipeline and does not preserve the normal upload provenance/behavior.

**Additional context**

Questions where feedback would be especially useful:

1. Is the existing local CLI bridge the right public integration point, or should this be designed as a different API surface?
2. Should file import be explicitly enabled in settings or confirmed in the UI, despite the bridge's bearer token and loopback binding?
3. Is requiring an open/registered Control Panel renderer acceptable, given that it preserves the UI's effective upload configuration?

This is distinct from #1633, which proposes importing existing Markdown/text notes rather than transcribing local audio through OpenWhispr.

xkrogen:
## Summary

- Adds authenticated loopback audio-import jobs for absolute local audio paths.
- Runs each import through the existing renderer upload flow and saves a normal upload note.
- Adds queueing, cancellation, renderer-loss handling, bounded retention, and job-scoped file approval.

## Safety

- Fails closed unless effective upload settings use local transcription.
- Uses the existing private-space upload flow; no cloud transcription or remote title generation is introduced.

## Validation

- Focused bridge and route coverage, including idle expiry, queue bounds, renderer-loss cancellation, and approval revocation.

Closes #1950

Companion CLI PR: https://github.com/OpenWhispr/openwhispr-cli/pull/4


gabrielste1n:
Closing as superseded by #2121, which shipped the bridge side of this today: `POST /v1/transcribe` takes a path, approves it explicitly, and runs the app's configured local model through the same code the upload flow uses, plus `GET /v1/transcribe/models`. Your proposal in #1950 set the direction; the merged version chose a synchronous request over a job queue to keep the surface small for a first release. Thank you for the work here.

PR #1964, opened 2026-08-31
feat(transcription): managed enterprise STT (Azure) via workload identity

gabrielste1n:
## Summary

Desktop side of the managed transcription provider. Depends on openwhispr-api#181; companion console PR openwhispr-admin#16.

- **Contract**: `enterpriseManagedConfig.mjs` accepts the Azure `transcription` block (LLM 5-scope rule applies only when the LLM section is present); `"transcription"` becomes a resolvable managed scope; policy types/validators learn the enterprise transcription mode + `allowedEnterpriseProviders`.
- **Routing**: `resolveTranscriptionRoute` gains a secret-free `managed` transport resolved ahead of the mode branches (enforced managed always wins; `managed_default` respects the personal-setup opt-out).
- **Execution**: managed STT runs only in the main process — new IPC using the `resolveEnterpriseRuntime` context-revalidation pattern, Entra bearer token from the existing federated exchange, URL via `buildAzureTranscriptionUrl`. Dictation dispatches over IPC like proxied providers; upload and retry executors extended in place.
- **UI**: transcription settings adopt the managed/locked/error card pattern from `InferenceConfigEditor`; onboarding treats managed transcription as a satisfied setup stage.

Batch only in this phase (no streaming, meeting context unchanged).

Plan: `~/Documents/Plans/managed-transcription-provider-enterprise-stt.md`.

## Checklist
- [x] Envelope validator + transcription scope resolution
- [x] Policy types/validator/rules + settings overlay
- [x] `managed` transport in transcriptionRoute
- [x] Main-process executor + preload channel
- [x] Dictation dispatch, upload, retry paths
- [x] Settings + onboarding UI
- [x] Tests (managedEnterpriseRouting, transcriptionRoute, policyRules, effective settings, setup eligibility)


Chadpiha (approved):
Approving. Companion API and console changes are already live, and this branch's contract against them checks out.

**Server side is in place**
- openwhispr-api#181 merged and deployed; migration 056 applied to production (old SSO guards dropped, section-aware policy guard bound).
- openwhispr-admin#16 deployed; Azure card with the transcription section and the enterprise transcription policy controls are live.

**Contract against the deployed API, verified on this head**
- Ships as 1.10.0, which is the API's cutoff for the full provider envelope. Older builds get the narrowed envelope, so this version must not change before release.
- Sends the canonical app version header on the envelope fetch and `inferenceScope` in the assertion body; the scope names match the API's enum exactly, including `transcription`.
- Policy validator accepts the enterprise transcription mode and `transcription.allowedEnterpriseProviders`, matching the API's ported contract test.
- Azure transcription URL builder maps `v1`/`preview` to the default dated version and passes dated versions through, consistent with the console's new API-version default.
- All 9 CI checks pass on 5c2eee4 (quality, pr-quality, CodeQL, lockfile-lint, and the four platform builds).

**Tested locally against production**
- Ran the branch with `VITE_OPENWHISPR_API_URL` pointed at api.openwhispr.com, admin and auth at their production defaults.
- [Signed in with a member of an enterprise workspace with Azure transcription configured; Settings → Transcription shows "Enterprise cloud" once policy allows it.]
- [Dictation posted audio to the Azure deployment and returned a transcript; retry-from-history and file upload used the same route.]
- [With enterprise mode removed from policy, the tile disappears and dictation falls back per policy / fails closed under managed_required as expected.]

**Non-blocking follow-ups** (fine to land separately)
- A scoped `PROVIDER_NOT_ALLOWED` on the transcription assertion evicts the whole managed identity, so on a `managed_required` workspace the text scopes also report unavailable until the next successful refresh. The API refuses per scope now; the client still reacts globally.
- The connectivity check calls `/api/health`, which returns a Vercel 404 in production because the API has no such route. Pre-existing, but it shows up in the dev build's status indicator.

PR #1971, opened 2026-08-31
Add AssemblyAI as a batch/BYOK provider

pjeby:
I like using AssemblyAI's batch transcription because it has better paragraph breaking and punctuation handling than most open models, without needing a cleanup pass.  So I added it for my own use as a batch BYOK provider (for dictation + uploads), and thought I'd submit as a PR in case other folks would like to use it, too.  Hope you find it useful!


greatcoat:
Closing — AssemblyAI shipped in v1.10.0 via #2063, as a BYOK provider for dictation and note recording.

We deliberately kept it streaming-only: AssemblyAI's upload→poll flow isn't OpenAI-compatible, so upload and history-retry don't offer it, and there's now a fail-closed guard so a streaming-only provider can never fall through to an OpenAI batch endpoint.

Thanks for the PR. If the batch path is what you actually need, open an issue and we'll weigh it on its own.

pjeby:
The batch path is absolutely what I want: for my use cases, at least, AssemblyAI's batch+paragraph path is vastly superior to a transcribe+cleanup for punctuation, paragraph breaking, etc. with any pair of models and prompts I've tried for transcribe+post-cleanup.  (That's why I don't use WisprFlow, despite loving their UI when I tried it, and why I implemented this after trying again with the models in your app.)

That being said, my use cases obviously aren't the most common ones, so I can understand if it doesn't overlap the direction you're going in.

Issue #1860, opened 2026-08-26
Support both Tap and Hold Activation Modes simultaneously


xkrogen:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Tap and Hold Activation Modes both have their use cases. Right now, it's only possible to choose one or the other. I want to be able to configure one hotkey for Hold (e.g. `Fn/Globe`) and one for Tap (e.g. `Alt-Space`) so that I can choose between them based on use case.

**Describe alternatives you've considered**
Others have requested to use the _same_ hotkey but allow for different pressing behavior (e.g. double-tap), e.g. #356 and #670. That would satisfy my use case as well, but I notice both requests have been open for several months, and I wonder if it is due to the implementation complexity of such a feature. I think the functionality I described, entirely separate hotkeys, would be simpler and thus perhaps we could get it implemented sooner.

IdrisGit:
I would love to know what is your workflow/usecase where having both tap and hold modes will help, it's to understand what sort of usecases it solves.

xkrogen:
@IdrisGit for me it's primarily dependent on length of my dictation. I'm migrating from Wispr Flow which has this functionality, and how I would use it is:
- Hold mode for brief dictations, up to a few sentences
- Tap mode for longer dictations, up to several minutes long

I would expect most users to use Hold if their dictations are typically short, and Tap if they're typically long. But I don't think I can be alone in making frequent use of both.

FWIW the latter I primarily use for two things: (a) longer brain dumps, where I _can_ hold the key the whole time though it's a bit annoying; (b) reviewing changes (eg a PR, document, or plan) where I need to be able to click around and navigate, which makes holding down a key very impractical.

IdrisGit:
AFAIK whisprflow supports both double tap and a dedicated hotkeys for both hold and tap, which one do you use more?

I can see the usecase being very helpful.

xkrogen:
I have both set up actually. I primarily use the double-tap for Tap mode (I find it to be quite intuitive and fits the muscle memory well), but even on Wispr Flow it can be a little finicky especially if the system is under high load, so sometimes I fall back to a dedicated hotkey.

xkrogen:
Sample PR for this functionality: https://github.com/OpenWhispr/openwhispr/pull/1961
(Using this in my local fork and working well so far)

Issue #670, opened 2026-04-26
Double tap support


aliaksandrsen:
Description:

It would be very useful to support triggering actions via a double-tap on a single key (for example, pressing a key twice quickly).

This interaction pattern is already familiar to users on macOS. For instance, system dictation can be activated by double-pressing the right Command key. This makes the feature intuitive and aligns with existing OS-level behavior.

Use case:

Quick activation of Openwhispr without occupying additional complex shortcuts
Reduced cognitive load compared to multi-key combinations
Better ergonomics for frequent actions

Suggested behavior:

Allow users to bind an action to a double-tap of a specific key
Configurable timing threshold for what counts as a "double tap"
Support for modifier keys as well (e.g. Ctrl, Command, Alt, etc.)

Example:

Double-tap Right Command → trigger Openwhispr

This would significantly improve usability, especially for users who prefer minimal and fast keyboard interactions.

xkrogen:
Seems like a dup of https://github.com/OpenWhispr/openwhispr/issues/356

IdrisGit:
Linking related reports: both request double-press/tap hotkeys. #362 also covers triple-press and Fn support, while #670 focuses on timing and modifiers. Please review them together.

<sub>Note: Drafted and posted by automation. Please flag any errors.</sub>

Issue #356, opened 2026-03-02
dual hotkey or modifier key for switching activation mode


winne617:
Currently, toggling between "tap to speak" and "hold to speak" requires going into Settings. It would be useful to support either two separate configurable hotkeys (one per mode) or a modifier key that temporarily switches the active mode. This would allow switching on the fly without interrupting workflow.

gabrielste1n:
@winne617 how do you feel if you just had to double tap the hotkey to activate "hands free" (same as tap) and then press and hold for push to talk?

winne617:
@gabrielste1n that would feel already a lot better;)

winne617:
Hi @gabrielste1n, picking this back up — the "double-tap for hands-free + press-and-hold for push-to-talk" approach you proposed sounds great to me. Is it still on the radar? This is honestly the one missing piece keeping me from switching over full-time. Happy to test a build. Thanks!

xkrogen:
Alternative proposal that I think would be simpler: https://github.com/OpenWhispr/openwhispr/issues/1860

PR #1869, opened 2026-08-26
Feat/arabic UI localization

B0-KA:
Adds Arabic (العربية) as the 11th supported language with full RTL layout handling.

### Changes

- **Translation:** Complete Arabic `translation.json` and `prompts.json`, following the existing locale conventions.
- **Locale registration:** Arabic is registered in the renderer, main process, settings language picker, onboarding, and retired-prompt hashes.
- **RTL and bidi:** The app direction follows the active locale; shared layouts use logical properties; user-authored and technical values retain explicit direction boundaries.
- **Native-review follow-up:** Canonical default folders (`Personal`, `Meetings`, `Videos`) keep their stored names but display localized labels; Notes tree hover actions reserve collision-free logical inline-end space; note dates follow the active app locale.

### Integration status

- Squashed and rebased onto current `main` by the maintainers (`maintainerCanModify`), then followed by two cleanup commits:
  - Added the 104 strings `main` gained after the fork (leaderboard, model cards, push-force-stop, streaming-only provider), removed 20 dead keys, retranslated Insights claim/enable copy, and resolved the `"a / b"` numerics to `"a من b"`.
  - Removed `rtl:space-x-reverse` on avatar stacks (Tailwind v4 `space-x` is already logical), moved the sidebar slide to `ltr:`/`rtl:` classes, isolated emails inline with `<bdi dir="ltr">` instead of block-level `dir="ltr"`, extracted `useUiLocale()`, reused `BidiInterpolatedText` for the SSO domain, and finished the logical-property swaps in files the PR and newer `main` code had missed.
  - Restored the Arabic `cleanupPrompt`, `translatePrompt`, `dictionarySuffix`, and `screenContextSuffix` to match the other nine locales (shared `fullPrompt` stays English), registered the hashes, and corrected the CLAUDE.md i18n rule accordingly.
- Deliberately kept from the original PR: routing now reads the assistant name through `getAgentName()` (default `"OpenWhispr"`), matching what the settings UI already displays.

### Verification

- `npm run typecheck`, `npm run lint`, Prettier, and `npm run i18n:check` pass.
- Full suite: **4,011 passed**, 0 failed, 196 skipped.
- The Arabic copy was reviewed for placeholder parity, six-form plural grammar, terminology consistency, and bidi punctuation; a post-rebase native visual pass has not been rerun.

### Original screenshots

Screenshot 2026-08-27 004141
Screenshot 2026-08-27 004233

PR #1978, opened 2026-09-01
feat(hotkeys): Tap/Hold for every hotkey and double-press hands-free mode

greatcoat:
## Intent

Make Hold and hands-free available to Dictation, Voice Agent and Translation. Hold is the default where the selected key and desktop can report a release; double tap starts hands-free, and the next press stops it. Meeting shortcuts keep their existing behaviour.

Existing Tap slots migrate once to Hold when supported. Unsupported keys or backends fall back to Tap with a reason. Settings presents the available gestures, a one-time card explains the migration, and a short tip offers hands-free after a long Hold session.

## Changes

- Route per-slot activation and hands-free gestures through the platform listeners, with cancellation and forced-stop handling that cannot leave a microphone running or stop another input kind.
- Keep the Linux regular-key default and native key-name parsing fixes. Onboarding presets remain owned by `onboarding/hotkeyPresentation.ts`, including the Windows Voice Agent suggestion.
- Keep tip/card controls and keycaps direction-aware, with hotkey and card strings in all eleven locale catalogs.
- Rebase onto main `1c123e8739f87c0a72d0b252bc9cff14e684b7f4` (1.10.2), preserving its Linux pill hover/retry fixes and meeting-ended pipeline removal. Include tip-card placement in Linux's new mouse-input region so card controls remain reachable through their exit animation.
- Keep the approved onboarding Hold/hands-free reveal in stacked PR #2195. It is not introduced here. PR #2070 is unchanged.

## Rebase audit

Both `package.json` and `package-lock.json` are byte-identical to the new main. Compared with the previous base `a7f22e07`, the only package changes are main's 1.10.0 → 1.10.2 version bump and `remark-gfm@^4.0.1`: 18 added lockfile package entries, no removed entries, and only root package metadata changed among existing entries.

All eleven catalogs were checked by key and value against the old base, old branch head and new main. Each is the exact union of main's changes and the branch's 20 changed hotkey/card keys (including the retired release-key reason). There were no competing edits to the same key. Main's new strings and intentional removals are preserved; catalogs were not manually spliced.

## Validation

- Rebased head: `110e8e2267d2037a889cddd9b03509fcdf5bdf2f`.
- Node 24.20.0 full suite at concurrency 4: 4,778 tests. Run 1: 4,541 pass, 3 fail, 233 skip, 1 todo (544.7 s). Run 2: 4,539 pass, 5 fail, 233 skip, 1 todo (936.7 s). Neither local full run was green.
- Local failures are in test/source files byte-identical to main: the native microphone-listener Swift compile reached 120 seconds; the second run also exceeded the 100 ms tar test deadline and a target-window activation deadline; two window-bounds tests intermittently exceeded their 700 ms helper deadline. The native test passed all 11 cases in isolation, tar and selection files passed in isolation, and the bounds file passed all 5 tests once but failed 2 in a later retry. The machine was under concurrent build load; timing sensitivity is an inference, not a platform-verification claim.
- Linux tip-card regression: observed RED before the fix, then GREEN; all 17 focused input-region and retry/auto-hide tests passed.
- `npm run quality-check`, `npm run i18n:check`, `npm run build:renderer` and `git diff --check` passed. Six existing lint warnings remain; a new unused binding warning was fixed and its file re-linted cleanly.
- All nine GitHub checks passed on this exact head: quality, pr-quality, lockfile-lint, both CodeQL checks, Linux, Windows, and macOS x64/arm64 builds. GitHub reports MERGEABLE; BLOCKED is the outstanding review requirement.

Windows and Linux have not been run in this session. Unit tests, CI builds and code-level capability checks are not platform verification. The Linux tip-card regression test exercises App's actual input-policy arguments and the real hook's IPC output; it does not run a Linux compositor.

The earlier macOS onboarding attempt on #2195 was invalidated by its missing Globe helper; fresh onboarding evidence belongs to #2195's review. No merge is requested by this update.

Fixes DES-451; refs #1860, #356 and #670.


greatcoat:
Pushed `b1858b76` — one focused commit fixing the B2 defect from the second pressure-test round, plus its regression test. Ready for a look and a re-test.

**The defect.** Double-tap Globe to latch hands-free, then `Fn+Left` inside the recent-latch window. On a cold mic nothing cancelled: the device opened ~3 s later and recorded indefinitely. The user thinks they cancelled and is left with a live microphone. A cold mic is the ordinary state for the first dictation after idle (`micReadyMs` 3056–3235 cold vs 37–65 warm), so this is the common path.

**Correction to the earlier diagnosis.** The handover blamed the `primeToken` ordering in `pressGesture.interruptGesture()`. That is not it, and I have not changed that file. The run's own log settles it — after the latch, the release takes the `Release without a registered press — ignored` path, so `handlePushQuickRelease` is never called and `primeToken` is `0`:

```
18:51:32.860  Double press latched hands-free recording {"inputKind":"dictation"}
18:51:32.910  [Globe] Release without a registered press — ignored
18:51:33.475  (interrupt lands)
18:51:36.165  Recording started with microphone   micReadyMs: 3228, usedPreparedCapture: false
```

`interruptGesture()` returns `cancel-recording` correctly and clears the latch; `usedPreparedCapture: false` proves the cancel *arrived* in the renderer. The gesture layer already had unit coverage asserting exactly this. The `primeToken && handsFree` overlap the handover describes is not reachable through any current caller — every path that could prime after a latch (`main.js` globe-up, `handleNativePushKeyUp`, `handleMacPushModifierUp`) guards on state a latch never establishes.

**Actual root cause — the renderer.** `performStartRecording` is parked awaiting the mic open, so `isRecording` is still `false`. Neither signal can reach it: `cancel-dictation-preparation` only cancels the *prepared capture*, and `cancel-hotkey-pressed` is gated on `isRecordingRef.current` in `App.jsx`. The start completes regardless and the recording runs.

This is the same class of bug the block immediately below already fixes for **stop** (`stopRequestedDuringStartRef`, "a stop that landed while the start was still awaiting the mic open was dropped… leaving a runaway recording"). Cancel never got the same treatment.

**Fix.** `cancelRequestedDuringStartRef`, set by the cancel-preparation handler and by `cancelRecording()` whenever the start lock is held, honored in `performStartRecording` the moment the mic lands. Checked *ahead* of the stop block — a cancel discards audio a stop would keep. It covers every cancel path that fires while the machine still sees no recording (the Fn-combo interrupt, a force-stopped push, the deferred quick-release cancel), not only this gesture.

**Verified on a dev build** against the production API, Globe + Hold, driven with CGEvents:

| Step | Result | Evidence |
|---|---|---|
| **B2 cold** | **pass** (was the defect) | latch `20:22:22.568`, interrupt +618 ms, mic opened `20:22:26.072` (`micReadyMs 3420`) and torn down **16 ms later**. No transcription, no paste. The same gesture at `ffed4f3b` left it live 16 s later. |
| B2 warm | pass | recording live 59 ms after the latch, cancelled 175 ms after the combo, no paste |
| B3 | pass | `Fn+Left` 7.5 s after the latch — recording survived, plain tap stopped and pasted |
| B5 | pass | arrow pressed while Globe held — survived the release, plain tap stopped and pasted |
| B6 | pass | quick tap + `Fn+Right` — no recording in 9 s, nothing pasted |

The regression test is RED at `ffed4f3b` with the bug's own signature (the mic opens after the cancel and nothing tears it down).

**Full suite:** 3250 pass / 185 skip. Four local failures, all reproduced on pristine `ffed4f3b` across five runs and none touching this code — `enterpriseIdentityStoreImports` (constant) plus `textEditMonitorSelection`/`WindowBounds`, which assert wall-clock deadlines (`Date.now() - start < 3000`) and flake under load. CI is the arbiter.

**One thing I did not fix, flagging rather than widening the diff:** `App.jsx`'s `cancel-hotkey-pressed` handler is gated on `isRecordingRef.current`, so **Escape** also cannot cancel a start that is still opening the mic. This fix makes that path work when the cancel arrives via `cancel-dictation-preparation`, but the bare Escape gate is a separate call worth making deliberately.

Not merging — that's yours.

greatcoat:
Round 2 pushed — three commits on top of `f4d7b17a`, folding in Saket's cross-platform review from the team thread (2026-09-02).

**`cf13f761` hands-free is a Hold-mode gesture only.** Saket asked why Tap needed a double press at all and Josh agreed: a single Tap press already starts an unlimited recording, so the double press only took the ordinary "stop" meaning away from a fast second press. The toggle path carries no gesture now; the companion pill no longer needs its opt-out. Fail-first test: a fast second Tap press must send a second `toggle-dictation` (was 1 with the latch, now 2).

**`9169d5a1` Hold for the assistant and translation slots on GNOME and KDE; converge an unsupported Hold to Tap.**
- *Linux (Saket's Ubuntu finding).* The PR had gated every non-dictation slot off Hold on any DE-native backend, but both backends already carry the release phase. GNOME: the portal wrapper binds every Hold slot in ONE session (the portal binds a session's shortcuts exactly once, so adding or dropping a slot recreates the session with the full set), routes `Activated`/`Deactivated` by shortcut id, and serialises register/unregister so two callers can never race two sessions into life. A slot on Hold binds through the portal and its gsettings toggle stands down; back on Tap does the reverse. KDE: `globalShortcutReleased` reaches every push-capable slot (never the meeting slot — its callback takes no phase and would start a second meeting on release). The hotkey manager owns the per-slot modes and `setSlotActivationMode` (capability check, GNOME rebind, rollback if the portal refuses), mirroring `setActivationMode` for dictation. Hyprland still binds only dictation and now says so in the row.
- *macOS (Saket's screenshot).* Changing a Hold dictation hotkey to a plain single key no longer errors "native listener not available". The hotkey wins, the mode converges to Tap, `update-hotkey` returns `activationMode: "tap"`, the IPC layer persists and broadcasts it, and the renderer says so in the saved alert.
- *Windows (Saket's open finding).* Not reproduced from the code: the assistant rides the same native machine as dictation Hold end to end. Added log lines at each gate so the next attempt is diagnosable — the testbook's track W lists the exact lines to send with the hotkey string.

**`a9860805` Settings says why Hold is unavailable.** The greyed-out Hold on a macOS single key only had a hover tooltip, and it read "Push-to-Talk native listener not available" — a Windows message. The row now prints the real reason underneath; strings in all ten locales.

**Not in this PR:** the meeting hotkey starting a new note on every press is pre-existing on `main` (`startManualMeeting` has no active-meeting guard). Saket and Josh both flagged it; it needs a product call (ignore the press while a meeting runs, or make the press end the meeting) and its own PR.

Testbook updated for round 2 (Tap track rewritten, E8 for the macOS convergence, new tracks L and W): https://claude.ai/code/artifact/76e410b9-7b55-461e-93b1-0eb173774531

greatcoat:
Pushed `7a3cd9b7` — Josh asked why a single key could not Hold on macOS when RightCommand can. The answer, and the fix:

**Why F9 could not Hold.** RightCommand is read by `macos-globe-listener` as a modifier flag change, so it has a release. F9 went through Electron's `globalShortcut`, a Carbon hot key — and a Carbon hot key hides *both* edges of the key from every monitor. A spike registered F9 as a hot key and watched with a global `NSEvent` monitor: no keyDown, no keyUp, only the hot-key event. No release source, no Hold.

**Fix: the listener owns plain keys on Hold, as the Windows/Linux hooks already do.** A plain key on a Hold slot is not registered as a hot key at all. The listener gains a keyboard event tap (a second spike proved a session tap sees and can swallow F9 with no hot key registered) that reports `KEY_DOWN`/`KEY_UP` for watched keys and consumes them so they never type. The tap exists only while a key is watched, ignores auto-repeat, and is torn down when the watch empties. Nothing is watched during hotkey capture, so `HotkeyInput` still sees the key. `hotkeyManager.isMacListenerOwnedKey` is the single decision point; both mode setters re-register a slot whose plain keys change owner with the mode; `main.js` routes the events through the same native push machine as Windows/Linux.

Verified on the compiled helper with synthetic CGEvents: `--watch-keys F9` → `KEY_DOWN:F9 / KEY_UP:F9`, F10 silent, live reconfiguration to F13 honoured, empty watch tears the tap down, EOF exits. The dead "single key can't Hold on macOS" string left all ten locales; the Hold→Tap convergence from the earlier round-2 commit stays for the cases that still cannot Hold (modifier-only combos on a DE-native Linux desktop).

The testbook's E1/E2/E8 now expect Hold to be selectable on F13/F9 and to work; the end-to-end round will prove it in the app.

greatcoat:
## Autonomous macOS end-to-end run at `7a3cd9b7` — 22 cells, 22 pass, 0 defects in the branch

Ran the round-2 matrix on a macOS dev build with no human at the keyboard: every gesture a synthetic `CGEvent`, every utterance `/usr/bin/say` played through the speakers into the built-in mic, every verdict read from the app's debug log plus a TextEdit paste target and the renderer DOM over CDP. Free-tier cloud account, `staging` profile wiped to a true first run, native helpers compiled in the worktree at this head.

**No change to the branch came out of this run — the code is unchanged at `7a3cd9b7`.** All three defects found were in the test rig and were fixed there.

### The round-2 headline holds

A plain single key genuinely Holds, on all three slots:

| Cell | Evidence |
|---|---|
| Hold on **F9** (dictation) | `[Push-to-Talk] macOS watched key-down` + `Native push key-down`, held 3.7 s, transcript pasted on release |
| Double press on **F9** | `Double press latched hands-free recording` with `"verdict": "latch"`, spoke hands-free, single press stopped and pasted |
| **F9 never types** | Nothing entered the paste target on any press while F9 was a Hold hotkey |
| Hotkey change on Hold (E8) | Changing the hotkey while on Hold kept Hold; switching to Tap made F9 a plain toggle, still typing nothing |
| Hold on **F14** (assistant) | `kind: "agent"`, answer delivered |
| Hold on **F15** (translation) | `kind: "translation"`, Spanish paste with a target language set |
| Capture mode | Capture recorded a bare F9/F13/F14/F15 with the watch off, as designed |

### The rest of the matrix

- **Tap has no gesture** (round-2 change): two presses 309 ms apart app-side → start then **stop**; no `Double press latched`, no `toggle suppressed` anywhere in the run. Same on a plain key at 202 ms.
- **Dictation Hold on `Ctrl+\``**: hold+speak → paste on release; quick tap (~130 ms) → panel hidden 403 ms later, nothing recorded; double press → latch → hands-free speech → single press stops and pastes; Escape while latched → `"active": false`, no paste; a press during processing → exactly one recording in the window.
- **Bare Globe/Fn**: `[Globe] Starting dictation (push hold)` / `Stopping dictation (push release)`. Fn+Left **0.35 s** after a fresh latch cancelled it (`active:false` at 0.93 s, nothing pasted); Fn+Left **5 s** in kept it running — 13 s of latched audio transcribed whole. Note for the testbook wording: the boundary is `RECENT_LATCH_INTERRUPT_MS = 1000`, i.e. **1 s, not 5 s** — an attempt at 1.4 s correctly did not cancel.
- **Assistant / translation**: Tap and Hold on `Cmd+.` and `Ctrl+Shift+T`, plus the double-press latch on the assistant. Answers: "Paris.", "2 + 2 = 4.", "A leap year has 366 days."; Spanish: "Buenos días, ¿cómo estás hoy?", "¡Muchas gracias por tu ayuda!".
- **Settings rows** (read structurally from the DOM): Activation Mode present under Dictation and Voice Assistant, absent under Translation until its hotkey was set and present immediately after, and never under Meeting. No "reason" text anywhere on macOS.
- **Cross-kind / mode flip**: with dictation latched, the assistant key was seen (`macOS watched key-down`), started no assistant session, and the dictation stopped and pasted; the next dictation press started fresh. Flipping Hold → Tap mid-latch stopped the recording and transcribed what it had.

### One clarification on scope

The "a hotkey that still cannot Hold converges to Tap with an alert" path is **unreachable on macOS** at this head: `supportsPushToTalk` returns false only for a modifier-only hotkey on a DE-native backend, a non-dictation slot on Hyprland, or a GNOME portal without push support. Since round 2 every macOS hotkey supports Hold, so that convergence is a Linux (track L) observation, not a macOS one.

### Rig defects found (none of them this PR)

1. **Synthetic function keys are invisible to `globalShortcut`.** F9, F13 and `Control+F13` registered `true` and never fired, which made Tap-on-a-plain-key look dead. A stock Electron app with no OpenWhispr code reproduced it exactly, while bare `A`, `` ` `` and `Control+\`` fired fine. A real Mac function key carries `kCGEventFlagMaskSecondaryFn`; a synthetic one posted without it never reaches the Carbon hot key behind `globalShortcut`, though the app's own event tap still sees it — which is precisely why this PR routes Hold on plain keys through the listener. Driver fixed; every Tap cell then passed.
2. **Pastes landing nowhere.** After driving the Settings UI, raising TextEdit with `activate` left its text view without first responder: the app logged `Paste completed` against the right target pid while the document stayed empty, and a synthetic `Cmd+V` failed identically. Affected cells were re-run after the fix.
3. **The dedicated free test account can no longer sign in** — its address is a `+` alias and the sign-in form rejects those (`AuthenticationStep.tsx`). That guard is pre-existing on `origin/main` (from #1854) and unrelated to this PR, but it does block unattended runs until the stored credential is replaced with a non-alias address.

### Still out of reach here

Windows Voice Assistant Hold (track W), the GNOME portal / KGlobalAccel release paths and the Hyprland disabled-with-reason row (track L), and the packaged signed artifact. A clean macOS run says nothing about any of them.

Testbook with the full cell-by-cell evidence: https://claude.ai/code/artifact/76e410b9-7b55-461e-93b1-0eb173774531

greatcoat:
2026-09-07: #1992 (hands-free tip after a long hold) was folded into this branch and closed. Its three commits are on this PR, and the whole stack was rebased onto main (head 327c4670). The Hold-only refinement that follows (gesture rows, migration card, retimed pill/panel motion) builds on this branch; spec at docs/superpowers/specs/2026-09-07-hold-and-hands-free-design.md.

greatcoat:
## Round 4 — rebased on main, macOS verified end to end; Windows, Linux and UI feedback wanted

### Rebase

This branch was `CONFLICTING`; it is now clean on top of current `main`. 35 commits replayed,
two conflicts, both in the same region of `src/App.jsx` where #1886 (liquid cancel button, pill
spacing, overlay smoothness) meets this branch. **Main won both:**

- the `gap-2` on `assistant-pill-presence` that #1886 removed stays removed
- `panelReturnResizeActive` stays in the `useMainWindowSizeOwner` destructure, because main uses
  it downstream in `resolvePillVisualSuppression`

Checked rather than assumed: the 51 files this branch touches that main does not are
**byte-identical** before and after the rebase.

`feature/hold-motion` (#2070) is stacked on this branch, so its diff will look noisy until it
gets its own rebase. That is expected and is being done separately — the whole point of this
pass was to test the gesture model **without** the motion layer in frame.

### Two changes from today's rig session

- The migration card's own press no longer warns *"We didn't catch that"*. That first press is
  usually an experimental tap that catches no speech, and scolding the user for trying the
  gesture the card is teaching reads as the card failing. Suppression is tied to the card being
  visible, and the card shows once ever, so the mute does too.
- The hands-free tip stays up **7s** instead of 6s.

### Verified on macOS today

Dev build on the rebased branch, driven by hand and by synthetic CGEvents.

- **All 8 cells** of dictation/agent × hold/double-press latch × floating icon show/auto-hide.
  Every cell started, ran, stopped and transcribed.
- **Globe/Fn interrupt**, all three cases. Tap then Globe+Arrow cancels the preparation. A latch
  interrupted within 1s is cancelled and discarded. A latch older than 1s survives. The
  `RECENT_LATCH_INTERRUPT_MS` cutoff discriminates correctly in both directions.
- **Hold migration card** renders, interpolates the live hotkey into the keycap, and is one-time.
- **Hands-free tip card** renders after a 121s hold, with its countdown arc, and the lifetime
  shown-count increments.

### What still needs a human on the box

- **Windows.** Nothing in this stack has ever run on Windows. Saket's Voice Agent Hold report
  from round 2 is still unreproduced.
- **Linux — the highest-risk item in the whole change.** The default non-darwin hotkey is
  `Control+Super`, which is modifier-only, and GNOME / Hyprland / KDE-Wayland refuse a
  modifier-only hotkey in push mode. The capability gate written for exactly this has **never
  executed on Linux**.
- **UI feedback**, specifically on the two cards, the read-only gesture rows in Settings, and
  what the pill does on a quick tap.

### Known and filed, not blocking

- **A quick tap prepares the microphone and then unwinds it.** By design — the prepared session
  is held for up to 400ms waiting for a possible second press. Harmless, but the mic does open
  on a gesture that records nothing.
- **A short press can still produce a corrupt clip the cloud rejects** ("Audio file might be
  corrupted or unsupported"). The validation path is untouched by this branch and the bug
  pre-dates it; Hold-by-default just makes it reachable. Reproduced today on the agent path.
- **The double-press floor has thin margin on real hardware.** Measured second-press gaps of
  146ms, 159ms and 177ms against a 150ms floor. The 146ms one was discarded as a duplicate
  delivery and cost a press. Worth a look if hands-free feels unreliable to anyone.

greatcoat:
## Linux hotkey default decided — web-grounded, and one finding is bigger than the default

Pushed as 6cf97eb2. Research pass before release, since this is a core pipeline. Full analysis is filed in Titan as `linux-hold-modifier-only-default`.

### The default

Linux moves to `Control+Super+Space`. Windows keeps `Control+Super`, where the low-level hook sees both edges.

`Control+Super` is modifier-only, and the three refusals have three different causes:

- **GNOME** — the [freedesktop shortcuts spec](https://specifications.freedesktop.org/shortcuts/latest/) defines a trigger as "a set of modifiers… together with a key identifier". There is no modifier-only form to send. Our `convertToPortalFormat` returning `""` is the correct reading, not an invention.
- **KDE** — KWin accepts one on Wayland but fuses press and release into a single event after the first modifier lifts.
- **Hyprland** — our own guard. `bindt`/`bindrt` are real edges and the conversion already degrades to `CTRL, Super_L`, so this refusal may be conservative rather than necessary. Unproven.

`Control+Shift+Space` was the runner-up and was rejected on in-app collisions: it inserts a non-breaking space in LibreOffice and Word, and triggers parameter hints in IDEs. A global grab shadows all of those. `Control+Super+Space` is not reserved, is not a GNOME or KDE default, and has no common in-app meaning.

Migration is **fresh installs only** — a stored hotkey is never rewritten.

### The bigger finding: GNOME below 48 cannot Hold on any hotkey

`supportsPushToTalk()` on GNOME returns portal availability, and the GlobalShortcuts backend only landed in `xdg-desktop-portal-gnome` for **GNOME 48** (March 2025). Below that, every hotkey converges to Tap.

| Distro | GNOME | Hold possible |
|---|---|---|
| Ubuntu 24.04 LTS | 46 | no |
| Ubuntu 25.04 / Debian 13 / Fedora 42 | 48 | yes |

Ubuntu 24.04 LTS is supported into 2029. Also, `isGnome()` matches `XDG_CURRENT_DESKTOP` and ignores the session type, so GNOME-on-X11 takes the GNOME path too and loses a release edge the raw listener would have delivered.

So the user-facing message matters more than the default. `getPushToTalkUnavailableReason` now separates the two causes: a hotkey with no regular key names the hotkey and suggests the fix; a desktop that cannot report a release says so and names what it needs. One shared message would send half of those users chasing a fix that does not exist.

### Guidance pruned

Three of five `LINUX_RECOMMENDED` entries were modifier-only. A fourth suggested a bare right-side modifier, which the portal cannot express at all — the Mumble maintainer implementing portal shortcuts records that "You can not distinguish between left/right modifiers on the keyboard" ([mumble#5976](https://github.com/mumble-voip/mumble/pull/5976)). Four of five were wrong on GNOME. Linux failure-path suggestions led with `Control+Super` for the same reason.

A new test asserts every Linux recommendation, example and default carries a regular key, so this class cannot come back.

### Double-press window: (150, 400] → (50, 500]

The floor only has to reject one physical press delivered twice. Measured duplicates arrive **~1 ms apart**; real double-presses measured 146, 159 and 177 ms, and the 150 ms floor discarded one of the three. The ceiling now matches the Windows double-click default (500 ms); GTK's is 400.

### Prior art — nobody has solved this

Wispr Flow ships exactly this gesture (hold, double-tap to lock) and `Ctrl+Win` on Windows, but has no Linux build. superwhisper: no Linux. Talon: X11 only, Wayland "not planned". Handy reaches our rule independently — on Windows and Linux it *requires* a modifier plus a regular key. Discord ships **no** default PTT keybind at all and is broken under Wayland; OBS relies on third-party portal plugins; Mumble's portal PTT is still WIP.

Shipping Tap as the honest Linux fallback is the norm, provided we say why. That is what the new strings do.

### Still unproven — needs a real Linux box

Everything Linux here is read from code and documentation.

1. That the GNOME 48 portal accepts `CTRL+LOGO+space` and delivers `Deactivated`.
2. That KGlobalAccel emits `globalShortcutReleased` for it.
3. Whether Hyprland's modifier-only refusal is necessary.
4. Whether a bare right-side modifier binds at all — `isModifierOnlyHotkey` misses it (no `+` in the string), so it passes the gate.

greatcoat:
@saketaryan — the Windows Voice Agent Hold finding from 09-02 is diagnosed and a fix is on this branch at `860514c8`. Could you rerun track W of the testbook against it? Nothing in this has run on Windows yet, so your retest is the proof.

**What was wrong.** The Windows key listener (`resources/windows-key-listener.c`) could not parse 27 of the 109 key names the Settings capture screen can emit: Enter, Backspace, the four arrows, Delete, PrintScreen, every numpad key and the media keys. Bound bare, the helper exited and Hold died for that slot. Bound with a modifier (say `Ctrl+Alt+Up`), it *silently* became a bare `Ctrl+Alt` listener that fired on the modifiers alone. Dictation escaped because its default `Ctrl+Win` is genuinely modifier-only. This is a gap in a new feature, not a regression — Voice Agent Hold has never shipped.

**What changed.** Both listeners now refuse an unknown name loudly instead of degrading, all the missing names parse, and a contract test compiles the real parser code against every capturable name so the two sides cannot drift again. The Voice Agent also gets a Windows default of `Win+Alt+Space` on fresh installs (it cannot pre-empt `Ctrl+Win`, is not on Microsoft's shortcut list, and AltGr cannot reach it).

**One thing before you start — the released listener binary is old.** `npm install` downloads `windows-key-listener.exe` from the `windows-key-listener-v1.0.0` release, which is only rebuilt on a push to `main`. A dev build on this branch runs the old binary unless you compile the fixed one:

```
cl /O2 /nologo resources\windows-key-listener.c /Fe:resources\bin\windows-key-listener.exe user32.lib
```

(from an MSVC developer prompt; check `resources\bin\windows-key-listener.exe` has today's timestamp). If `cl` gives you any trouble, tell me — that would be the first MSVC compile of this code and I want to know.

**Track W steps** (https://claude.ai/code/artifact/76e410b9-7b55-461e-93b1-0eb173774531):

- **W1** — Settings → Voice Assistant → set a hotkey and **write down the exact string** (e.g. `Ctrl+Alt+A` vs `F9`), then Activation Mode → Hold. Expect Hold selectable. **Please also tell me which hotkey you had bound on 09-02** if you remember it — if it was the onboarding suggestion `Ctrl+Shift+Space`, that one parsed fine even before this fix and the cause of your run is somewhere else, so it is the single most useful data point.
- **W2** — Hold the assistant hotkey, speak, release. If the assistant does not process: Settings → open the logs folder and copy the lines around the press. Expect, in order: `[WindowsKeyManager] Starting key listener` for that key · `KEY_DOWN detected` · `[Push-to-Talk] Native key-down` with `slot: voiceAgent, mode: push` · `Native push key-down` · `prepare-dictation` / `start-dictation`. The first missing line, or a `blocked by the panel state` / `handled by the gesture` line, is the finding — paste those lines plus the hotkey string here.
- **W3** — Bind **dictation** to the same style of hotkey on Hold and hold-speak-release. Dictation failing too points at the key type and the hook; dictation working while the assistant fails points at the assistant flow after `start-dictation`.
- **W4 (new)** — Bind the Voice Assistant to `Ctrl+Alt+Up` on Hold, then hold and release **only Ctrl+Alt** without the arrow. Expect nothing to happen. Before the fix this started a recording, because the chord had silently become `Ctrl+Alt`.
- **W5 (new)** — Bind it to a bare `Delete` or `num5` on Hold and hold-speak-release. Expect the assistant to process. Before the fix the listener exited on that name.

Thank you — and if you have the native Windows machine now rather than Parallels, say which one you ran on.

greatcoat:
Update: the branch is rebased onto today's `main` (40 commits replayed, four conflicts resolved by hand) and force-pushed. The fix described above is now at **`420dab72`** rather than `860514c8`; everything else in the instructions is unchanged, including the local `cl` compile of the listener before track W.

Two things moved during the rebase that are worth knowing if you read the diff: the onboarding refresh from #1984 owns the assistant suggestion now, so the Windows default `Win+Alt+Space` is applied through `resolveOnboardingAssistantHotkey` in the onboarding presentation module, and the recommended-pattern lists the refresh removed from the validator stay removed.

greatcoat:
## Rebase onto 1.10.0 — 2026-09-14

Head is now `02584275` on `main` `a7f22e07` (two passes: 22 conflicts against `4335972b`, then 15 more after main moved mid-session). Every conflict was resolved by reading main's change first; onboarding and the app shell are main's, this branch keeps only its per-slot Tap/Hold, gesture, tip and card behaviour. Three overlaps needed a real decision — each is its own commit with the reasoning:

- **Autorepeat gate** (#2122's `createHotkeyRepeatGate`) replaces the branch's 150 ms debounce inside the shared hotkey callback, so all three slots get it.
- **macOS "a plain key cannot hold" guard** (#2122) is superseded: on this branch a plain key on Hold is owned by the listener's key watch and never goes through `globalShortcut`, so the toggle-on-autorepeat bug it fixed cannot occur. The guard, its test and its reason string (`holdNeedsReleaseKey`, all 11 locales) retire. Onboarding's fallback-to-Tap logic is untouched.
- **#2102 force-stop reasons** ride the renamed native push machine: a forced stop (ceiling or settings reset) never arms the hands-free tip and cancels a not-yet-recording preparation immediately, as the macOS compound path already did. `pushToTalkForceStop.test.js` pins both.

Also in this push:

- `getDefaultVoiceAgentHotkey()` moves out of `src/utils/hotkeys.ts` into onboarding's `hotkeyPresentation.ts` as `getDefaultAssistantOnboardingHotkey(platform)` — onboarding is the only place the Voice Agent preset is ever applied, so it owns it. Values unchanged (Win+Alt+Space on Windows).
- The four card/gesture-row files import from `src/components/icons` (#2117 dropped `lucide-react`); the Hold chip takes `MicVocal`, the icon onboarding's own Tap/Hold control uses for Hold.
- Arabic (#1869 landed after this branch's strings) gets the 19 Hold/tip/card keys; the ten gesture-row and migration-card strings that were English placeholders in the other nine locales are translated (one translator + one independent verifier per locale).
- The tip cards' dismiss button and badge padding use logical classes (`end-`, `ps-/pe-`), so the X no longer sits on the badge in RTL.
- One duplicate dropped: main's post-window Hold re-check in `main.js`, which the branch's `demoteUnsupportedDictationHold` + settled-mode listener already cover.

Verification: `quality-check` green; suite 4,696 tests / 0 failures / 220 Electron-ABI skips on Node 24 (`--test-concurrency=4`; the default concurrency starves files on a loaded machine). **Nothing here has run on Windows or Linux** — the Track W retest (compile `resources/windows-key-listener.c` with `cl` first) still stands.

Decision recorded with Josh: this branch keeps the Hold-only Settings model; onboarding's Tap/Hold card and teaching the double-tap gesture go to a separate follow-up PR.

PR #1979, opened 2026-09-01
fix(clipboard): preserve dictation history when clipboard delivery fails

### LINKED ISSUE #1972: Dictations lost when clipboard on, auto-paste off, and non-preview STT model
**Describe the bug**
When a non-preview transcription model is used with "Automatic pasting" OFF and "Keep transcription in clipboad" ON, dictations are not saved in the history nor do they appear in the clipboard.  (The audio file also doesn't appear in the audio save directory.)  The dictation is 100% lost if preview is off, and even with an available preview, the transcription is not "kept in clipboard" or retained in history even with all retention settings on.

**To Reproduce**

1. Configure settings as described above
2. Record a dictation (e.g. by paired hotkeys in tap mode)
3. Wait for transcription to finish
4. Try to paste - no new data on clipboard; check history and audio folder: no text or audio is present.

**Expected behavior**
The transcription should be in the clipboard, history, and audio folder, regardless of preview.  The data should never be lost, regardless of settings (if retention or keep in clipboard are requested).  It should be possible to retain transcriptions even when auto-paste is turned off, and the lack of a preview for your STT model should not result in silent data loss.

- OS: Windows 10
- Version 1.9.2

greatcoat:
## Summary

OpenWhispr can finish understanding a recording, then lose it while trying to copy the result. The dictation window intentionally does not take focus, so a renderer clipboard write can fail before the history save starts.

Audio follows the history row: it is written only after history persistence returns an ID. When the clipboard error skipped history, the retained audio had nowhere to attach and was lost as well.

This PR:

- starts history/audio persistence independently of clipboard delivery and explicitly observes its result;
- routes both clipboard-only delivery and failed selection-edit fallback through the existing main-process `writeClipboard` bridge;
- catches and logs clipboard failures separately from persistence failures;
- preserves the existing auto-paste, selection-edit, Assistant, preview, clipboard-restoration, and retention branches.

## Scope

The completion handler is shared by macOS, Windows, and Linux. The existing bridge uses native clipboard writes on macOS and Windows and the X11, Wayland, and KDE/XWayland handling already implemented for Linux.

The affected non-preview models include OpenAI `whisper-1`; Groq `whisper-large-v3` and `whisper-large-v3-turbo`; xAI `grok-stt`; Mistral `voxtral-mini-latest`; and Gemini `gemini-3.5-transcribe`. OpenWhispr Cloud, Custom, and self-hosted transcription also converge on this path. Preview/streaming-capable results use the same completion handler and are covered by the regression test.

## Tests

- `node --import tsx --test test/helpers/useAudioRecordingClipboardPersistence.test.js` — 5/5 passed; the original four cases failed 4/4 against `abfaf2ba` before the implementation.
- `npm run format:check` — passed.
- `npm run typecheck` — passed.
- `npm run build:renderer` — passed.
- `npm test` on Node 24 — all five clipboard-persistence cases passed; 3,194 tests passed overall. Three unrelated macOS native text-monitor timing cases failed under full-suite load; both affected files passed directly, 21/21. The 185 SQLite cases were skipped because postinstall built the native module for Electron rather than host Node.

## Manual release QA

Runtime QA was not performed in this implementation session. For each platform below, keep another app focused, turn Automatic pasting off, Keep transcription in clipboard on, and retention/audio retention on; then verify final clipboard text, exactly one history row, and its matching audio file.

- [x] macOS
- [ ] Windows 10
- [ ] Windows 11
- [ ] Linux X11
- [ ] Linux Wayland, including KDE/XWayland where available
- [x] Repeat with failed selection-edit fallback and confirm the history/audio save survives clipboard failure
- [ ] Repeat one non-preview model and one preview-capable model, with preview both available and unavailable
- [x] Check adjacent settings: auto-paste on + keep on, auto-paste off + keep off, and retention off

## Credits

#1934 by @IdrisGit established the tested replacement baseline: both affected renderer writes should use the main-process bridge and persistence must start independently. #1022 originally identified and exercised that bridge for Linux Wayland.

Fixes #1972


Chadpiha (approved):
Approved — no blocking issues.
I traced the full renderer → preload → main-process clipboard path and the transcription/audio persistence path. Persistence now starts independently of clipboard delivery, both affected clipboard branches use the existing cross-platform main-process bridge, and failures are observed without risking transcription loss.
Regression coverage is red/green, the stricter test harness passes, and the full Node 24 suite completed with 3,197 passed and 0 failed. Format, typecheck, renderer build, CodeQL, and platform CI checks are also clean.
Windows and Linux runtime QA remains worthwhile before release, but I don’t consider it a merge blocker.

PR #1981, opened 2026-09-01
fix(enterprise): harden AWS Bedrock error handling

greatcoat:
## Summary

- add a Bedrock-specific retry policy with six total attempts and exponential backoff plus jitter for 503, 429, timeout, and audited network failures
- keep authentication, permission, invalid-model, and configuration failures non-retryable; preserve the configured AWS region with no silent failover
- preserve HTTP status, AWS exception type, request ID, nested cause, and actionable user-facing diagnostics across Check connection and dictation cleanup
- make retries, backoff, model resolution, and in-flight cleanup requests cancellable and bounded by application timeouts
- report the raw-dictation fallback only after an actual paste, with separate quieter UI and expandable or copyable technical details

## Review findings addressed

- unwrap RetryError.lastError before classification
- handle real AI SDK error shapes, including case-insensitive x-amzn-errortype and data code or type fields
- prevent explicit invalid credential providers from falling through to unrelated environment credentials
- retain the resolved managed Bedrock region in diagnostics
- cancel during runtime resolution and retry backoff without starting another request
- suppress fallback status and AutoLearn monitoring for clipboard-only or no-paste outcomes

## Verification

- focused Bedrock, cancellation, credential, paste, fallback, and preload matrix: 70 passed, 0 failed
- full serialized suite: 3,414 tests; 3,228 passed, 0 failed, 185 skipped, 1 todo
- TypeScript, ESLint, Prettier, and renderer production build passed
- fresh whole-branch review plus scoped final-fix re-review found no remaining Critical or Important issues

No intentional behavior changes were made to Azure, Vertex, Agent, translation, or non-cleanup dictation routes.


Chadpiha:
Traced this end-to-end from the user journey rather than hunk-by-hunk, with particular attention to whether anything that worked before regresses. No critical or blocking defects. I found four important issues and a handful of minors, and pushed fixes for all of them to this branch (`a809ac0d`, `0d279f96`, `fab0029e`) so the PR is mergeable as it stands.

### Verified working, not assumed

- **The retry/cancel gate actually engages.** `inferenceScope: "dictationCleanup"` genuinely reaches the handler on both BYOK and managed paths (`audioManager` → `resolveManagedScope`, which always re-emits the scope → `enterprise.ts` spread → handler gate). Worth stating explicitly, since the whole feature is dead code if that string doesn't arrive.
- **Error shapes match the installed SDKs.** Checked against `@ai-sdk/amazon-bedrock@4.0.129`, `@ai-sdk/provider@3.0.14`, and `ai@6.0.219` rather than from memory: `APICallError.statusCode`, lowercase `responseHeaders`, `RetryError.lastError`, and `maxRetries: 0` not wrapping. `AbortSignal.timeout` produces a `TimeoutError` that classifies as retryable; a user abort classifies as non-retryable.
- **`runAbortableOperation` does not leak an unhandled rejection.** `Promise.race` attaches handlers to both inputs, so a late rejection from the abandoned operation is absorbed — confirmed empirically. The operation is abandoned rather than cancelled, which is wasted work but harmless.
- **Azure and Vertex are behaviorally unchanged.** The new `async getEnterpriseAIModel` is awaited at every call site including `enterprise-stream-start`, and `mappingConfig` only diverges for Bedrock.
- **Cancellation is leak-free**: listener add/remove balanced, registry completed on success, error, and abort, with the `isDestroyed()` post-registration race closed.

### Fixed in the pushed commits

1. **The cleanup-failure toast was suppressed whenever the text didn't auto-paste.** `recordCleanupFailure` sat inside `if (pasteSucceeded)`, so a user with auto-paste off — or on the macOS clipboard-only fallback — silently lost the entire notice, including the actionable "SSO expired → `aws sso login`" guidance. That's the exact silent-failure class this PR exists to fix. It now records on every delivery outcome, with the paste result choosing the secondary line.
2. **Retries multiplied the timeout budget roughly sixfold.** Each attempt built its own `AbortSignal.timeout` inside the operation closure, so a hung endpoint could hold a dictation paste for ~6 minutes where a single request previously timed out at 60s, and Check Connection could spin ~3.4 minutes with no cancel affordance. `runBedrockRequest` now composes the signals itself and enforces one deadline for the whole call.
3. **Non-Bedrock cleanup failures showed a raw English string as the toast title.** Unkeyed provider messages ("fetch failed", HTTP bodies) replaced the localized headline in all ten locales. The localized title is restored and the raw cause moved to the description, where the error surface already offers to copy it.
4. **Only 503 retried among 5xx.** A transient `InternalServerException` or a gateway 502 failed outright with generic copy, where both AWS standard retry mode and the AI SDK's own `isRetryable` treat ≥500 as retryable. Added a `serverError` kind, and 504 is now classified as the timeout it actually is.

Minors, also fixed: the "Technical details" summary ignored its `onDark` prop (~2.3:1 contrast on the dark toast in light app theme); the expired-credential branch handed static-key users an `aws sso login --profile default` command for a profile name we invented; AccessDenied lost the self-service "enable model access in the console" step that solo BYOK users need; a 400 context-length overflow was reported as a bad model ID; the destructive toast offered a second copy button that copied an instruction sentence, with a hardcoded English aria-label; `retryable` was missing from the `testEnterpriseConnection` type; and cancellation is now registered for every Bedrock scope, so a cancelled translation or note-formatting request stops instead of running to its timeout (retries stay cleanup-only).

Test coverage grew by 9 to close gaps: destroyed-sender abort, the real `sleepWithAbort` abort branch with a timer spy, the secret-key-only credential guard, and 5xx classification.

### Verification

Full suite **3,435 tests — 3,249 pass, 0 fail**, 185 expected locally-skipped DB tests, 1 todo. `typecheck`, `lint`, `format:check`, `i18n:check`, and the renderer production build all clean. All 30 new translation keys are present with matching interpolation placeholders in all ten locales.

Chadpiha (approved):

PR #1984, opened 2026-09-01
feat(onboarding): refresh setup flow and preserve progress

boseq:
## Summary

- refresh authentication, permissions, keyboard shortcut, and local-model setup screens to match the updated onboarding designs
- persist the local onboarding session and restore the exact eligible step after restarting, while retaining safe in-progress configuration
- clarify downloaded, downloading, selectable, and selected model states and recommend Cmd + K on macOS
- add focused coverage for authentication, route restoration, model selection, permissions, and shortcut confirmation

## Validation

- 62 focused onboarding and hotkey tests pass
- TypeScript typecheck passes
- locale consistency check passes
- ESLint passes with no errors
- Prettier and git diff checks pass
- production renderer build and packaged local-production smoke test pass


Chadpiha (approved):
Approved, thanks @boseq 

Reviewed end-to-end: the route/session machine, the hotkey path from capture through IPC to registration, provider/model setup, auth + permissions, all 10 locales, and the tests. 3809/3816 pass, 0 fail; tsc, ESLint, Prettier and i18n:check clean.

Good work. The session parse boundary is total and idempotent — all 16 step ids against 5 route shapes land on-route and stay there. Download states read real on-disk truth. And the Agent → Assistant rename is complete rather than partial: the only "agent" left in any locale is "agentic coding model", which is correct.

Fixed on this branch
The last six commits are mine, so they need a second reviewer — an adversarial re-review already caught a real defect in my first attempt at the first one.

A hotkey confirmed on the shipped build was silently replaced. Sessions written before resume existed have no confirmation flags, so dictationHotkeyConfirmed read back false, Right Option was substituted, and finalizeOnboarding persisted it on routes that never show the hotkey step again. An Enterprise session resuming on setup-choice auto-finalizes from an effect, so it took zero clicks. Fixed at the parse boundary by inferring the flag from step position; the resolver is unchanged. (It couldn't be fixed there: dictationKey isn't evidence of user choice, because hotkeyManager auto-persists GLOBE before onboarding runs.)
Resuming into email verification left no controls for 60s, every launch — the resend cooldown starts on every mount and hides Resend + Back, and auth draws no shell footer.
The recommendation row never rendered on first entry, so two of the three macOS recommendations were reachable only by clearing the chord the step had just offered.
Log out was offered to users with no account — the legacy migration labels any pre-v2 session "account", and the action wipes progress unconfirmed.
The Notes step lost its forward action while workspaces resolved, leaving only Back where a disabled, loading Continue used to be.
"Back to sign in" landed on the create-account form, because authMode restored as "sign-up".
Plus coverage for two blind spots: the local-model resume write could be deleted with every test green, and the "no secrets" assertion ran against a literal the test built itself — so writing the API key and Corti client id into the real BYOK draft passed both the suite and tsc. That's the regression b6f65cb3 fixed. Both now assert where the session serializes, and every fix is mutation-verified.

PR #1995, opened 2026-09-02
fix(dictation): stop restacking the pill on every live transcript chunk

### LINKED ISSUE #1262: Live Transcription Preview redrawing and repositioning every half second.
**Describe the bug**
The live transcription window is constantly redrawing and repositioning itself. It's jumping around whenever I use it and I am dictating text. 

**To Reproduce**

1. Enable Live Transcription Preview under the Speech to Text settings
2. Use the hotkey to start listening.
3. Talk

**Expected behavior**
The windows should be in a stable position when dictating.

**Screenshots**

- OS: Windows 11
- Version 1.7.6

**Additional context**
This worked in 1.7.2

hsusul:
## Summary
- Live transcription preview still called `showInactive()` and re-applied always-on-top on **every** partial result, even after the renderer started buffering/measuring text.
- On Windows that restacks the already-visible dictation pill, which is the remaining cause of the overlay jumping around while you talk.
- Surface and z-order the window only when it is hidden; later chunks just send `preview-text`.

Fixes #1262

## Test plan
- [ ] Enable Live Transcription Preview, start dictation, speak for several seconds: the pill should stay put (grow for new lines is fine; it should not hop or flicker).
- [ ] Confirm a hidden pill still appears on the first preview update.
- [ ] `ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/windowManagerAssistantPanel.test.js`

### DIFF
--- a/src/helpers/windowManager.js
+++ b/src/helpers/windowManager.js
@@ -1319,8 +1319,13 @@ class WindowManager {
     if (!this.mainWindow || this.mainWindow.isDestroyed()) return;
     this.mainWindow.webContents.send("preview-text", text);
     this._sendAgentDictationPreview("preview-text", text);
-    this.mainWindow.showInactive();
-    this.enforceMainWindowOnTop();
+    // Partial results arrive many times per second. Re-showing the already
+    // visible pill and re-applying always-on-top makes Windows restack and
+    // reposition the overlay on every chunk (#1262). Surface it only once.
+    if (!this.mainWindow.isVisible()) {
+      this.mainWindow.showInactive();
+      this.enforceMainWindowOnTop();
+    }
   }
 
   appendTranscriptionPreview(text) {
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks @hsusul, this was right and it has landed via #2185 with your commit and authorship intact. The premise checked out: every live-transcript partial re-showed the visible pill and re-applied always-on-top, and on Windows and Linux each `showInactive()` also re-fired the window's `show` handler, so each chunk cost a ShowWindow plus two SetWindowPos calls. Guarding on `isVisible()` matches how `showDictationPanel` and `showAgentDictationPill` already behave, and nothing ever lowers the pill's on-top state.

Two small edits were made on the way in: a rebase onto main to resolve the conflict in `test/helpers/windowManagerAssistantPanel.test.js` (main had replaced the single "surfaces a hidden pill window" test with a platform trio; your two new tests sit right after the companion-mirroring test), and a shorter comment above the guard without the "reposition" wording, since `setAlwaysOnTop` on Windows is a SetWindowPos with SWP_NOMOVE. I could not push the rebase to your branch because it carries main's workflow-file commits, which the maintainer token is not allowed to push to a fork, hence the separate PR. Closing this one in favour of #2185.

PR #1996, opened 2026-09-02
fix(macos): resolve ffmpeg and ffprobe from Nix-managed prefixes

### LINKED ISSUE #912: [BUG] ffmpeg not detected on macOS managed via Nix-Darwin

hsusul:
## Summary
- GUI-launched Electron never sources `~/.zshrc`, so nix-darwin ffmpeg at `/run/current-system/sw/bin` is invisible even though it is installed.
- Shared resolver now probes Homebrew plus Nix prefixes (`current-system`, `~/.nix-profile`, default profile) for ffmpeg, and production `PATH` setup on macOS includes the same dirs so yt-dlp can find sibling `ffprobe`.
- Does not spawn a login shell; keeps the previous Homebrew/system candidates.

Fixes #912

## Test plan
- [ ] On a nix-darwin Mac with ffmpeg only on `/run/current-system/sw/bin`, launch the packaged app from Finder (not a terminal) and dictate: local conversion should find ffmpeg.
- [ ] URL/file flows that need ffprobe should succeed when ffprobe sits next to that ffmpeg.
- [ ] `node --test test/helpers/unixToolBinDirs.test.js`

### DIFF
--- a/main.js
+++ b/main.js
@@ -362,9 +362,12 @@ const AUTH_BRIDGE_HOST = "127.0.0.1";
 const AUTH_BRIDGE_PORT = parseAuthBridgePort();
 const AUTH_BRIDGE_PATH = "/oauth/callback";
 
-// Set up PATH for production builds to find system tools (whisper.cpp, ffmpeg)
+// Set up PATH for production builds to find system tools (whisper.cpp, ffmpeg,
+// ffprobe). GUI launch never sources shell rc, so Nix prefixes are otherwise
+// invisible even when the binaries exist on disk (#912).
 function setupProductionPath() {
   if (process.platform === "darwin" && process.env.NODE_ENV !== "development") {
+    const { getUnixToolBinDirs } = require("./src/helpers/unixToolBinDirs");
     const commonPaths = [
       "/usr/local/bin",
       "/opt/homebrew/bin",
@@ -372,10 +375,11 @@ function setupProductionPath() {
       "/bin",
       "/usr/sbin",
       "/sbin",
+      ...getUnixToolBinDirs({ platform: "darwin" }),
     ];
 
     const currentPath = process.env.PATH || "";
-    const pathsToAdd = commonPaths.filter((p) => !currentPath.includes(p));
+    const pathsToAdd = [...new Set(commonPaths)].filter((p) => !currentPath.includes(p));
 
     if (pathsToAdd.length > 0) {
       process.env.PATH = `${currentPath}:${pathsToAdd.join(":")}`;
--- a/src/helpers/ffmpegUtils.js
+++ b/src/helpers/ffmpegUtils.js
@@ -4,6 +4,7 @@ const os = require("os");
 const path = require("path");
 const debugLogger = require("./debugLogger");
 const { createAbortError } = require("./abortError");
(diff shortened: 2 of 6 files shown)


gabrielste1n:
Thanks for the write-up on how a launchd PATH differs from a login shell, @hsusul; that part is accurate. We're closing this because the situation it addresses doesn't arise in OpenWhispr.

The app ships its own ffmpeg. `ffmpeg-static` is unpacked from the asar (`electron-builder.json` `asarUnpack`) and `scripts/afterPack.js` fails the build on every platform if the unpacked binary is missing. All three resolvers this PR touches (`getFFmpegPath` in `src/helpers/ffmpegUtils.js`, `src/helpers/whisper.js` and `src/helpers/whisperServer.js`) return that bundled binary before the Homebrew or Nix candidate list is consulted, so in a packaged build the added directories are never reached, and in `npm run dev` the process inherits the terminal PATH and `node_modules/ffmpeg-static` exists anyway.

On ffprobe: the app never invokes it. The error quoted in #912 is the HTTP response body of the reporter's self-hosted OpenAI-compatible transcription server, passed through verbatim by `src/helpers/ipcHandlers.js`; nothing in the Electron process environment reaches that server. yt-dlp gets `--ffmpeg-location` pointing at the bundled directory and resolves both `ffmpeg` and `ffprobe` there without a PATH fallback.

Smaller notes for future reference: the helper re-lists `/opt/homebrew/bin` and `/usr/local/bin` that `main.js` already has, which is what forced the `new Set` dedupe, and the tests assert that a constant array contains constants rather than exercising resolution.

PR #1998, opened 2026-09-02
fix(transcription): pass measured audio duration to speaker merge in file transcription (#1997)

### LINKED ISSUE #1997: fix(transcription): pass measured audio duration to speaker merge in file transcription
### Affected component

`src/services/fileTranscription.ts` (`transcribeFileWithSpeakers`)

### Describe the bug

When transcribing an audio file from disk with speaker diarization enabled, the caller typically does not provide an upfront duration (`durationSeconds` is `undefined`). The local diarizer measures the audio file and returns its true duration in `diar.durationSeconds`, which is stored in `measuredDuration`:

```typescript
const measuredDuration = durationSeconds || (diar?.success && diar.durationSeconds) || null;
const result = { ...transcribed, durationSeconds: measuredDuration };
```

However, when merging speaker segments with the transcribed text, `mergeSpeakerText` is called with `durationSeconds || 0` rather than `measuredDuration || 0`:

```typescript
const merged = await window.electronAPI.mergeSpeakerText?.(
  diar.segments,
  result.text,
  durationSeconds || 0 // Passes 0 when durationSeconds is undefined
);
```

When `mergeSpeakerText` receives `duration = 0`, `mergeSpeakersWithText` falls back to `maxSegmentEnd` (the end timestamp of the last detected speech segment). If the audio file has trailing music, silence, or outro after the last detected speech segment, the total duration used for character-proportion calculations is truncated, skewing the sentence midpoint calculations and misaligning speaker attribution across the transcript.

### To Reproduce

1. Call `transcribeFileWithSpeakers` without specifying `durationSeconds`, where the diarizer returns `durationSeconds = 120` and `segments` ending at `80s`.
2. Inspect the arguments received by `window.electronAPI.mergeSpeakerText`.
3. Notice the duration argument is `0` instead of `120`.

### Expected behavior

`mergeSpeakerText` should receive the measured duration (`measuredDuration || 0`) so sentence midpoints accurately reflect the full duration of the audio.

### Environment

- OpenWhispr: main (v1.9.2+)
- OS: All platforms
- Node.js: >= 24

### User impact

Imported audio files with diarization enabled experience misaligned speaker attribution when audio duration is measured by the diarizer rather than passed in advance.

### Root-cause hypothesis

`transcribeFileWithSpeakers` passes the raw `durationSeconds` argument instead of `measuredDuration` to `mergeSpeakerText`.

### Proposed focused scope

Update `transcribeFileWithSpeakers` in `src/services/fileTranscription.ts` to pass `measuredDuration || 0` to `window.electronAPI.mergeSpeakerText`, and add a regression test in `test/services/fileTranscriptionTimestamps.test.js`.

hsusul:
## Description

Fixes audio duration propagation in `transcribeFileWithSpeakers` so that when an audio file's duration is measured by the local diarizer (rather than passed upfront), the measured duration is passed to `mergeSpeakerText`.

### Root Cause
When transcribing audio files with speaker diarization enabled, the caller often does not provide an upfront duration (`durationSeconds` is `undefined`). The diarizer measures the converted audio file and populates `measuredDuration` from `diar.durationSeconds`.

However, `mergeSpeakerText` was called with `durationSeconds || 0` instead of `measuredDuration || 0`. Because `0` was passed to the main-process merge handler, `mergeSpeakersWithText` fell back to `maxSegmentEnd` (the end timestamp of the last detected speech segment). When the audio file had trailing silence, music, or outro past the last detected speech segment, sentence midpoints were scaled against a truncated duration, skewing the sentence-to-speaker assignment across the transcript.

### Fix
Pass `measuredDuration || 0` to `window.electronAPI.mergeSpeakerText` in `transcribeFileWithSpeakers`.

## Verification
- Added test coverage in `test/services/fileTranscriptionTimestamps.test.js` confirming that the diarizer's measured duration flows to `mergeSpeakerText`.
- Ran tests: `node --test test/services/fileTranscriptionTimestamps.test.js` (passed)

Closes #1997

### DIFF
--- a/src/services/fileTranscription.ts
+++ b/src/services/fileTranscription.ts
@@ -210,7 +210,7 @@ export async function transcribeFileWithSpeakers(
     const merged = await window.electronAPI.mergeSpeakerText?.(
       diar.segments,
       result.text,
-      durationSeconds || 0
+      measuredDuration || 0
     );
     if (merged?.success && merged.text) return { ...result, text: merged.text };
   } catch {
--- a/test/services/fileTranscriptionTimestamps.test.js
+++ b/test/services/fileTranscriptionTimestamps.test.js
@@ -61,4 +61,25 @@ test("timestamps opt-in reaches the BYOK IPC call and segments flow back", async
   assert.equal(receivedOptions.timestamps, true);
   assert.deepEqual(withSpeakers.segments, segments);
   assert.equal(withSpeakers.durationSeconds, 120);
+
+  let mergedDurationPassed = null;
+  window.electronAPI.diarizeAudioFile = async () => ({
+    success: true,
+    durationSeconds: 150,
+    segments: [{ speaker: "speaker_0", start: 0, end: 100 }],
+  });
+  window.electronAPI.mergeSpeakerText = async (_segments, text, duration) => {
+    mergedDurationPassed = duration;
+    return { success: true, text: `[Speaker 1] ${text}` };
+  };
+
+  const localDiarizeConfig = { ...byokConfig(), cloudTranscriptionProvider: "groq" };
+  const withMeasuredDuration = await transcribeFileWithSpeakers(
+    "/tmp/audio.webm",
+    localDiarizeConfig,
+    { enabled: true, localModelsReady: true, numSpeakers: null }
+  );
+  assert.equal(mergedDurationPassed, 150);
+  assert.equal(withMeasuredDuration.durationSeconds, 150);
+  assert.equal(withMeasuredDuration.text, "[Speaker 1] Hello.");
 });


gabrielste1n:
Thanks, @hsusul. You're right that picked and dropped files reach the merge with a duration of 0 (`UploadAudioView` only sets `durationSeconds` on the URL path, and `batchQueueStore` only fills it from a URL download), and that the diarizer's measured duration is available at that point.

We're closing this anyway, because the effect on attribution runs the other way from the description. `mergeSpeakersWithText` in `src/helpers/speakerMerge.js` places each sentence at its character-proportional position on `[0, totalDuration]` and picks the nearest segment, and transcript text only covers speech. When a file has trailing silence or outro music, the `maxSegmentEnd` fallback spreads the text over the actual speech span, which is the accurate scale; the full file duration stretches it over the silent tail instead. Example: speaker A 0 to 40 s, speaker B 40 to 80 s, silence 80 to 120 s, ten equal sentences. With 80 the split is 5/5; with 120 the midpoints land at 6, 18, 30, 42 ... 114, so three sentences go to A and seven to B. So for the scenario in the issue the change makes attribution worse, it is a no-op without trailing non-speech, and it only helps when the diarizer misses the end of speech.

The test passes and would fail without the change, but it checks which number is forwarded rather than how speakers end up labelled. A change that maps characters onto total speech time (the sum of segment durations) would remove the need for a file-level duration on both paths; that would be welcome as a separate PR.

PR #2000, opened 2026-09-02
fix(utils): prevent duplicate date group headers in groupItemsByDate (#1999)

### LINKED ISSUE #1999: fix(utils): prevent duplicate date group headers in groupItemsByDate
### Affected component

`src/utils/dateGrouping.ts` (`groupItemsByDate`)

### Describe the bug

`groupItemsByDate` groups list items (such as conversations in `ConversationList` and notes in `OverviewNoteList`) into date buckets: Today, Yesterday, Previous 7 days, and Older.

Currently, it creates a new date group whenever the calculated bucket label does not match the previous item's label:

```typescript
if (!current || current.label !== label) {
  current = { label, items: [] };
  groups.push(current);
}
current.items.push(item);
```

If items are not strictly monotonically ordered by timestamp (for instance, when notes or conversations are updated concurrently, reordered, or contain mixed timestamps), `groupItemsByDate` creates duplicate date groups with the same label (e.g. two separate "Today" sections: `[{ label: "Today", items: [...] }, { label: "Yesterday", items: [...] }, { label: "Today", items: [...] }]`).

When rendered in `ConversationList` and `OverviewNoteList`, this causes duplicate section headers, broken visual hierarchy, and duplicate key warnings in the virtualized list.

### To Reproduce

1. Call `groupItemsByDate` with items where items in the same date bucket are non-consecutive (e.g. `[todayItem1, yesterdayItem, todayItem2]`).
2. Observe that the returned array contains multiple groups with `label: "Today"`.

### Expected behavior

`groupItemsByDate` should consolidate all items for each date bucket into a single group under the canonical date order (Today -> Yesterday -> Previous 7 days -> Older) while preserving relative item ordering within each bucket.

### Environment

- OpenWhispr: main (v1.9.2+)
- OS: All platforms
- Node.js: >= 24

### User impact

Users viewing conversation history or overview note lists see duplicate date headers (e.g., multiple "Today" groups) if items are updated or reordered.

### Root-cause hypothesis

`groupItemsByDate` only checks if the current item matches the immediately preceding item's bucket rather than collecting into bucket maps.

### Proposed focused scope

Refactor `groupItemsByDate` in `src/utils/dateGrouping.ts` to bucket items into the canonical date categories, and add unit tests in `test/utils/dateGrouping.test.js`.

hsusul:
## Description

Fixes `groupItemsByDate` to collect items into canonical date buckets (Today, Yesterday, Previous 7 days, Older) rather than generating duplicate date group headers when items belonging to the same date bucket are non-consecutive.

### Root Cause
In `src/utils/dateGrouping.ts`, `groupItemsByDate` pushed a new `DateGroup` whenever an item's calculated date bucket label differed from the immediately preceding item's label:

```typescript
if (!current || current.label !== label) {
  current = { label, items: [] };
  groups.push(current);
}
```

If input items were not strictly sorted monotonically by timestamp (e.g., when conversations or overview notes are updated asynchronously, reordered, or contain mixed timestamps), this created duplicate date groups with identical labels (e.g. `[{ label: "Today" }, { label: "Yesterday" }, { label: "Today" }]`), resulting in duplicate headers in `ConversationList` and `OverviewNoteList` as well as duplicate React keys in virtualized lists.

### Fix
Pre-allocate canonical date buckets (Today -> Yesterday -> Previous 7 days -> Older) in a `Map` and append items to their corresponding bucket, then return non-empty groups in canonical order while preserving item order within each group.

## Verification
- Added a regression test in `test/utils/dateGrouping.test.js` confirming non-consecutive items consolidate into single canonical date groups without duplicate headers.
- Ran tests: `node --test test/utils/dateGrouping.test.js` (5/5 passed)

Closes #1999

### DIFF
--- a/src/utils/dateGrouping.ts
+++ b/src/utils/dateGrouping.ts
@@ -23,8 +23,16 @@ export function groupItemsByDate<T>(
   const weekAgo = new Date(today);
   weekAgo.setDate(weekAgo.getDate() - 7);
 
-  const groups: Array<DateGroup<T>> = [];
-  let current: DateGroup<T> | null = null;
+  const bucketOrder = [
+    t("chat.today"),
+    t("chat.yesterday"),
+    t("chat.previousWeek"),
+    t("chat.older"),
+  ];
+  const bucketMap = new Map<string, T[]>();
+  for (const label of bucketOrder) {
+    bucketMap.set(label, []);
+  }
 
   for (const item of items) {
     const date = normalizeDbDate(getDate(item));
@@ -41,11 +49,15 @@ export function groupItemsByDate<T>(
       label = t("chat.older");
     }
 
-    if (!current || current.label !== label) {
-      current = { label, items: [] };
-      groups.push(current);
+    bucketMap.get(label)!.push(item);
+  }
+
+  const groups: Array<DateGroup<T>> = [];
+  for (const label of bucketOrder) {
+    const bucketItems = bucketMap.get(label)!;
+    if (bucketItems.length > 0) {
+      groups.push({ label, items: bucketItems });
     }
-    current.items.push(item);
   }
 
   return groups;
--- a/test/utils/dateGrouping.test.js
+++ b/test/utils/dateGrouping.test.js
@@ -84,3 +84,26 @@ test("SQLite zoneless timestamps are grouped like their UTC instant", async (t2)
   assert.equal(groups.length, 1);
   assert.equal(groups[0].label, "Today");
 });
+
+test("consolidates non-consecutive items into single canonical groups without duplicate headers", async (t2) => {
+  const { groupItemsByDate } = await load();
+  t2.mock.timers.enable({ apis: ["Date"], now: NOON_JUNE_15 });
+
+  const items = [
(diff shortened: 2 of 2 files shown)


gabrielste1n:
Merging, thanks @hsusul. The bug is real and this is the right place to fix it. `updated_at` mixes `CURRENT_TIMESTAMP` and ISO strings between local writes and cloud-synced rows, so SQLite's string `ORDER BY` can interleave days (`2026-09-13T02:00:00.000Z` sorts above `2026-09-13 20:00:00`), and `updateNoteInStore` swaps an edited note in place without re-sorting. Either one yields duplicate `Today`/`Yesterday` headers, and `OverviewNoteList` keys groups by label, so React reported duplicate keys there too. Fixing at the source would need a data migration, so making the grouping order-independent is the right scope.

I pushed one commit on top of yours that keeps the behaviour but indexes into four bucket arrays instead of a `Map` keyed by translated label (no non-null assertions, labels computed once) and rewords the JSDoc, which still said the input was newest-first. Your test fails on main and passes with the change, typecheck is clean, and the full suite is green on main plus this change.

PR #2006, opened 2026-09-03
fix(build): prevent download scripts from silently failing as release count grows past 50

### LINKED ISSUE #2005: Bug: Fresh Windows builds silently skip windows-key-listener download, breaking Hold/Push-to-Talk
# Issue Title

```
Bug: Fresh Windows builds silently skip windows-key-listener download, breaking Hold/Push-to-Talk
```

# Issue Body

**Describe the bug**

Fresh Windows builds from source produce an application where Hold (Push-to-Talk) activation mode does nothing. Tap mode works fine. The hotkey is registered, but holding it has no effect — no recording starts, no visual feedback, nothing happens.

The root cause is that `scripts/download-windows-key-listener.js` silently fails to download `windows-key-listener.exe` during `prebuild:win`. This native binary is required for Hold mode on Windows — without it, `WindowsKeyManager.resolveListenerBinary()` returns `null`, no native key listener process is spawned, and the hold keydown/keyup events are never captured.

The download fails because `fetchLatestRelease()` in `scripts/lib/download-utils.js` (line 101) only queries the first 50 GitHub releases via `GET /releases?per_page=50`. The release `windows-key-listener-v1.0.0` is now at position 56 of 88 total releases, so it has fallen off the first page and the prefix search returns `null`. The script then exits with **exit code 0** and a console warning — it does not fail the build.

This same pagination issue will eventually affect 5 other download scripts that use `{ tagPrefix: "..." }`:

| Script | Tag | Position (of 50) | Status |
|---|---|:---:|---|
| `download-windows-key-listener.js` | `windows-key-listener-v1.0.0` | 56 | ❌ **Already broken** |
| `download-windows-fast-paste.js` | `windows-fast-paste-v1.0.0` | 42 | ⚠️ Breaks in ~8 more releases |
| `download-text-monitor.js` | `windows-text-monitor-v1.0.0` | 34 | ⚠️ Breaks in ~16 more releases |
| `download-meeting-aec-helper.js` | `meeting-aec-helper-v1.0.0` | 23 | ⚠️ Breaks in ~27 more releases |
| `download-windows-mic-listener.js` | `windows-mic-listener-v1.1.0` | 5 | ✅ Safe for now |
| `download-windows-system-audio-helper.js` | `windows-system-audio-helper-v1.1.0` | 2 | ✅ Safe for now |

**To Reproduce**

1. Clone the repository fresh (no cached `resources/bin/` binaries).
2. Run `node scripts/download-windows-key-listener.js` and observe the output:
   ```
   [windows-key-listener] Fetching latest release...
   [windows-key-listener] Could not find a release matching prefix: windows-key-listener-v
   [windows-key-listener] Push-to-Talk will use fallback mode (compile locally or tap mode)
   ```
3. Run `npm run prebuild:win && npm run build:renderer && electron-builder --win` — the build completes with exit code 0 (no error).
4. Launch the built application from `dist\win-unpacked\OpenWhispr.exe`.
5. Open Settings → set activation mode to **Hold**.
6. Hold the dictation hotkey — nothing happens. Release it — nothing happens.
7. Switch activation mode back to **Tap** — tap works fine.

**Expected behavior**

`windows-key-listener.exe` should be downloaded successfully regardless of how many releases exist in the repository. Hold/Push-to-Talk should work on fresh builds.

- OS: Windows 10 / Windows 11
- Version: 1.9.2 (also affects all prior versions when built from source with >50 releases on the repository)

**Additional context**

The fix is straightforward: switch all affected download scripts from `{ tagPrefix: "..." }` (paginated prefix search across 50 releases) to `{ tag: "exact-tag-name" }` (direct lookup via `/releases/tags/{tag}`). The `fetchLatestRelease` function in `download-utils.js` already supports this — it uses a completely different GitHub API endpoint that requires no pagination and always resolves regardless of release count.

```javascript
// Before (broken — depends on 50-release pagination window)
const release = await fetchLatestRelease(REPO, { tagPrefix: "windows-key-listener-v" });

// After (fixed — direct tag lookup, always works)
const release = await fetchLatestRelease(REPO, { tag: "windows-key-listener-v1.0.0" });
```

The reason this hasn't been noticed in CI or by existing developers is that:
1. The script exits with code 0 (no build failure).
2. Developers who have built previously already have the binary cached in `resources/bin/` — the script early-exits when the file exists.
3. Developers with MSVC or MinGW installed have the `compile:winkeys` step build it locally from source as a fallback.

B0-KA:
# PR Title

```
fix(build): prevent download scripts from silently failing as release count grows past 50
```

# PR Body

### Problem

Closes #2005

Download scripts for prebuilt Windows helper binaries use `fetchLatestRelease(REPO, { tagPrefix: "..." })` to find their release on GitHub. This calls:

```
GET /repos/OpenWhispr/openwhispr/releases?per_page=50
```

…and searches the first page (50 releases) for a tag matching the prefix. As the repository has grown to 88+ releases, older binary releases have fallen past position 50 and can no longer be found.

**`windows-key-listener-v1.0.0`** (position 56) is already unreachable, causing Hold/Push-to-Talk to silently break on any fresh Windows build from source. The remaining helper binaries will break one by one as more releases are published.

### Fix

Switch all affected scripts from `{ tagPrefix: "..." }` (paginated prefix search) to `{ tag: "exact-tag" }` (direct tag lookup). The `fetchLatestRelease` function in `download-utils.js` already supports this via the GitHub endpoint `GET /repos/{owner}/{repo}/releases/tags/{tag}`, which returns the release directly without pagination.

### Changes

| File | Change |
|---|---|
| `scripts/download-windows-key-listener.js` | Pin to `windows-key-listener-v1.0.0`, use `{ tag }` |
| `scripts/download-windows-fast-paste.js` | Pin to `windows-fast-paste-v2.0.0`, use `{ tag }` |
| `scripts/download-windows-system-audio-helper.js` | Pin to `windows-system-audio-helper-v1.1.0`, use `{ tag }` |
| `scripts/download-meeting-aec-helper.js` | Pin to `meeting-aec-helper-v1.0.0`, use `{ tag }` |
| `scripts/download-text-monitor.js` | Pin to `windows-text-monitor-v1.0.0` and `linux-text-monitor-v1.0.0`, use `{ tag }` |
| `scripts/download-windows-mic-listener.js` | Switch from `{ tagPrefix }` to `{ tag }` (already reads exact version from C source via `readPinnedTag()`) |

### How It Was Verified

1. Deleted all helper binaries from `resources/bin/`.
2. Ran each download script individually — all 6 downloaded successfully via the exact tag endpoint.
3. Ran a full clean build (`npm run prebuild:win && npm run build:renderer && electron-builder --win`).
4. Confirmed `windows-key-listener.exe` appears in the signing log (previously absent):
   ```
   • signing with signtool.exe  path=dist\win-unpacked\resources\bin\windows-key-listener.exe
   ```
5. Launched the built application, switched to Hold mode, verified Push-to-Talk works.

### Before (broken)

```
[windows-key-listener] Fetching latest release...
[windows-key-listener] Could not find a release matching prefix: windows-key-listener-v
[windows-key-listener] Push-to-Talk will use fallback mode (compile locally or tap mode)
```

### After (fixed)

```
[windows-key-listener] Fetching latest release...
[windows-key-listener] Successfully downloaded windows-key-listener-v1.0.0 (146KB)
```

### DIFF
--- a/scripts/download-meeting-aec-helper.js
+++ b/scripts/download-meeting-aec-helper.js
@@ -18,7 +18,7 @@ const {
 } = require("./lib/download-utils");
 
 const REPO = "OpenWhispr/openwhispr";
-const TAG_PREFIX = "meeting-aec-helper-v";
+const PINNED_TAG = "meeting-aec-helper-v1.0.0";
 const VERSION_OVERRIDE = process.env.MEETING_AEC_HELPER_VERSION || null;
 const BIN_DIR = path.join(__dirname, "..", "resources", "bin");
 
@@ -105,7 +105,7 @@ async function main() {
     console.log("\n[meeting-aec-helper] Fetching latest release...");
   }
   const release = await fetchLatestRelease(REPO, {
-    tagPrefix: VERSION_OVERRIDE || TAG_PREFIX,
+    tag: VERSION_OVERRIDE || PINNED_TAG,
   });
 
   if (!release) {
--- a/scripts/download-text-monitor.js
+++ b/scripts/download-text-monitor.js
@@ -22,7 +22,7 @@ const BIN_DIR = path.join(__dirname, "..", "resources", "bin");
 const PLATFORM_CONFIG = {
   linux: {
     label: "linux-text-monitor",
-    tagPrefix: "linux-text-monitor-v",
+    pinnedTag: "linux-text-monitor-v1.0.0",
     archiveName: "linux-text-monitor-linux-x64.tar.gz",
     binaryName: "linux-text-monitor",
     versionEnv: "LINUX_TEXT_MONITOR_VERSION",
@@ -30,7 +30,7 @@ const PLATFORM_CONFIG = {
   },
   win32: {
     label: "windows-text-monitor",
-    tagPrefix: "windows-text-monitor-v",
+    pinnedTag: "windows-text-monitor-v1.0.0",
     archiveName: "windows-text-monitor-win32-x64.zip",
     binaryName: "windows-text-monitor.exe",
     versionEnv: "WINDOWS_TEXT_MONITOR_VERSION",
@@ -45,7 +45,7 @@ async function main() {
     return;
   }
 
-  const { label, tagPrefix, archiveName, binaryName, versionEnv, compileHint } = config;
+  const { label, pinnedTag, archiveName, binaryName, versionEnv, compileHint } = config;
   const versionOverride = process.env[versionEnv] || null;
   const forceDownload = process.argv.includes("--force");
   const outputPath = path.join(BIN_DIR, binaryName);
@@ -61,11 +61,11 @@ async function main() {
   } else {
     console.log(`\n[${label}] Fetching latest release...`);
   }
-  const tagToFind = versionOverride || tagPrefix;
-  const release = await fetchLatestRelease(REPO, { tagPrefix: tagToFind });
+  const tagToFind = versionOverride || pinnedTag;
+  const release = await fetchLatestRelease(REPO, { tag: tagToFind });
 
   if (!release) {
-    console.error(`[${label}] Could not find a release matching prefix:`, tagPrefix);
+    console.error(`[${label}] Could not find release:`, tagToFind);
     console.log(`[${label}] Auto-learn correction monitoring will be disabled`);
     return;
   }
--- a/scripts/download-windows-fast-paste.js
+++ b/scripts/download-windows-fast-paste.js
@@ -21,7 +21,7 @@ const {
 } = require("./lib/download-utils");
 
 const REPO = "OpenWhispr/openwhispr";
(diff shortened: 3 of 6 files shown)


boseq:
Thanks @B0-KA, and sorry for the overlap. Your write-up in #2005 was spot on, and this PR fixes it too: with the exact tags, the key listener downloads again on a fresh Windows build.

We're closing it in favour of #2147, which fixes the same 50-release limit once inside `fetchLatestRelease`. That way every helper that looks up its release by tag prefix keeps working without a tag bump in each script. #2147 now also fails a Windows build that's missing `windows-key-listener.exe`, so this can't ship quietly again. The Hold side of your report is in #2239: when the listener is missing, the app now disables Hold and refuses hotkeys that can't fire instead of silently accepting them.

PR #2011, opened 2026-09-03
feat: add analytics leaderboard

boseq:
## Summary

- place **Leaderboard** inside **Insights** as a peer tab to **Your Usage**, preserving sidebar space
- keep **Insights Sync** as its own local analytics-consent setting and **Leaderboard participation** as a separate account preference
- when a user joins while Sync is off, obtain explicit Sync consent first (including when no rows are queued), then join without coupling either setting's later lifecycle
- unlock leaderboard access for every signed-in plan through workspace, invitation, and verified-domain relationships rather than billing gates
- cover the setup funnel: create a workspace, accept an invitation, request to join a same-domain workspace, or invite teammates when the viewer administers a workspace
- render production rankings with a top-three podium, paginated table, five metrics, weekly history, rank jump, multiple scopes, and an hourly client refresh interval
- add sparse-board invitation nudges plus viewer-only image copy/download and X/LinkedIn sharing

## Production behavior

- the Join surface discloses that account participation publishes name, email, and activity to every eligible workspace and work-domain leaderboard; transcripts remain private
- turning Insights Sync on or off never joins or leaves leaderboards; leaving a leaderboard never disables Insights Sync
- a failed participation read shows a retry surface, and a failed join preserves any accepted or pre-existing Sync consent
- simultaneous participation reads are coalesced, duplicate joins are suppressed, and account changes cannot apply a stale response to the replacement account
- a server-confirmed Join survives same-account token rotation; only ambiguous request failures retain the compensating Leave
- an offline leaderboard leave is stored under an independent per-account key and retried later without gating analytics uploads
- analytics queue reads, cloud requests, and acknowledgements are pinned to one account/auth generation; revoking consent stops the next batch of an active multi-batch drain
- failures to count pending rows cannot silently bypass the consent prompt, and a failed foreground upload is not reported as synchronized
- the personal Insights summary keeps using the device IANA time zone; leaderboard weeks and streaks use the API's UTC boundary
- successful API payloads are validated before render, policy failures are non-retryable, and expired/account-required responses offer sign-in
- share exports contain only the viewer's rank and participant total; teammate names, activity values, workspace names, and employer domains are not exported
- an older API that lacks the participation GET is treated as unconfigured so existing Insights uploads continue during a staggered rollout

## Deployment order

1. Apply API migration 065.
2. Deploy [OpenWhispr/openwhispr-api#185](https://github.com/OpenWhispr/openwhispr-api/pull/185).
3. Release this Desktop change.

The API writes the new `client_platform` column, so it must not run before migration 065. Desktop has a compatibility fallback for an older API's missing participation GET, but the leaderboard itself requires the new API routes.

## Security and privacy

- access is derived server-side; Desktop never supplies members or ranking values
- workspace growth actions are shown only when the viewer can create a workspace or administer an existing one; leaderboard UI no longer redirects ordinary members to referral rewards
- account/domain/workspace response counts describe leaderboard participants, not all OpenWhispr accounts at a domain
- local Sync consent is isolated at account boundaries so a replacement signed-in account does not inherit the prior account's setting

## Known product decisions

- leaderboard participation is one account preference shared by every eligible leaderboard scope; the Join disclosure names that scope explicitly
- the API aggregates leaderboard data on request and sends an hourly refresh hint; there is no server response cache or rollup table in this PR

## Verification

- `npm run format:check`
- `npm run typecheck`
- `npm run i18n:check`
- `npm run build:renderer`
- `npm test` — 3,630 tests: 3,433 passed, 0 failed, 196 environment-skipped, 1 existing todo
- focused consent, account-switch, upload-fencing, participation, and durable-leave regressions — 59/59 passed


Chadpiha (approved):
Approving.

**Verified:** 3,693 tests / 0 failures, `quality-check`, `i18n:check`, `npm audit`, all four
platform builds, and CodeQL clean after the assertion rewrite in `leaderboardStates.test.js`.
Branch merged (not rebased) with `main` and currently 0 behind.

**Blockers fixed on-branch:** an undeliverable opt-out permanently disabling the sync throttle;
domain rosters carrying member email addresses; and a participation write whose outcome was
misclassified — a 2xx with an unreadable body was silently reverting a join that had actually
landed. Also fixed: the solo-scope join dead-end, page-turn blanking, and the unvalidated-session
spinner with no way out.


**Knowingly shipping with:**
- Consent is one global switch covering both workspace and work-domain boards, and opting in publishes the account's full uploaded history retroactively. Product decision, not a defect —
  worth settling before real users opt in.
- No manual/E2E pass on the leaderboard surfaces yet.

Disclosure: several of the fixes above are my own commits on this branch.

PR #2012, opened 2026-09-03
fix(dictation): keep Linux floating controls usable

vsem-azamat:
## Problem

On Linux, the floating dictation controls could become unreachable after the pointer left the window:

- Electron ignores the `forward` option passed to `setIgnoreMouseEvents` on Linux. Once the window returned to click-through mode, it stopped receiving the next `mouseenter` needed to make it interactive again.
- The hover-only cancel button participated in the right-anchored flex layout. Revealing it widened the group and moved the recording pill 36 px out from under the pointer.
- Long dictation errors expanded the native window toward the full work area and had no unconditional dismiss control.

## Fix

- Keep normal hit-testing enabled for the Linux floating window, matching the existing Windows reliability fallback.
- Position the cancel button outside layout flow on the display-inward side; bridge its hover region without changing pill geometry.
- Add an always-visible dismiss button to dictation errors.
- Cap the error surface at 240 px and scroll overflow inside it; size the native window from the rendered capped height.

## Linux hit-testing trade-off

The compact Linux window remains hit-testable because Electron cannot forward pointer movement from an ignored window there. This matches the existing Windows fallback; the pill-state bounds remain limited to `176 × 120 px` normally and `208 × 120 px` with the hover cancel visible.

### Stable recording control

Recording pill hover geometry before and after

- Before: pill `x = 104 → 68` on hover (`−36 px`).
- After: pill `x = 104 → 104` on hover.

### Bounded, dismissible error

Long dictation error before and after

- Before: long error expanded the native window to `466 × 1080 px`; no dismiss button.
- After: the same payload is capped at `466 × 264 px`, scrolls internally, and exposes Dismiss.

## Behavior flow

```mermaid
flowchart LR
  subgraph Before
    A[mouseleave] --> B[ignore mouse events + forward]
    B --> C[Linux does not forward mouse movement]
    C --> D[next mouseenter never reaches renderer]
    E[show cancel in flex row] --> F[group width grows]
    F --> G[right-anchored pill shifts 36 px]
    H[long error scrollHeight] --> I[work-area-sized native window]
  end

  subgraph After
    J[Linux floating window] --> K[normal hit-testing stays enabled]
    L[absolute cancel on inward side] --> M[pill geometry stays fixed]
    N[rendered height capped at 240 px] --> O[internal scroll + Dismiss]
  end
```

## Verification

- Targeted helper/window tests: `45 passed`.
- TypeScript typecheck: passed.
- Renderer production build: passed.
- Actual Linux Electron window:
  - repeated hover cycles kept cancel reachable;
  - pill `x` stayed at `104` before and during hover;
  - Dismiss removed the error;
  - long error window capped at `466 × 264 px`;
  - normal pill restored to `176 × 120 px`.


vsem-azamat:
Closing: the core Linux hover fix landed upstream in #2161 (v1.10.1). Thanks!

PR #1521, opened 2026-08-10
fix(cli): respond with HTTP 400 validation_error on route validation errors

hsusul:
## Summary
Fixes an issue where CLI bridge routes that throw a validation error (with `err.code === "VALIDATION"`) trigger a generic 500 `internal_error` response instead of an HTTP 400 `validation_error` response.

## Problem & Root Cause
In `src/helpers/cliBridge.js`, helper validation functions (such as `requireWordList`, search query validation in `/v1/notes/search`, and `/v1/dictionary/update`) throw errors annotated with `err.code = "VALIDATION"`.

However, `CliBridge._sendError(res, err)` only checked `if (err.code === "NOT_FOUND")`. Unhandled validation errors were passed to the default 500 handler (`sendV1Error(res, 500, "internal_error", ...)`), causing client-side input validation errors to be reported as HTTP 500 internal server errors.

## Behavior Before & After
- **Before:** A request to `GET /v1/notes/search` without a `q` parameter or `POST /v1/dictionary/update` with invalid parameters returned HTTP status 500 with `{"error":{"code":"internal_error","message":"..."}}`.
- **After:** Returns HTTP status 400 with `{"error":{"code":"validation_error","message":"..."}}`.

## Scope & Non-Goals
- **Scope:** Correct error mapping in `_sendError` for `VALIDATION` error code to return HTTP status 400 and code `"validation_error"`.
- **Non-Goals:** Does not alter route logic, database operations, or existing 404/500 error paths.

## Tests Added
- Added unit tests in `test/helpers/cliBridgeDictionary.test.js` verifying that `_handleRequest` returns HTTP 400 and `validation_error` when route validation fails (e.g. missing search query or invalid payload array).

## Validation Commands & Results
- `ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/cliBridgeDictionary.test.js` (PASSED 9/9)
- `ELECTRON_OVERRIDE_DIST_PATH=/tmp npm test` (PASSED 1103/1103)
- `npm run typecheck` (PASSED)
- `npm run lint` (PASSED)
- `npm run format:check` (PASSED)
- `npm run i18n:check` (PASSED)
- `npm run build:renderer` (PASSED)
- `git diff --check` (PASSED clean)

Fixes #1519

Issue #1519, opened 2026-08-10
fix(cli): CLI bridge responds with HTTP 500 instead of HTTP 400 for route validation errors


hsusul:
### Component
CLI Bridge (`src/helpers/cliBridge.js`)

### Description
When a CLI bridge request triggers a route validation error (such as a missing search query on `GET /v1/notes/search` or invalid parameters on `POST /v1/dictionary/update`), route helpers throw an error with `err.code = "VALIDATION"`. However, `_sendError(res, err)` in `src/helpers/cliBridge.js` only handles `err.code === "NOT_FOUND"`. Unhandled validation errors fall through to the default 500 handler, causing the bridge to return HTTP status 500 `internal_error` instead of HTTP status 400 `validation_error`.

### Reproduction
1. Send a request to a CLI bridge route that fails validation, e.g. `GET /v1/notes/search` without a `q` parameter.
2. Observe the HTTP response:
   - Status code: `500`
   - Body: `{"error":{"code":"internal_error","message":"Search query is required"}}`

### Expected Behavior
Route validation errors carrying `err.code === "VALIDATION"` should return:
   - Status code: `400`
   - Body: `{"error":{"code":"validation_error","message":"Search query is required"}}`

### Actual Behavior
Returns status code `500` with error code `"internal_error"`.

### Environment
- OpenWhispr main branch (1.8.1)
- Platform: macOS / Linux / Windows

### User Impact
Clients consuming the CLI bridge receive server error 500 responses when submitting invalid request parameters, misclassifying client-side input errors as internal application crashes.

### Proposed Fix
Update `_sendError` in `src/helpers/cliBridge.js` to handle `err.code === "VALIDATION"` by sending HTTP 400 with `"validation_error"`. Add regression tests covering HTTP validation responses across routes.

I intend to submit a PR to fix this issue.

PR #2015, opened 2026-09-03
fix(cli): validate query parameters limit and folder_id in CLI bridge endpoints (#2014)

### LINKED ISSUE #2014: fix(cli): validate query parameters limit and folder_id in CLI bridge endpoints
### Affected component

`src/helpers/cliBridge.js`

### Describe the bug

In `src/helpers/cliBridge.js`, endpoints that accept integer query parameters (`GET /v1/notes/list`, `GET /v1/notes/search`, and `GET /v1/transcriptions/list`) use unguarded `Number(query.get(...))`:

```javascript
const limit = query.get("limit") ? Number(query.get("limit")) : 100;
const folderId = query.get("folder_id") ? Number(query.get("folder_id")) : null;
```

When a client passes non-numeric or non-positive values (e.g. `?limit=abc`, `?limit=-1`, `?folder_id=abc`):
1. `Number("abc")` produces `NaN`.
2. In `getNotes(noteType, limit, folderId)`, `folderId != null` evaluates to `true` for `NaN`. SQLite executes `WHERE folder_id = NaN`, which matches no rows and silently returns an empty array with HTTP 200 rather than rejecting the invalid request.
3. Passing non-positive or `NaN` limits passes `NaN` or negative integers directly into the database query's `LIMIT ?` clause.

In accordance with the CLI bridge error contract (#1519 / #1521), invalid parameters should be rejected with HTTP 400 (`validation_error`).

### To Reproduce

1. Send `GET /v1/notes/list?limit=invalid` or `GET /v1/notes/list?folder_id=invalid` to the CLI bridge.
2. The server responds with HTTP 200 and `{ data: [] }` rather than HTTP 400 `validation_error`.

### Expected behavior

Invalid query parameters for `limit` and `folder_id` should throw a `VALIDATION` error resulting in HTTP 400 `validation_error`.

### Environment

- OpenWhispr: main (v1.9.2+)
- OS: All platforms
- Node.js: >= 24

### User impact

CLI and automation scripts supplying malformed or non-integer query parameters silently receive empty responses or run malformed database queries rather than receiving an actionable HTTP 400 validation error.

### Proposed focused scope

Add a `parsePositiveIntQuery(query, key, defaultValue)` helper to `src/helpers/cliBridge.js` that validates integer query parameters and throws a `VALIDATION` error if malformed, and add unit test coverage in `test/helpers/cliBridgeQueryValidation.test.js`.

hsusul:
## Description

Validates positive integer query parameters (`limit` and `folder_id`) in the CLI bridge endpoints (`GET /v1/notes/list`, `GET /v1/notes/search`, and `GET /v1/transcriptions/list`) to return HTTP 400 (`validation_error`) when invalid values are provided.

### Root Cause
In `src/helpers/cliBridge.js`, endpoints converted query parameters directly via `Number(query.get(...))`.
When non-integer or non-numeric values were passed (e.g. `?limit=abc`, `?limit=-1`, `?folder_id=abc`):
1. `Number("abc")` evaluated to `NaN`.
2. In `getNotes(noteType, limit, folderId)`, `folderId != null` evaluated to `true` for `NaN`. The SQL query executed `WHERE folder_id = NaN`, matching no rows and silently returning an empty array with HTTP 200 rather than rejecting the invalid query.
3. `limit` values of `NaN` or negative numbers flowed unguarded into SQLite `LIMIT ?` clauses.

### Fix
Added `parsePositiveIntQuery(query, key, fallback)` helper to validate that query parameters, when supplied, parse to positive integers; otherwise, it throws a `VALIDATION` error that maps to HTTP 400 (`validation_error`), aligning with the CLI bridge error contract (#1519 / #1521).

## Verification
- Added unit tests in `test/helpers/cliBridgeQueryValidation.test.js` covering valid parameters, default values, non-numeric strings, negative integers, zero, and floats across all three endpoints.
- Ran tests: `node --test test/helpers/cliBridgeQueryValidation.test.js` (6/6 passed)

Closes #2014

### DIFF
--- a/src/helpers/cliBridge.js
+++ b/src/helpers/cliBridge.js
@@ -84,6 +84,18 @@ function parseIdParam(value) {
   return id;
 }
 
+function parsePositiveIntQuery(query, key, fallback) {
+  const raw = query.get(key);
+  if (raw === null || raw === "") return fallback;
+  const num = Number(raw);
+  if (!Number.isInteger(num) || num <= 0) {
+    const err = new Error(`Query parameter '${key}' must be a positive integer`);
+    err.code = "VALIDATION";
+    throw err;
+  }
+  return num;
+}
+
 function unwrapMutationResult(result, label) {
   if (!result?.success || !result[label]) {
     throw new Error(result?.error || `Failed to write ${label}`);
@@ -300,8 +312,8 @@ class CliBridge {
       exact("GET", "/v1/health", () => ({ data: { ok: true, version: 1 } })),
       exact("GET", "/v1/notes/list", ({ query }) => {
         const noteType = query.get("note_type") || null;
-        const limit = query.get("limit") ? Number(query.get("limit")) : 100;
-        const folderId = query.get("folder_id") ? Number(query.get("folder_id")) : null;
+        const limit = parsePositiveIntQuery(query, "limit", 100);
+        const folderId = parsePositiveIntQuery(query, "folder_id", null);
         const notes = db.getNotes(noteType, limit, folderId);
         return { data: notes, has_more: false, next_cursor: null };
       }),
@@ -312,7 +324,7 @@ class CliBridge {
           err.code = "VALIDATION";
           throw err;
         }
-        const limit = query.get("limit") ? Number(query.get("limit")) : 20;
+        const limit = parsePositiveIntQuery(query, "limit", 20);
         const notes = db.searchNotes(q, limit);
         return { data: notes, has_more: false, next_cursor: null };
       }),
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Merged, thanks @hsusul. The problem is real: the CLI forwards `--folder <id>` as-is and only rejects `--limit` when it is negative or non-finite, so `--limit 1.5` reached the bridge and SQLite raised "datatype mismatch" as a 500, `--folder abc` silently returned an empty list, and a raw `?limit=-1` removed the row cap. The cloud API already answers these with a 400, so the local bridge now matches.

One small follow-up landed separately in #2184 so that `parsePositiveIntQuery` reuses `parseIdParam` instead of re-implementing the same check two lines above it. Behaviour and your tests are unchanged, and the full suite is green.

PR #2017, opened 2026-09-03
fix(meeting): ignore late stderr from inactive child in WindowsLoopbackAudioManager (#2016)

### LINKED ISSUE #2016: fix(meeting): ignore late stderr from inactive child in WindowsLoopbackAudioManager
### Affected component

`src/helpers/windowsLoopbackAudioManager.js`

### Describe the bug

In `WindowsLoopbackAudioManager.prototype.start`, stdout and process lifecycle event handlers guard against callbacks from replaced or stopped processes:

```javascript
child.stdout.on("data", (chunk) => {
  if (this.process !== child) return;
  this.onChunk?.(chunk);
});

child.on("error", (error) => {
  if (this.process === child) {
    this.process = null;
  }
  ...
});
```

However, `child.stderr.on("data")` does not check whether `this.process === child`:

```javascript
child.stderr.on("data", (chunk) => {
  this._consumeStderr(chunk, (message) => {
    ...
  });
});
```

When `stop()` is called or when a capture session is restarted, `this.process` is cleared or replaced with a new child. If the terminating helper flushes stderr lines during process teardown, its stderr output continues to be consumed by `this._consumeStderr`. This corrupts `this.stderrBuffer` on the active manager instance and causes spurious `onError` or `onWarning` callbacks to fire on the active/new session.

This contrasts with the macOS counterpart (`src/helpers/audioTapManager.js`), which guards both stdout and stderr with `if (this.process !== child) return;`.

### To Reproduce

1. Call `windowsLoopbackManager.start(...)`.
2. Stop the session via `windowsLoopbackManager.stop()`.
3. Emit a stderr chunk from the old process.
4. Observe that the old process's stderr is parsed and processed on the manager instance.

### Expected behavior

`child.stderr.on("data")` should check `if (this.process !== child) return;`, matching `stdout` and `audioTapManager.js`.

### Environment

- OpenWhispr: main (v1.9.2+)
- OS: Windows
- Node.js: >= 24

### User impact

Users on Windows using system audio loopback capture can receive unexpected error notices or corrupted stderr buffers if a previous helper process writes to stderr while shutting down.

### Root-cause hypothesis

Missing `if (this.process !== child) return;` check in `child.stderr.on("data")`.

### Proposed focused scope

Add `if (this.process !== child) return;` to `child.stderr.on("data")` in `src/helpers/windowsLoopbackAudioManager.js`, and add a regression test in `test/helpers/windowsLoopbackAudioManager.test.js`.

hsusul:
## Description

Guards `child.stderr.on("data")` with `if (this.process !== child) return;` in `WindowsLoopbackAudioManager.prototype.start` so that late stderr chunks from stopped or replaced helper processes do not corrupt the active manager's state or trigger spurious error/warning callbacks.

### Root Cause
In `src/helpers/windowsLoopbackAudioManager.js`, `child.stdout.on("data")` checks `if (this.process !== child) return;`, but `child.stderr.on("data")` was missing this guard.
When a capture session stopped or restarted (`this.process = null` or replaced with a new child), any lingering stderr output written during helper teardown was fed into `this._consumeStderr`. This corrupted `this.stderrBuffer` on the active instance and invoked `onError` or `onWarning` on the wrong/new session. This aligns with the pattern used in the macOS loopback manager (`src/helpers/audioTapManager.js`).

### Fix
Added `if (this.process !== child) return;` at the top of `child.stderr.on("data")`.

## Verification
- Added a unit test in `test/helpers/windowsLoopbackAudioManager.test.js` confirming that stderr chunks emitted by an inactive child after process teardown or replacement are ignored without invoking `onError` or mutating `stderrBuffer`.
- Ran tests: `node --test test/helpers/windowsLoopbackAudioManager.test.js` (10/10 passed)

Closes #2016

### DIFF
--- a/src/helpers/windowsLoopbackAudioManager.js
+++ b/src/helpers/windowsLoopbackAudioManager.js
@@ -174,6 +174,7 @@ class WindowsLoopbackAudioManager {
       });
 
       child.stderr.on("data", (chunk) => {
+        if (this.process !== child) return;
         this._consumeStderr(chunk, (message) => {
           if (message.type === "start") {
             finish(resolve);
--- a/test/helpers/windowsLoopbackAudioManager.test.js
+++ b/test/helpers/windowsLoopbackAudioManager.test.js
@@ -18,9 +18,12 @@ function setPlatform(platform) {
 
 afterEach(() => setPlatform(originalPlatform));
 
-function loadManagerClass() {
+function loadManagerClass(mocks = {}) {
   delete require.cache[managerModulePath];
   Module._load = function loadWithMocks(request, parent, isMain) {
+    if (mocks[request]) {
+      return mocks[request];
+    }
     if (request === "./debugLogger") {
       return {
         info() {},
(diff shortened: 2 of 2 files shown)


gabrielste1n:
Merging as is, thanks @hsusul. The stderr handler was the one stdio path in `src/helpers/windowsLoopbackAudioManager.js` without the `this.process !== child` guard that the stdout handler and the macOS tap manager already have, and the condition is reachable: the exit handler nulls `this.process` on every stop and crash while Node can still deliver buffered stderr afterwards. Your test fails against main and passes with the guard, and the full suite is green on main plus this change.

For the record, the graceful stop paths already null the callbacks before late stderr can arrive, so the case this fixes is a helper crash where exit lands before the final flushed JSON error, which produced a duplicate error toast. `src/helpers/linuxPortalAudioManager.js` has the same unguarded handler if you'd like a follow-up.

PR #2018, opened 2026-09-03
fix(tinfoil): move the default chat model from glm-5-2 to glm-5-3

greatcoat:
**Problem** — Tinfoil emailed a deprecation notice: GLM 5.2 is replaced by GLM 5.3, action required by 2026-09-10. The desktop app hardcodes `glm-5-2` as the Tinfoil default and seeds it first in the registry. A live probe of `https://inference.tinfoil.sh/v1/models` today already returns no `glm-5-2` at all (chat ids served: `deepseek-v4-flash`, `glm-5-3`, `glm-5-3-flash`, `kimi-k3`, `llama3-3-70b`, `gemma4-31b`, `gpt-oss-120b`).

**Root cause** — `pickDefaultTinfoilModel` looks up a constant id and falls back to `models[0]`. With the constant retired, every reconciled Tinfoil selection (and every fresh default) lands on whatever Tinfoil happens to list first, `deepseek-v4-flash` right now, instead of the intended flagship. New installs also show the retired `glm-5-2` seed until the first catalog fetch.

**Fix**
- `DEFAULT_MODEL_ID` → `glm-5-3`; description key map follows.
- Registry seed entry → `glm-5-3` / `GLM-5.3`, description updated to Tinfoil's current copy (agentic coding, 1M context).
- `tinfoil_glm_5_2` → `tinfoil_glm_5_3` in all ten locales with the updated description.
- Catalog test fixture renamed for consistency.

Deliberately untouched:
- `reconcileSelectedModels` and the `models[0]` fallback — the fallback is correct when the default itself is ever retired; only the constant was stale.
- `glm-5-3-flash`, `kimi-k3`, `deepseek-v4-flash` — the catalog is fetched live so they already appear; adding seed entries and translations for every served model is a separate decision.
- Persisted user selections — the existing live-catalog reconcile already migrates a stored `glm-5-2` to the default on the next hourly fetch, now to `glm-5-3`.

**Tests** — new `test/helpers/tinfoilDefaultModel.test.js`, written first and watched fail on `origin/main`:
```
actual: 'deepseek-v4-flash', expected: 'glm-5-3'   (default lookup)
actual: 'glm-5-2',           expected: 'glm-5-3'   (registry seed)
```
Covers: default picked even when listed second, `models[0]` fallback still works when the default is gone, seed leads with the default and no longer ships `glm-5-2`. The existing translation-coverage test guards that the renamed key resolves in `en` and all other languages.

Full suite: 3457 tests, 3266 pass, 187 skipped (Electron-ABI DB tests), 3 failures are the known local-environment set unrelated to this change (`activateTargetPid`, window-rect parse, bare `localStorage` global on Node 25). Prettier and eslint clean on touched files.

**Provenance** — Tinfoil deprecation email from Sacha, 2026-09-03.


Chadpiha (approved):
## Approving ✅

Reviewed end-to-end and independently verified the premise, the failure, and the fix.

### Verified independently

A live probe of `GET https://inference.tinfoil.sh/v1/models` confirms the PR's account: `glm-5-2` and
`deepseek-v4-pro` are both gone, `glm-5-3` is served (`type: "chat"`, `reasoning: true`,
`context_window: 1048576` — so the "1M context" copy is accurate), and the chat list genuinely leads
with `deepseek-v4-flash`, which is what `models[0]` was landing on.

I reproduced the bug and the fix on both sides with a harness test — persisted
`cleanupProvider=tinfoil`, `cleanupModel=glm-5-2`, offline store load, then a catalog refresh:

| | `main` | this branch |
|---|---|---|
| after store load | `glm-5-2` | `glm-5-3` |
| after catalog refresh | `deepseek-v4-flash` ❌ | `glm-5-3` ✅ |

The race the second commit describes is real: `tinfoilClient.ts:34` fires
`void refreshTinfoilModels().catch(() => {})` and returns immediately, so the first request of a
session goes out on the stale model. The startup sweep beats it.

Also checked: no cross-repo coordination needed — the API and admin console policy schemas reference
the `tinfoil` *provider id* only, never model ids. i18n parity holds across all 10 locales, and the
new descriptions are correct idiomatic renderings of the new English, not stale copies of the old.

### Follow-up commit (cd82749a)

Five things the first two commits left open:

- **Retirement coverage.** The table mapped only `glm-5-2`, but `deepseek-v4-pro` and `kimi-k2-6` are
  also retired at Tinfoil, and all three shipped as *seed* entries (the first two through v1.7.6), so
  users who simply took a default can be sitting on one. Each now maps to its own family's successor.
- **The sentinel trap.** The sweep short-circuits on a per-provider sentinel, so an id added under a
  key a released build already writes reaches nobody. That constraint is now documented where the
  table is edited, and snapshotted so the decision lands in the diff.
- **The switch notice.** The sweep repaired selections before `reconcileSelectedModels` could see
  anything stale, silently suppressing the existing "Tinfoil retired X. Switched to Y." toast. The
  sweep now reports what it rewrote and the store announces Tinfoil moves itself, once per model.
- **A crash path.** The sweep wrote at module scope unguarded, unlike `sweepRetiredCustomPrompts`
  directly above it — a rejected `localStorage` write threw out of settings-store module evaluation.
  Now guarded per provider; the sentinel stays unset so the next launch retries.
- **Default resolution.** `defaultLlmModel` and the onboarding step each repeated a
  `"defaultModel" in provider` narrowing dance; `reconcileRetiredCloudModelSelections` hand-rolled the
  same lookup by list position. One shared `pickDefaultModelId` now serves all three.

### Two consequences worth a second opinion

1. **The notice can now fire at cold start**, where before it only ran mid-session after a fetch.
   `TinfoilModelSwitchToastListener` answers it with `showDictationPanel({ reposition: true })`, so
   affected users see the pill move to the active display once — and with `floatingIconAutoHide` on,
   see it appear until the toast clears. One-time and self-healing, but new behaviour.
2. **A replacement outside the curated seed is named by id** in the toast ("Switched to kimi-k3"),
   because the sweep resolves names offline. Pinned by a test rather than papered over with a
   display-name table that would duplicate what the catalog owns.

### Known gap

`ProviderSetupStep.tsx:290` is uncovered — reverting it to `models[0]` leaves the suite green. It's
one of the two paths the second commit names as broken, so worth a mounted-component test at some
point, though not a blocker.

### Verification

3477 tests / 3289 pass / **0 fail** / 187 skipped on the pinned Node 24, plus `typecheck`, `lint`,
`format:check`, `i18n:check` and `build:renderer` all clean. Merges cleanly with `main`.

One note on CI: the earlier `pr-quality` failure was `npm audit` getting a 503 from
registry.npmjs.org — infrastructure, not this branch, which touches neither `package.json` nor
`package-lock.json`. The `Run tests` step in that same job passed.

Issue #2008, opened 2026-09-03
GLM-5.3 / GLM-5.3-Flash Generate Notes degrades on longer transcripts via self-hosted/Ollama OpenAI-compatible route


christian-taillon:
**Describe the bug**

When using GLM-5.3 or GLM-5.3-Flash for Note Formatting through an Ollama/OpenAI-compatible self-hosted endpoint, **Generate Notes can return the meeting transcript instead of summarized notes**.

The behavior appears sensitive to transcript length. A short test transcript generates appropriate notes, while a normal-length meeting returns essentially the full transcript as the enhanced output.

I reproduced this with both GLM-5.3 and GLM-5.3-Flash using the default or effectively default Generate Notes prompt.

**To Reproduce**

1. Configure **Note Formatting** to use an Ollama/OpenAI-compatible self-hosted endpoint.
2. Select `glm-5.3` or `glm-5.3-flash`.
3. Leave **Disable thinking output** enabled.
4. Open or record a meeting with a normal-length transcript.
5. Click **Generate Notes**.
6. Observe that the enhanced output may reproduce the transcript instead of generating summarized meeting notes.

For comparison, the same configuration works with a very short test transcript.

**Expected behavior**

Generate Notes should transform the transcript into concise, structured meeting notes regardless of normal transcript length.

**Screenshots**

Can provide if useful. 
```enhanced_note
## Meeting Context
Invited participants: first@domain.com, second@domain.com, ... 
## Meeting Transcript
Speaker 1: ...
```
**Desktop (please complete the following information):**

* OS: Debian GNU/Linux 13.6 (trixie), Linux 6.12.107+deb13-amd64
* Desktop: Wayland / Mutter
* Browser: N/A, Electron desktop application
* Version: OpenWhispr 1.9.2
* Electron: 41.10.5
* Node: 24.18.0
* Models: GLM-5.3 and GLM-5.3-Flash
* Endpoint: Ollama/OpenAI-compatible self-hosted endpoint

**Smartphone (please complete the following information):**

* Not applicable

**Additional context**

One possible lead is GLM reasoning compatibility, although I have not confirmed this as the root cause.

GLM-5.3 does not currently appear in `modelFamilyConstraints.ts`. With **Disable thinking output** enabled, the generic LAN request shaping can use a reasoning effort of `none`. GLM-5.3-Flash is an always-reasoning model whose supported effort levels are `low`, `high`, and `max`.

[...]

PR #2019, opened 2026-09-03
fix(reasoning): floor GLM-5.3 thinking suppression at low effort

IdrisGit:
**GLM-5.3** and **GLM-5.3-Flash** always reason and accept only `low/high/max` effort. 

This adds a `glm-5.3` entry to the shared model-family constraints with `suppressValue: "low"` (same precedent as gpt-oss), with no `cleanupValue` so the default path sends nothing and the server applies its `max` default.

related #2008

### DIFF
--- a/src/services/ai/modelFamilyConstraints.ts
+++ b/src/services/ai/modelFamilyConstraints.ts
@@ -9,7 +9,7 @@
  * like thinkingSuppressionDialects.
  */
 export interface ModelFamilyConstraints {
-  family: "gpt-oss" | "qwen" | "magistral";
+  family: "gpt-oss" | "qwen" | "magistral" | "glm-5.3";
   reasoningEffort?: {
     /** Value that best approximates "thinking off" for reasoning_effort. */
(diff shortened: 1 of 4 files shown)


christian-taillon:
One detail from my reproduction in #2008 that may be relevant here: **Disable thinking output was already OFF** when I reproduced the failure with GLM-5.3 / GLM-5.3-Flash.

From the current control flow, it looks like `suppressThinking()` is only reached when `disableThinking === true` (aside from models explicitly forced through the registry). If so, the new `suppressValue: "low"` in this PR would not affect the requests from my failing reproduction.

So I think this PR is still a useful fix for GLM-5.3 when thinking suppression is enabled, but it may not address the original issue I reported and might be worth considering while we are updating how the model is being handled. 

My failing setup is **Ollama Cloud via its OpenAI-compatible `/v1/chat/completions` endpoint**. That makes me wonder if there may also be a provider-specific GLM-5.3 issue in the normal, non-suppressed request/response path.

PR #2025, opened 2026-09-04
fix: local note generation failed on every meeting — context sized from a 6x-too-large KV estimate

futuregerald:
## Summary

Generating notes with a local model failed on **every** meeting with `Prompt is too long for this model: about 2654 tokens against a budget of 1228`, and the queue behind it failed identically. The model was being started with a 2048-token context — the hard floor.

Four defects compounded. Two more were found by review before merge.

| | before | after |
|---|---|---|
| resolved context | 2,048 | **16,384** |
| prompt budget | 1,228 tokens | **9,830 tokens** |
| the 2,654-token prompt that failed | refused | **fits** |

This branch also carries two earlier fixes already released as 1.16.1 (memory-aware context budget, and keeping audio for unprocessed meetings). It ships **1.17.0**.

### 1. The KV-cache estimate was 5.5–6.3× too large

`ggufMetadata.js` priced every layer as growing with `n_ctx`. Sliding-window models do not work that way. Measured by running the bundled `resources/bin/llama-server-darwin-arm64` against the real `google_gemma-4-E4B-it-Q4_K_M.gguf` with `LLAMA_ARG_FIT=off` — the app's own flag:

| `--ctx-size` | measured KV | old estimate | new estimate |
|---|---|---|---|
| 4,096 | 164 MiB | 420 MiB (2.6×) | 392 MiB (2.4×) |
| 32,768 | **612 MiB** | 3,360 MiB (5.5×) | 1,176 MiB (1.9×) |
| 131,072 | 2,148 MiB | 13,440 MiB (6.3×) | 3,864 MiB (1.8×) |

`llama.cpp`'s own log shows why: `llama_kv_cache_iswa: creating non-SWA KV cache, size = 32768 cells` over **4** layers, and a **fixed** `SWA KV cache, size = 2560 cells` over 20 layers — the sliding layers do not grow with the context at all. The reader now takes `attention.key_length`, `sliding_window`, `sliding_window_pattern` and the SWA variants, and `resolveContextSize` searches powers of two instead of dividing, because the cost is no longer linear in the context.

The estimate stays deliberately conservative (1.8–2.4× over measured). Under-estimating is the direction that caused a machine-wide swap stall in August.

### 2. Fixing the estimate alone was not enough

A first draft of the plan claimed the corrected formula would lift the context to 8192 on its own. A spike disproved that before any code was written, and the plan reviewer reached the same conclusion independently: the flat 256 MB `MIN_KV_BUDGET` was the binding constraint, reached whenever the model is larger than the available-memory share. From the reported machine's log:

```
availableMemBytes: 5292244992   modelBytes: 5405168384   totalMemBytes: 25769803776
kvBudgetBytes:     268435456    contextSize: 2048        source: "available-bound"
```

`0.8 × 5.29GB − 5.41GB` is negative, so the budget pinned to the floor. Refusing the cache 0.8 GB next to a 5.4 GB weights allocation the app makes without hesitation is incoherent, so the floor now scales with the weights already committed, capped by the polite total-RAM allowance so a large model on a small machine cannot use its own size to claim more.

### 3. The budget was enforced against an estimate that was never checked

`llamaServer.js` stored the estimate; `modelManagerBridge.js` refused prompts against it. `llama-server` publishes the real answer at `GET /props`, and nothing read it. It is now adopted once the server reports healthy — any failure leaves the estimate in place, since by then the server is already serving.

### 4. The pipeline truncated instead of chunking

`postCallPipelineManager.js` sent `text.slice(0, 8000)` in one call — roughly the first 25 minutes of a meeting — and silently discarded the rest. The multi-pass runner that solves this already existed and was wired only to the manual "regenerate notes" action; `transcriptPassChunker.js` names this exact line in its own header comment as the thing it was built to cure.

A locally served model now routes through it. Cloud models keep the single call but lose the truncation — a cloud note should not end at 25 minutes either.

The local branch is chosen with `MainProcessInference.resolveProvider`, not a raw `provider === "local"`. Settings has been observed persisting a model family (`"gemma"`) into the provider field, and now that the cloud branch no longer truncates, sending a local model down it would put an entire transcript in one call — worse than the bug being fixed. Caught in review.

### 5. Hardening against corrupt model files (found by review)

Model files are downloaded. Three review rounds each found a distinct way a malformed header could be priced **cheaper** than an honest one, unlocking the full trained 131,072 context — the ~14 GB allocation that stalled a machine in August:

- a `sliding_window_pattern` claiming more sliding layers than `block_count` made the growing-layer count negative
- negative optional dimensions (`key_length` and friends) shrank the per-token cost
- `head_count` exceeding `embedding_length` floors the derived head dimension to **0**, and `keyLength ?? headDim` keeps a literal zero — pricing both sides of every cell at nothing. Reachable from a real file.

Plus: `contextLength: Infinity` made the halving loop run forever (reproduced — the test run hung and had to be killed); a fractional `swaLayerCount` under-priced; and an honest model trained below 2048 was being routed to the 8192 fallback, eight times its trained context.

A header whose geometry does not add up is now treated as unreadable.

> One fix was mine to make and mine to break: returning `0` for an arithmetic overflow told the resolver's search that an enormous context was **free**, so `contextLength: 1e308` resolved to 8.99e307. My own fuzzer caught it. Overflow now returns `Infinity` (priced, unaffordable); only `NaN` returns zero (unpriceable).

## Test plan

- **`npm test`** — 772 tests, 767 pass, 0 fail (5 skips are Linux-only sandbox tests). Baseline before this work was 742/737.
- **`npm run lint`** — clean.
- **Zero pre-existing assertions were changed.** `git diff` on `test/helpers/llamaContext.test.js` shows three deletions: two import lines and the `writeGguf` signature. All 10 original `resolveContextSize` tests, including the August incident test, pass untouched.
- **Verified against the real model and the real logged memory numbers**, not just fixtures: 2048 → 16384, budget 1228 → 9830, and the exact 2,654-token prompt from the report now fits.
- **Fuzzed 14,304 mutated headers across three machine profiles** (the reported machine, a tight one, a roomy one) asserting that no corrupt header is priced more cheaply, or granted more context, than an honest one. **Zero violations.**
- **Mutation-checked**, each guard disabled alone, red pasted at the time:

  | mutation | result |
  |---|---|
  | old all-layers-grow KV formula | 5 fail |
  | weights-proportional floor removed | 2 fail |
  | `/props` reconciliation removed | 1 fail |
  | `slice(0, 8000)` restored | 1 fail |
  | raw `provider === "local"` | 1 fail |
  | corrupt-pattern distrust → the weaker clamp | 1 fail |
  | overflow reads as free | 1 fail |
  | `MAX_TRAINED_CONTEXT` bound removed | 1 fail |
  | `swaLayerCount` integer check removed | 1 fail |
  | derived `headDim >= 1` removed | 1 fail |
  | `headDim` dropped from validation | 1 fail |

  Mutation testing also caught that four of my own new tests were too weak to fail, and that one guard was only exercised on a machine profile where the memory budget hid the bug. Both fixed.

- **Blast radius walked** after every code-mutating round. Untested callers (`modelManagerBridge.js:514-515`, `ipcHandlers.js:3837`) verified by hand: they read only `contextSize` and `source`, both still present, and the new constructor field is optional — the two test files that construct the pipeline without it exercise the fallback and stay green.

### Not covered by tests

`kvGeometry` uses `gguf.keyLength || headDim` rather than `??`. Mutating it back to `??` fails nothing, because the `headDim` validation rejects every input that could reach it. It is kept as the correct operator for reading untrusted files, not because a test demands it — recorded rather than left to imply coverage it does not have.

## Rollback

Two independent commits.

- Reverting `414c8158` alone restores the truncating single call, leaving the corrected context sizing in place.
- Reverting `97edc513` restores the old estimate.

Nothing here touches the database, migrations, or persisted settings, so a revert needs no data repair. No new user-facing strings, so no translation files changed.

## Out of scope

- **The `vm_stat` memory probe** (`systemMemory.js`) understates by roughly 2× — `memory_pressure` reports 61% free where the probe sums 7.20 GB, and llama.cpp's Metal device reported 18,185 MiB free. It is a real contributor, but the KV floor makes it non-fatal without touching it, and it was tuned against its own incident. Worth doing separately.
- **`modelAlreadyResident`** (`llamaContext.js`) is dead — never passed by either production caller, asserted only by a test. Not a cause of this bug; a cold start collapses too.

https://claude.ai/code/session_012RuAAu3MkYb6GHHVQxKtXX


futuregerald:
Opened by mistake against the upstream repository — this work belongs on my fork, where PR #40 already tracks the branch. Closing; apologies for the noise.

PR #2026, opened 2026-09-04
fix: bump the lockfile version alongside package.json

futuregerald:
## Summary

`lockfile-drift` failed on #40 and the check was not read before merging. `main` currently has `package.json` at 1.17.0 and `package-lock.json` at 1.16.1.

## Background

The 1.17.0 bump touched `package.json` only. The CI job regenerates the lockfile with a pinned npm and fails on any diff — and the two version fields were the entire diff:

```
-  "version": "1.16.1",     +  "version": "1.17.0",
-      "version": "1.16.1", +      "version": "1.17.0",
```

Regenerated with the same command the job runs, `npx --yes npm@11.17.0 install --package-lock-only --ignore-scripts`, so this cannot itself drift.

## Test plan

- `git diff` is exactly the two version fields — nothing else moved.
- All **19** `libc` fields are still present. That is what this guard is really for: npm below 11.11.0 strips them silently and `npm ci` still succeeds, so nothing else would catch it.
- Regenerating a second time produces no further diff, i.e. CI's `git diff --exit-code` passes.

## Rollback

Single commit, `git revert`. No source, schema, or dependency change — only the two version fields.


futuregerald:
Opened against the upstream by mistake — belongs on my fork. Closing.

Issue #472, opened 2026-03-18
[Bug] v1.6.5 records audio in mono, causing whisper-server transcription failure on Linux


rearwindowvertigo:
Description
After upgrading from v1.6.2 to v1.6.5 on Linux (Ubuntu 24.04, GNOME Wayland), dictation transcription silently fails. The app plays start/stop sounds normally, but no transcription is produced — instead, the previous clipboard contents are pasted.
Root Cause
v1.6.5 records audio in mono (channelCount: 1), whereas v1.6.2 records in stereo (channelCount: 2). The bundled whisper-server-linux-x64 binary cannot process mono WebM audio, returning {"error":"failed to read audio data"}.
This means the whisper server shipped with v1.6.5 is incompatible with the audio format that v1.6.5 itself produces.
Steps to Reproduce

Install OpenWhispr v1.6.5 on Ubuntu 24.04 (GNOME Wayland)
Use any mono microphone input (e.g., Jabra Evolve2 40, Apple Studio Display mic, or any USB headset mic)
Press F8 (or configured hotkey) to start dictation
Speak a sentence, press F8 again to stop
Observe: no transcription appears; previous clipboard content is pasted instead

Evidence
Audio format comparison (via ffprobe):
# v1.6.2 recording (working):
Stream #0:0(eng): Audio: opus, 48000 Hz, stereo, fltp (default)

# v1.6.5 recording (broken):
Stream #0:0(eng): Audio: opus, 48000 Hz, mono, fltp (default)
Direct whisper-server test with v1.6.5 mono file:
bash$ curl -s -X POST http://127.0.0.1:8178/inference \
    -F "file=@~/.config/open-whispr/audio/OpenWhispr-2026-03-18-20-34-06-37.webm" \
    -F "response_format=json"
{"error":"failed to read audio data"}
Same file converted to stereo WAV — transcription succeeds:
bash$ ffmpeg -i broken-mono-file.webm -ac 2 /tmp/test-stereo.wav -y
$ curl -s -X POST http://127.0.0.1:8178/inference \
    -F "file=@/tmp/test-stereo.wav" \
    -F "response_format=json"
{"text":" Thanks so much for making all of these changes..."}
v1.6.2 logs show stereo recording:
[INFO][audio][renderer] Recording started with microphone {
  label: 'Default',
  deviceId: 'default...',
  sampleRate: 48000,
  channelCount: 2
}
Expected Behavior
[...]

PR #2029, opened 2026-09-04
fix(transcription): re-encode to WAV for Custom STT endpoints and retries (microsoft/mai-transcribe-2)

### LINKED ISSUE #2028: Custom cloud STT always uploads audio/webm; providers that accept only WAV/MP3/FLAC (e.g. microsoft/mai-transcribe-2 via OpenRouter) fail with an opaque 400
**Describe the bug**

The Custom (OpenAI-compatible) cloud STT provider always uploads the raw MediaRecorder blob, which on macOS/Electron is WebM/Opus. Providers whose upstream only accepts WAV/MP3/FLAC therefore reject every dictation with a 400, and the surfaced message gives the user nothing to act on:

```
Transcription failed
API Error: 400 {"error":{"message":"Provider returned 400","code":400}}
```

Concrete case: OpenRouter + `microsoft/mai-transcribe-2`. OpenRouter relays to Azure, and Microsoft's docs for MAI-Transcribe state the input must be "an audio file (less than 300 MB in size) in one of these formats: **WAV, MP3, or FLAC**" — WebM is not accepted.
https://learn.microsoft.com/en-us/azure/ai-services/speech-service/mai-transcribe

This is **not** a recurrence of #769. That issue was OpenRouter rejecting `multipart/form-data` outright when their endpoint was JSON-only; OpenRouter has since added multipart support and it works. The transport is fine now — the *container* is the problem, which is a different failure with the same HTTP status.

It is, however, exactly the case called out in #769 on 2026-07-07 and never implemented:

> "`audio/webm` recordings are either accepted by the provider or converted before upload, with a clear error when the provider rejects the container"

**Evidence**

Same one second of 220 Hz tone, generated once and re-encoded into different containers, sent to `POST https://openrouter.ai/api/v1/audio/transcriptions` with `model=microsoft/mai-transcribe-2`:

| # | File | Container | Status | Body |
|---|------|-----------|--------|------|
| 1 | `probe.wav` (32044 B) | WAV | **200** | `{"text":"","usage":{"seconds":1,"cost":0.0000277…}}` |
| 2 | `probe.webm` (8983 B) | WebM/Opus | **400** | `{"error":{"message":"Provider returned 400","code":400}}` |
| 3 | `OpenWhispr-2026-09-04-09-58-21-12.webm` (121295 B) | WebM/Opus | **400** | `{"error":{"message":"Provider returned 400","code":400}}` |
| 4 | same tone as MP3 | MP3 | **200** | success |
| 5 | same tone as FLAC | FLAC | **200** | success |

Legs 1 and 2 are the identical audio from the identical source, differing only in container, which rules out duration, file size, content and silence detection and leaves the container as the sole variable. Leg 3 is a real OpenWhispr recording, failing the same way. WAV, MP3 and FLAC all pass — exactly matching Microsoft's documented list.

*It is the content, not the labeling.* The endpoint sniffs the actual bytes and ignores both the filename and the part's `Content-Type`:

| Sent | Result |
|------|--------|
| WAV bytes, `filename=x.webm`, `type=audio/webm` | **200** |
| WebM bytes, `filename=x.wav`, `type=audio/wav` | **400** |
| WebM bytes, no type override | **400** |

So a fix that only renames the upload or rewrites the MIME type will not work — the audio has to actually be transcoded.

*Replication.* Six clean 400s on WebM bytes across four different labelings, and 3/3 200s on WAV. Note when reproducing: firing these back-to-back trips OpenRouter's rate limiter and returns `Provider returned 429`, which is easy to mistake for the 400 under test. Space the requests out.

*Pre-empting the "corrupt file" reading.* Decoding the real recording (leg 3), ffmpeg warns `Error parsing Opus packet header` on one packet, though it recovers and decodes all 7.5 seconds. That is not the cause: the leg 2 probe is a clean ffmpeg-generated WebM/Opus file with no such warning and it fails identically.

**To Reproduce**

1. Settings → AI Models → Speech-to-Text → **Dictation** tab
2. Choose **Cloud Providers**, then the **Custom** tab
3. Endpoint URL: `https://openrouter.ai/api/v1`
4. API key: an OpenRouter key with credits
5. Model: `microsoft/mai-transcribe-2`
6. Press the dictation hotkey, speak, release

Result: `API Error: 400 {"error":{"message":"Provider returned 400","code":400}}`

**Expected behavior**

Either the audio is converted to a container the endpoint accepts, or the failure explains that the endpoint rejected the container and which formats it wants. Right now a correctly configured endpoint is indistinguishable from a broken one.

**Desktop**

- OS: macOS 26.5.2 (build 25F84)
- OpenWhispr version: 1.9.2
- Hardware: Apple Silicon (M3), arm64-only bundle
- Install: DMG

**Additional context**

*Where it comes from*

The recorder is constructed with no `mimeType`, so Chromium picks WebM/Opus:

```js
// src/helpers/audioManager.js:1384
const recorder = adoption?.recorder ?? new MediaRecorder(micStream);
```

and the upload path passes the blob through unchanged, despite the variable name:

```js
// src/helpers/audioManager.js:3339
const optimizedAudio = audioBlob;
```

```js
// src/helpers/audioManager.js:3386-3411
const mimeType = optimizedAudio.type || "audio/webm";
const extension = mimeType.includes("webm") ? "webm" : /* ... */ "webm";
formData.append("file", optimizedAudio, `audio.${extension}`);
```

There is no conversion step and no per-endpoint format setting, so every Custom endpoint gets WebM regardless of what it supports.

Confirmed at the output rather than only in the source: `ffprobe` on the recordings the app itself saved reports `opus, 48000 Hz, mono` in a `matroska,webm` container, tagged `encoder: Chrome` — i.e. exactly Chromium's default MediaRecorder encoding, reaching the upload path untouched.

*Ruled out while narrowing this down*

- `stream` — only appended when the provider is literally `"openai"` (`shouldStreamTranscription`, `audioManager.js:2962`), so it is absent here.
- `language=auto` — `getBaseLanguageCode()` returns `undefined` for `"auto"` (`src/utils/languageSupport.ts:27-30`), so no `language` field is sent in the default configuration.
- `prompt` — only appended when the custom dictionary is non-empty; the failure reproduces with an empty dictionary.
- Error truncation — OpenWhispr is *not* hiding anything. `audioManager.js:3506-3513` reads the full body and includes it verbatim. The unhelpful `"Provider returned 400"` string is OpenRouter's own, not a formatting loss on this side.

*Suggested directions (raising rather than assuming, since there's a real trade-off)*

1. **Convert only when needed.** ffmpeg is already resolved for local Whisper via `getFFmpegPath()` in `src/helpers/whisper.js:243`, so a WebM→WAV step needs no new dependency.
2. **Convert-on-reject and retry once.** Keeps WebM's bandwidth advantage for endpoints that accept it, and costs a transcode only on the failing path.
3. **Per-endpoint upload-format setting** on the Custom tab (`webm` default / `wav` / `mp3`), explicit and cheap.
4. **At minimum, a better message.** When a Custom-endpoint request 4xxs and the upload was WebM, say so and name WAV/MP3/FLAC.

Note that relabeling is not among the options — per the table above, the upstream sniffs content, so only a real transcode (or recording to a supported container in the first place, via a `mimeType` argument to `MediaRecorder`) resolves it.

Leg 4 above matters for the trade-off. My first instinct was that conversion would be costly — Opus runs about 4 KB/s against roughly 32 KB/s for 16 kHz mono PCM, so blanket WAV would inflate uploads ~8x and eat into the 25 MB BYOK cap. But **MP3 is accepted too**, and mono MP3 at 64 kbps is about 8 KB/s, so converting to MP3 rather than WAV costs roughly 2x rather than 8x. That makes option 1 or 2 cheap enough to be the obvious default, and largely removes the reason to push the decision onto users via option 3.

Happy to open the PR for whichever direction you prefer.

marselsel:
Fixes #2028.

## Why

Chromium's `MediaRecorder` only offers WebM/Opus (and MP4/AAC) on the desktop platforms we ship, so every dictation reaches the upload path in a container some backends refuse. Most OpenAI-compatible transcription services accept it — but some decode the bytes themselves and reject anything outside WAV/MP3/FLAC. Azure's MAI-Transcribe, reached through OpenRouter, is one, and it fails every dictation with `{"error":{"message":"Provider returned 400"}}`.

Measured against `microsoft/mai-transcribe-2`, the same one second of 220 Hz tone re-encoded into different containers:

| Container | Status |
|---|---|
| WAV | 200 |
| MP3 | 200 |
| FLAC | 200 |
| WebM/Opus | 400 |

And the rejection is by **content**, not labelling — so a rename-only fix would not work:

| Sent | Result |
|---|---|
| WAV bytes, `filename=x.webm`, `type=audio/webm` | 200 |
| WebM bytes, `filename=x.wav`, `type=audio/wav` | 400 |

Full evidence, replication counts and ruled-out alternatives are in #2028.

## What

- `src/utils/audioContainer.ts` — `needsWavConversion()` predicate, a pure `encodeWav()` (interleaved 16-bit PCM in RIFF), `downmixToMono()`, and a `convertToWav()` wrapper over `decodeAudioData`.
- `audioManager.js` re-encodes immediately before the multipart body is built, leaving the proxied-provider path above it untouched.
- `ipcHandlers.js` does the same for **retries**. `retry-transcription` re-uploads stored audio from the main process and never reaches the renderer, so without this every retry of a dictation that failed for this reason failed again — which is the exact recovery a user reaches for after hitting the bug.
- `ffmpegUtils.js` gains `convertBufferToWav()`, a buffer-in/buffer-out wrapper over the existing path-based `convertToWav()`. `whisperServer` and `parakeetServer` were each open-coding the same write-temp / convert / read / unlink dance.

Deliberate choices, happy to change any of them:

- **Scoped to `provider === "custom"`.** Built-in providers accept Opus today, so they keep sending the smaller upload. Only Custom can reach a backend with this restriction.
- **An allowlist, not a blocklist.** The predicate converts anything that is not already WAV/MP3/FLAC, rather than enumerating known-bad containers. MP4/AAC is a real case — `MediaRecorder` emits it on some platforms, the extension switch in `audioManager.js` already handles it, and AAC is no more acceptable to these backends than Opus. Confirmed: the same recording as AAC-in-MP4 returns 400.
- **16 kHz mono, matching the main process.** `ffmpegUtils.convertToWav` already defaults to exactly this, and speech models resample to it internally, so both paths now produce interchangeable uploads. `decodeAudioData` honours an `OfflineAudioContext`'s sample rate but *not* its channel count, so the mixdown has to be explicit.
- **The mono downmix is load-bearing, not a nicety.** `getAudioConstraints` requests `channelCount: 2` unconditionally (`audioManager.js:913`) — stereo is required because mono WebM breaks silence detection on Linux/PipeWire (#472). Any platform whose mic honours that constraint therefore records two channels, and without `downmixToMono()` the upload would carry two copies of the same voice. A macOS mic that only delivers mono hides this. On PipeWire it is worse than "stereo hardware only": @mazixs reports that *every* device there negotiates 2ch at 48 kHz, including a physically mono webcam mic, because PipeWire satisfies the non-exact constraint by upmixing — so the downmix matters on essentially any Linux capture.
- **`OfflineAudioContext`, not a live `AudioContext`.** It neither opens an output device nor inherits whatever rate the hardware happens to run at, so the output rate is deterministic rather than machine-dependent.
- **Two converters rather than IPC.** ffmpeg is only reachable from the main process and Web Audio only from the renderer. Plumbing either across IPC costs more than the small amount of duplication, and both targets are pinned to the same format.
- **Fails open.** A conversion error logs and uploads the original bytes, on both paths, so this can only widen what works.
- **WAV, not MP3.** MP3 would be smaller but needs an encoder dependency. At 16 kHz mono the size is no longer a concern at dictation length — see below.

## Size

Re-encoding inflates the upload, so the ceiling matters. Measured on a real 4.86 s recording:

```
source WebM/Opus            78,663 B
decoded at 48 kHz mono     466,604 B     (5.9x source)
encoded at 16 kHz mono     155,564 B     (2.0x source)
```

Downsampling on the way in cuts the output exactly 3x, which moves the point where a converted dictation would cross a 25 MB endpoint cap from roughly 4.6 minutes to roughly 13.7 minutes.

On a platform that actually delivers the requested stereo, the saving is 6x rather than 3x:

| | bytes/sec |
|---|---|
| decoded at 48 kHz, mono | 96,000 |
| decoded at 48 kHz, stereo | 192,000 |
| encoded at 16 kHz mono | 32,000 |

## Tests

`test/utils/audioContainer.test.js` — 6 cases covering the provider/container predicate (including the MP4 case and the empty-recording guard), RIFF header fields, clamping of out-of-range samples (`decodeAudioData` can return values past full scale, which a naive multiply wraps to the opposite rail), multi-channel interleaving, the mono downmix, and the empty-input guard.

`test/helpers/retryTranscriptionHandler.test.js` — 4 further cases on the retry path: a stored WebM is converted before it reaches a Custom endpoint, audio already in WAV is left alone, built-in providers are untouched, and a conversion failure falls open to the stored container. Checked by mutation — disabling the branch fails the first case rather than passing vacuously. Conversion is mocked, so the suite does not shell out to ffmpeg.

## Verification

Run on macOS 26.5.2, Apple Silicon, against `microsoft/mai-transcribe-2` via OpenRouter, using a real OpenWhispr recording rather than synthetic audio:

| Path | Bytes | Status |
|---|---|---|
| unmodified WebM (what ships today) | 78,663 | **400** |
| renderer — `audioContainer.convertToWav` | 155,564 | **200** |
| retry — `ffmpegUtils.convertBufferToWav` | 155,598 | **200** |

Both conversion paths returned the same transcript, word for word, and both produced 16 kHz mono 16-bit WAV that validates under `ffprobe` and Python's `wave`. The retry leg exercised the shipped `ffmpegUtils` code rather than a reimplementation.

One detail worth recording: `ffmpeg` reports `Error parsing Opus packet header` on OpenWhispr's own recordings. `decodeAudioData` handles them anyway, so the fail-open branch is not silently swallowing the fix — but it is why the tests cover a real capture and not just generated tone.

Gates run locally against Node 25 (the repo requires >= 24, which is what made the toolchain awkward to install on a first attempt):

- `npm run quality-check` — eslint under both configs, prettier, and `tsc --noEmit`: clean.
- `npm test` — 3,468 tests, 3,460 pass, 6 skipped, **1 fail**.

That one failure is `settingsStore imports without a bare localStorage global` in `test/stores/enterpriseIdentityStoreImports.test.js`. It reproduces on `main` at f03e027, before either commit here, so it is pre-existing and not from this branch — flagging it because CI on this PR is still pending maintainer approval and will surface it as red once it runs.

## Independent verification on Linux

Everything above was measured on macOS, outside Electron. @mazixs — unaffiliated with this PR, running a self-hosted GigaAM v3 endpoint that accepts WebM and was never hit by the bug — ran this branch at 1d86e56 on Wayland/PipeWire and covered the two gaps it could not. Their full report is in #2028; the figures below are theirs, not mine.

One 27.66 s capture, same audio through each path:

| Path | Bytes | What the server received |
|---|---|---|
| WebM/Opus (ships today) | 196,205 | `opus: 48000Hz, 2ch` |
| renderer `convertToWav` | 885,164 | `wave: 16000Hz, 1ch` |
| retry, bundled ffmpeg | 885,198 | `wave: 16000Hz, 1ch` |

Transcripts from all three are word-for-word identical to a WAV control of the same source sent straight to the endpoint. The retry leg resolved the downloaded `ffmpeg-static` 7.0.2 binary rather than a system one, and the captured invocation carries the per-process sequence number added here. `needsWavConversion` returned true for `custom` and false for `openai`, so the scoping holds. Both new test files pass.

Two details corroborate rather than merely agree with the macOS numbers: the 34-byte difference between the two encoders is identical on both platforms despite different file sizes, and 885,164 − 44 is exactly 27.66 s of 16 kHz mono 16-bit PCM.

Their WAV-to-Opus ratio reads 4.5x rather than the ~2x quoted above; they flag this themselves as an artefact of a virtual source feeding identical mono to both channels, which Opus codes at ~57 kbit/s. Real stereo capture on that machine runs ~129 kbit/s and lands on ~2x.

Still not covered:

- A **packaged** build. That run was from source, so the `app.asar.unpacked` branch of `getFFmpegPath()` remains untested — although the bundled binary it resolves to has now been exercised.
- The hotkey and pill UI. The native Linux key listener is not built without the predev step, so `AudioManager` was driven directly; everything from `startRecording()` onward is the shipped path.

### DIFF
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -4,6 +4,7 @@ import logger from "../utils/logger";
 import { isAzureOpenAIEndpoint } from "../utils/urlUtils";
 import { withSessionRefresh } from "../lib/auth";
 import { getBaseLanguageCode, getLanguageLabel } from "../utils/languageSupport";
+import { convertToWav, needsWavConversion } from "../utils/audioContainer";
 import {
   applyChineseScript,
   mergeWhisperPrompt,
@@ -3381,9 +3382,36 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
         return { success: true, text, rawText: proxyText, source, timings };
       }
 
+      // Some Custom endpoints decode the upload and reject anything that isn't
+      // WAV/MP3/FLAC (Azure MAI-Transcribe via OpenRouter, for one), which
+      // Chromium's WebM/Opus recordings always are. Re-encode for those rather
+      // than failing the dictation; a conversion failure falls through to the
+      // original bytes so this can only widen what works.
+      let uploadAudio = optimizedAudio;
+      if (needsWavConversion(provider, optimizedAudio.type)) {
+        try {
+          uploadAudio = await convertToWav(optimizedAudio);
+          logger.debug(
+            "Re-encoded recording to WAV for custom endpoint",
+            {
+              fromType: optimizedAudio.type,
+              fromSize: optimizedAudio.size,
+              toSize: uploadAudio.size,
+            },
+            "transcription"
+          );
+        } catch (conversionError) {
+          logger.warn(
+            "WAV re-encode failed; uploading original container",
+            { error: conversionError?.message, type: optimizedAudio.type },
+            "transcription"
+          );
+        }
+      }
+
       const formData = new FormData();
       // Determine the correct file extension based on the blob type
-      const mimeType = optimizedAudio.type || "audio/webm";
+      const mimeType = uploadAudio.type || "audio/webm";
       const extension = mimeType.includes("webm")
         ? "webm"
         : mimeType.includes("ogg")
@@ -3401,13 +3429,13 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
         {
           mimeType,
           extension,
-          optimizedSize: optimizedAudio.size,
+          optimizedSize: uploadAudio.size,
           hasApiKey: !!apiKey,
         },
         "transcription"
       );
 
-      formData.append("file", optimizedAudio, `audio.${extension}`);
+      formData.append("file", uploadAudio, `audio.${extension}`);
       formData.append("model", model);
 
       if (language) {
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -3388,7 +3388,7 @@ registerProcessor("pcm-streaming-processor", PCMStreamingProcessor);
       // than failing the dictation; a conversion failure falls through to the
       // original bytes so this can only widen what works.
       let uploadAudio = optimizedAudio;
-      if (needsWavConversion(provider, optimizedAudio.type)) {
+      if (needsWavConversion(provider, optimizedAudio.type, optimizedAudio.size)) {
         try {
           uploadAudio = await convertToWav(optimizedAudio);
           logger.debug(
--- a/src/helpers/ffmpegUtils.js
+++ b/src/helpers/ffmpegUtils.js
@@ -186,6 +186,31 @@ function convertToWav(inputPath, outputPath, options = {}) {
   });
 }
 
+// Buffer in, buffer out. convertToWav works on paths, but the retry and upload
+// paths hold recordings in memory, and every caller that bridged that gap was
+// writing the same temp-file dance by hand.
+async function convertBufferToWav(audioBuffer, options = {}) {
+  const { getSafeTempDir } = require("./safeTempDir");
+  const tempDir = getSafeTempDir();
+  const stamp = `${Date.now()}-${process.pid}`;
+  const inputPath = path.join(tempDir, `ow-reencode-${stamp}.input`);
+  const outputPath = path.join(tempDir, `ow-reencode-${stamp}.wav`);
+
+  try {
+    fs.writeFileSync(inputPath, audioBuffer);
+    await convertToWav(inputPath, outputPath, { sampleRate: 16000, channels: 1, ...options });
+    return fs.readFileSync(outputPath);
+  } finally {
+    for (const f of [inputPath, outputPath]) {
+      try {
+        if (fs.existsSync(f)) fs.unlinkSync(f);
+      } catch {
+        // ignore cleanup errors
+      }
+    }
+  }
+}
+
 function parseWavFormat(wavBuffer) {
   if (!isWavFormat(wavBuffer)) return null;
 
@@ -459,6 +484,7 @@ module.exports = {
   isWavFormat,
   parseWavFormat,
   convertToWav,
+  convertBufferToWav,
   splitAudioFile,
   parseFfmpegDuration,
   wavToFloat32Samples,
--- a/src/helpers/ffmpegUtils.js
+++ b/src/helpers/ffmpegUtils.js
@@ -186,13 +186,19 @@ function convertToWav(inputPath, outputPath, options = {}) {
   });
 }
 
+let reencodeSequence = 0;
+
 // Buffer in, buffer out. convertToWav works on paths, but the retry and upload
 // paths hold recordings in memory, and every caller that bridged that gap was
 // writing the same temp-file dance by hand.
 async function convertBufferToWav(audioBuffer, options = {}) {
   const { getSafeTempDir } = require("./safeTempDir");
   const tempDir = getSafeTempDir();
-  const stamp = `${Date.now()}-${process.pid}`;
+  // process.pid is constant within a process and Date.now() only has
+  // millisecond resolution, so two concurrent conversions -- a user retrying
+  // several history entries at once -- would otherwise share both temp paths
+  // and race each other's output.
+  const stamp = `${Date.now()}-${process.pid}-${++reencodeSequence}`;
   const inputPath = path.join(tempDir, `ow-reencode-${stamp}.input`);
   const outputPath = path.join(tempDir, `ow-reencode-${stamp}.wav`);
 
--- a/src/helpers/ipcHandlers.js
+++ b/src/helpers/ipcHandlers.js
@@ -5875,6 +5875,7 @@ class IPCHandlers {
             ? preferredLanguage.split("-")[0]
             : undefined;
         const { resolveTranscriptionRoute } = await import("./transcriptionRoute.ts");
+        const { convertBufferToWav, isWavFormat } = require("./ffmpegUtils");
         // Renderer pre-flight owns policy; retry re-routes stored audio through
         // whatever is selected NOW.
         const route = resolveTranscriptionRoute({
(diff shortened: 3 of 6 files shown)


Chadpiha (approved):
Approved, thank you for the contribution @marselsel !

I reviewed the fresh-dictation and retry flows end to end. The follow-up commit prevents cancelled conversions from uploading stale audio and preserves the original container when WAV expansion would exceed the provider limit.

PR #2031, opened 2026-09-04
fix(windows): quote Expand-Archive fallback paths that contain apostrophes

### LINKED ISSUE #2030: Windows zip fallback fails when the user path contains an apostrophe
**Describe the bug**

When Windows `tar` cannot extract a zip, `extractZipWindows` falls back to:

```
Expand-Archive -Force -Path '${zipPath}' -DestinationPath '${destDir}'
```

PowerShell treats `'` as the end of a single-quoted string. A path like `C:\Users\O'Brien\...` splits the command. Expand-Archive then errors with `Cannot find drive` and the helper throws `Zip extraction failed`.

`gpuBinaryManager` uses this helper for GPU binary zips. Any download that extracts under the user profile hits the same fallback.

Tar-first already covers a lot of special characters. This is only the PowerShell fallback.

**To Reproduce**

1. Put a zip under a directory whose name contains `'`.
2. Force the fallback (tar reject or timeout).
3. Run the `Expand-Archive -Path '...'` command `src/helpers/downloadUtils.js` builds today.

On this Win11 box (`origin/main` `f03e027e`):

```
OLD_RC 1
OLD_STDERR New-Item : Cannot find drive. A drive with the name 'Brien-ow-extract\binary.zip -DestinationPath C' does not exist.
```

**Expected behavior**

The fallback extracts the zip. `-LiteralPath` plus doubled apostrophes (`O''Brien`) works here:

```
NEW_RC 0
NEW_FILES ...\out-new\test-file.txt
```

**Desktop**

- OS: Windows 11 zh-TW
- App: OpenWhispr `origin/main` `f03e027e`

**Additional context**

CHANGELOG already notes tar instead of Expand-Archive for special characters, and an older download-script fix for brackets. The runtime fallback still interpolates raw paths into `-Path '...'`.

stantheman0128:
Windows zip extract tries `tar` first, then PowerShell `Expand-Archive`. The fallback still interpolates `-Path '${zipPath}'`. An apostrophe in the profile path closes that string early, so Expand-Archive looks for a drive named `Brien-...` and GPU/binary install fails (`gpuBinaryManager` uses this helper).

This keeps `tar` as the primary path. The fallback now uses `-LiteralPath` and doubles apostrophes inside the single-quoted command.

Fixes #2030

## Test plan

- [x] `node --import tsx --test test/helpers/extractArchive.test.js` on Win11
- [x] Command-string test for `O'Brien` (`O''Brien` in the `-Command`)
- [x] Live extract under `%TEMP%\O'Brien-extract-*` with tar forced to fail
- Ubuntu CI `npm test` (the live apostrophe extract is skipped there)

## Evidence

HEAD command against `C:\Users\stans\AppData\Local\Temp\O'Brien-ow-extract\binary.zip`:

```
OLD_RC 1
OLD_STDERR New-Item : Cannot find drive. A drive with the name 'Brien-ow-extract\binary.zip -DestinationPath C' does not exist.
```

Same zip with `-LiteralPath` and doubled `'`:

```
NEW_RC 0
NEW_FILES ...\out-new\test-file.txt
```

After the helper change, the Windows live test extracted `hello from test`. 6 passed in that file; the remaining fail is the Linux unzip corrupt-zip case (pre-existing on Win32, CI is ubuntu).

## AI assistance

Drafted with Cursor. I reproduced the fallback on this Win11 box and wrote the quoting change plus tests.

### DIFF
--- a/src/helpers/downloadUtils.js
+++ b/src/helpers/downloadUtils.js
@@ -432,6 +432,10 @@ async function checkDiskSpace(directory, requiredBytes) {
   }
 }
 
+function escapePowerShellSingleQuoted(value) {
+  return String(value).replace(/'/g, "''");
+}
+
 async function extractZipWindows(zipPath, destDir) {
   try {
     await runSystemTar(zipPath, destDir);
@@ -440,19 +444,15 @@ async function extractZipWindows(zipPath, destDir) {
     debugLogger.info("tar extraction failed, trying PowerShell", { error: error.message });
   }
 
+  const command =
+    `Expand-Archive -Force -LiteralPath '${escapePowerShellSingleQuoted(zipPath)}' ` +
+    `-DestinationPath '${escapePowerShellSingleQuoted(destDir)}'`;
+
   return new Promise((resolve, reject) => {
-    execFile(
-      "powershell",
-      [
-        "-NoProfile",
-        "-Command",
-        `Expand-Archive -Force -Path '${zipPath}' -DestinationPath '${destDir}'`,
-      ],
-      (psError) => {
-        if (psError) reject(new Error(`Zip extraction failed: ${psError.message}`));
-        else resolve();
-      }
-    );
+    execFile("powershell", ["-NoProfile", "-Command", command], (psError) => {
+      if (psError) reject(new Error(`Zip extraction failed: ${psError.message}`));
+      else resolve();
+    });
   });
 }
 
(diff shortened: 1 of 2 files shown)


Chadpiha (approved):
Reviewed the extraction fallback and traced its callers through the GPU download/install flows. Escaping both paths and using -LiteralPath addresses the reported apostrophe-path failure while preserving tar-first behavior and existing extraction routes. No blocking findings.
Thanks @stantheman0128 !

PR #2033, opened 2026-09-04
fix(settings): inherit dictation local STT provider for upload and meeting

### LINKED ISSUE #2022: Upload transcription ignores the dictation local provider on fresh installs
## Summary

`selectResolvedUploadTranscription` falls back to the dictation settings for every local model field but **not** for the provider that selects between them. Combined with onboarding never mirroring the provider into the upload scope, a fresh install that onboards on Parakeet or Cohere ends up with audio upload pointed at Whisper `base` — a model the user never downloaded.

## The inconsistency

`src/stores/settingsStore.ts:2518-2534`. The comment directly above says unset values fall back to the base dictation settings:

```ts
useLocalWhisper: state.uploadUseLocalWhisper,
whisperModel: state.uploadWhisperModel || state.whisperModel,        // falls back
localTranscriptionProvider: state.uploadLocalTranscriptionProvider,  // does NOT
parakeetModel: state.uploadParakeetModel || state.parakeetModel,     // falls back
cohereModel: state.uploadCohereModel || state.cohereModel,           // falls back
```

Every model field honours the comment. The provider — the field that decides *which* of those models is used — doesn't. When `uploadLocalTranscriptionProvider` is unset, `readLocalProvider` (`settingsStore.ts:167`) defaults it to `"whisper"` rather than inheriting the user's actual choice.

## How a fresh install hits it

`setCloudTranscriptionForAllScopes` (`settingsStore.ts:2306`) mirrors the dictation config into the meeting and upload scopes. It sets `uploadUseLocalWhisper`, `uploadTranscriptionMode` and the upload cloud fields — but never `uploadLocalTranscriptionProvider`.

Onboarding's local path calls it with exactly that (`OnboardingFlow.tsx:562`, `onboarding/BackgroundModelDownloadTray.tsx:79`):

```ts
setCloudTranscriptionForAllScopes({ useLocalWhisper: true });
```

So for a new user who picks Parakeet during onboarding:

- `uploadUseLocalWhisper` → `true` ✅
- `uploadLocalTranscriptionProvider` → never written, defaults to `"whisper"` ❌
- `uploadWhisperModel` → unset, falls back to global `whisperModel`, whose default is `"base"` (`settingsStore.ts:1228`)

Result: uploads resolve to Whisper `base`, which was never downloaded. Best case the upload view shows `NoProviderView`; if the user happens to have *any other* Whisper model on disk, `providerReady` is true and it fails at transcribe time with `Whisper model "base" not downloaded`.

The existing migration (`migrateUploadTranscription`, `settingsStore.ts:453`) papers over this for upgrading users — it copies `localTranscriptionProvider` across once — so this only bites fresh installs, which is probably why it hasn't been reported directly.

## Relationship to #1970

Same root cause, different symptom. #1970's reporter hit the *upgrade* side: the migration pinned their upload provider to Parakeet, then changing the dictation model had no effect on uploads and the upload view gave no hint the setting was separate. #1991 fixes the discoverability half by making the model label a link to the upload-scoped settings. This issue is the fresh-install half, which that link doesn't help with because the resolved provider is wrong before the user ever looks.

## Suggested fix

One line brings the provider in line with its own comment and with every neighbouring field:

```ts
localTranscriptionProvider:
  state.uploadLocalTranscriptionProvider || state.localTranscriptionProvider,
```

This needs a small amount of care rather than a blind edit:

- `readLocalProvider` currently returns `"whisper"` for an unset key, so `||` never sees a falsy value. The stored default has to become genuinely empty (or the resolver has to read the raw key) for the fallback to engage.
- Decide whether `setCloudTranscriptionForAllScopes` should also mirror the provider explicitly. Doing both is belt-and-braces; doing only the resolver keeps the "unset means inherit" rule in one place, which I'd prefer.
- Check the meeting scope for the same asymmetry while in there.

Worth a test in `test/stores/` pinning that a fresh profile with `localTranscriptionProvider: "nvidia"` and no `upload*` keys resolves uploads to Parakeet, not Whisper.

Refs #1970, #1991

hsusul:
## Summary
- Audio upload (and meeting) stored their local STT provider as Whisper whenever the scoped key was missing, even if onboarding chose Parakeet or Cohere.
- Unset meeting/upload keys now inherit `localTranscriptionProvider`, and `setCloudTranscriptionForAllScopes` writes that provider into both scopes the same way it already mirrors `useLocalWhisper`.
- Fresh installs that pick Parakeet no longer resolve uploads to Whisper `base`.

Fixes #2022

## Test plan
- [ ] Fresh install, pick Parakeet in onboarding, upload an audio file: upload uses Parakeet, not Whisper `base`.
- [ ] Same path with Cohere.
- [ ] Opening Upload settings and explicitly choosing Whisper still sticks.
- [ ] `node --test test/helpers/transcriptionModelMemory.test.js`

### DIFF
--- a/src/stores/settingsStore.ts
+++ b/src/stores/settingsStore.ts
@@ -174,6 +174,15 @@ function readLocalProvider(key: string): LocalTranscriptionProvider {
   return stored === "nvidia" || stored === "cohere" ? stored : "whisper";
 }
 
+// Meeting/upload keys defaulted to "whisper" even when never stored, so a
+// fresh Parakeet/Cohere install resolved uploads against Whisper `base`.
+function readScopedLocalProvider(scopeKey: string): LocalTranscriptionProvider {
+  if (isBrowser && localStorage.getItem(scopeKey) === null) {
+    return readLocalProvider("localTranscriptionProvider");
+  }
+  return readLocalProvider(scopeKey);
+}
+
 function readBoolean(key: string, fallback: boolean): boolean {
   if (!isBrowser) return fallback;
   const stored = localStorage.getItem(key);
@@ -1470,7 +1479,7 @@ export const useSettingsStore = create<SettingsState>()((set, get) => ({
   })(),
   meetingUseLocalWhisper: readBoolean("meetingUseLocalWhisper", false),
   meetingWhisperModel: readString("meetingWhisperModel", ""),
-  meetingLocalTranscriptionProvider: readLocalProvider("meetingLocalTranscriptionProvider"),
+  meetingLocalTranscriptionProvider: readScopedLocalProvider("meetingLocalTranscriptionProvider"),
   meetingParakeetModel: readString("meetingParakeetModel", ""),
   meetingCohereModel: readString("meetingCohereModel", ""),
   meetingCloudTranscriptionProvider: readString("meetingCloudTranscriptionProvider", ""),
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Merging as is, thanks @hsusul. Confirmed the fresh-install path on main: `migrateUploadTranscription` latches with nothing to copy, onboarding's local step only mirrored `useLocalWhisper` into the upload and meeting scopes, and `readLocalProvider` turned the missing scoped key into `whisper`, so a fresh Parakeet or Cohere install resolved uploads to Whisper `base`. Your test fails on main and passes with the fix, and the full suite is green on main plus this change.

One note for the record: the scoped provider inherits at store init rather than live in the resolver like the model fields do, so a dictation provider changed mid-session in Settings reaches an untouched upload or meeting scope on the next launch. UI and routing read the same field, so they never disagree.

PR #2034, opened 2026-09-04
fix(linux): drop override-redirect while the assistant panel needs the keyboard

### LINKED ISSUE #2027: Linux: Voice agent panel can't receive keyboard input (overlay is an override-redirect window)
**Describe the bug**

The voice agent result panel can't take keyboard focus on Linux. I can click its text input and the caret appears, but everything I type goes to whatever app was focused before the panel opened. In my case that was a terminal, and I watched my message get typed into it while the cursor sat blinking in the panel. Clicking buttons, dragging the panel and selecting text inside it all work normally, and the panel never appears in Alt-Tab.

**To Reproduce**

1. On Linux (GNOME/Wayland, so the app runs under XWayland), press agent keyboard shortcut and dictate a question to the voice agent.
2. Wait for the panel with the answer to appear.
3. Click the "Type a message..." input at the bottom. A caret appears in it.
4. Type anything.
5. The text lands in the previously focused application instead of the panel.

**Expected behavior**

**Desktop:**

- OS: Ubuntu, GNOME, Wayland session (XDG_SESSION_TYPE=wayland), app forced onto XWayland
- Browser: N/A, desktop app
- Version: OpenWhispr 1.9.2 (.deb), Electron 41.10.5

**Additional context**

xprop run while the panel was open and visible on screen:

0x600004 "Voice Recorder": 932x1124
  _NET_WM_WINDOW_TYPE = _NET_WM_WINDOW_TYPE_NORMAL
  WM_PROTOCOLS = WM_DELETE_WINDOW, _NET_WM_PING, _NET_WM_SYNC_REQUEST
  Map State: IsViewable
  Override Redirect State: yes

Override-redirect looks like the cause. The window is invisible to the window manager, so it's missing from _NET_CLIENT_LIST, it never advertises WM_TAKE_FOCUS, and mutter has no way to give it keyboard focus. That also explains why Alt+Tab doesn't know about it. Mouse events still land because X delivers those by pointer position rather than focus, which is why the buttons work but typing doesn't.

_NET_WM_WINDOW_TYPE comes back as NORMAL, so whatever picks the window type is doing the right thing here. The dictation pill (0x60002e, type TOOLBAR) is override-redirect as well.

hsusul:
## Summary
- On Linux X11/XWayland, always-on-top frameless windows are override-redirect, so the WM cannot give the Voice Assistant panel keyboard focus.
- While the panel is open, Linux now drops always-on-top and skip-taskbar so the follow-up input (and Escape) actually receive keys. Overlay chrome is restored when the panel closes.
- macOS and Windows keep the existing always-on-top overlay.

Fixes #2027

## Test plan
- [ ] On GNOME Wayland (XWayland app): open Voice Assistant, click the follow-up field, type — text should stay in the panel, not the previous app.
- [ ] Escape still dismisses the panel.
- [ ] After closing, dictation pill still floats above other windows.
- [ ] `ELECTRON_OVERRIDE_DIST_PATH=/tmp node --test test/helpers/assistantPanelChrome.test.js test/helpers/windowManagerAssistantPanel.test.js`

### DIFF
--- a/src/helpers/assistantPanelChrome.js
+++ b/src/helpers/assistantPanelChrome.js
@@ -1,0 +1,18 @@
+/**
+ * Overlay chrome for the dictation/assistant window.
+ *
+ * On Linux X11/XWayland, Chromium maps always-on-top frameless windows to
+ * override-redirect. The WM then cannot give the assistant panel keyboard
+ * focus, so typing lands in the previously focused app (#2027).
+ */
+function getAssistantPanelChrome({ platform, open }) {
+  if (!open) {
+    return { focusable: false, alwaysOnTop: true, skipTaskbar: true };
+  }
+  if (platform === "linux") {
+    return { focusable: true, alwaysOnTop: false, skipTaskbar: false };
+  }
+  return { focusable: true, alwaysOnTop: true, skipTaskbar: true };
+}
+
+module.exports = { getAssistantPanelChrome };
--- a/src/helpers/windowManager.js
+++ b/src/helpers/windowManager.js
@@ -40,6 +40,7 @@ const {
   WINDOW_SIZES,
   WindowPositionUtil,
 } = require("./windowConfig");
+const { getAssistantPanelChrome } = require("./assistantPanelChrome");
 const AGENT_DICTATION_PILL_SIZE = Object.freeze({ ...WINDOW_SIZES.BASE });
 const { centeredBounds, clampedBounds } = require("./onboardingWindowBounds");
 const { ONBOARDING_DEMO_KINDS, isOnboardingInputAllowed } = require("./onboardingInputPolicy");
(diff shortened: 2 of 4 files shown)


gabrielste1n:
Thanks for taking on #2027, @hsusul. The override-redirect diagnosis in the issue is right and this is a real problem on Linux. We're closing this PR because the mechanism it changes isn't the one that produces override-redirect, so the panel would still not receive keys.

On X11 and XWayland, Chromium marks a window override-redirect in `X11Window::Initialize` when the window type is a popup, menu or tooltip, or when the window is not activatable (`if (!activatable_ || override_redirect_) req.override_redirect = true;`). Always-on-top and frameless never feed into that decision. Our main window is non-activatable because `MAIN_WINDOW_CONFIG` in `src/helpers/windowConfig.js` sets `focusable: false`, which Electron maps to `activatable = kNo`. `setAlwaysOnTop(false)` only toggles `_NET_WM_STATE_ABOVE`, `setSkipTaskbar` only touches `_NET_WM_STATE_SKIP_TASKBAR`, `activatable_` is fixed at creation, and `X11Window::Activate` returns early for a non-activatable window, so after this change `focus()` in `setAssistantPanelOpen` is still a no-op and typing still lands in the previous app.

Two smaller things for whatever comes next: the new `assistantPanelChrome.js` returns a `focusable` value that `windowManager.js` never reads, and the updated assertion in `test/helpers/windowManagerAssistantPanel.test.js` branches on the host's `process.platform`, which main's version of that file now avoids via its `withPlatform` helper (that is also where the merge conflict is).

The fix most likely lives at window creation: make the Linux X11 main window activatable (keeping `showInactive()` so plain dictation doesn't steal focus) or recreate the window when the panel opens, while preserving the Sway behaviour from #1718. It would need a verified run on GNOME/XWayland. Happy to review a pass along those lines.

PR #2037, opened 2026-09-05
fix(enterprise): hydrate managed identity in the dictation window over IPC

greatcoat:
**Problem:** The dictation window never resolves a session, so managed Azure STT resolves `active: true` in Settings but `active: false` in the only window that records; dictation silently falls through to the personal lane.

**Fix:** Expose the main process's validated account scope over IPC (`get-active-account-scope` + an `active-account-scope-changed` broadcast) and hydrate the dictation window's policy and managed identity from it, through the same store entry points `useAuth` already uses. No window capability changes.

## Problem

Reproduced on a live rig (2026-09-05) on top of #1964. With managed Azure STT configured and enforced, the two renderer windows disagree:

```js
const mt = await import("/services/managedTranscription.ts");
JSON.stringify({ active: mt.isManagedTranscriptionActive(), res: mt.getManagedTranscriptionResolution() });
// control panel (?panel=true) -> active: true
// dictation window (bare URL)  -> active: false
```

Settings shows "Managed by your organization"; the dictation window, which is where audio is actually transcribed, does not consider itself managed and records through the personal lane. This breaks the feature for every user on every host, and it fails silently.

Probing the network capability directly in each window:

```js
await fetch("https://api.openwhispr.com/api/workspaces");
// control panel    -> 401 (a real response)
// dictation window -> TypeError: Failed to fetch
```

## Root cause

The dictation window (`MAIN_WINDOW_CONFIG`, `src/helpers/windowConfig.js`) runs with `sandbox: true` and `webSecurity` enforced, while the control panel (`CONTROL_PANEL_CONFIG`) overrides `webSecurity: false`; Better Auth's client uses renderer `fetch`, so its cross-origin session request fails in the dictation window, `useAuth` never reports a signed-in session there, and `refreshManagedEnterpriseIdentity` (`src/hooks/useAuth.ts`, reached only from a resolved session) never runs in that window.

This is explicitly **not** "the stores fail to initialise": `policyStore` and `enterpriseIdentityStore` reach the API over `window.electronAPI` (IPC) and work perfectly in the dictation window. The only thing that window lacks is the identity (`accountId` + credential generation) to ask with.

## Fix

Option (a) from the handover: hand the dictation window the one thing it is missing over IPC, the way every other piece of this feature already crosses the boundary.

**Main process** (`src/helpers/ipcHandlers.js`, `src/helpers/accountScopeBinding.js`):

- `get-active-account-scope` returns `{ accountId, authGeneration } | null`. The account id comes from the persisted account-scope binding through the same hash-checked `resolveBootAccountScope` the boot restore uses, so a binding validated under a rotated bearer resolves to `null` rather than a stale account; the generation is the token store's current one, which is what `get-workspace-policy` and `get-managed-enterprise-config` already gate on. The main process holds both halves before any window resolves a session (the binding is on disk and restored at boot).
- `set-active-account-scope` broadcasts `active-account-scope-changed` with the new scope after persisting, or `null` after a validated sign-out clears it. The token-store subscription broadcasts `null` when the bearer is cleared, so a sign-out that never reaches the handler still clears every window.

**Preload / types**: `getActiveAccountScope`, `onActiveAccountScopeChanged`, and an `ActiveAccountScope` type next to `setActiveAccountScope`.

**Renderer** (`src/lib/accountScopeMirror.ts`, started from `src/AppRouter.jsx` in the dictation window only):

- Subscribes to the broadcast, then reads the scope once at boot; if a broadcast landed while the read was in flight, the read's older answer is ignored.
- A scope is applied through the same two entry points `useAuth` uses for a validated session: `policyStore.fetchPolicy(accountId, authGeneration)` and `refreshManagedEnterpriseIdentity(accountId, authGeneration, workspaceId)`. The active workspace is the control panel's choice, read from the shared `activeWorkspaceId` localStorage key it writes (every window of the origin shares it, the same mechanism `settingsStore` already uses for cross-window sync) and re-applied on that key's `storage` event.
- A `null` scope clears policy and identity.
- No timers or polling: the boot read plus three existing events (scope broadcast, storage event, and the managed-config broadcast the identity store already subscribes to on first refresh) cover every transition.

Behaviours covered:

- Pill boots after a settled sign-in: boot read returns the scope, hydration completes without any user action.
- Pill boots before sign-in, or before the control panel has resolved the workspace: the scope arrives by broadcast; the workspace arrives by storage event; nothing is managed until both are known.
- Sign-out: `null` broadcast, policy and identity cleared in the pill at once.
- Credential rotation (new generation): the main process clears its managed cache as before; the pill keeps its previous identity until the control panel re-validates and re-broadcasts, then clears and refreshes in one tick. Any managed call made in the gap carries the old generation and is rejected by the main process (`AUTH_CONTEXT_CHANGED`), so the gap fails closed rather than falling through; the persisted enforced-scope hint holds transcription closed while the refresh is loading.

Why the pill does not call `useWorkspaceStore.refresh()` itself: the cloud proxy fences every authenticated request on a validated credential generation (`AUTH_CONTEXT_UNVALIDATED` otherwise), which only the control panel's session path commits. Listing workspaces is the control panel's job; the pill mirrors its choice.

This adds **no capability** to the dictation window: `webPreferences` are unchanged, and the new IPC read returns only an account id and an integer generation that the main process already validated.

## Alternatives considered

- **(b) Move the managed decision into the main process entirely.** The Entra token already lives there and `managed-transcribe` is already an IPC call, so the renderer arguably should not be deciding whether it is managed at all. This is the cleaner long-term shape, and it should be on record. It is the largest diff and touches `transcriptionRoute.ts`, so it is out of scope for a fix on top of #1964; this change is compatible with doing it later (the IPC scope becomes an input to that decision rather than to the renderer stores).
- **(c) `webSecurity: false` on `MAIN_WINDOW_CONFIG`.** One line and it would work, but it disables the same-origin policy on an always-on-top window that is currently `sandbox: true`, and the only cross-origin request that window needs is the one session read this fix replaces. Rejected: the fix should add information to the pill window, not capability.
- A previous session's `AppRouter.jsx` scaffold (a `useEffect` polling `localStorage` for the usage account id up to ten times and then calling the same store entry points) is deliberately not shipped; it served only to prove that IPC-driven hydration brings the pill to `isManagedTranscriptionActive() === true`.

## Base branch

Targeted at `feat/managed-transcription-provider` alongside the sibling fixes from the same rig session. The cause lives in files #1964 does not touch (`windowConfig.js`, `useAuth.ts`, `AppRouter.jsx`), and the fix's own hunks do not overlap #1964's changes in the three files both touch (`preload.js`, `ipcHandlers.js`, `electron.ts`).

The diff **applies cleanly to `main`** as well (verified with `git apply --check` against `origin/main` via a temporary index). Author's choice: keep it on this PR so managed STT lands working, or rebase it onto `main` as a standalone fix; it needs no changes for either.

## Deliberately untouched

- `src/helpers/windowConfig.js`: no capability change; the pill stays `sandbox: true` with `webSecurity` enforced.
- `src/hooks/useAuth.ts`: the control panel keeps hydrating from the real session; the pill's `useAuth` still reports signed-out, which is existing behaviour and not what this defect is about.
- `src/stores/enterpriseIdentityStore.ts`, `src/stores/policyStore.ts`: reused through their existing entry points, unchanged.
- `src/stores/workspaceStore.ts`: logic unchanged; only the `activeWorkspaceId` key is exported so the mirror reads the same key the store writes.
- `src/helpers/transcriptionRoute.ts`: option (b) territory.
- `DatabaseManager.activeAccountId`: not exposed; the binding file is the self-verifying source and the boot restore already reads it the same way.

## Tests

Fail-first. Each new test was run alone before the implementation and captured RED:

- `test/helpers/accountScopeIpc.test.js` (new; registers the real handlers against a fake `this`, real `accountScopeBinding` in a temp userData dir, bearer state under test control):
  `TypeError: handlers.get(...) is not a function` (4 of 5; the fifth — a rejected scope request broadcasts nothing — passed vacuously against the old handler).
- `test/helpers/accountScopeBinding.test.js` (extended): `TypeError: binding.resolveActiveAccountScope is not a function`.
- `test/helpers/accountScopeRestoreContract.test.js` (extended, source-pinned like its neighbours because the token-store subscription is constructor wiring the handler harness cannot reach): `AssertionError [ERR_ASSERTION]: the no-token branch broadcasts the cleared scope` (`actual: false, expected: true`).
- `test/lib/accountScopeMirror.test.js` (new; Vite SSR harness with a stubbed `electronAPI`, real policy/identity/workspace stores, and the same `isManagedTranscriptionActive()` probe the live check uses): `Error: Failed to load url /lib/accountScopeMirror.ts ... Does the file exist?` (all 4).

After the implementation, alone: 5/5, 5/5, 4/4, 6/6 (4 tests + 2 subtests). The mirror tests cover: boot with a settled scope reaches `kind: "managed"`, `provider: "azure"`, `deployment: "gpt-4o-transcribe"` with the expected context; a scope that arrives after boot by broadcast and a workspace that arrives by storage event; a new credential generation re-resolving policy and identity with that generation; a `null` scope clearing both stores synchronously; a broadcast landing during the boot read not being overwritten by the read; and `stop()` detaching both listeners.

Full suite (`npm test`, this branch at the commit): 3473 tests, 3285 pass, 185 skipped (the usual Electron-ABI DB skips), 1 todo, 2 fail. Both failures are pre-existing environmental cases, not regressions:

- `test/stores/enterpriseIdentityStoreImports.test.js` "settingsStore imports without a bare localStorage global": the known Node 25 global-`localStorage` failure; this change does not touch `settingsStore`.
- `test/helpers/textEditMonitorWindowBounds.test.js` "a reported window rect is parsed" (`actual: null` for a real window rect read from the OS under suite load). The file alone, three times: 5/5, 4/5, 5/5, with the only miss being the already-known load-flaky "reads are cached per pid for one press, then re-read"; "a reported window rect is parsed" passed in all three solo runs. Nothing in this change is near `textEditMonitor`.

The known-flaky `textEditMonitorSelection` "activateTargetPid resolves false for an unmapped PID" did not fail in this run.

`npx prettier --check` clean on every touched file; `eslint` clean under the `src/` config for the renderer files (`src/helpers/*.js` is ignored by both eslint configs in this repo); `tsc --noEmit` reports only the 4 pre-existing missing-optional-dependency errors (`canvas-confetti`, `@tiptap/extension-mention`, `@tiptap/suggestion`).

## Live verification (still to run)

On a clean boot, signed in to a workspace with managed Azure STT enforced, with no manual step, in both windows:

```js
const mt = await import("/services/managedTranscription.ts");
JSON.stringify({ active: mt.isManagedTranscriptionActive(), res: mt.getManagedTranscriptionResolution() });
```

- Before: `active: true` in `?panel=true`, `active: false` in the bare window.
- After: both `true`, with the same `context.accountId`, `context.workspaceId`, and `context.authGeneration` in each.
- Then one real dictation must log `source: "azure-managed"`.
- Worth adding: sign out in the control panel and re-run the probe in the bare window (`active: false`, no restart); sign back in (`active: true` again, no restart).

## Provenance

Found while proving managed Azure STT end to end on 2026-09-05; on top of #1964.

### DIFF
--- a/preload.js
+++ b/preload.js
@@ -215,6 +215,11 @@ contextBridge.exposeInMainWorld("electronAPI", {
   getSpaces: () => ipcRenderer.invoke("db-get-spaces"),
   setActiveAccountScope: (accountId, expectedAuthGeneration) =>
     ipcRenderer.invoke("set-active-account-scope", accountId, expectedAuthGeneration),
+  getActiveAccountScope: () => ipcRenderer.invoke("get-active-account-scope"),
+  onActiveAccountScopeChanged: registerListener(
+    "active-account-scope-changed",
+    (callback) => (_event, scope) => callback(scope)
+  ),
   deleteAccountData: (accountId, expectedAuthGeneration) =>
     ipcRenderer.invoke("delete-account-data", accountId, expectedAuthGeneration),
   updateSpace: (id, updates) => ipcRenderer.invoke("db-update-space", id, updates),
--- a/src/AppRouter.jsx
+++ b/src/AppRouter.jsx
@@ -9,6 +9,7 @@ import BackgroundModelDownloadTray from "./components/onboarding/BackgroundModel
 import { LEGACY_ONBOARDING_STEP_KEY, ONBOARDING_SESSION_KEY } from "./components/onboarding/flow";
 import { useAuth } from "./hooks/useAuth";
 import { useTheme } from "./hooks/useTheme";
+import { mirrorActiveAccountScope } from "./lib/accountScopeMirror";
 import { usePolicyStore } from "./stores/policyStore";
 import { resolveSettledControlPanelWindowMode } from "./utils/controlPanelWindowMode.ts";
 import { isControlPanelWindow } from "./utils/windowContext.ts";
@@ -79,6 +80,13 @@ function MainApp() {
         .catch(() => {});
     }
   }, [autoSyncReady, isControlPanel]);
+
+  useEffect(() => {
+    // The dictation window cannot resolve a session (see mirrorActiveAccountScope),
+    // so its policy and managed identity follow the main process's account scope.
+    if (!isDictationPanel) return;
+    return mirrorActiveAccountScope();
+  }, [isDictationPanel]);
 
   useEffect(() => {
     if (!authLoaded) return;
--- a/src/helpers/accountScopeBinding.js
+++ b/src/helpers/accountScopeBinding.js
@@ -45,6 +45,15 @@ function resolveBootAccountScope({ token, binding }) {
   return binding.accountId;
 }
 
+// The scope a window without a resolvable session hydrates from (the dictation
+// window has no cross-origin fetch, so its session never resolves): the same
+// validated binding boot restores, paired with the credential generation the
+// policy and managed-config handlers gate on.
+function resolveActiveAccountScope({ token, generation, binding }) {
+  const accountId = resolveBootAccountScope({ token, binding });
+  return accountId ? { accountId, authGeneration: generation } : null;
+}
+
 function read() {
   try {
     const parsed = JSON.parse(fs.readFileSync(bindingFile(), "utf8"));
@@ -77,4 +86,12 @@ function clear() {
   }
 }
 
-module.exports = { clear, evaluateScopeRequest, hashToken, persist, read, resolveBootAccountScope };
+module.exports = {
+  clear,
+  evaluateScopeRequest,
+  hashToken,
+  persist,
+  read,
+  resolveActiveAccountScope,
+  resolveBootAccountScope,
+};
--- a/src/helpers/ipcHandlers.js
+++ b/src/helpers/ipcHandlers.js
@@ -641,6 +641,7 @@ class IPCHandlers {
       if (!token) {
         this.databaseManager.setActiveAccountId(null);
         accountScopeBinding.clear();
+        broadcastToWindows("active-account-scope-changed", null);
       }
       broadcastToWindows("auth-token-state-changed", {
         generation,
@@ -1983,8 +1984,19 @@ class IPCHandlers {
       this.databaseManager.setActiveAccountId(accountId);
       if (accountId !== null) accountScopeBinding.persist(accountId, state.token);
       else accountScopeBinding.clear();
+      broadcastToWindows(
+        "active-account-scope-changed",
+        accountId !== null ? { accountId, authGeneration: state.generation } : null
+      );
       return { success: true };
     });
+
+    ipcMain.handle("get-active-account-scope", () =>
+      accountScopeBinding.resolveActiveAccountScope({
+        ...tokenStore.getState(),
+        binding: accountScopeBinding.read(),
+      })
+    );
 
     ipcMain.handle("delete-account-data", async (_event, accountId, expectedGeneration) => {
       const state = tokenStore.getState();
--- a/src/lib/accountScopeMirror.ts
+++ b/src/lib/accountScopeMirror.ts
@@ -1,0 +1,57 @@
+import { useEnterpriseIdentityStore } from "../stores/enterpriseIdentityStore";
+import { usePolicyStore } from "../stores/policyStore";
+import { ACTIVE_WORKSPACE_KEY, refreshManagedEnterpriseIdentity } from "../stores/workspaceStore";
+import type { ActiveAccountScope } from "../types/electron";
+
+/**
+ * Hydrates the workspace policy and managed enterprise identity in a window
+ * that cannot resolve its own session. The dictation window is sandboxed with
+ * web security enforced, so Better Auth's cross-origin session fetch fails
+ * there and the session-driven hydration in useAuth never runs. The control
+ * panel validates the session and the main process persists that scope; this
+ * mirrors it back over IPC, both the value at boot and every later change
+ * (re-validation under a new credential generation, sign-out). The active
+ * workspace is the control panel's choice, read from the localStorage key it
+ * writes, which every window of this origin shares.
+ */
+export function mirrorActiveAccountScope(): () => void {
+  let scope: ActiveAccountScope | null = null;
+  let broadcastReceived = false;
+
+  const refreshIdentity = () => {
+    if (!scope) return;
+    refreshManagedEnterpriseIdentity(
+      scope.accountId,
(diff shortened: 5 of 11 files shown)


greatcoat:
Live-verified on a macOS dev-build rig at the PR head (`a2211d74`), production API,
virgin `staging` channel profile, signed into a real account in the OpenWhispr
workspace (`6cc0ae2e…`) whose managed Azure provider is `managed_default`.

**Two-window check — pass.** Both renderers report an identical managed resolution:

| | `?panel=true` (Settings) | `/` (dictation pill) |
|---|---|---|
| `isManagedTranscriptionActive()` | `true` | `true` |
| accountId / workspaceId | identical | identical |
| `authGeneration` / `generation` / `providerVersion` | 1 / 11 / 9 | 1 / 11 / 9 |

**One real dictation — pass.** `source: "azure-managed"`, `model: gpt-4o-transcribe`,
transcription 2208 ms, transcript pasted into the target app.

**Revocation — pass.** Signing out in the control panel dropped the bare pill window
to `active: false` within ~2 s, in the same process, with no restart.

**Survives a clean restart — pass.** After a full stop/start both windows came back
`active: true` with the same context.

For provenance: I measured the four results above on this branch only. The pre-fix
baseline (pill at `active: false`, `hasConfig: false`) is carried from the 2026-09-04
investigation rather than re-measured here.

---

**Confirmed alongside: the managed text leg still 404s on the Foundry host**, as
expected since this branch does not carry #2036. Same dictation: transcription
succeeded, then `ENTERPRISE_ERROR` → `404 Resource not found` → `Deployment not
found. Verify the deployment name in Azure portal.` for `gpt-4.1-mini`, then
`REASONING_FAILED { fallbackToCleanup: true }` → `USING_STANDARD_CLEANUP`. That is
the `/openai` vs `/openai/v1` base-URL defect #2036 fixes, now reproduced on live
hardware.

Minor and separate: the paste line still labels that result
`source: "azure-managed-reasoned"` even though reasoning failed and standard cleanup
ran — the label overstates what happened.

One unrelated, pre-existing issue also turned up during this run (sign-out leaves a
multi-workspace account with no active workspace, which disables managed AI until it
is re-picked). It is not caused by this PR — verified by diffing `workspaceStore.ts`
between base `123f9fe7` and this head — and is tracked separately. Flagging it only
because signing out mid-test will otherwise look like a regression here.

PR #2040, opened 2026-09-05
fix(meeting): recover system audio capture when the macOS tap is stranded

### LINKED ISSUE #1990: Meeting audio capture silently stops mid-call on macOS (Firefox / Google Meet)
**Describe the bug**

During a meeting transcription session on macOS, system audio capture stops partway through the call with no error shown anywhere in the app. Audio and transcription work correctly for a while after starting the recording, then transcription of the other participants goes quiet for the rest of the call. The recording indicator stays active the whole time, so nothing in the UI suggests capture has failed.

**To Reproduce**

1. Join a Google Meet call in Firefox on macOS.
2. Start meeting transcription in OpenWhispr (system audio permission already granted).
3. Stay on the call for a while with audio playing normally at first.
4. At some point during the call, system audio capture stops being transcribed. My own voice may or may not continue to transcribe; the other participants' audio does not.
5. No error, toast, or warning appears in the app when this happens.

**Expected behavior**

Meeting transcription should either keep capturing system audio for the whole call, or surface a visible warning/recovery action as soon as capture stops, the same way the initial-silence warning does at the start of a recording.

- OS: macOS
- Browser: Firefox
- Version: 1.9.2

**Additional context**

No debug log yet (debug logging is off by default). From the packaged app and `main` source: `resources/macos-audio-tap.swift` sets up its Core Audio process tap once at startup and never re-checks device-alive or route-change state afterward, and the only system-audio silence watchdog (45s) disarms for good after the first audible chunk, so a mid-call route change could silently stop audio with nothing to catch it. Firefox's audio backend suspends and reopens its output unit during silence more readily than Chromium's, which may explain the Firefox link. Can attach a debug log from a fresh occurrence if that helps.

alekc:
On macOS the system-audio tap is built once against the devices present at creation and never follows the machine afterwards, so a route change or a device going away can stop capture while the helper process stays alive: it does not exit, nothing raises an error, and silence from a dead tap is indistinguishable from a quiet call. That is why #1990 goes unnoticed until the recording is over, and why the existing one-shot silence warning does not catch it: that warning only covers a session that never produced audio, and it stops watching as soon as any arrives. This adds a watchdog in main that restarts the helper process when chunks stop arriving for six seconds or when the tap reports its device changed, capped at three attempts and surfaced to the user either way, plus HAL property listeners in the Swift helper so a route change is reported rather than waiting for the stall window.

Two deliberate limits are worth a reviewer's attention. A call that goes quiet only warns and never restarts, because a meeting where nobody is speaking looks exactly like a dead tap from the audio alone and restarting a working capture costs real audio. The delivery check is opt-in per strategy and currently only the macOS tap sets it, since that helper emits a chunk every period whatever is playing, while the Windows and Linux loopback helpers may legitimately go idle.

CI never compiles the Swift helper, so I built it locally (arm64, target macosx14.2) and smoke-ran the binary: six seconds of capture on an idle machine gave 270kB of PCM and clean start/stop events on stderr, with the emit path serialised now that the listener queue writes there too.

Fixes #1990


Chadpiha (approved):
Thanks for this @alekc !

Four review findings i picked up which i have resolved: replacement-device warnings are retained, startup interruptions survive microphone permission waits, transcript timestamps stay aligned through reconnects, and quiet warnings clear when audio resumes.

Regression coverage includes recovery, replay, teardown, and warning lifecycle behavior. Quality checks and the renderer build pass; 3,433 tests passed, with 196 skipped and one TODO. No further blocking issues found.

PR #2045, opened 2026-09-05
fix(analytics): handle Date objects and defaults in calculateStreaks and activity days (#2044)

### LINKED ISSUE #2044: fix(analytics): handle Date objects and defaults in calculateStreaks and activity days
### Affected component

`src/helpers/analytics.js`

### Describe the bug

In `src/helpers/analytics.js`:

1. In `calculateStreaks(ascendingDays, today)`:
   - `today` has no default parameter (unlike `summarizeAnalyticsDays(days, today = localDateKey())`).
   - When `today` is omitted, `dayNumber(today)` runs `Date.parse("undefinedT00:00:00Z")`, which produces `NaN`.
   - The expiration check `dayNumber(today) - dayNumber(days[days.length - 1]) > 1` evaluates `NaN > 1` as `false`.
   - As a result, stale streaks (even from months or years in the past) are never expired and incorrectly reported as active (`currentStreakDays: 1`).
   - Similarly, if `today` is passed as a `Date` instance (as is accepted by `buildAnalyticsActivityDays` and standard JS date functions), `dayNumber` interpolates `${date}T00:00:00Z`, which parses to `NaN` and similarly fails to expire stale streaks.

2. In `buildAnalyticsActivityDays(daily, today = new Date())`:
   - `daily` lacks a default parameter. Calling `buildAnalyticsActivityDays()` throws `TypeError: Cannot read properties of undefined (reading 'map')` instead of returning the 6-month empty activity template.

3. In `localDateKey(date = new Date())`:
   - If `date` is passed as an epoch timestamp (number) or ISO string, calling `date.getFullYear()` throws `TypeError: date.getFullYear is not a function`.

### To Reproduce

```javascript
const { calculateStreaks, buildAnalyticsActivityDays } = require("./src/helpers/analytics.js");

// 1. Omitted today doesn't expire a month-old streak:
calculateStreaks(["2026-08-01"]);
// Returns { currentStreakDays: 1, longestStreakDays: 1 } instead of { currentStreakDays: 0, longestStreakDays: 1 }

// 2. Date instance doesn't expire a month-old streak:
calculateStreaks(["2026-08-01"], new Date("2026-09-05T12:00:00Z"));
// Returns { currentStreakDays: 1, longestStreakDays: 1 } instead of { currentStreakDays: 0, longestStreakDays: 1 }

// 3. buildAnalyticsActivityDays with no arguments throws:
buildAnalyticsActivityDays();
// TypeError: Cannot read properties of undefined (reading 'map')
```

### Expected behavior

- `calculateStreaks` should default `today` to `localDateKey()`.
- `dayNumber` should support both date strings (`"YYYY-MM-DD"`) and `Date` instances.
- `buildAnalyticsActivityDays` should default `daily = []`.
- `localDateKey` should accept `Date` instances, timestamps, or date strings.

hsusul:
### Summary of changes

This PR fixes three date and default handling issues in `src/helpers/analytics.js`:

1. **Stale streak expiration with omitted `today` or `Date` instances**:
   - In `calculateStreaks(ascendingDays, today)`, `today` previously lacked a default parameter. When omitted, `dayNumber(today)` evaluated to `NaN`, causing `dayNumber(today) - dayNumber(lastDay) > 1` to evaluate to `false`. Consequently, historical streaks never expired and were reported as active streaks.
   - When callers passed a `Date` instance (as accepted by `buildAnalyticsActivityDays`), `dayNumber` interpolated `${value}T00:00:00Z`, also parsing to `NaN` and failing to expire stale streaks.
   - **Fix**: Added default `today = localDateKey()` and updated `dayNumber` to correctly handle `Date` instances (as well as timestamps and ISO strings) by computing UTC calendar day numbers.

2. **Omitted `daily` in `buildAnalyticsActivityDays`**:
   - `buildAnalyticsActivityDays(daily, today = new Date())` previously threw a `TypeError: Cannot read properties of undefined (reading 'map')` if `daily` was omitted.
   - **Fix**: Defaulted `daily = []`.

3. **Robustness in `localDateKey`**:
   - Updated `localDateKey` to handle date strings (preserving `YYYY-MM-DD` keys directly to avoid UTC midnight timezone rollback) and timestamp numbers.

### Verification

- Added unit tests in `test/helpers/analytics.test.js` verifying:
  - Stale streak expiration when `today` is omitted (defaults to current date).
  - Stale streak expiration and active streak counting when `today` is provided as a `Date` instance.
  - `buildAnalyticsActivityDays` returns the 6-month empty activity template when `daily` is omitted.
  - `localDateKey` formats date strings, `Date` objects, and numeric timestamps.
- Ran `node --test test/helpers/analytics.test.js` (10/10 passed).
- Ran `node --test test/services/analyticsService.test.js` (17/17 passed).

Closes #2044

### DIFF
--- a/src/helpers/analytics.js
+++ b/src/helpers/analytics.js
@@ -30,13 +30,18 @@ export function countSpokenWords(text) {
 }
 
 export function localDateKey(date = new Date()) {
-  const year = date.getFullYear();
-  const month = String(date.getMonth() + 1).padStart(2, "0");
-  const day = String(date.getDate()).padStart(2, "0");
+  if (typeof date === "string" && /^\d{4}-\d{2}-\d{2}$/.test(date)) {
+    return date;
+  }
+  const d = date instanceof Date ? date : new Date(date);
+  if (Number.isNaN(d.getTime())) return "";
+  const year = d.getFullYear();
+  const month = String(d.getMonth() + 1).padStart(2, "0");
+  const day = String(d.getDate()).padStart(2, "0");
   return `${year}-${month}-${day}`;
 }
 
-export function buildAnalyticsActivityDays(daily, today = new Date()) {
+export function buildAnalyticsActivityDays(daily = [], today = new Date()) {
   const byDate = new Map(daily.map((bucket) => [bucket.date, bucket]));
   const startDate = new Date(
     today.getFullYear(),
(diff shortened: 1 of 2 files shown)


gabrielste1n:
Thanks for the tests, @hsusul. We're closing this one: none of the inputs it guards against can reach these helpers.

- `calculateStreaks` has one caller, `summarizeAnalyticsDays` in `src/helpers/analytics.js`, which always passes a `YYYY-MM-DD` string for `today`.
- `buildAnalyticsActivityDays` is only called from the Insights heatmap with `summary.daily`, which is always an array; the cloud summary is validated in `src/services/AnalyticsService.ts` before it reaches the component.
- Every `localDateKey` caller (`src/helpers/audioManager.js`, `src/helpers/database.js`) passes a `Date` instance.

The change also moves the wrong way on failure handling: `localDateKey` would return `""` for an unparseable value and `"1970-01-01"` for `null`, and that string is written straight into `analytics_events.local_date`, so a future bug would silently corrupt streaks and the heatmap instead of failing at the insert as it does today. Closing #2044 with this for the same reason. If you find a path in the app that actually delivers one of these values, please open an issue with the trace and we'll fix it there.

PR #2051, opened 2026-09-07
fix: give Audio Upload its own self-hosted transcription URL

### LINKED ISSUE #2049: Self-hosted Server URL is shared between Dictation and Audio Upload despite separate settings tabs
Describe the bug
In Settings → Speech-to-Text, OpenWhispr provides separate configuration tabs for:
- Dictation
- Note Recording
- Audio Upload
Each tab has its own UI for selecting the Speech-to-Text engine and configuring a Self-hosted Server URL.
However, the Self-hosted Server URL for Dictation and Audio Upload appears to use the same underlying setting.
If I change the Server URL under Dictation, the URL under Audio Upload changes to the same value automatically, and vice versa.
This makes it impossible to route Dictation and Audio Upload workloads to different self-hosted transcription servers.
Steps to reproduce
1. Open Settings → Speech-to-Text → Dictation.
2. Select Self-hosted Server.
3. Set Server URL to, for example:
   http://192.168.11.83:8090
4. Switch to Audio Upload.
5. Select Self-hosted Server.
6. Change its Server URL to another server, for example:
   http://192.168.11.84:8090
7. Return to Dictation.
Actual behavior
The Dictation Server URL has also changed to:
http://192.168.11.84:8090
The two tabs appear to share the same Self-hosted Server URL configuration.
Expected behavior
Dictation and Audio Upload should retain independent Self-hosted Server configurations.
For example:
Dictation
http://192.168.11.83:8090
Audio Upload
http://192.168.11.84:8090
Ideally, the entire Self-hosted configuration should be independent per workload, including:
- Server URL
- Model
- other workload-specific STT parameters
Why separate endpoints are important
We are deploying OpenWhispr in an office environment with approximately 20 employees.
We intentionally use separate transcription servers for different workloads:
Server A — Dictation
- short audio requests
- high concurrency
- optimized for minimum latency
- interactive workload
Server B — Audio Upload
- long audio recordings
- significantly heavier processing
- speaker diarization
- batch/background workload
Running both workloads against the same server is inefficient. A long transcription/diarization task can consume GPU resources and increase the latency of interactive dictation for all employees.
Our intended architecture is:
OpenWhispr Dictation → Fast ASR Server A
OpenWhispr Audio Upload → ASR / Diarization Server B
The OpenWhispr UI already presents Dictation and Audio Upload as separate Speech-to-Text configurations, so we expected the Self-hosted Server settings in these tabs to be stored independently.
Suggested implementation
Store separate Self-hosted configurations for each workload, for example:
dictation.remoteTranscriptionUrl
audioUpload.remoteTranscriptionUrl
and preferably separate model settings as well.
For backward compatibility, existing installations could initially copy the current shared Self-hosted configuration into both profiles.
Environment
- macOS
- OpenWhispr desktop
- Self-hosted Speech-to-Text
- Approximately 20-client office deployment

STiFLeR7:
Fixes #2049

## Root cause
Settings → Speech-to-Text presents separate tabs for Dictation, Note Recording, and Audio Upload, each with its own self-hosted server config UI. But Audio Upload's Server URL read/wrote the same shared `remoteTranscriptionUrl` key that Dictation uses — `UploadSettings.tsx` and `UploadAudioView.tsx` both referenced the raw shared field directly instead of an upload-specific one. Changing the URL in one tab silently changed it in the other, exactly as reported (confirmed by reproducing the repro steps against `main`).

Every other upload-context field (`uploadWhisperModel`, `uploadCloudTranscriptionProvider`, etc.) already has its own `upload*`-prefixed key with a documented "audio upload gets its own context" migration (`UPLOAD_TRANSCRIPTION_PAIRS`) — `remoteTranscriptionUrl` was simply missing from that list. The meeting context already has the analogous `meetingRemoteTranscriptionUrl` field (`state.meetingRemoteTranscriptionUrl || state.remoteTranscriptionUrl`), so this change follows that exact, already-established pattern rather than inventing a new one.

## Fix
- Added `uploadRemoteTranscriptionUrl` (state field, setter, default read from localStorage) mirroring `meetingRemoteTranscriptionUrl`.
- Added it to `UPLOAD_TRANSCRIPTION_PAIRS` for backward-compat migration consistency with the other upload fields.
- `selectResolvedUploadTranscription` (the selector the actual upload transcription runtime consumes) now resolves `uploadRemoteTranscriptionUrl || remoteTranscriptionUrl`, so existing single-server users see no change until they explicitly set an Audio Upload-specific URL.
- `UploadSettings.tsx` (Settings UI) and `UploadAudioView.tsx` (runtime config builder) now read/write the upload-specific field instead of the shared one.

Scoped to the reported Server URL field only — left `remoteTranscriptionModel` shared, since that's already the established behavior for the meeting context too (no per-context model override exists there either), so decoupling it here would be inconsistent with the rest of the codebase rather than fixing a reported bug.

## Test evidence
Added `test/helpers/uploadRemoteTranscriptionUrlIsolation.test.js`, which reproduces the exact issue repro (set Dictation's URL, then Audio Upload's URL, assert Dictation's is unchanged) and confirms the fallback for existing single-server setups. Red-before-green verified: fails on `main` with `dictation-server:8090` clobbered to `undefined`/overwritten, passes with this fix.

```
✔ audio upload self-hosted URL is isolated from dictation's
```

Also ran the full adjacent settings-store test suite (`policyEffectiveSettings`, `settingsStoreLocalProviderMigrations`, `settingsStoreRetiredGroqModels`, `transcriptionModelMemory`) — all 22 tests pass, no regressions. `tsc --noEmit` and `eslint` both clean on the touched files.

### DIFF
--- a/src/components/notes/UploadAudioView.tsx
+++ b/src/components/notes/UploadAudioView.tsx
@@ -288,6 +288,7 @@ export default function UploadAudioView({ onNoteCreated, onOpenSettings }: Uploa
     cloudTranscriptionBaseUrl,
     cloudTranscriptionMode,
     transcriptionMode,
+    remoteTranscriptionUrl,
   } = useSettingsStore(
     useShallow((settings) =>
       selectResolvedUploadTranscription(selectPolicyEffectiveSettings(settings, policyState))
@@ -295,7 +296,6 @@ export default function UploadAudioView({ onNoteCreated, onOpenSettings }: Uploa
   );
   const uploadAllowedByPolicy = useTranscriptionContextAllowed("upload");
 
-  const remoteTranscriptionUrl = useSettingsStore((s) => s.remoteTranscriptionUrl);
   const remoteTranscriptionModel = useSettingsStore((s) => s.remoteTranscriptionModel);
 
   const setUploadTranscriptionMode = useSettingsStore((s) => s.setUploadTranscriptionMode);
--- a/src/components/settings/UploadSettings.tsx
+++ b/src/components/settings/UploadSettings.tsx
@@ -34,8 +34,8 @@ export function UploadTranscriptionPanel() {
     uploadCloudTranscriptionBaseUrl,
     setUploadCloudTranscriptionBaseUrl,
     setUploadCloudTranscriptionMode,
-    remoteTranscriptionUrl,
-    setRemoteTranscriptionUrl,
+    uploadRemoteTranscriptionUrl,
+    setUploadRemoteTranscriptionUrl,
     remoteTranscriptionModel,
     setRemoteTranscriptionModel,
   } = useSettingsStore();
@@ -146,8 +146,8 @@ export function UploadTranscriptionPanel() {
       {effectiveTranscriptionMode === "self-hosted" && (
         <SelfHostedPanel
           service="transcription"
-          url={remoteTranscriptionUrl}
-          onUrlChange={setRemoteTranscriptionUrl}
+          url={uploadRemoteTranscriptionUrl}
+          onUrlChange={setUploadRemoteTranscriptionUrl}
           model={remoteTranscriptionModel}
           onModelChange={setRemoteTranscriptionModel}
         />
--- a/src/stores/settingsStore.ts
+++ b/src/stores/settingsStore.ts
@@ -454,6 +454,7 @@ const UPLOAD_TRANSCRIPTION_PAIRS: ReadonlyArray<[string, string]> = [
   ["cloudTranscriptionBaseUrl", "uploadCloudTranscriptionBaseUrl"],
   ["cloudTranscriptionMode", "uploadCloudTranscriptionMode"],
   ["transcriptionMode", "uploadTranscriptionMode"],
+  ["remoteTranscriptionUrl", "uploadRemoteTranscriptionUrl"],
 ];
 
 function migrateUploadTranscription() {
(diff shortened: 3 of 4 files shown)


boseq (approved):
Tested on macOS with two mock self-hosted servers: Dictation and Audio Upload now keep separate URLs and models, and requests go to the right server. I pushed one commit (122fe4d) so an empty Upload URL no longer quietly falls back to the dictation server, plus a render test for the Upload tab.

boseq:
@Chadpiha heads up: 122fe4d removes the fallback from c54895c, so an empty Audio Upload URL or model no longer uses dictation's server.

Why: with the fallback, a fresh install that only set a dictation server showed blank URL and model fields on the Upload tab, but uploads still went to the dictation server. Same after clearing the fields. For the #2049 setup (fast dictation server, separate heavy upload server) that's the silent cross-routing the issue is about.

Trade-off: the self-hosted server is now the one upload setting that doesn't inherit from dictation. Cloud provider, model and local model still do. The one-time copy from c54895c means existing profiles keep uploading where they do today and now see that server in the tab. An empty URL shows the upload page's "No transcription provider set up" prompt instead.

The other option was to keep the fallback and show dictation's URL greyed out in the empty field. I didn't go that way because the field already shows an example URL there, so an inherited one would read like another example, and there'd be no way to leave uploads without a server. Happy to switch if you'd rather keep it inheriting.

Tested on 122fe4d, macOS dev build, two mock servers, starting from a profile that had one shared server:
- After the upgrade the Upload tab showed the existing server, and an upload went there.
- After setting Upload to server B, uploads went to B with the upload model, dictation still went to A, and both survived a restart.
- On a fresh profile with an empty Upload URL, the upload page showed the prompt and no request was sent.

Issue #2059, opened 2026-09-07
Windows: no supported way to run elevated, so dictation into elevated windows silently fails


davidbhoward:
**Is your feature request related to a problem? Please describe.**

On Windows, dictating into a window owned by an elevated process (an administrator terminal, or anything started with "Run as administrator") silently does nothing. The hotkey fires, the audio records, the transcription completes and lands in history, but no text reaches the focused window and no error appears anywhere.

The cause is UIPI: Windows refuses synthetic input sent from a lower-integrity process to a higher-integrity window, and does not report the rejection to the sender. All three Windows delivery paths are user-mode synthetic input, so all three are blocked:

- `windows-fast-paste.exe` uses `SendInput` (`resources/windows-fast-paste.c:152`)
- the PowerShell fallback uses `[System.Windows.Forms.SendKeys]::SendWait('^v')` (`src/helpers/clipboard.js:1314`)
- `nircmd.exe` is the remaining fallback

A maintainer already identified this mechanism in #1092 ("Windows silently blocks simulated keystrokes sent from a normal app into an elevated one... Making that failure loud instead of silent is on us"). That issue was closed after the reporter said things seemed better, so the elevation path was never addressed.

The part with no remedy today is that OpenWhispr cannot be run elevated in a supported way. `autoStart.js` implements launch at login through `setLoginItemSettings()`, which writes `HKCU\...\Run`. Explorer starts `Run` entries with the user's filtered token, and no flag changes that. It is a property of the mechanism, not a missing option.

Measured here by reading each process's token (`GetTokenInformation` / `TokenElevation`):

| process | elevated |
|---|---|
| OpenWhispr main, started from the `Run` key | no |
| `explorer.exe` | no |
| administrator terminal | yes |

Starting OpenWhispr elevated makes dictation into the administrator terminal work immediately, with nothing else changed.

**Describe the solution you'd like**

A Windows-only "Launch as administrator" option beside "Launch at login" that registers a Scheduled Task with `RunLevel=Highest` and `LogonType=Interactive`, carrying the existing `--hidden` flag. It would ask for elevation once when enabled and never prompt again.

Verified that this shape does what is needed, using a throwaway task with that exact principal:

```json
{ "elevated": true, "user": "<machine>\\<user>", "sessionId": 1 }
```

Elevated, in the interactive session, and no UAC prompt at any point. Session 1 matters: `LogonType=Interactive` is what keeps it on the user's desktop.

Notes for whoever picks this up:

[...]

Issue #1092, opened 2026-07-08
Cant STT Into Notepad ++


djnaydee:
**Describe the bug**
Doing speech-to-text directly into Notepad++ doesn't work most of the time, especially in these newer versions like the 1.7.4. The software just ignores it. Similar to how most STT works with the built-in notepad software, which stopped working several months ago, but that's a different conversation.

**To Reproduce**
Just try using speech-to-text in Notepad++, the latest version.

**Expected behavior**
Nothing at all shows up in your document.

**Desktop (please complete the following information):**
Windows 11, fully up to date as of 7-7-26

gabrielste1n:
Thanks for the report — and sorry for the frustration. Some context on what's likely happening, plus two quick tests that would pin it down.

**What's going on under the hood:** transcription is almost certainly working — the failing step is the paste. OpenWhispr puts the text on your clipboard, then simulates a Ctrl+V keystroke into the focused window. When that keystroke doesn't land, the app currently can't tell: it assumes the paste worked and restores your previous clipboard a moment later, so the text vanishes without a trace. Making that failure loud instead of silent is on us — but the right fix depends on *why* the keystroke isn't landing on your machine.

### Test 1: Is Notepad++ running as administrator?

Windows silently blocks simulated keystrokes sent from a normal app (like OpenWhispr) into an elevated one — no error is shown to anyone. Notepad++ ends up elevated more often than people realize: if you've ever let it relaunch as admin to save a protected file (like `hosts`), that elevated instance sticks around and opens all your later files too. Same if it's launched from an admin terminal.

- Check: Task Manager → **Details** tab → right-click the column header → *Select columns* → enable **Elevated** → look at `notepad++.exe`.
- Or simpler: close every Notepad++ window, reopen it normally from the Start menu (not "Run as administrator"), and try dictating.

If dictation works in a non-elevated Notepad++, that's our answer.

### Test 2: Does it depend on how long you dictate?

There's another open Windows 11 report (#829) where the simulated Ctrl+V silently fails for longer dictations (roughly 200+ characters) while short ones work — across multiple apps. In the same Notepad++ window, try a very short dictation ("testing one two three") and then a couple of full sentences. Do the short ones paste while the long ones don't?

### Meanwhile, a workaround

[...]

PR #2060, opened 2026-09-07
feat(autostart): elevated launch at login on Windows (mechanism only)

davidbhoward:
Part of #2059. **Draft on purpose** — this is the mechanism only, with no UI. I would rather check the approach is one you want before writing the settings toggle and translating four new strings into ten locales.

> ⚠️ **Stacked on #2058.** This branch is based on that fix, so the diff here contains both commits. Only the second one (`feat(autostart): add the elevated Windows login mechanism`) belongs to this PR. If #2058 lands first this rebases to a single commit.

## Why

On Windows, dictation into any window owned by an elevated process is discarded silently — no error, no log line, transcription still lands in history. UIPI blocks synthetic input from a lower-integrity process into a higher-integrity window and reports nothing to the sender. All three delivery paths are affected: `SendInput` in `windows-fast-paste.exe`, the `SendKeys::SendWait('^v')` fallback in `clipboard.js`, and `nircmd.exe`.

A maintainer identified this mechanism in #1092 ("Making that failure loud instead of silent is on us"), but there is currently no way to run OpenWhispr elevated in a supported way. `setLoginItemSettings()` writes `HKCU\...\Run`, and Explorer starts `Run` entries with the user's filtered token. No flag changes that; it is a property of the mechanism.

## What is here

`windowsElevatedAutostart.js`, mirroring how `linuxAutostart.js` stands in on a platform where `setLoginItemSettings()` cannot express what is needed. It registers a Scheduled Task with `RunLevel=Highest` and `LogonType=Interactive`. Task Scheduler runs as SYSTEM and mints the token itself rather than requesting elevation, so the app starts elevated at login with **no UAC prompt**. Creating or deleting the task needs administrator rights once, which is the single prompt the user sees when they flip the switch.

Decisions live in `autoStartPolicy.js`, pure and unit-tested:

- `resolveAutoStartMechanism()` enforces that the `Run` entry and the task are **mutually exclusive**. Both start the app at login, so running both leaves two instances racing the single-instance lock.
- `needsHiddenFlagMigration()` is suppressed while the task owns startup. Without that it would restore the `Run` entry that switching to the task just removed, on every launch.
- `resolveAutoStartState()` reports `elevated`, and counts the task as launch-at-login so the switch does not read as off while the task is what starts the app.

Three task settings are load-bearing and easy to get wrong:

| setting | default | why it is overridden |
|---|---|---|
| `ExecutionTimeLimit` | `PT72H` | Task Scheduler would kill a tray app after three days, with no visible cause |
| `DisallowStartIfOnBatteries` | `true` | autostart would silently not happen on battery |
| `LogonType` | — | `InteractiveToken` keeps it in the user's session; `Password`/`S4U` run with no desktop to type into |

`schtasks /XML` also rejects UTF-8, so the file is written UTF-16LE with a BOM to match the encoding its prolog declares.

## Verification

Generated XML registered through real `schtasks` on Windows 11, with a throwaway task name:

```
CREATE status: 0   SUCCESS
PROPS: RunLevel=Highest LogonType=Interactive Trigger=MSFT_TaskLogonTrigger
       ExecTimeLimit=PT0S Batteries=True
RUN:   { "elevated": true, "user": "<machine>\\<user>", "sessionId": 1 }
DELETE status: 0
```

Elevated, in the interactive session (session 1, not session 0), no prompt. Separately confirmed that starting OpenWhispr elevated makes dictation into an administrator terminal work immediately with nothing else changed.

`test/helpers/autoStartPolicy.test.js`: 24 pass. One existing assertion changed, `"a startup item Windows will actually launch reads as enabled"`, because `resolveAutoStartState` now also returns `elevated`. `elevated` is Windows-only, so the macOS assertions are untouched.

## What is deliberately missing

- `autoStart.js` dispatch to the new mechanism
- IPC handler + `preload.js` method
- the `SettingsPage` toggle and its four i18n keys across ten locales
- a docs entry

## Questions before I finish it

1. Is a scheduled task the mechanism you want, or would you rather start with making the failure loud (#1092) and treat elevation separately?
2. Task naming: currently `OpenWhispr Launch At Login (<username>)`. Task names are machine-global and the principal is per-user, so two accounts on one machine would otherwise collide. Happy to change the shape.
3. `workspace-policy.json` suggests managed deployments may want to gate this. Should it respect a policy flag?
4. A standard (non-administrator) user cannot create a `RunLevel=Highest` task at all. Currently that surfaces as a failed result; I would rather show a specific message, but that needs a string.

Happy to close this if the direction is wrong — nothing here is load-bearing yet.

For transparency, I worked through this with Claude Code (Opus 5). The verification output and code references are real output from my machine.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -109,6 +109,7 @@ OpenWhispr is an Electron-based desktop dictation application that uses whisper.
   - `HIDDEN_LAUNCH_FLAG` (`--hidden`) is how a login launch tells the app to start in the tray. Windows has no native equivalent (`openAsHidden` is macOS-only and a no-op on macOS 13+), so the flag rides on the login item's `args`; Linux puts it on the autostart entry's `Exec`; macOS uses `wasOpenedAtLogin` instead
   - On Windows, read the state from `executableWillLaunchAtLogin`, never from `openAtLogin`: `openAtLogin` only compares the `Run` value against the current executable and args and ignores the `StartupApproved` key that Task Manager and Settings write when a user disables a startup app
   - Reads and writes must pass identical `args`, or `openAtLogin` always reports false
+  - `getLoginItemLookupPath()` supplies the Windows `path`, **quoted**. Electron parses both that path and every `Run` value with `CommandLine::FromString` and compares the parsed programs, so an unquoted `C:\Program Files\...` truncates to `C:\Program` and compares equal to any other unquoted `Run` entry under `C:\Program Files`. Unquoted, `executableWillLaunchAtLogin` reports an unrelated app's startup entry as ours, stays true after our own entry is deleted, and returns true for a path that does not exist. Passing no `path` is not a way out: Electron then falls back to the equally unquoted `GetProcessExecPath()`
 - **linuxAutostart.js**: Launch-at-login on Linux via an XDG autostart entry
   - `app.setLoginItemSettings()` is a no-op on Linux, so the entry is written directly to `$XDG_CONFIG_HOME/autostart/open-whispr.desktop`, matching the executable name electron-builder packages under
   - `Exec` resolves from `$APPIMAGE` first: `process.execPath` is the ephemeral AppImage FUSE mount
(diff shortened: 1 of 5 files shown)


davidbhoward:
Closing in favour of https://github.com/OpenWhispr/openwhispr/pull/2061, which branches from `main` and contains only this feature's commit. This one was accidentally based on the #2058 branch, so its diff carried that fix too and the two changes could not be reviewed independently.

PR #2066, opened 2026-09-08
fix(perf): prevent high cpu usage in idle ui states

Scott-Poole:
**Root cause:** The pulse-glow animation appears to cause a repaint on every animation frame, leading to high CPU usage in idle states. CSS animations combined with backdrop-blur had a similar demand for repaints.

**Solution:** Remove the offending pulse-glow animation. Remove backdrop-blur from some components where either UX is not impacted or idle ui performance benefits.

**Setup:**
Distributor ID:	Ubuntu
Description:	Ubuntu 24.04.4 LTS
Release:	24.04
Codename:	noble

**Reproducing the CPU usage spikes:**

Open `top` and monitor OpenWhispr’s CPU usage while visiting these three UI states:

  1. Hotkey recording: Settings > Hotkeys > Dictation Hotkeys > click to change
  2. Speech-to-text: Settings > Speech-to-text > Dictation (local, OpenAI, Base)
  3. Google Calendar integration loading: Integrations > Connect to Google Calendar (do not connect in browser, return to OpenWhisper)

  In each case, CPU usage can reach 100% and remain elevated while the UI waits without further interaction.

Before changes:
image

After fix:
image

  **Other testing activities after fix:**

  - Visit every menu, settings menu, and submenu.
  - Leave each page idle long enough for CPU usage in `top` to stabilize.
  - Confirm there is no sustained high CPU load after the UI settles.

### DIFF
--- a/src/components/CliIntegrationCard.tsx
+++ b/src/components/CliIntegrationCard.tsx
@@ -39,7 +39,7 @@ export default function CliIntegrationCard({ isPaid, onUpgrade }: CliIntegration
   };
 
   return (
-    <div className="rounded-lg border border-border/50 dark:border-border-subtle/70 bg-card/50 dark:bg-surface-2/50 backdrop-blur-sm p-4">
+    <div className="rounded-lg border border-border/50 dark:border-border-subtle/70 bg-card/50 dark:bg-surface-2/50 p-4">
       <div className="flex items-center gap-2 mb-4">
         <LogoTile src={logo} alt="OpenWhispr" />
         <div className="w-9 h-9 rounded-lg bg-white dark:bg-surface-raised shadow-[0_0_0_1px_rgba(0,0,0,0.04)] dark:shadow-none dark:border dark:border-white/5 flex items-center justify-center shrink-0">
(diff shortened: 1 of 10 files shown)


yfsfan:
**Additional macOS reproduction result**
I tested this on an Apple Silicon M4 Mac with local Whisper Large and Qwen3.5 9B.
When OpenWhispr is idle on the main UI, CPU/GPU usage drops back to normal and the fans stop.
However, simply opening the model selection/settings page causes OpenWhispr Helper and OpenWhispr Helper (Renderer) CPU/GPU usage to immediately increase significantly, even when no transcription or LLM inference is running.
Closing the model settings page while leaving OpenWhispr running immediately drops the CPU/GPU usage again.
I repeated this several times and the behavior is reproducible. This suggests the high idle load is specifically related to rendering or polling on the model selection/settings UI, rather than the local Whisper/LLM models remaining loaded in memory.
I also confirmed that keeping the local models loaded does not by itself cause sustained high CPU/GPU usage when the settings page is closed.

Scott-Poole:
Merge conflicts addressed. The issue is still present on main.

PR #2068, opened 2026-09-08
feat(models): add LFM2.5 2.6B to the local model list

### LINKED ISSUE #2032: Add LFM2.5 2.6B to local models list

I run OpenWhispr fully locally and I want a small model that handles cleanup, rewrite, and agent steps well without a cloud call. The models list already has the older LFM2 2.6B but not the new LFM2.5 2.6B that Liquid AI released on August 4. Right now I have to sideload it as a custom model in llama-server, which is annoying.

Add LFM2.5 2.6B in the local models list

I tried LFM2.5 1.2B local. It's fast but doesn't correctly remove filler words for me.

**Additional context**

N/A.

greatcoat:
Closes #2032.

## Problem

The local model list already carries LFM2.5 at 1.2B, 8B MoE, 350M and 230M — but the only **2.6B** entry is the older **LFM2** generation. The 2.6B sweet spot of the current family was reachable only by sideloading the GGUF into llama-server by hand, which is what the issue reports.

## Root cause

Nothing broken — the entry was simply never added when LFM2.5 landed. Local models are fully data-driven from `modelRegistryData.json` (`getModelProvider`, the settings-store migrations and the picker all read the registry), so no code path hard-codes the catalog and none needed changing.

## Fix

One registry entry plus its description in all 11 locales. 22 added lines, nothing removed, no code touched.

| field | value | why |
|---|---|---|
| `sizeBytes` | `1674455040` | Live `content-length` of the published GGUF. Feeds the pre-download disk-space check in `modelManagerBridge`, so it has to be exact. |
| `size` | `1.7GB` | Decimal rounding — the only convention the repo actually encodes (`gemmaMtpRegistry.test.js`: `${(bytes/1e9).toFixed(1)}GB`), and what Finder and the browser report for the transfer. |
| `contextLength` | `131072` | See note below. |
| `quantization` | `q4_k_m` | Matches every ≥1B entry in the block. |

**Compatibility.** `LFM2.5-2.6B` is `model_type: lfm2` with 30 layers, hidden 2048, intermediate 10752 — structurally the same as the LFM2 2.6B already shipping. The pinned llama.cpp (`b9763`) already runs four LFM entries, so the bundled runtime needs no change.

**One deliberate inconsistency.** `contextLength` is `131072` (what Liquid AI publish) while its four siblings carry `32768` despite their own configs declaring 128K. The field is documented in `modelManagerBridge.js` as *"the full trained context"* and is clamped to `SERVER_CONTEXT_SIZE` (16384) at runtime, so the value is truthful and behaviour-neutral either way. The stale sibling values are left alone rather than swept into this PR.

## Tests

- `test/locales/translationCoverage.test.js` — gates the new `descriptionKey` in `en` and its presence in all 10 other locales.
- Ran the registry/locale suite: **22/22 pass** (`translationCoverage`, `modelRegistryProviderFallback`, `settingsStoreLocalProviderMigrations`, `localModelSelections`, `modelListBaseCandidates`, `pendingLocalModels`).
- Verified through the real accessor that the model surfaces from `modelRegistry.getAllProviders()`, routes `local`, does not shadow the existing LFM2 2.6B, and builds the download URL `https://huggingface.co/LiquidAI/LFM2.5-2.6B-GGUF/resolve/main/LFM2.5-2.6B-Q4_K_M.gguf` — which returns `200` with a `content-length` matching `sizeBytes` byte for byte.

No new test ships with this: the catalog invariants are already covered generically, and a bespoke test for one data row would just restate the data.

## Known gap, not addressed here

The issue asks for cleanup, rewrite **and agent steps**. Local tool-calling is gated on `estimateModelSizeB(model) >= 4` ([`useChatStreaming.ts:277`](https://github.com/OpenWhispr/openwhispr/blob/main/src/components/chat/useChatStreaming.ts#L277)), and that estimator correctly reads `2.6` from this id — so the agent will run **without tools**. That floor is pre-existing and already applies to every LFM entry including the recommended 1.2B, so this is not a regression. Worth a separate look though: Liquid AI post-trained this model specifically for agentic work and claim parity with models 4× larger on tool use, which is exactly the case the 4B floor was written to exclude.


greatcoat:
## Self-review findings

Ran `/code-review high` on this PR and verified every finding against the vendor's files and our own code rather than taking them at face value. Two were real and are now fixed; one was directionally right but its suggested fix was wrong; the rest are recorded below.

### Fixed in 1b2cc8d5

- **`size` understated the download.** It read `1.6GB` (GiB rounding, copied from the sibling LFM2 2.6B). The only convention the repo actually *encodes* is decimal — `gemmaMtpRegistry.test.js` pins `${(bytes/1e9).toFixed(1)}GB`, giving **1.7GB**, which is also what Finder and the browser report. Now `1.7GB`. (The sibling LFM2 2.6B is `1.5GB` where the same rule gives `1.6GB`; left alone, out of scope.)
- **The description made a claim nothing supports.** "stronger cleanup and rewrite quality" → **"Liquid AI's agentic 2.6B, always reasons before answering"**. Liquid AI recommend this model for *"agentic workloads, tool use, data extraction, RAG, and long-context workflows"* and explicitly not for knowledge-heavy tasks. Updated in all 10 locales.

### The finding worth reading — this model always thinks

Verified from the vendor's own model card: *"LFM2.5-2.6B is a pure reasoning model that always thinks before it answers. It adds a `<think>` tag directly in the chat template."* Its `chat_template.jinja` ends `{{- "<|im_start|>assistant\n<think>" -}}` whenever `add_generation_prompt` is set, and we launch llama-server with `--jinja`, so the embedded template applies and the forced `<think>` reaches every turn. Its four LFM siblings all end with a plain `<|im_start|>assistant\n` — this is the first entry in the local catalog that forces a reasoning pass.

**We have no lever to turn it off.** `suppressThinking` (`thinkingSuppressionDialects.ts`) sends `think: false` (Ollama's switch — llama-server ignores it) and `chat_template_kwargs: { enable_thinking: false }` (Qwen's convention). LFM2.5's template declares **no `enable_thinking` variable at all** — its only related knob is `preserve_thinking`, which governs whether *prior* turns keep their reasoning. Both switches are no-ops here.

**The review's suggested fix — adding `supportsThinking: true` — is not applied, deliberately.** For a *local* model that flag does exactly two things: it lets the two no-op switches above be sent, and it surfaces the disable-thinking toggle in `InferenceConfigEditor.tsx:180-181`. Setting it would put a control in front of the user that provably cannot work. (`ReasoningService.ts:833` reads `getCloudModel`, so it never applies to local models.) Better to leave the flag off than to ship a toggle that lies.

**Consequence worth a decision before merge.** On the dictation-cleanup path `calculateMaxTokens` floors at 512 tokens for any text under 256 chars. If the forced reasoning block exhausts that budget, `finish_reason` is `length`, `message.content` is empty, and `llamaServer.js:589` falls back to `message.reasoning_content` — pasting the model's internal monologue into the user's document as their dictated text. The truncation guard three lines above only fires when `requireCompleteOutput` is set, which the cleanup path does not set. Candidate mitigations, neither in this PR: guard the `content || reasoning_content` fallback on `finish_reason`, or pass llama-server a reasoning flag for this model.

### Recorded, not acted on

- **Agent tools stay off.** `estimateModelSizeB >= 4` (`useChatStreaming.ts:277`) reads `2.6` from this id, so the agent runs without tools — the third of the three things #2032 asks for. Pre-existing floor, applies to every LFM entry including the recommended 1.2B. Notable that Liquid AI post-trained this model for agentic work and claim tool-use parity with models 4× larger, which is exactly the case the floor excludes.
- **Two near-identical picker rows.** "LFM2.5 2.6B" now renders directly above "LFM2 2.6B", differing by one character — the very confusion #2032 opens with. Renaming the older shipped entry is out of scope here.
- **DSpark drafter not wired.** Liquid AI publish a 328M drafter claiming ~2.6× faster decoding. The registry models drafters already, but `llamaServer` passes `--spec-type draft-mtp` which is MTP-specific and DSpark is not, and `gemmaMtpRegistry.test.js` asserts no non-QAT entry carries drafter fields. Follow-up.
- **`contextLength: 131072` kept.** It is what Liquid AI publish for this model and what its own `config.json` declares. The field is clamped to 16384 at runtime, so it is behaviour-neutral either way.

greatcoat:
Closing — benchmarking the model before merge showed it's the wrong fit for the use case that motivated #2032.

LFM2.5 2.6B is a pure reasoning model: its chat template opens `<think>` unconditionally, and there is no lever to stop it. `enable_thinking: false` is ignored (that's a Qwen-template variable; LFM2.5's template has no equivalent), and llama.cpp's `--reasoning off` has no effect for the same reason.

Measured against the dictation-cleanup prompt on Apple Silicon / Metal:

- **Output quality is good** when it's allowed to finish thinking.
- **Latency is 19–102s per dictation**, against ~2s for the models already in the list.
- At the current token budget it frequently never reaches an answer at all — thinking consumes the whole budget, `content` comes back empty, and `llamaServer.js` then falls back to `reasoning_content`, pasting the model's scratchpad into the user's document.
- Six runs of one medium dictation needed 1517 / 2398 / 3192 / 3214 / 4095 / 4096 tokens — two hit the ceiling, so no budget increase reliably fixes it.
- `--reasoning-budget 0` makes it fast but silently wrong: it returned uncleaned text and, in one case, a spurious `<|tool_call_start|>` block.

It's also below the 4B floor for local tool use, so it can't serve the "agent steps" half of the request either.

Recommendation given on #2032 instead: LFM2 2.6B for cleanup, Gemma 3 4B for translation / note formatting / chat — both already in the catalogue.

PR #2069, opened 2026-09-08
feat(analytics): backfill local Insights history

boseq:
## Summary

- Reconcile retained, completed local transcriptions into the existing content-free analytics event table.
- Page through only missing event IDs in bounded SQLite transactions; analytics events themselves are the checkpoint, so there is no metadata table, migration version, or durable cursor.
- Reuse each transcription client ID, assigning one only to old local rows that never had one, so later cloud sync deduplicates the same dictation.
- Preserve retention timestamps, local clear boundaries, and existing deletion tombstones.
- Re-run reconciliation before analytics reads and after a successful transcription retry so later-eligible history is picked up.

## Accuracy and safety policy

Historical text uses the current spoken-word counter, with processed text as the fallback when raw text is blank. Missing duration stays unknown, and ambiguous provider attribution stays unknown. Before any Insights clear, legacy completion timestamps are accepted as an approximation. After a clear, only exact client-captured occurrence timestamps proven to be newer than the boundary are eligible; ambiguous legacy rows stay skipped rather than reviving cleared counters.

## Why no migration

The local backfill writes into the analytics tables already introduced by the leaderboard stack. Progress is inferred from existing event IDs and the scan is safe to repeat, so extra schema and permanent completion state would add failure modes without adding correctness.

## Stack

This draft is based on the current `codex/analytics-leaderboard` head from https://github.com/OpenWhispr/openwhispr/pull/2011. Keep it unmerged until PR 2011 lands, then retarget this PR to `main`.

## Verification

- npm run quality-check
- npm run i18n:check
- npm run build:renderer
- npm test: 3,624 tests; 3,423 passed, 0 failed, 200 platform/native skips, 1 todo
- Focused analytics and participation suite: 39 passed, 0 failed
- Electron SQLite analytics suite with skips forbidden: 11 passed, 0 failed


boseq:
Superseded by #2075, which uses the same reviewed commits from an OpenWhispr-owned head branch while remaining a draft stacked on #2011.

PR #2070, opened 2026-09-08
feat(motion): retime the pill and panel transitions on one shared set of springs

greatcoat:
## Problem

Closing the assistant panel took about 1.15 s and looked broken: the content faded, then empty frames, then the shell collapsed, then the pill sat alone in the wrong colour for ~300 ms, then its mark morphed, then the window snapped with a 16 px jump. The pill itself appeared and vanished with a hard cut. Streaming replies arrived in blocks. The live transcript's unfinalised tail used a stepped `background-clip` shimmer. Several chained steps were driven by bare timers guessing at when a CSS transition would finish, and the native window resized while transitions were still visible.

## What this does, in plain terms

Every one of those transitions is retimed against one shared set of springs. A "spring" here is an animation that overshoots slightly and settles, like a real spring — more natural than a fixed curve. CSS cannot express one, so each spring's shape is computed once in JavaScript, sampled into a CSS `linear()` easing, and handed to CSS as a custom property. Chained steps now wait for the browser to say a transition genuinely ended, with a timeout only as a backstop, instead of guessing. And the operating-system window never changes size while something is visibly moving: it grows before, and shrinks only once the motion has landed.

Stacked on #1978 — its base is `feature/hotkey-activation-modes`, not `main`. Spec: `docs/superpowers/specs/2026-09-07-hold-and-hands-free-design.md` §3. Plan: `docs/superpowers/plans/2026-09-07-hold-and-hands-free-motion.md`. Both are on the base branch.

| Transition | Before | Now |
|---|---|---|
| Listening entrance | 300 ms width/height on a fixed curve | same stages, re-eased on the morph spring at 360 ms; the waveform slides in by transform |
| Stop → idle | window shrank on a bare 340 ms timer | shrinks on the pill's own `transitionend`; a non-pill shrink no longer waits at all |
| Panel open | rectangle revealed from a corner point | the shell springs open **from the pill's own 40 px circle**; the pill glides to the footer on the same spring |
| Streaming reply | text appended in blocks | each word rises in over 320 ms, 28 ms apart; settled text is never re-rendered |
| Copy tick | 150 ms colour switch back | tick pops, 6 s hold, then a true 320 ms crossfade — both labels overlap, no blank frame |
| Panel close | staged, ~1.15 s | **one 440 ms morph** to the pill's circle; measured 578.8 ms end to end |
| Companion pill on close | cut dead | fades 160 ms on the same clock |
| Hide | `hide()` hard cut | "zoop": 200 ms scale toward the pill's centre, opacity over the last 40 %, window hides after |
| Show | `showInactive()` hard cut | window shows transparent, pill unzoops over 260 ms (measured 3.7 % overshoot) |
| Live transcript tail | stepped `background-clip` shimmer | opacity only, 62 % → 100 % on commit; entrance became event-driven |
| Decision 8 | the pill turned into the dictation logo and *then* vanished | the agent's leaf is held while the window leaves; the morph runs after it is hidden |

Nothing was removed. Superseded rules are deliberately retained and labelled, per the plan.

## Reduced motion

`prefers-reduced-motion` collapses everything to 1 ms as before. That took real care here: the app's global rule forces `opacity` **into** `transition-property` with `!important` while stripping `transform` and `clip-path`, so any chain waiting on a transform or clip-path `transitionend` would wait forever. Every such chain on this branch short-circuits explicitly instead. Overrides were also kept from collapsing opacity fades that the global rule deliberately preserves.

## Tests

TDD throughout. **3,888 tests, 3,689 passing, 197 skipped, 1 environmental failure** that reproduces on the untouched base (`settingsStore imports without a bare localStorage global` — Node 25 defines `localStorage` globally; passes in CI). `typecheck`, `lint` and `i18n:check` clean.

Each of the 10 tasks was implemented, then independently reviewed, then fix-rounded until its reviewer signed off; a final whole-branch review followed, with its own fix wave and re-review. Reviewers repeatedly went past static analysis — building the app's real CSS bundle and measuring computed styles and frame-by-frame opacity in headless Chrome. That is how the panel close's 578.8 ms, the unzoop's 3.7 % overshoot, the copy tick's single-window crossfade, and the reduced-motion property lists were established rather than assumed.

Notable things the review process caught that the tests did not:
- the per-word rise animated only the **first** word of any bullet list, heading or blockquote — and since replies are mostly lists, the headline feature did nothing for most real content. It failed silently because every fixture was a plain paragraph.
- the streaming split cut inside lists and code fences, so a numbered item rendered as "1." then jumped to "3.".
- the settled prefix could go backwards, un-settling and re-animating text the user had already read.
- the companion window never received the fresh-mount gate, so its transcript entrance played from the wrong place.
- a held agent mark could be stranded indefinitely if a toast blocked the hide.

## Frame-level capture — run, and it passes

Captured on a dev build with `ow-ui-capture` (CDP screencast + an in-page rAF probe), hotkeys rebound to F13–F15. **All eleven rows pass, and every capture had zero frames over 34 ms** (worst in-transition frame 20.4 ms). Selected numbers:

- **Listening entrance** — 417 ms hold, then the spring peaks at +1.6 % and settles; the window grows *before* anything moves. Holds the budget easily (18.7 ms worst) despite animating layout.
- **Panel open** — the clip genuinely starts at the pill's 40 px circle, overshoots −2.79 %, settles ~450 ms.
- **Panel close** — one monotonic clip sweep, +173 → +623 ms, shell opacity 1 in every sample, window shrinks at +640 ms. The old staged sequence is gone.
- **Hide / show** — zoop 1.0 → 0.04 in 215 ms with opacity leaving over the last ~40 %; unzoop peaks 1.0364 and settles +272 ms.
- **Live transcript** — the four entrance beats land identically *to the millisecond* across two runs; the tail interpolates 0.62 → 1.00 over 233 ms, monotonic.
- **Decision 8** — per-frame pixel evidence: the agent violet is present in every frame through the morph, the shrink, the wait and the zoop. No ring flash.
- **Reduced motion** — passes for every row.
- **Compositions** — close → shrink → zoop → hide, and a recording start during a hide or a close, all pass. Every native resize sits between two transitions, never inside one.

Two compositions were **not** captured and are not claimed: *hide during a close* and *re-open during a close* are unreachable as user gestures (the panel owns Escape while mounted, main refuses a hide while the panel owns the window, and `openPanel()`'s two callers are both gated). The live transcript's +80 ms gate fallback also never fired in the run and cannot be forced from the UI.

The capture found one defect — the copy tick's label sat left in its now constant-width button — **fixed in this PR** (`0b737b8e`).

## Two things for you to call, neither a failure

1. **The zoop's "2 px squash" is invisible.** The end pose is `matrix(0.04, 0, 0, 0.03)` — 1.6 × 1.2 px on a 40 px pill, 0.4 px of differential — and the per-frame ratio never separates by more than ~0.9 px. It reads as a uniform collapse into a dot. It does not look wrong; it just isn't a squash. If it's meant to be felt, the differential has to be applied earlier in the collapse, at a larger scale.
2. **With auto-hide off, the close colour fade occupies only the first 66 ms of a 450 ms morph** (4 frames, 15 % in). There is no purple-pill stall and no post-close ring flash — the flicker the redesign targeted is gone — but "the colour fading inside it" is in practice "the colour settles as the morph starts".

Also worth knowing: **the listening entrance re-eases the existing width/height transition rather than using the spec's literal `clip-path` reveal on a fixed 92×40 box** — clip-path clipped the pill's border. Recorded as an approved interpretation in the plan; the capture confirms it holds the frame budget and reads like the filmstrip.

`src/App.jsx` has no test harness anywhere in this repo and `checkJs` is off, so its wiring is verified by review rather than tests. Decision logic was pushed into pure, unit-tested helpers wherever possible to shrink that surface.


greatcoat:
Closing: the motion layer is not being pursued further. The gesture model it was stacked on lives in #1978; this branch is kept for reference.

PR #2075, opened 2026-09-08
feat(analytics): backfill local Insights history

boseq:
## What this does

Insights is empty for every dictation a user made before analytics existed. This reconstructs the missing content-free counters from the transcription history already on disk, so the chart reflects their actual usage rather than the day they upgraded.

A versioned high-water checkpoint in local SQLite means a completed history is scanned once and later launches inspect only the new tail; an older row becoming eligible rewinds precisely to that row. Each bounded page and its cursor advance commit in a single SQLite transaction, so a failed page keeps its prior checkpoint and stays retryable.

Reconstructed rows reuse each transcription's client id, so the desktop, the cloud, and this device all deduplicate the same dictation. They upload as `counter_version: 0` and sort behind live events, so an API that cannot store them leaves history pending without ever blocking current analytics.

## Requires the API first

Ship after **OpenWhispr/openwhispr-api#198** is deployed. Until then the deployed API rejects `counter_version: 0` and the feature is inert — rejected rows stay pending and retry once the capable API is live, so the ordering is a latency boundary, not a data-loss one.

## Privacy and consent

Reconstruction reads the same transcripts the "keep local history" switch governs, so it answers to the same switch — and to a managed workspace's `localHistoryMode` policy:

- A pass will not start until the renderer has reported the switch, and stops at the next batch boundary if it is turned off mid-scan.
- It waits for a managed policy to resolve before treating that switch as consent. `policyStore` starts idle and resolves to the user's personal default while a policy is still being fetched — a default, not an answer — and a scan finishes in milliseconds against a network round trip. Signed out there is no policy to wait for, so the feature is unaffected for personal users.
- Deleted transcriptions never resurrect as counters, cleared history stays cleared, and reconstructed rows carry no transcript content.

## Leaderboard fairness

Reconstructed history is marked `reconciled` server-side and excluded from every leaderboard ranking, so retaining a long transcript archive cannot outrank someone who reinstalled last month. Legitimately backdated *live* events — a laptop closed for two months draining its queue — still count in full.

## Review fixes in this branch

The four commits after the feature each fix a defect found in review, test-first:

| | Fix |
|---|---|
| `d9b3e98e` | Cloud pulls inverted the history list. `transcriptions.timestamp` is sorted as TEXT and the API emits ISO 8601, so `"T"` (0x54) outranked the space local writes use — every cloud row sorted above every local dictation from the same UTC day, repeating the date header. The rule now lives in `dbTimestamp.toDbTimestamp`, shared by both writers. |
| `6f66f816` | The consent gate was read once and then the scan ran to completion. It is now re-read after each yield. |
| `e1b85b17` | Version-0 rows were re-POSTed at full ambient cadence against an API that cannot store them — roughly 288 doomed requests per user per day. History now backs off for an hour after a batch comes back without it, covering both an explicit rejection and an older response that simply omits the rows. Live events are unaffected. |
| `e76bf250` | Closed the policy-hydration race described above. |

## Verification

4260 tests pass, 0 fail. `quality-check`, `lint` and `i18n:check` clean. No new user-facing strings.

Beyond the suite, the reconstruction engine was probed directly against a real SQLite database: deleted transcriptions never resurrect; draining twice is idempotent; the cursor rewinds when a row becomes eligible later; an exactly-full page loses nothing at its boundary; a forced mid-page failure rolls the checkpoint back together with the inserts; and cleared history stays cleared even for mis-shaped legacy timestamps.

## Rebase note

Rebased onto `main`. The original base branch `codex/analytics-leaderboard` was squash-merged as #2011, so replaying the 18 development commits individually would have meant resolving that squash 18 times into intermediate states that were never tested. They are squashed into the first commit here; the four review fixes are preserved separately. The net change is identical — verified file by file — and is what the suite above covers.

### DIFF
--- a/src/components/InsightsView.tsx
+++ b/src/components/InsightsView.tsx
@@ -218,11 +218,13 @@ function MetricCard({
 }
 
 function YourUsage({
+  accountId,
   dataRetentionEnabled,
   isLoaded,
   onSyncErrorChange,
   syncActive,
 }: {
+  accountId: string | null;
   dataRetentionEnabled: boolean;
   isLoaded: boolean;
   onSyncErrorChange: (error: boolean) => void;
@@ -266,7 +268,7 @@ function YourUsage({
         if (!(await syncService.syncAnalyticsNow())) {
           throw new Error("Insights uploads are not enabled for this account");
         }
-        const account = await getAccountAnalyticsSummary();
+        const account = await getAccountAnalyticsSummary(accountId);
         if (requestId !== requestIdRef.current) return;
         setSummary(account);
       } else {
@@ -280,7 +282,7 @@ function YourUsage({
     } finally {
       if (requestId === requestIdRef.current) setLoading(false);
     }
-  }, [isLoaded, syncActive]);
+  }, [accountId, isLoaded, syncActive]);
 
   useEffect(() => subscribeToAnalyticsRefresh(load, syncActive), [load, syncActive]);
 
@@ -391,7 +393,7 @@ function YourUsage({
 
 export default function InsightsView({ onSignIn }: InsightsViewProps) {
   const { t } = useTranslation();
-  const { isLoaded, isSignedIn } = useAuth();
+  const { isLoaded, isSignedIn, user } = useAuth();
   const authValidated = hasValidatedAuthContext();
   const { dataRetentionEnabled: personalDataRetentionEnabled, insightsSyncEnabled } = useSettings();
   const dataRetentionEnabled = usePolicyStore((policyState) =>
@@ -485,6 +487,7 @@ export default function InsightsView({ onSignIn }: InsightsViewProps) {
 
         <TabsContent value="usage" className="mt-6 flex flex-1 flex-col">
           <YourUsage
+            accountId={user?.id ?? null}
             dataRetentionEnabled={dataRetentionEnabled}
             isLoaded={isLoaded}
             onSyncErrorChange={setSyncError}
--- a/src/helpers/analytics.js
+++ b/src/helpers/analytics.js
@@ -3,6 +3,13 @@
 // via Node's require(esm) with module-syntax detection.
 const DAY_MS = 86_400_000;
 export const ANALYTICS_ACTIVITY_MONTH_COUNT = 6;
+// Bump only when changed eligibility rules require old transcription rows to
+// be reconsidered. Each local database keeps its own scan cursor per version.
+export const ANALYTICS_HISTORY_BACKFILL_VERSION = 1;
+// Version 0 is deliberately below every exact counting rule. Historical rows
+// reconstructed from legacy desktop storage can therefore be upgraded by the
+// API's richer server history without overwriting a live event.
+export const ANALYTICS_HISTORICAL_COUNTER_VERSION = 0;
 export const ANALYTICS_COUNTER_VERSION = 2;
 
 const CJK_SCRIPT_PATTERN =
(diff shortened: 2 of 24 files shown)


Chadpiha (approved):
Approving the change itself. Two conditions before it can actually land.

## Blocking

1. **Ship only after openwhispr-api#198 is deployed.** Until then the feature is
   completely inert: the deployed API rejects `counter_version: 0`, so no
   reconstructed row reaches the cloud.

## Verified

- 3882 pass / 0 fail (6 unrelated Linux-sandbox skips), with the DB-backed tests
  genuinely executing rather than skipping. `quality-check`, `lint` and `i18n:check`
  clean. CI green.
- Local reconstruction holds up under direct probing against a real SQLite DB:
  deleted transcriptions never resurrect; draining twice is idempotent; the cursor
  rewinds precisely when a row becomes eligible later; an exactly-full page loses
  nothing at the boundary; a forced mid-page failure rolls the checkpoint back
  together with the inserts (no `await` inside any better-sqlite3 transaction);
  cleared history stays cleared including mis-shaped timestamps.
- One dictation is counted once across all three writers (desktop local, desktop
  upload, server reconciliation), keyed on `client_transcription_id`.
- Upload ordering is genuinely correct, not just in the tested shape: with a v0 row
  dated *newer* than two live rows, the queue still returns live-first.
- No new user-facing strings, so no i18n work owed.

## Fixed during review

1. **Cloud pulls inverted the history list.** `upsertTranscriptionFromCloud` bound
   the API's `created_at` verbatim into `transcriptions.timestamp`; the API emits
   ISO 8601 and local writes use a space separator, and that column is sorted as
   TEXT — so `'T'` (0x54) outranked `' '` (0x20) and every cloud row sorted above
   every local dictation from the same UTC day, with `HistoryView` repeating the
   date header. The rule now lives in `dbTimestamp.toDbTimestamp`, shared by both
   writers — and those two are the only writers of the column, with nothing
   updating it in place, so the invariant holds at every write site.
2. **The consent gate stopped being enforced mid-scan.** The backfill read the
   local-history switch once and then scanned to the end.
3. **Version-0 rows were re-POSTed at full ambient cadence forever** against an API
   that can't store them — ~288 doomed requests per user per day with no backoff.
   History now backs off for an hour after a batch comes back without it, covering
   both refusal shapes, while live events keep flowing.

## Follow-ups (not blocking)

- **No user-facing copy explains the server-side fill.** A 50k-dictation user watches
  account totals climb for hours with nothing saying why. The local half has
  `insights.claimInclude`; the server half has nothing.
- `getPendingAnalyticsEvents`' pending read has no supporting index, and this PR is
  what makes the pending set as large as the whole dictation history.
- On a queue of pure history the backoff can't self-clear early, since no batch is
  sent to carry a capable answer — recovery there waits out the full hour.
- Pre-existing, worth a separate fix: the retention sweep compares `created_at`
  lexicographically against `datetime('now', '-N days')`, so a cloud-pulled row
  purges up to a day late.

## Not verified

Neither side has had a manual pass — everything here is automated tests, PGlite and
SQLite probes. Before the release I'd want one real run: open the history list on an
account with cloud-pulled *and* locally-recorded dictations from the same day and
confirm the headers don't repeat, and watch one reconciliation chain drain against a
real Postgres.

gabrielste1n:
The red `pr-quality` check here is not this branch. Two tests in `test/helpers/updater.test.js` from #2120 never set a platform and fail on the Linux runner; every branch has been red since it merged. Fix is #2125. Re-run CI here once that lands.

Reviewed end to end alongside OpenWhispr/openwhispr-api#198 (which needed a rebase and one SQL fix, both pushed there). Nothing to change on this side. Merge order stays: API migrations `067`–`070` → API deploy → this PR.

PR #192, opened 2026-02-05
Add Mistral Voxtral Transcription Support

iTubeGamer:
# Add Mistral Voxtral Transcription Support

## Summary

This PR adds Mistral as a cloud transcription provider using Voxtral models alongside the existing OpenAI and Groq options.

Mistral's transcription endpoint (`/v1/audio/transcriptions`) requires an `x-api-key` header instead of `Authorization: Bearer`, which prevents it from working through the current Custom tab. This PR adds Mistral as a dedicated provider following the Groq integration pattern.

Custom dictionary support is included via Mistral's `context_bias` parameter (equivalent to OpenAI's `prompt` field), ensuring feature parity across all transcription providers.

Tested against `main` (v1.3.4).

## Changes

- Added Mistral to the transcription provider registry with `voxtral-mini-latest`.
- Added a Mistral provider tab and API key input in the transcription settings UI.
- Added Mistral API key persistence and IPC wiring.
- Added an IPC proxy for Mistral transcription to avoid browser CORS restrictions.
- Added provider-aware defaults so Mistral selects `voxtral-mini-latest`.
- Added custom dictionary support via Mistral's `context_bias` parameter (up to 100 tokens).
- Updated `.env.example` and README with `MISTRAL_API_KEY`.
- Added an npm `lint` script for convenience.

## UI

image

## Files Modified

| File | Change |
|------|--------|
| `src/models/modelRegistryData.json` | Add Mistral transcription provider |
| `src/helpers/audioManager.js` | Mistral model defaults, proxy handling, and dictionary forwarding |
| `src/components/TranscriptionModelPicker.tsx` | Add Mistral provider tab and key input |
| `src/components/SettingsPage.tsx` | Pass Mistral API key props |
| `src/components/OnboardingFlow.tsx` | Pass Mistral API key props |
| `src/hooks/useSettings.ts` | Add Mistral API key state management |
| `src/helpers/ipcHandlers.js` | Add Mistral key IPC handlers, proxy endpoint, and context_bias support |
| `src/helpers/environment.js` | Add Mistral key environment methods |
| `preload.js` | Expose Mistral key methods to renderer |
| `src/types/electron.ts` | Add Mistral key and proxy method TypeScript types |
| `.env.example` | Document `MISTRAL_API_KEY` |
| `README.md` | Document `MISTRAL_API_KEY` |
| `package.json` | Add `lint` script |
| `CHANGELOG.md` | Document Mistral provider addition |

## Testing

[...]

PR #2077, opened 2026-09-08
Support gpt-live-transcribe sessions in the OpenAI realtime client

gabrielste1n:
## Why

OpenAI deprecated `gpt-4o-mini-transcribe`, today's managed realtime dictation model, on 2026-08-26 with removal on 2027-02-26. `gpt-live-transcribe` is the live successor. This PR makes the desktop client **capable** of running it when the server mints such a session. **Nothing switches here**: no registry entry, no picker change, no default change, and the server-side env gate (openwhispr-api #193) stays unset.

## What

- **Model-aware turn detection.** `gpt-live-transcribe` rejects every `turn_detection` config ("Turn detection is not supported for this transcription model") and never emits `speech_started`/`speech_stopped`. For that model `session.update` sends `turn_detection: null`; legacy models keep the existing `server_vad` block byte-for-byte (pinned by test).
- **The server's model is authoritative for managed sessions.** In the preconfigured (cloud) branch the client adopts the model from the `session.created` echo. No change to `realtimeTokenProviders`, IPC or callers.
- **Stop commits the turn.** With no VAD, a turn only completes once the client sends `input_audio_buffer.commit`. `disconnect()` already committed; it now waits on the committed turn's terminal event (completed, including empty; failed; or socket death) instead of wrapping `onFinalTranscript`, so a silent dictation or a dead socket no longer stalls for the whole budget. No commit is sent when no audio was sent (an empty commit is a server error).
- **Live budget.** OpenAI acks the commit only after transcribing every appended byte, so the live model gets a 30 s commit budget; legacy keeps 3 s. Partials keep flowing to the renderer meanwhile and the renderer's partial-text fallback still applies on timeout.
- **Note recording is fenced off.** New test asserts no streaming registry entry starts with `gpt-live-transcribe` and that a stale selection falls back to the default in both BYOK and managed modes. Server-side, #193 excludes two-stream sessions from the override as well.

## Verified live (real API, shipped class, real 12.7 s recording)

| Run | Result |
|---|---|
| live, BYOK, 24 kHz | 52 partials (first at 2.2 s), full final, stop → resolved 9.0 s |
| live, BYOK, 16 kHz capture + 50 ms frames (exact dictation path) | 53 partials, full final, 9.6 s |
| live, preconfigured secret minted with #193's body | model adopted from echo, no `session.update` sent, full final, 3.3 s |
| legacy `gpt-4o-mini-transcribe` | VAD events, 13 partials, correct final, trailing commit → `commit_empty`, stop 299 ms |

Full suite 3767 (3570 pass, 0 fail), lint, typecheck, prettier clean.

## Do not flip the gate yet

At today's throughput `gpt-live-transcribe` transcribes at roughly 0.3–0.8x real time. The `committed` ack for a 12.7 s clip arrived 5.6 to 34 s after stop; a 3 s clip in 1.6 s; a 25 s clip in 45 s. `delay=minimal` was slower, and pauses or trailing silence did not help. With the previous 3 s budget the first run returned an empty final and would have pasted a partial missing two sentences. So dictation stop on the live model currently costs the renderer's 2 s settle ceiling plus several seconds of commit wait. Legacy is not immune today either (a VAD-open trailing turn took 5–6 s), but it is the better experience right now. Recommendation: merge to be ready, keep `OPENAI_REALTIME_TRANSCRIPTION_MODEL` unset until OpenAI's live throughput improves, and re-run the canary before flipping.

## Related

- openwhispr-api #193: token endpoint support and the env gate (dictation sessions only).
- openwhispr #2076 / openwhispr-api #192: the batch successor `gpt-transcribe`.

### DIFF
--- a/src/helpers/openaiRealtimeStreaming.js
+++ b/src/helpers/openaiRealtimeStreaming.js
@@ -3,6 +3,11 @@ const debugLogger = require("./debugLogger");
 
 const WEBSOCKET_TIMEOUT_MS = 15000;
 const DISCONNECT_TIMEOUT_MS = 3000;
+// gpt-live-transcribe acks a commit only once every appended byte is
+// transcribed, and it ran at 0.3–0.8x real time in testing, so its final trails
+// the stop by a large fraction of the dictation length (partials keep streaming
+// meanwhile).
+const LIVE_TRANSCRIBE_COMMIT_TIMEOUT_MS = 30000;
 const SAMPLE_RATE = 24000;
 const COLD_START_BUFFER_MAX = 3 * SAMPLE_RATE * 2; // 3 seconds of 16-bit PCM
 const KEEPALIVE_INTERVAL_MS = 15000;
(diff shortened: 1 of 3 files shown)


henmedia:
gpt-transcribe should also be added

PR #2080, opened 2026-09-08
test: pin the Oppulence API contract and repair the backend smoke test

yoanyombapro1234:
The desktop app parses live API responses with zod schemas that throw on a mismatch, so a server-side field rename would silently disable API key verification or note sync in a shipped build, with nothing to catch it.

This adds contract tests covering the four endpoints the app actually calls, using the app's own schemas:

- `/v1/voice/api-key-verifiers` parsed with `ApiKeyVerifierSnapshotSchema`, plus the 401 path
- `/v1/voice-sync/items` round-trip, asserting ciphertext comes back untouched
- `/v1/capture-artifacts` replay, asserting the outbox's retries stay idempotent

All four pass against a live API. They skip unless `OPPULENCE_API_URL` and `OPPULENCE_API_TOKEN_URL` are set, so the default suite stays hermetic (3471 pass, 0 fail).

Also fixes `distributions/oppulence-voice/scripts/backend-smoke.ts`, which could never run: its top-level await was invalid under the CommonJS transform. It now passes. The contract tests sidestep the same trap by checking reachability inside each test rather than at module scope.

### Known gap (not addressed here)

`extensions/oppulence-cloud/publicApi.ts` references `/v1/notes/*`, `/v1/folders/*`, `/v1/spaces/list`, `/v1/transcriptions/*` and `/v1/usage`, and the API returns **404** for all of them. Those paths need either an API implementation or removal from the client.


yoanyombapro1234:
Opened against the wrong repository by mistake; this belongs on the PlaybookMediaLLC fork.

PR #2083, opened 2026-09-09
feat(meeting): allow self-hosted transcription for Note Recording

rodaddy:
## Summary

Enables the Self-hosted mode for Note Recording. Today the routing
helper throws "not supported for Note Recording" and the settings card is
disabled with a Coming soon badge. A self-hosted OpenAI-compatible server
is a batch HTTP endpoint, not a realtime socket, so this reuses the
chunked pipeline that Local Note Recording already runs (buffer PCM,
downsample, WAV, 5 s cadence, the same RMS gate and mic holdback) and
POSTs each chunk to `<url>/audio/transcriptions` with the configured
model and language.

Opt-in only: nothing routes here unless the user selects Self-hosted for
Note Recording.

## How it works

- `meetingTranscriptionRouting.resolveMeetingTranscriptionOptions`
  resolves `{ provider: "self-hosted", endpoint, model, language }`. The
  URL is normalized the same way the dictation route does (a pasted
  `/v1/audio/transcriptions` or `/audio/transcriptions` suffix is
  stripped). It still throws, now only when the URL is empty, so the
  fail-closed behavior from #1280 holds.
- `meetingRecordingStore` passes the meeting-scoped URL (falling back to
  the dictation URL) and the global self-hosted model.
- `MeetingSettings` enables the mode and renders the existing
  `SelfHostedPanel` for URL and model.
- Main process: `prepare`/`start` accept `self-hosted`; the buffered-chunk
  path calls a small pure `selfHostedMeetingTranscription.transcribeSelfHostedChunk`
  (multipart POST, injected fetch) instead of decoding locally. A debug
  log line records per-chunk request time and result.
- `meetingStreamingProviders` adds `self-hosted` to the allowed set; the
  closure tests treat it like `local` (no socket, no token).

No locale changes: the mode label and the panel strings already exist.

## Relationship to #1337 and #1280

#1337 solves the same gap by routing meeting chunks through the batch
transcription router. This change takes a narrower path: it plugs the
self-hosted endpoint into the existing local meeting chunk loop and
touches no routing shared with Dictation or Upload. Ten files, one new
helper with its own tests. Either approach retires the Coming soon
badge; this one is offered as the smaller diff.

Refs #1280 (the custom Cloud Providers endpoint for Note Recording is a
separate problem and is not touched here).

## Verification

macOS 27.0 arm64, Node 24, branch on `main` at a2d0ab27. Endpoint:
llama-swap serving `parakeet-tdt-0.6b-v2` on a LAN GPU box.

```
npm run quality-check   # format:check + typecheck: pass
npm run lint            # 0 errors, 0 warnings
npm run i18n:check      # [i18n] Locale keys and placeholders are consistent.
npm test                # 3782 tests, 3584 pass, 1 fail
```

The one failure is `activateTargetPid resolves false for an unmapped PID`
in `test/helpers/textEditMonitorSelection.test.js`. It fails identically
on an untouched checkout of `main` (a2d0ab27) on this machine (times out
at ~3 s), so it is unrelated to this change.

Live run (packaged build of this branch):

Two-minute note recording (03:17:37 to 03:19:45 UTC) with a podcast playing through the speakers, Self-hosted selected for Note Recording, debug logging on. Every chunk the pipeline sent got a 200 with text back:

```
source  chunks  full 5 s chunks  returned text  request ms (min / avg / max)
system      26               23             26          71 / 116 / 250
mic         17                6             17          65 / 119 / 402
```

Consecutive full system-audio chunks arrived 4.92 to 5.08 s apart (the pipeline's 5 s buffer), full chunks exactly 160044 bytes (5.000 s of 16 kHz mono WAV), a shorter flush chunk at stop. Mic chunks are fewer because the existing RMS gate skips silence and the holdback logic suppresses bleed from the speakers; both behave the same as on the Local path.

First and last entries from the debug log (`Self-hosted meeting chunk transcribed` is the new debug line in this PR):

```
2026-09-09T03:17:42.881Z system 153644 146 true 55
2026-09-09T03:17:47.868Z system 160044 134 true 42
2026-09-09T03:17:52.812Z system 160044 76 true 51
...
2026-09-09T03:19:43.020Z mic 160044 67 true 55
2026-09-09T03:19:45.601Z system 78752 250 true 17
```
(columns: time, source, wavBytes, requestMs, success, textLength)

## Screenshots

The note's Transcript view after the two-minute run above (system audio
chunks transcribed by the self-hosted endpoint; the speaker label is a
user-assigned name from the existing diarization flow):

Transcript from the self-hosted note recording run

Settings, Speech-to-Text, Note Recording with Self-Hosted selected (the
existing `SelfHostedPanel` for URL and model; server address replaced
with a placeholder):

Note Recording settings with Self-Hosted selected

### DIFF
--- a/src/components/settings/MeetingSettings.tsx
+++ b/src/components/settings/MeetingSettings.tsx
@@ -8,6 +8,7 @@ import { InferenceModeSelector, SettingsRow } from "../ui/SettingsSection";
 import type { InferenceModeOption } from "../ui/SettingsSection";
 import { Toggle } from "../ui/toggle";
 import TranscriptionModelPicker from "../TranscriptionModelPicker";
+import SelfHostedPanel from "../SelfHostedPanel";
 import type { InferenceMode } from "../../types/electron";
 import { useStartOnboarding } from "../../hooks/useStartOnboarding";
 import { getStreamingTranscriptionProviders } from "../../models/ModelRegistry";
@@ -58,6 +59,11 @@ export function MeetingTranscriptionPanel() {
     meetingCloudTranscriptionBaseUrl,
     setMeetingCloudTranscriptionBaseUrl,
     setMeetingCloudTranscriptionMode,
+    meetingRemoteTranscriptionUrl,
+    setMeetingRemoteTranscriptionUrl,
+    remoteTranscriptionUrl,
+    remoteTranscriptionModel,
+    setRemoteTranscriptionModel,
   } = useSettingsStore();
   const {
     modes: transcriptionModes,
@@ -90,8 +96,6 @@ export function MeetingTranscriptionPanel() {
         label: t("settingsPage.transcription.modes.selfHosted"),
         description: t("settingsPage.transcription.modes.selfHostedDesc"),
         icon: <Network className="w-4 h-4" />,
-        disabled: true,
-        badge: t("common.comingSoon"),
       },
     ],
     "transcription",
@@ -100,7 +104,6 @@ export function MeetingTranscriptionPanel() {
   );
   const handleTranscriptionModeSelect = (mode: InferenceMode) => {
     if (!isModeAllowed(mode)) return;
-    if (mode === "self-hosted") return;
     if (mode === "openwhispr" && !isSignedIn) {
       startOnboarding();
       return;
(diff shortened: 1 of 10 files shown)


pranavajk1:
Tested this on macOS against whisper.cpp behind a LiteLLM gateway (OpenAI `/v1/audio/transcriptions` wire format, Bearer auth). The pipeline works well — 5 s chunks, ~350–530 ms round trip per chunk, transcript kept up, speaker labels unaffected.

One blocker for authenticated endpoints: `transcribeSelfHostedChunk` never sends an `Authorization` header, so every chunk gets a 401 and the transcript stays empty. I opened a small follow-up against this branch that threads the existing Dictation self-hosted key (`CUSTOM_TRANSCRIPTION_API_KEY`) through as `Bearer`, with tests: rodaddy/openwhispr#1

PR #2087, opened 2026-09-09
feat(settings): resolve API keys from env vars and login shell

akbarkanso:
## Why

Custom (and other BYOK) providers require an API key next to the endpoint URL. Today that key has to be pasted into Settings. People who already export keys in `~/.zshrc` / `~/.bashrc` should be able to reference them instead of duplicating the secret in the app.

## What

API key fields accept a whole-field `$VAR` / `${VAR}` instead of the secret.

- Settings and IPC keep the raw reference. It is expanded only on the request path.
- Only OpenWhispr secret names are expanded (`OPENAI_API_KEY`, `GROQ_API_KEY`, `CUSTOM_TRANSCRIPTION_API_KEY`, …). `$PATH` and arbitrary names like `$KEY_TEST` resolve to empty, so they are never sent as `Authorization: Bearer $VAR`.
- `$OPENAI_API_KEY` in the OpenAI field itself is a self-reference and is not stored as a working key. Use it in a *different* field (custom endpoint), or leave the OpenAI field empty.
- On Unix, empty known secrets are filled from a trusted login shell (`/bin/zsh`, `/usr/bin/zsh`, `/bin/bash`, `/usr/bin/bash`) so dock launches see `~/.zshrc` / `~/.bashrc` exports. Untrusted `SHELL` paths (e.g. `/tmp/zsh`) are not executed. Imports are never written to Settings or `.env`.
- Windows is unchanged (no shell import).

Custom provider is the original case: set the endpoint URL, put `$OPENAI_API_KEY` in the custom key field (or leave it empty and `export CUSTOM_TRANSCRIPTION_API_KEY` in zsh).

## Verification

- `node --test test/helpers/envKeyResolve.test.js test/helpers/enterpriseRuntimeResolve.test.js test/helpers/secretKeys.test.js` — 37 pass
- i18n keys added in all 10 locale files

## Still owed

Manual smoke: custom endpoint with `$OPENAI_API_KEY`, empty field + zsh export, confirm the import does not appear as a saved Settings secret.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,7 +7,18 @@ VITE_OPENWHISPR_API_URL=
 
 # OpenAI API Configuration
 # Get your API key from: https://platform.openai.com/api-keys
+# Keys exported in your shell (e.g. ~/.zshrc) are imported at startup when
+# the matching Settings field is empty. A key saved in Settings still wins.
+# Custom endpoints can also store a reference instead of the secret itself:
+#   CUSTOM_TRANSCRIPTION_API_KEY=$OPENAI_API_KEY
+# dotenv does not expand $VAR here. Only a whole-field $VAR / ${VAR} is
+# resolved at read time, and only if the name is an OpenWhispr secret
+# (OPENAI_API_KEY, GROQ_API_KEY, … — not $MY_COMPANY_KEY).
 OPENAI_API_KEY=your_openai_api_key_here
+
+# Optional: API key for a custom OpenAI-compatible transcription endpoint.
+# Leave unset to require no auth, or set / reference another variable.
+# CUSTOM_TRANSCRIPTION_API_KEY=$OPENAI_API_KEY
 
 # Optional: Customize the Whisper model
 # Available models: whisper-1 (default), whisper-1-large, whisper-1-large-v2
--- a/preload.js
+++ b/preload.js
@@ -388,6 +388,7 @@ contextBridge.exposeInMainWorld("electronAPI", {
 
   // BYOK API keys (get/save for every provider in the secretKeys manifest)
   ...secretKeyApi,
+  resolveSecretRef: (value) => ipcRenderer.invoke("resolve-secret-ref", value),
 
   // Clipboard functions
   checkAccessibilityPermission: (silent) =>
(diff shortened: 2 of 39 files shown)


akbarkanso:
## `get-*Key` consumer audit (raw IPC contract)

Settings/IPC getters return **raw** `$VAR` on purpose. Credential-use sites must resolve (or main-process `resolveSecretRef` / `usableSecret`).

### Converted (resolve on use)
- `audioManager.js` — transcription request path
- `ReasoningService.ts` — `getApiKey` + Authorization headers
- `OpenAICompatiblePanel.tsx` — model list fetch
- `fileTranscription.ts` — BYOK file upload (`resolveApiKey` + main `resolveSecretRef`)
- `inferenceProviders/lan.ts`, `inferenceProviders/openai.ts` — custom/LAN keys
- `ipcHandlers.js` `test-provider-connection` — resolves `apiKey` / Corti client secrets
- `providerConnectionTest.js` — `usableSecret` drops leftover `$VAR` before Bearer
- Enterprise runtime — `resolveManualEnterpriseRuntime`

### Intentionally raw
- `settingsStore.ts` `initializeSettings` — hydrates the Settings UI
- `getDictationKey` / `getVoiceAgentKey` / `getTranslationKey` — hotkeys, not API secrets

### Belt-and-braces
Any leftover `$VAR` is treated as missing (`usableSecret` / `resolveSecretRef`) instead of sent as `Authorization: Bearer $OPENAI_API_KEY`.

akbarkanso:
## Follow-up audit (`3efd0035`)

### CJS/ESM
- Renderer: ESM `src/helpers/envRef.ts`
- Main: CJS `src/helpers/envRef.cjs`
- `settingsStore` no longer imports CJS `secretKeys.js`

### `get-*Key` consumers
**Credential-use (resolved):** audioManager, ReasoningService, OpenAICompatiblePanel, fileTranscription + BYOK handler, lan/openai inference, enterprise runtime, test-provider-connection (including Bedrock/Azure/Vertex fields), AssemblyAI/Deepgram realtime (`getAssemblyAIKey`/`getDeepgramKey` via `_getKey` + `usableSecret`).

**Intentionally raw:** `initializeSettings` (Settings UI hydration); dictation/voice-agent/translation **hotkeys**.

**Out of scope:** AssemblyAI/Deepgram have no Settings `$VAR` field; they resolve in main if stored as a ref.

### `transcribe-audio-file-byok`
Confirmed: `apiKey: rawApiKey` in the destructure, `const apiKey` inside the `try`, all header uses in that block. Wrapped with `usableSecret`.

PR #2088, opened 2026-09-09
fix: accept plus-addressed emails at sign-in

### LINKED ISSUE #1700: Desktop app rejects "+" email aliases accepted during website registration
**Describe the bug**
Login fails when using an email address that contains a "+" alias tag (plus-addressing, e.g. `user+alias@domain.com`). The website registration flow accepts and successfully creates an account with this email format, but the desktop app rejects login attempts using the exact same email address.

**To Reproduce**

1. Go to the OpenWhisper website to download the app
2. Sign up using an email address with a plus alias (e.g. `user+alias@domain.com`)
3. Complete registration successfully
4. Download and install the app
5. Open the OpenWhisper desktop app
6. Attempt to log in using the same email address (`user+alias@domain.com`) and password
7. See error "Email aliases with '+' are not supported"

**Expected behavior**
Either the website should not allow registration with an email containing a "+" alias, or the desktop app's login should accept it. The two flows should be consistent, right now one accepts it and the other doesn't.

**Screenshots**
Image

- Browser: Google Chrome 151.0.7922.138 (latest)
- Device: MacBook Pro
- OS: macOS 26.6.1

boseq:
## Summary

The desktop email gate rejected any address whose local part contains `+` before it ever reached the server:

```js
const localPart = email.trim().split("@")[0];
if (localPart?.includes("+")) {
  setError(t("auth.errors.plusAliasUnsupported"));
  return;
}
```

The website and API accept plus-addressed emails, so those accounts exist and simply could not sign in from the app — the report in #1700.

Two things made it misfire:

1. The gate sits in `handleEmailContinue`, the shared email entry point for **both sign-in and sign-up**. It was meant to block signup; it blocked login too.
2. `AuthenticationStep` is the only email-auth component in the app — onboarding, `SignInDialog`, `CompactAuthenticationFlow` and `ReauthenticationScreen` all route through it. So a plus-addressed user was locked out of every in-app email auth surface, re-authentication included.

**It was a total in-app email-auth lockout, not just a sign-in failure.** "Forgot password?" only renders under `authMode === "sign-in"`, and `authMode` is only ever set by `handleEmailContinue` — so a plus-addressed user could not even request a password reset. And because `ReauthenticationScreen` routes through the same component, an existing plus-addressed *paying* user whose session expired had no way back into the app at all.

The check was added in 5f133001 to curb signup abuse via plus-addressing. That vector is real (free tier is 2000 cloud words, and a referral pays out once the referred account crosses the same threshold, with self-referral guarded only by `userId !== referrerId`), but a desktop-client check was the wrong place for it. The API does carry a plus-address rule — `signupSchema` in `lib/schemas.ts`, with the same error message as 5f133001 — but it has **zero consumers**: `api/auth/signup.ts` is a shim that forwards the raw body to Better Auth's `/sign-up/email`, whose validator is a stock `z.email()` with no plus restriction (zod's email validator accepts `+`), and no `before` hook filters it either. Nothing enforces it, which is exactly why website registration succeeds.

To be precise about what removing this costs: the gate *did* block plus-addressed sign-in, so it was the only thing standing between a website-registered alias and the desktop client, and taking it out does widen desktop-side exposure to that vector. But a client-side control is bypassable by construction — registering on the web already sidesteps it — so it was buying the appearance of enforcement at the price of locking out legitimate accounts that already exist. Enforcement of this kind belongs in the server signup path, and it must not retroactively lock out addresses already in use. Filed as OpenWhispr/openwhispr-api#195, which also covers deleting the orphaned refine so nobody wires it up later and reintroduces this bug server-side.

**A narrower fix was considered and rejected:** moving the plus check into `handleSubmit` under `authMode === "sign-up"` would unblock login while keeping the signup deterrent. It buys nothing. The website funnel would still accept plus signups, so the two flows stay inconsistent — the exact complaint in #1700 — and a client-side signup check is sidestepped by just registering on the web.

The check is now the shared `EMAIL_REGEX` from `src/utils/validation.ts`, the same loose shape check `ReferralDashboard` and `ShareNoteDialog` already use: plus signs, dots and hyphens pass, obviously malformed input is stopped client-side, and the server stays the source of truth.

### One behavior change worth calling out

`EMAIL_REGEX` requires a dotted domain, so **single-label addresses — `user@corp`, `admin@localhost` — are now rejected client-side** where the old plus-only gate let them through to `/api/check-user` (whose `checkUserSchema` is `z.string().min(1)`, no format check). Against the hosted server this is consistent, since Better Auth's own `z.email()` rejects them too, so the round trip was only ever going to fail — but `src/lib/emailAuthDiscovery.ts` explicitly accommodates self-hosted auth servers, and one with a different backend could legitimately accept an intranet address. The test pins this as deliberate; say the word if it should be relaxed to `/^[^\s@]+@[^\s@]+$/` for this call site instead.

For context on how much the regex is actually doing: the only call site is a `<form>` submit whose field is `<Input type="email" … required>` with no `noValidate`, so Chromium's own constraint validation already blocks `user`, `user@`, `@example.com`, whitespace and non-ASCII IDN before `handleEmailContinue` runs. The single-label case above is the one genuinely new rejection, and therefore the only input that can ever surface the new `invalidEmail` string.

### The rest of the plus path

I checked for a second failure waiting behind the gate — there isn't one:

- `/api/check-user` is POST + JSON body matching `LOWER(email) = $1` exactly, so it finds plus-addressed accounts
- **nothing in the desktop app, the API, or the website strips a plus tag.** Better Auth's only normalization is `.toLowerCase()`, so `user+tag@x.com` is stored and looked up intact as a distinct account — the silent account-mismatch failure this could have been does not exist
- every client site that puts an email in a URL percent-encodes it, so `+` travels as `%2B` and never decodes to a space. There are **three** query-string sites, not two: `src/lib/auth.ts` (SSO, `URLSearchParams`), `src/components/EmailVerificationStep.tsx` (`encodeURIComponent`), and `src/services/NoteSharingService.ts` (share-dialog principal search, `encodeURIComponent`). The `mention:` hrefs and the hashed gravatar URL are email-derived and also safe
- the SSO login-expectation fingerprint hashes the address (SHA-256) rather than carrying it in the URL, so it is unaffected

Verified against the deployed API (`origin/main`), not just a local checkout — a live `POST /api/check-user` returns `200 {"exists":false,"sso":{"available":false,"required":false}}` for `probe+alias@example.com` and `probe+tag+more@example.co.uk`, byte-identical to a plain control address.

Supersedes #1701, which had the right idea but was withdrawn by its author; this drops the unrelated Prettier reformatting that PR carried.

## Changes

- `src/components/AuthenticationStep.tsx` — replace the plus-address rejection in `handleEmailContinue` with the shared `EMAIL_REGEX` check and a generic invalid-email error
- `src/locales/*/translation.json` — swap `auth.errors.plusAliasUnsupported` for `auth.errors.invalidEmail` across all 10 locales. The phrasing is reused from each locale's existing `noteEditor.share.dialog.error.invalidEmail`, except in de, ru and zh-CN, where that string carries the share dialog's conventions rather than the auth namespace's: German needed the Sie register (every other `auth.*` string uses it, and `auth.sso.emailRequired` on the same screen is literally the Sie form of this sentence), zh-CN needed 邮箱 over 电子邮件 to match all thirteen sibling strings, and ru needed "электронной почты" spelled out as `invalidCredentials` does one line above
- `test/components/authenticationStep.test.js` — behavioural regression pin for #1700 through the existing harness: a plus-addressed alias reaches `discoverEmailAuth` unchanged and shows no error, and a malformed address never reaches discovery
- `test/utils/validation.test.js` — accept/reject coverage for the shared regex (plus variants, the other shapes users type, malformed input, and the `.trim()` seam callers rely on)
- Merged `main` — the one conflict was an import-ordering collision with the new `useDebouncedCallback` import; both are kept

No IPC, API, or platform-specific surface touched; no layout changes to the sign-in UI.

## Testing

- `npm test` — **4046 tests, 3849 passed, 0 failed** (196 skipped are the known better-sqlite3 ABI skips)
- `npm run quality-check` — exit 0 (`format:check` + `typecheck`); the 4 lint warnings are pre-existing, from `test/helpers/settingsStoreMeetingFollowFlags.test.js` on `main`
- `npm run i18n:check` — "Locale keys and placeholders are consistent"
- `prettier --check` on all changed files — clean

The regression pin was mutation-checked against three reintroductions of the bug, each failing a named assertion:

| Mutation | Result |
|---|---|
| the literal `localPart.includes("+")` gate restored | ✖ *a plus-addressed alias must reach discovery unchanged* |
| the bug reintroduced indirectly, via a module-scope `hasUnsupportedAlias` helper | ✖ *a plus-addressed alias must reach discovery unchanged* |
| the shape check dropped entirely | ✖ *a malformed address must never reach discovery* |

An earlier revision of this PR pinned the fix by grepping the `handleEmailContinue` source text instead. That was unsound in both directions and has been replaced: the helper-indirection mutation above left all five of those tests green with the bug fully back, and `assert.doesNotMatch(body, /\+/)` failed on ordinary string concatenation — the idiom the same file already uses 40 lines above the gate — under a message claiming plus special-casing. The extraction's `"\n  const "` end anchor also silently widened the captured body from 1196 to 1445 characters when the adjacent `errorMessageIncludes` was written as a `function` declaration, with nothing asserting against over-capture.

**Not manually exercised against a live account.** Two journeys need a real plus-addressed account created on the website:

1. **Accept a workspace invitation while signed out** — `ControlPanel` → `AcceptInvitationModal` → `SignInDialog`, the only mount of that dialog, and the surface most likely to carry a `+` address
2. **Re-authentication after session expiry** — `ReauthenticationScreen`, the hard-lockout case

Plus one negative check: confirm `user@corp` now shows "Enter a valid email address", and decide whether that is intended.

Fixes #1700


Chadpiha (approved):
Approving — right diagnosis, minimal fix, in the correct place. Tracing the dead signupSchema to its zero consumers and probing the live /api/check-user rather than just reading the handler is what made this quick to review.

The production change is untouched. Two notes:

The bug was worse than the title. "Forgot password?" only renders under authMode === "sign-in", which only handleEmailContinue sets, and ReauthenticationScreen routes through the same component — so plus-addressed users couldn't reset their password either, and an existing paying one whose session expired was locked out entirely.

I replaced the regression pin and pushed it. Pinning the reintroduction was the right instinct, but the grep didn't hold: a module-scope hasUnsupportedAlias helper reintroduces #1700 exactly, keeping EMAIL_REGEX.test( inline, and all five tests stayed green. Separately, doesNotMatch(body, /\+/) fails on ordinary concatenation — the idiom the file already uses ~40 lines up. The component harness already mocks discoverEmailAuth, so the behavioural version was ~18 lines: the alias reaches discovery unchanged, a malformed address doesn't. Mutation-checked three ways. Also merged main and fixed de/ru/zh-CN, where the verbatim reuse imported the share dialog's register into auth (the German was the only du string in the subtree).

Issue #1909, opened 2026-08-28
openwhispr fails to install via winget win11


pleabargain:
**Describe the bug**
winget fails to upgrade OpenWhispr to version 1.9.0 with a 404 "Not found" error when attempting to download from the GitHub release URL. The download request returns HTTP status code 0x80190194 (404).

**To Reproduce**
1. Open PowerShell 7.6.5 (or later)
2. Run `winget upgrade --all`
3. Allow the upgrade of OpenWhispr from 1.8.3 to 1.9.0
4. See error: "Download request status is not success. 0x80190194 : Not found (404)."

**Expected behavior**
OpenWhispr 1.9.0 should download and install successfully from the GitHub release URL without HTTP errors.

**Screenshots**

<img width="822" height="130" alt="Image" src="https://github.com/user-attachments/assets/e0d73e4f-1c2f-44bb-9d8a-05318876bcf8" />

```
(2/2) Found OpenWhispr [OpenWhispr.OpenWhispr] Version 1.9.0
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
Downloading https://github.com/OpenWhispr/openwhispr/releases/latest/download/OpenWhispr-Setup-1.9.0.exe
An unexpected error occurred while executing the command:
Download request status is not success.
0x80190194 : Not found (404).
```

**Desktop (please complete the following information):**
- OS: Windows
- Package Manager: winget (Windows Package Manager)
- PowerShell Version: 7.6.5

**Additional context**
- The Go upgrade in the same batch completed successfully
- The issue appears to be with the GitHub release URL for OpenWhispr 1.9.0 no longer being accessible or incorrectly configured in the winget package manifest
- Suggest verifying the release URL exists and is publicly accessible at: `https://github.com/OpenWhispr/openwhispr/releases/latest/download/OpenWhispr-Setup-1.9.0.exe`

IdrisGit:
Linking related reports: both describe Winget 404 failures caused by versioned files using the `/releases/latest/` URL. #1355 covers the older manifest, while #1909 reproduces it on version 1.9.0. Please review them together.

<sub>Note: Drafted and posted by automation. Please flag any errors.</sub>

greatcoat:
[...]

Issue #1355, opened 2026-07-27
Invalid winget download link breaks installing via winget  (+ releases not published to winget)


iki:
**Describe the bug**

Winget download link references latest release resulting in 404 for older versions, when there's a new release on GitHub.

**To Reproduce**

1. Run `winget install -e --no-upgrade --id OpenWhispr.OpenWhispr`

    - Error:
        ```
        Found OpenWhispr [OpenWhispr.OpenWhispr] Version 1.7.5
        This application is licensed to you by its owner.
        Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
        Downloading https://github.com/OpenWhispr/openwhispr/releases/latest/download/OpenWhispr-Setup-1.7.5.exe
        An unexpected error occurred while executing the command:
        Download request status is not success.
        0x80190194 : Not found (404).
        ```

**Expected behavior**

✅  The link should be `https://github.com/OpenWhispr/openwhispr/releases/download/v$VER/OpenWhispr-Setup-$VER.exe` so it works even when $VER is not latest release.

❌ The current link `https://github.com/OpenWhispr/openwhispr/releases/latest/download/OpenWhispr-Setup-$VER.exe` breaks when $VER is not latest release. Like now with [v1.7.6](https://github.com/OpenWhispr/openwhispr/releases/tag/v1.7.6) on GitHub and [v1.7.5](https://github.com/OpenWhispr/openwhispr/releases/tag/v1.7.5) on winget.

**Desktop (please complete the following information):**

- OS: Windows

**Additional steps**

Also, it would be great to use [WinGet Releaser](https://github.com/vedantmgoyal9/winget-releaser) to submit PR to [microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs) when new [release](https://github.com/OpenWhispr/openwhispr/releases) is published.

Required steps to setup before first run on next release:

1. [Create a Classic Personal Access Token (PAT)](https://github.com/settings/tokens) with `public_repo` scope, copy the token and use it to [create a repository action secret](https://github.com/OpenWhispr/openwhispr/settings/secrets/actions) `WINGET_TOKEN`.

2. [Fork microsoft/winget-pkgs](https://github.com/microsoft/winget-pkgs/fork) to your own account or organization. The bot needs this fork to stage the changes before creating a PR.

IdrisGit:
[...]

PR #2089, opened 2026-09-09
fix(ci): publish winget manifests from CI with tag-pinned installer URLs

boseq:
Fixes the winget 404 reported in #1909 (and #1355).

## The bug

Every winget manifest since 1.7.5 uses GitHub's `latest` alias with a version-pinned filename:

```
https://github.com/OpenWhispr/openwhispr/releases/latest/download/OpenWhispr-Setup-1.9.0.exe
```

`/releases/latest/download/` resolves to whatever the newest release is, so that path 404s the moment the next release ships. Verified:

| URL | Result |
|---|---|
| `latest/download/OpenWhispr-Setup-1.9.2.exe` (1.9.2 *is* latest) | 206 |
| `latest/download/OpenWhispr-Setup-1.9.0.exe` | **404** |
| `download/v1.9.0/OpenWhispr-Setup-1.9.0.exe` (pinned) | 206 |

The project has never owned this package. The oldest manifest (1.7.3, submitted by a different contributor) uses the correct pinned URL and still installs today; everything from 1.7.5 on was submitted by a third-party bot whose tooling emits the `latest` form.

## Why it bites users

The bot lags each release by 18 hours to 2.7 days, and during that lag the newest manifest winget offers is already broken:

| Release | Published | winget PR opened | Lag |
|---|---|---|---|
| 1.9.0 | 08-24 20:29Z | 08-27 12:11Z | 2.7 days |
| 1.9.1 | 08-27 22:31Z | 08-28 16:41Z | 18 hours |
| 1.9.2 | 08-29 18:23Z | 09-01 09:16Z | 2.6 days |

#1909 was filed 08-28 06:22Z — winget was still offering 1.9.0 while GitHub had moved to 1.9.1 eight hours earlier. Exactly the reported symptom.

## The fix

Submit the manifests ourselves on `release: published`. `winget-releaser` feeds komac the release assets' `browser_download_url`, which is always the tag-pinned form, so the URL structurally cannot go stale. Firing on the release event also collapses the submission lag to minutes.

Notes on the implementation:

- The `Get release version` step mirrors `update-nix.yml` so a failed submission can be retried via `workflow_dispatch`. Its `^\d+\.\d+\.\d+$` guard also rejects prerelease tags, so no separate condition is needed.
- `release-tag` is passed explicitly rather than `version`: the action's `release-tag` default is `github.ref_name` on a manual dispatch, which is `main`, not a tag.
- `fork-user` is omitted — it defaults to `github.repository_owner`, which is what we want.
- `installers-regex` is validated against the real v1.9.2 asset list: it matches only `OpenWhispr-Setup-1.9.2.exe`, excluding the portable `OpenWhispr-1.9.2.exe` and the `.blockmap`.
- `GH_REPO` stands in for a checkout, which this job does not otherwise need. Permissions are `contents: read` (`update-nix.yml` needs write only because it pushes a branch here).

## Prerequisites before this can run

1. Classic PAT with `public_repo` scope, stored as the `WINGET_TOKEN` repo secret.
2. A fork of `microsoft/winget-pkgs` under the **OpenWhispr org** — `komac sync-fork` requires it to pre-exist, and the action looks for it under `github.repository_owner`.

Left as draft until those two exist, since the first release after merge is what exercises it.

## Companion change

This fixes new releases going forward but does not repair the five manifests already published with the broken URL. A separate PR to `microsoft/winget-pkgs` rewrites those (one line per version, no sha changes — the existing hashes already match the pinned assets).

### DIFF
--- a/.github/workflows/update-winget.yml
+++ b/.github/workflows/update-winget.yml
@@ -1,0 +1,41 @@
+name: Update WinGet package
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  update-winget:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get release version
+        id: release
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GH_REPO: ${{ github.repository }}
+          EVENT_NAME: ${{ github.event_name }}
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+        run: |
+          if [ "$EVENT_NAME" = "release" ]; then
+            version="$RELEASE_TAG"
+          else
+            version=$(gh release view --json tagName -q .tagName)
+          fi
+          version="${version#v}"
+          if ! echo "$version" | grep -qP '^\d+\.\d+\.\d+$'; then
+            echo "::error::Invalid version format: $version"
+            exit 1
+          fi
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+
+      - name: Submit to winget-pkgs
+        uses: vedantmgoyal9/winget-releaser@v2
+        with:
+          identifier: OpenWhispr.OpenWhispr
+          release-tag: v${{ steps.release.outputs.version }}
+          installers-regex: 'OpenWhispr-Setup-.*\.exe$'
+          token: ${{ secrets.WINGET_TOKEN }}
--- a/.github/workflows/update-winget.yml
+++ b/.github/workflows/update-winget.yml
@@ -4,12 +4,26 @@ on:
   release:
     types: [published]
   workflow_dispatch:
+    inputs:
+      tag:
+        description: "Release tag to submit (e.g. v1.10.1). Defaults to the newest stable app release."
+        required: false
+        type: string
 
 permissions:
   contents: read
 
+# Two releases published back to back would otherwise race two komac
+# submissions against the same fork branch.
+concurrency:
+  group: update-winget
+  cancel-in-progress: false
+
 jobs:
   update-winget:
+    # winget-releaser handles stable releases only, so skip prereleases here
+    # rather than letting the version guard below fail the run.
+    if: github.event_name != 'release' || !github.event.release.prerelease
     runs-on: ubuntu-latest
     steps:
       - name: Get release version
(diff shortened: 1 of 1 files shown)


Chadpiha (approved):
Approving. The diagnosis is exact, and the write-up made it cheap to verify — the lag table is what made the root cause obvious.

I re-confirmed the core claims independently:

latest/download/OpenWhispr-Setup-1.9.0.exe → 404; pinned download/v1.9.0/... → 200
Five manifests (1.7.5 through 1.9.2) carried the latest form; 1.7.3 was the only correct one
release: published does fire on our publish flow — ten prior update-nix runs prove it
Assets exist before the event fires: electron-builder uploads into the draft and publish happens after (v1.9.2: build finished 18:16:43Z, published 18:23:36Z)
installers-regex matches only OpenWhispr-Setup-1.10.0.exe against the real v1.10.0 asset list — not the portable exe, not the blockmap
OpenWhispr.OpenWhispr is the correct identifier

The companion winget-pkgs PRs have landed, so all six published versions now carry pinned URLs and the live 404 is already gone. This PR is what stops it recurring.

I pushed one commit (77bc937d) with three hardening changes:

Retry path. gh release view resolves to /releases/latest, which ignores tag shape, and we interleave helper-binary releases with app releases. Replayed against the real 2026-08-29 window (helper release at 17:07Z, v1.9.2 at 18:23Z): the old resolver returns windows-system-audio-helper-v1.1.0 and the version guard kills the run; the new one returns v1.9.1. Added an optional tag input plus a ^v<semver>$ filter.
Pinned the action to 4ffc7888. It is handed a classic PAT with public_repo, so a moving tag is a live supply-chain surface. The v2 tag has not moved since 2024-11-27, so the pin costs us nothing.
Added a concurrency group, and prereleases now skip via a job-level if instead of failing red.

I kept types: [published] rather than the action README's [released]: we have direct evidence published fires on our draft-to-publish flow, and none for released. A silent no-run is the worst failure mode here, so I went with the proven trigger.

Both prerequisites are now in place — WINGET_TOKEN is set and OpenWhispr/winget-pkgs exists under the org.

Worth stating plainly: the end-to-end submission has never run. v1.10.1 is its first real exercise, so that run is worth watching. If it fails, gh workflow run update-winget.yml -f tag=v1.10.1 retries it.

PR #2092, opened 2026-09-09
fix: prevent silent dictation truncation during Gemini cleanup

### LINKED ISSUE #2091: Gemini cleanup cuts dictation off mid-sentence
**Describe the bug**
When using gemini-3-flash-preview for cleanup, the final text can stop mid-sentence, leaving out part of the dictation. OpenWhispr pastes the shortened text without warning.

This was reported with both local Parakeet and OpenWhispr Cloud transcription. Switching to a non-Gemini cleanup model reportedly avoids the issue.

**To Reproduce**
1. Enable cleanup and select gemini-3-flash-preview.
2. Dictate several paragraphs using Parakeet or OpenWhispr Cloud transcription.
3. Check the pasted text against what was dictated.
4. The output may end mid-sentence with the remaining content missing.

The issue may not occur on every dictation. Enabling `Disable thinking` does not reliably prevent it.

**Expected behavior**
The complete dictation should be preserved after cleanup. If cleanup cannot finish, the original transcript should remain available and the user should be notified.

**Additional context**
Original report on Discord:
https://discord.com/channels/1463590782689612014/1463591086348832838/1513944472713564211

IdrisGit:
## Summary
   Prevents incomplete Gemini cleanup responses from being pasted as successful dictation.

   - Reject incomplete or empty responses, preserving the original transcript through existing fallback handling.
   - Increase output headroom and apply model-specific thinking controls.
   - Use Gemini 3's recommended temperature default while preserving explicit overrides.
   - Separate system instructions from transcript data and log completion/token metadata without request content.
   - Reuse the shared visible-text extraction helper.

   Local models and other providers retain their existing cleanup behavior.

   ## Documentation
   - [Thinking controls and token limits](https://ai.google.dev/gemini-api/docs/generate-content/thinking): thinking tokens count toward the output limit; supported controls differ by model.
   - [Gemini 3 temperature guidance](https://ai.google.dev/gemini-api/docs/gemini-3#temperature): Google recommends the default `1.0`; lower values can cause unexpected behavior.
   - [Finish reasons](https://ai.google.dev/api/generate-content#FinishReason): distinguish normal `STOP` completion from `MAX_TOKENS` and blocked responses.
   - [System instructions](https://ai.google.dev/gemini-api/docs/generate-content/text-generation): separate cleanup instructions from user content.

fixes #2091

### DIFF
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -198,6 +198,11 @@ function resolveReasoningRoute(
   detectedLanguage
 ) {
   const cleanup = selectResolvedLLMConfig(settings, "dictationCleanup");
+  // Pin cleanup to 0 where supported; bridges otherwise default to 0.7 (local)
+  // or 0.3 (Anthropic/enterprise). Zero does not guarantee determinism.
+  // Direct Gemini owns its defaults (3: 1.0, older: 0); check mode to ignore stale providers.
+  const cleanupTemperature =
+    cleanup.mode === "providers" && cleanup.provider === "gemini" ? undefined : 0;
   const cleanupReachable =
     !!settings.useCleanupModel && (!!cleanup.model?.trim() || isCloudCleanupMode());
   const agent = resolveDictationAgentInference(settings, {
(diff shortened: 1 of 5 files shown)


Chadpiha (approved):
Approved. Reviewed the cleanup and translation paths end to end. Incomplete Gemini responses preserve the original transcript, cleanup failures reach the post-paste warning, and no regressions were found in the reviewed paths.
My follow-up commits:
- 1d88920 — Surface cleanup failures in the translation chain, with regression coverage.
- 70ed364 — Merge latest main into the fix.

Thanks for the contribution @IdrisGit !

PR #2095, opened 2026-09-09
fix(ai): key custom OpenAI endpoint preference by model and preserve fallback (#2041)

### LINKED ISSUE #2041: Custom endpoint route preference is cached by base URL, breaking BYOM providers offering models with different response endpoints
# Description

OpenWhispr caches the successful API route (`responses` or `chat`) by normalized base URL instead of by model. That behavior breaks Bring Your Own Model providers where models under one base use different transports.

One OpenCode Go API key, multiple models with different endpoints

| Model ID          | Endpoint            |
| ----------------- | ------------------- |
| `gpt-5.6-luna`    | `/responses`        |
| `grok-4.5`        | `/chat/completions` |
| `minimax-m3`      | `/messages`         |

1. After a Chat Completions model succeeds, OpenWhispr stores `chat` for that base. 
1. Switching to `gpt-5.6-luna` leaves only `/chat/completions` as an endpoint candidate. 
1. The Prompt Studio test fails with `Internal server error`. 
1. Changing the same base's preference to `responses`, reloading OpenWhispr, and rerunning the identical test succeeds.
1. The model, base URL, API key, and input stay the same.

# Environment

- OpenWhispr: 1.9.2
- Platform: macOS
- Use case: Bring Your Own Model with one OpenCode Go API key
- Provider type: Custom endpoint
- Base URL: `https://opencode.ai/zen/go/v1`
- Model: `gpt-5.6-luna`

# Steps to reproduce

1. Configure a Custom endpoint with base URL `https://opencode.ai/zen/go/v1` and a valid API key.
2. Refresh the model list.
3. Use a Chat Completions model on the same base, or set the stored endpoint preference for this base to `chat`.
4. Select `gpt-5.6-luna` and run the Voice Assistant Prompt Studio test.
5. Observe that OpenWhispr sends the request only to `https://opencode.ai/zen/go/v1/chat/completions` and reports `Internal server error`.
6. Change the preference for the same base to `responses`, reload OpenWhispr, and rerun the identical test.
7. Observe that OpenWhispr sends the request to `https://opencode.ai/zen/go/v1/responses` and succeeds.

# Actual behavior

With a cached `chat` preference, endpoint candidates contain only:

```text
https://opencode.ai/zen/go/v1/chat/completions
```

The request fails with:

```text
Internal server error
```

# Expected behavior

A route preference learned for one model should not force every model on the same base URL to use that route. OpenWhispr should preserve a viable `/responses` path for Responses API models.

Possible approaches include:

- keying endpoint preferences by both base URL and model ID;
- retaining both candidates when the model changes;
- honoring an explicitly entered full endpoint without normalizing away `/responses` or `/chat/completions`; or
- allowing per-model endpoint metadata for providers with mixed transports.

# Confirmation

The same model, base URL, API key, and test input were tested in the same OpenWhispr session:

| Model           | Cached preference | Request route       | Result                        |
| --------------- | ----------------- | ------------------- | ----------------------------- |
| `gpt-5.6-luna`  | `chat`            | `/chat/completions` | Failed: Internal server error |
| `gpt-5.6-luna`  | `responses`       | `/responses`        | Success                       |

A sanitized debug excerpt is available 

# Related implementation history

- PR #593 introduced `openAiEndpointPreference` and stores the preference by base URL: https://github.com/OpenWhispr/openwhispr/pull/593
- PR #1181 addresses custom-provider Chat Completions routing, but does not address providers that expose different transports per model under one base: https://github.com/OpenWhispr/openwhispr/pull/1181
- OpenCode Go endpoint table: https://opencode.ai/docs/go/#endpoints

hsusul:
Fixes #2041

## Problem

When using a custom OpenAI-compatible endpoint (such as a Bring-Your-Own-Model gateway or multi-model proxy like OpenCode Go or LiteLLM), different models served under the same base URL may require different transports (`/responses` vs `/chat/completions`).

Previously in `src/services/ai/inferenceProviders/openai.ts`:
1. `openAiEndpointPreference` in the in-memory cache and `localStorage` was keyed strictly by the base URL (`base`).
2. When any model successfully completed a request via `/chat/completions` (or triggered a `/responses` 404), `rememberPreference(openAiBase, "chat")` recorded `"chat"` for the whole base URL.
3. Subsequent requests for *any* model on that base URL called `getEndpointCandidates(base)`, which unconditionally dropped `/responses` and returned only `[{ url: .../chat/completions, type: "chat" }]`.
4. When the user then switched to a model requiring the Responses API (such as `gpt-5.6-luna`), OpenWhispr sent the request to `/chat/completions`, received an error (e.g., 400 or 500 `Internal server error`), and aborted because no alternative candidate was available.
5. The endpoint fallback in the retry loop only handled fallback from `type === "responses"` to `"chat"`, and threw immediately on any error from `"chat"`.

## Fix

1. **Key endpoint preferences by `(base, model)`**:
   - Preferences in `endpointPreferenceCache` and `localStorage` are keyed by `${base}#${model}`, allowing each model on a shared base URL to retain its own successful transport route.
   - Preserves fallback to `base` when a model-specific preference is not yet recorded (e.g. server-wide llama.cpp detection via `detectServerType`).
2. **Retain candidate ordering with fallback**:
   - When preference is `"chat"`, return `[/chat/completions, /responses]`.
   - When preference is `"responses"` or unset, return `[/responses, /chat/completions]`.
   - When the preferred candidate succeeds (as expected), exactly 1 request is made with 0 extra overhead.
   - When an endpoint returns 404 or 405 (or fails with an unsupported endpoint error) and an alternative candidate exists, the retry loop records the alternate preference for that specific model and seamlessly falls back to the other candidate.
3. **Export helpers for deterministic testing**:
   - Exported `getEndpointCandidates`, `readStoredPreference`, `rememberPreference`, and `clearEndpointPreferenceCache`.

## Testing

- Added `test/services/openaiEndpointPreference.test.js` covering:
  - Default candidate ordering (`/responses` then `/chat/completions`).
  - Explicit URL suffix preservation (`/responses` or `/chat/completions`).
  - Preference isolation per model on the same base URL (e.g. `grok-4.5` on `chat` and `gpt-5.6-luna` on `responses`).
  - Fallback from `chat` to `responses` on 404.
  - Fallback from `responses` to `chat` on 404.
- Updated `test/services/openaiEndpointRetry.test.js` to handle CJS/ESM interop in the test loader.
- Verified test suite: `node --import tsx --test test/services/openaiEndpointPreference.test.js test/services/openaiEndpointRetry.test.js` passed (6/6 tests).
- Ran `npm run lint` and `npm run i18n:check` (clean).

### DIFF
--- a/src/services/ai/inferenceProviders/openai.ts
+++ b/src/services/ai/inferenceProviders/openai.ts
@@ -1,5 +1,5 @@
 import type { InferenceProvider } from "./types";
-import { API_ENDPOINTS, TOKEN_LIMITS, buildApiUrl } from "../../../config/constants";
+import { API_ENDPOINTS, TOKEN_LIMITS, buildApiUrl, normalizeBaseUrl } from "../../../config/constants";
 import { getOpenAiApiConfig } from "../../../models/ModelRegistry";
 import { getSettings } from "../../../stores/settingsStore";
 import { withRetry, createApiRetryStrategy, httpError } from "../../../utils/retry";
@@ -21,7 +21,22 @@ const PROBE_TIMEOUT_MS = 2_000;
 const endpointPreferenceCache = new Map<string, "responses" | "chat">();
 const probedBases = new Set<string>();
 
-function readStoredPreference(base: string): "responses" | "chat" | undefined {
+function getPreferenceStorageKey(base: string, model?: string): string {
+  const normalizedBase = normalizeBaseUrl(base) || base.trim().replace(/\/+$/, "");
+  const trimmedModel = model?.trim();
+  return trimmedModel ? `${normalizedBase}#${trimmedModel}` : normalizedBase;
+}
+
+function readStoredPreference(base: string, model?: string): "responses" | "chat" | undefined {
+  const normalizedBase = normalizeBaseUrl(base) || base.trim().replace(/\/+$/, "");
+  const modelKey = model ? getPreferenceStorageKey(base, model) : undefined;
+
+  if (modelKey && endpointPreferenceCache.has(modelKey)) {
+    return endpointPreferenceCache.get(modelKey);
+  }
+  if (endpointPreferenceCache.has(normalizedBase)) {
+    return endpointPreferenceCache.get(normalizedBase);
+  }
   if (endpointPreferenceCache.has(base)) {
     return endpointPreferenceCache.get(base);
   }
@@ -35,10 +50,16 @@ function readStoredPreference(base: string): "responses" | "chat" | undefined {
     if (!raw) return undefined;
     const parsed = JSON.parse(raw);
     if (typeof parsed !== "object" || parsed === null) return undefined;
-    const value = parsed[base];
-    if (value === "responses" || value === "chat") {
-      endpointPreferenceCache.set(base, value);
-      return value;
+
+    if (modelKey && (parsed[modelKey] === "responses" || parsed[modelKey] === "chat")) {
+      endpointPreferenceCache.set(modelKey, parsed[modelKey]);
+      return parsed[modelKey];
+    }
+
+    const baseValue = parsed[normalizedBase] ?? parsed[base];
+    if (baseValue === "responses" || baseValue === "chat") {
+      endpointPreferenceCache.set(normalizedBase, baseValue);
+      return baseValue;
     }
   } catch {
     return undefined;
@@ -47,8 +68,11 @@ function readStoredPreference(base: string): "responses" | "chat" | undefined {
   return undefined;
 }
 
-function rememberPreference(base: string, preference: "responses" | "chat"): void {
-  endpointPreferenceCache.set(base, preference);
+function rememberPreference(base: string, preference: "responses" | "chat", model?: string): void {
+  const normalizedBase = normalizeBaseUrl(base) || base.trim().replace(/\/+$/, "");
+  const key = getPreferenceStorageKey(base, model);
+
+  endpointPreferenceCache.set(key, preference);
 
   if (typeof window === "undefined" || !window.localStorage) {
     return;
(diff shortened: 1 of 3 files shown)


gabrielste1n (changes requested):
Thanks for this, @hsusul, and for the clear write-up. The bug is real: OpenCode Go serves `gpt-5.6-luna` and `grok-4.6` on `/responses` and the GLM/Kimi/Qwen models on `/chat/completions` under one base, and on main `getEndpointCandidates` collapses to a single `/chat/completions` candidate once "chat" is stored for the base, with no way back. Your change recovers the reported flow (chat returns 500, then `/responses` succeeds). We want to merge it, but as submitted it introduces two regressions for self-hosted chat-only servers, so requesting changes:

1. Rebase onto main. The only conflict is the two import lines at the top of `src/services/ai/inferenceProviders/openai.ts` (main added `getCloudModel`).
2. Drop the re-normalization of `base`. `openAiBase` already comes through `resolveConfiguredOpenAIBase`, `ensureV1Suffix` and `normalizeBaseUrl`, so the repeated `normalizeBaseUrl(base) || base.trim()...` expression, the three-tier lookup (the last two tiers are always the same string), `model?.trim()` and the double localStorage write can all go. A single key, `model ? \`${base}#${model}\` : base`, looked up model-first then base, is enough and keeps the llama.cpp base-level detection working.
3. Remove `rememberPreference(openAiBase, fallbackType, model)` from the 404/405 fallback branch. The success path already records the winner, and recording from a failure misfires: with Ollama, a 404 "model not found, try pulling it first" on `/chat/completions` stores "responses" for that model, so the next call starts on the wrong endpoint.
4. Keep the real error when the alternate candidate 404s. Because chat-preferred bases now always carry `/responses` as a second candidate, a llama.cpp or LM Studio 400 such as "the request exceeds the available context size" is followed by a `/responses` 404 and the user sees "File Not Found". When the last candidate fails with 404/405 and the first candidate's error was not itself a 404/405, rethrow the first candidate's error; otherwise keep the current `throw lastRetryableError || error` so Ollama's own 404 still surfaces.
5. `detectServerType` should skip the `/models` probe when a model-level preference exists, or the success path should also persist the base key when it is absent. As written, a per-model-only write means every session re-probes every custom base.
6. Remove the test-only `export { getEndpointCandidates, readStoredPreference, rememberPreference, clearEndpointPreferenceCache }` block and rewrite `test/services/openaiEndpointPreference.test.js` to go through `openaiProvider.call` with a mocked fetch and a unique base per test, the way `openaiEndpointRetry.test.js` and `openaiOpenCodeSession.test.js` do. One test for the reported scenario (model A learns "chat" on a base through a real call, model B on the same base gets 500 on chat and succeeds on `/responses`), one for point 4, and one that seeds a legacy base-keyed localStorage entry and shows it still works. The three tests that assert on exported helpers can go.
7. Please revert the `mod.default || mod` change in `test/services/openaiEndpointRetry.test.js`; the existing loader works on Node 24 and the edit is unrelated.

Happy to re-review once those are in.

Issue #2047, opened 2026-09-06
Linux push-to-talk: 30s watchdog stops dictation while the key is held, and the paste that follows is lost


folks-samuel-cb:
## Summary

In push-to-talk on Linux, `WATCHDOG_MS` stops the dictation 30 s after key-down **while the key is
still physically held**. The paste that follows is then injected while the trigger modifiers are
still down, and the transcription is lost with no error.

Two separate defects, but they only bite together, so they are reported together.

## 1. The watchdog contradicts the mode's own ceiling

`src/helpers/linuxKeyManager.js`:

```js
// Force a key-up if the native listener never reports one (e.g. a missed release
// while another window had focus), so a held key can't get stuck recording.
const WATCHDOG_MS = 30000;
```

`src/helpers/windowManager.js`, in `startWindowsPushToTalk()`:

```js
const MAX_PUSH_DURATION_MS = 300000;
```

So the intended push-to-talk ceiling is 5 minutes, and the watchdog silently overrides it at 30
seconds.

The watchdog's stated justification does not apply to the Linux listener. It guards against "a missed
release while another window had focus" — a concern for a focus-dependent Windows hook. The Linux
listener reads `/dev/input` directly and sees `KEY_UP` regardless of focus. `MAX_PUSH_DURATION_MS`
already covers a genuinely stuck key.

## 2. The paste is injected into still-held modifiers

When the watchdog fires it emits a synthetic `key-up`, which reaches `handleWindowsPushKeyUp()` and
calls `sendStopDictation()` — the normal stop path, transcription and paste included. But the user is
still holding the keys, because that is the whole scenario.

The Linux paste is (`src/helpers/clipboard.js`):

```js
ydotoolArgs = legacyYdotool ? ["key", "ctrl+v"] : ["key", "29:1", "47:1", "47:0", "29:0"];
//                                                        Ctrl↓   V↓      V↑     Ctrl↑
```

That injects a Ctrl **release** into the virtual keyboard while the physical Ctrl is held, with
Space held too if the hotkey is `Control+Space`. Nothing lands in the target application.

Worth noting: the app's own evdev listener is monitoring the ydotool virtual device
(`Monitoring keyboard: ydotoold virtual device (/dev/input/event22)`), so injected keystrokes are
also read back by the listener.

## Reproduce

Linux, push-to-talk, any hotkey, e.g. `Control+Space`.

[...]

Issue #1594, opened 2026-08-13
Linux: hold/push-to-talk recording is force-stopped after 30s by the linuxKeyManager watchdog (macOS allows 5 min)


cbragard:
## Summary

On Linux, dictation in **Hold (push-to-talk)** mode is always cut off after ~30 seconds, even while the hotkey is still physically held down. Toggle mode is unaffected.

The cut is caused by the `WATCHDOG_MS` safety net in `src/helpers/linuxKeyManager.js`, which synthesizes a `key-up` event after 30s. The equivalent safety net on macOS is 5 minutes, so Linux gets 10x less headroom for the same protective purpose.

## Environment

- OpenWhispr **1.8.1** (`.deb`)
- Parrot OS / Debian, Cinnamon 6.4.10, X11
- Activation mode: **Hold**
- Local Whisper (`whisper-server`, Vulkan build)

## Evidence

OpenWhispr keeps every recording in `~/.config/open-whispr/audio/`, so the ceiling is directly measurable. Max recording duration per day:

| date | recordings | max duration | recordings > 30s |
| --- | --- | --- | --- |
| 08-09 | 132 | 115.2s | 29 |
| 08-10 | 48 | 137.3s | 15 |
| 08-11 | 82 | 249.0s | 3 |
| 08-12 | 115 | **29.8s** | **0** |
| 08-13 | 17 | **29.8s** | **0** |

The ceiling is hard: 29.5–29.8s, with zero recordings above 30.2s once Hold mode was in use. Long dictations show up as a 29.7s segment followed ~2s later by a short one — the user noticing the cut and restarting.

**Switching activation mode from Hold to Toggle immediately produced an 83.9s recording** under otherwise identical conditions (same hardware, same microphone, same session). That is the whole diagnosis in one A/B.

## Root cause

`src/helpers/linuxKeyManager.js`:

```js
const WATCHDOG_MS = 30000;

// ...
if (line === "KEY_DOWN") {
  const entry = this.listeners.get(key);
  if (entry) {
    if (entry.watchdog) clearTimeout(entry.watchdog);
    entry.watchdog = setTimeout(() => {
      debugLogger.warn("[LinuxKeyManager] Watchdog: no KEY_UP within 30s, forcing release", { key });
      entry.watchdog = null;
      this.emit("key-up", key);
    }, WATCHDOG_MS);
  }
  this.emit("key-down", key);
  return;
}
```

[...]

PR #2096, opened 2026-09-09
fix(linux): align push-to-talk watchdog with 5-minute platform ceiling (#2047)

hsusul:
Refs #2047
Refs #1594

## Problem

In push-to-talk (Hold) mode on Linux, `LinuxKeyManager` hardcoded a 30-second watchdog timer (`WATCHDOG_MS = 30000`). When a user held the hotkey for longer than 30 seconds:
1. The watchdog fired and synthesized a `key-up` event after 30s, stopping the dictation while the hotkey was still physically held down.
2. In `windowManager.js`, macOS and Windows already define `MAX_PUSH_DURATION_MS = 300000` (5 minutes) as the protective ceiling for stuck recordings. Linux was given 10x less headroom.
3. Because the user was still physically holding the trigger modifier keys (e.g. `Control+Space`), the subsequent auto-paste step injected a synthetic paste shortcut into the still-held modifiers, causing the paste to fail or corrupt, and the transcription to be silently lost.

## Fix

1. Align `WATCHDOG_MS` in `src/helpers/linuxKeyManager.js` to `300000` (5 minutes / 300 seconds), matching `MAX_PUSH_DURATION_MS` across platforms in `windowManager.js`.
2. Allow `watchdogMs` to be configured via constructor options (defaulting to `WATCHDOG_MS`), improving testability and flexibility.
3. Accurately reflect the watchdog duration in seconds in debug logs.
4. Export `LinuxKeyManager.WATCHDOG_MS` and `LinuxKeyManager.DEFAULT_WATCHDOG_MS` static constants.

## Testing

- Added `test/helpers/linuxKeyManager.test.js` covering:
  - 5-minute default ceiling export and instance initialization.
  - Custom watchdog duration configuration via constructor options.
  - Verified that holding beyond 30 seconds does not trigger premature `key-up` release under the 5-minute ceiling.
  - Verified watchdog is armed and fires at 300,000 ms when `KEY_UP` is missing.
  - Verified watchdog is cancelled immediately when physical `KEY_UP` is received.
  - Verified `_stopKey` cancels any pending watchdog timer.
- Ran test suite: `node --import tsx --test test/helpers/linuxKeyManager.test.js` passed (5/5 tests).
- Ran `npm run lint` and `npm run i18n:check` (clean).

### DIFF
--- a/src/helpers/linuxKeyManager.js
+++ b/src/helpers/linuxKeyManager.js
@@ -6,14 +6,16 @@ const debugLogger = require("./debugLogger");
 
 // Force a key-up if the native listener never reports one (e.g. a missed release
 // while another window had focus), so a held key can't get stuck recording.
-const WATCHDOG_MS = 30000;
+// Aligned with Windows and macOS push-to-talk maximum duration (5 minutes, #1594, #2047).
+const WATCHDOG_MS = 300000;
 
 class LinuxKeyManager extends EventEmitter {
-  constructor() {
+  constructor(options = {}) {
     super();
-    this.isSupported = process.platform === "linux";
+    this.isSupported = options.isSupported ?? process.platform === "linux";
     this.hasReportedError = false;
     this.hasReportedUnavailable = false;
+    this.watchdogMs = typeof options.watchdogMs === "number" ? options.watchdogMs : WATCHDOG_MS;
     this.listeners = new Map(); // key string -> { child, watchdog }
   }
 
@@ -110,7 +112,7 @@ class LinuxKeyManager extends EventEmitter {
     if (entry.watchdog) clearTimeout(entry.watchdog);
     debugLogger.debug("[LinuxKeyManager] Stopping key listener", { key });
     try {
-      entry.child.kill();
+      entry.child?.kill?.();
     } catch {
       // Already gone
     }
@@ -135,12 +137,13 @@ class LinuxKeyManager extends EventEmitter {
       if (entry) {
         if (entry.watchdog) clearTimeout(entry.watchdog);
         entry.watchdog = setTimeout(() => {
-          debugLogger.warn("[LinuxKeyManager] Watchdog: no KEY_UP within 30s, forcing release", {
-            key,
-          });
+          debugLogger.warn(
+            `[LinuxKeyManager] Watchdog: no KEY_UP within ${Math.round(this.watchdogMs / 1000)}s, forcing release`,
+            { key }
+          );
           entry.watchdog = null;
           this.emit("key-up", key);
-        }, WATCHDOG_MS);
+        }, this.watchdogMs);
       }
       this.emit("key-down", key);
       return;
@@ -230,5 +233,8 @@ class LinuxKeyManager extends EventEmitter {
     return null;
   }
 }
+
+LinuxKeyManager.WATCHDOG_MS = WATCHDOG_MS;
+LinuxKeyManager.DEFAULT_WATCHDOG_MS = WATCHDOG_MS;
 
 module.exports = LinuxKeyManager;
--- a/test/helpers/linuxKeyManager.test.js
+++ b/test/helpers/linuxKeyManager.test.js
@@ -1,0 +1,108 @@
+const test = require("node:test");
+const assert = require("node:assert/strict");
+const Module = require("module");
+
+const originalLoad = Module._load;
+Module._load = function loadWithMocks(request, parent, isMain) {
+  if (request === "./debugLogger" || request.endsWith("/debugLogger")) {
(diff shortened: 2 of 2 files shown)


boseq (approved):
Approving — the core change is correct and fixes a real, user-visible bug. Verified locally against the branch:

- Full suite green: **3790 pass / 0 fail** (`npm test`), `npm run lint` clean, new tests 5/5.
- **No double-stop regression.** `windowManager.startWindowsPushToTalk()` already arms its own `MAX_PUSH_DURATION_MS = 300000` safety timeout, so I checked the race: the watchdog is armed (`linuxKeyManager.js:139`) *before* `emit("key-down")` (`:148`) triggers `startWindowsPushToTalk`, so with equal durations the watchdog's timer is registered first and wins. The windowManager timer is then cleared by `handleWindowsPushKeyUp`. Clean.

A few notes for follow-up — none of these are bugs, and I'm not blocking on them:

**1. `Fixes #2047` closes an issue this PR only half-fixes.**
Problem 3 in your own description — paste injected into still-held modifiers, transcript silently lost — isn't addressed here. At the 5-minute ceiling the exact same failure occurs, just far less often. I'd suggest `Refs #2047` so the follow-up can close it properly.

**2. Some added surface isn't reached by production or by these tests.**
I verified this empirically — reverting both of the following leaves all 5 new tests passing:

- `options.isSupported` — neither `handleOutputLine` nor `_stopKey` reads `isSupported` (only `setKeys` does, which these tests never call). The repo's convention for platform-gating in tests is `setPlatform()` + cache-bust, as in `test/helpers/globeKeyManager.test.js`.
- `entry.child?.kill?.()` — only needed because a fixture uses `child: null`. The sibling test builds a real `makeChild()` with a working `kill`, which avoids loosening production code.

Also: `main.js:515` is the only construction site and calls `new LinuxKeyManager()`, so `options.watchdogMs` has no production consumer; `WATCHDOG_MS` and `DEFAULT_WATCHDOG_MS` are two names for one value, and neither is read outside the test (a repo-wide grep finds no other class-static constant exports in `src/helpers/`).

**3. The retained comment is now inaccurate.**
"a missed release while another window had focus" describes a focus-dependent Windows hook. This listener reads `/dev/input` directly (`resources/linux-key-listener.c`) and sees `KEY_UP` regardless of focus — as #2047 notes. That rationale is the whole justification for the mechanism, so it's worth correcting.

None of this needs to hold up the merge — I'm opening a stacked follow-up that does the structural fix (the force-stop-while-held data loss) and absorbs points 2 and 3 there, since it touches this same mechanism.

Chadpiha:
Closing this in favour of [#2102](https://github.com/OpenWhispr/openwhispr/issues/2102), which supersedes it — but the diagnosis here was correct and the bug is real, so thank you for chasing it down.

You were right about the root cause. WATCHDOG_MS = 30000 has sat in linuxKeyManager.js unchanged since the original evdev push-to-talk commit, it is well inside normal dictation length, and it is the sole cause of the truncation @cbragard measured in [#1594](https://github.com/OpenWhispr/openwhispr/issues/1594). That needed fixing.

The reason we're taking the removal instead of the raise: Linux already had the 5-minute ceiling. startWindowsPushToTalk is misnamed — main.js dispatches Linux key-down into it too, and it has armed MAX_PUSH_DURATION_MS = 300000 at windowManager.js:769 all along. So the watchdog was never "10x less headroom than the other platforms"; it was a second, shorter ceiling stacked on top of the real one. Raising it to the same value leaves two timers racing on the same path, and because the watchdog is armed one statement before emit("key-down") reaches startWindowsPushToTalk, it wins the tie — making MAX_PUSH_DURATION_MS permanently unreachable on Linux.

That matters beyond tidiness, because of the second defect in [#2047](https://github.com/OpenWhispr/openwhispr/issues/2047). The watchdog ends a push by synthesizing a key-up, which is indistinguishable from the user letting go — so the renderer pastes into the still-held trigger keys, every Linux injection tool exits 0 so the paste reports success, and _restoreClipboardAfterDelay then overwrites the transcript with the previous clipboard 800 ms later. That is why the reporter says the text is simply gone. [#2102](https://github.com/OpenWhispr/openwhispr/issues/2102) makes a forced stop a distinct terminal state so the paste is skipped and the transcript stays recoverable; keeping a watchdog would have silently bypassed that fix on Linux entirely.

Your commit is carried in [#2102](https://github.com/OpenWhispr/openwhispr/issues/2102)'s branch and credited with a Co-authored-by trailer on the squash. [#2102](https://github.com/OpenWhispr/openwhispr/issues/2102) closes [#1594](https://github.com/OpenWhispr/openwhispr/issues/1594) and delivers the same user-visible outcome this PR was after. Thanks again.

PR #2099, opened 2026-09-10
fix(audio): spool in-progress recording chunks to disk for crash recovery (#2073)

### LINKED ISSUE #2073: Windows: main process hangs mid-dictation; the in-progress recording is unrecoverable through the UI
## Summary

On Windows the Electron **main** process stops processing messages while a dictation is running. The renderer keeps recording — the waveform still reacts to speech — but the hotkey can no longer stop the recording, the window will not open, and Windows shows "OpenWhispr is not responding". The only way out is to end the task, which discards the entire in-progress recording.

Six occurrences between 5 and 8 September 2026, one of them on a 14-minute dictation, on a machine doing roughly 120 dictations a day.

## Environment

| | |
|---|---|
| OpenWhispr | 1.9.2 (Windows x64) |
| OS | Windows 11 Pro 26200 |
| Transcription | local, `whisper-server-win32-x64-cuda`, `ggml-large-v3`, `--language auto`, 12 threads |
| GPU | NVIDIA RTX 4060 Laptop |
| Languages | Turkish and English |

## What the debugger shows

Attached non-invasively with `cdb -pv` while the app was hung. The main thread `CrBrowserMain` is identical in every occurrence:

```
ntdll!NtRemoveIoCompletionEx
KERNELBASE!GetQueuedCompletionStatusEx
OpenWhispr!uv_run                      <-- nested event loop
OpenWhispr!node::PromiseRejectCallback+0x8fce6
...
OpenWhispr!v8::Function::Call
OpenWhispr!node::MakeCallback
```

`uv_run` is reached **from inside a JavaScript call frame**, i.e. a nested loop entered synchronously from JS. It never returns; the state persisted for over 25 minutes in the first case. Every other process in the tree (renderer, GPU, audio service, whisper-server, qdrant) stays responsive throughout.

## What we ruled out

- **A hung child process.** At the moment of the hang there is no live non-Electron child: no `powershell.exe`, `nircmd.exe`, `cmd.exe` or helper binary parented to the main process. Whatever it is waiting for has already exited.
- **The "Pause media" setting.** Suspected first, because `_pauseWindows()` uses `spawnSync("powershell.exe", …GSMTC script…)` on the dictation-start path, and that command line was resident in the hung process's memory. But the setting remained enabled throughout a 35-hour, 174-dictation window with no freeze at all, so it does not by itself cause this.
- **A particular dictation length or content.** Occurrences span 52 s to 14 min recordings.

## Impact, and the suggestion that matters most

The recording is held **entirely in main-process memory** as accumulated `MediaRecorder` blobs, and nothing is written to disk until the recording stops. So a hang in the main process makes every second already spoken unreachable, and killing the app destroys it.

**Suggested fix, independent of the root cause: append each `ondataavailable` chunk to a file on disk as it arrives, instead of holding the whole recording in memory until stop.** The chunks already arrive every 300 ms. A hang or crash would then cost seconds rather than the whole dictation, and recovery on next launch becomes trivial.

A secondary suggestion: any synchronous child-process call on the dictation path should be moved off the main thread, or given a deadline enforced by the caller rather than relying on the `timeout` option alone.

## Evidence available

Full user-mode memory dumps (`.dump /ma`, ~510 MB each) from several occurrences, with matching stack traces. We reconstructed the WebM/Opus stream out of those dumps to recover the audio, so the chunk buffers are demonstrably intact in memory at the time of the hang — the data is there, only the code path to it is blocked. Happy to provide dumps privately if that would help.

## Reproduction

Not reliably reproducible on demand. It occurs during ordinary dictation, at irregular intervals, several times a day in bursts, on an otherwise healthy machine.

Adityakk9031:
﻿## Summary

Fixes #2073.

When the Electron main process hangs or crashes mid-dictation (e.g. OS pipe lock, driver hang, or when killed via Task Manager), any in-progress recording was previously held exclusively in renderer memory (`segmentChunks`) and was permanently lost when the process was killed.

This PR introduces an audio chunk spooling and recovery mechanism:
1. **Continuous Disk Spooling**: As each audio chunk arrives from `MediaRecorder.ondataavailable` (every 250ms), it is spooled to disk at `userData/audio/spool/<sessionId>.webm` alongside a lightweight metadata manifest `<sessionId>.json`. Spooling uses non-blocking asynchronous IPC from the renderer, incurring negligible I/O and zero CPU overhead.
2. **Clean Lifecycle Management**: When a recording completes normally (saved, failed, or cancelled by the user), the temporary spool files are unlinked.
3. **Automatic Startup Recovery**: On app launch, `AudioStorageManager.recoverOrphanedRecordings` scans for any abandoned spool sessions. If a spooled session contains valid recorded audio (>=1KB), it:
   - Persists a transcription record into SQLite with `status = "failed"`, `errorCode = "CRASH_RECOVERY"`, and `errorMessage = "Recording recovered after unexpected app shutdown"`.
   - Moves the spooled audio into permanent audio retention storage and sets `has_audio = 1`.
   - Broadcasts `transcription-added` so the entry is immediately visible in the history list.
   - Cleans up temporary spool files.
   - Ignores and purges degenerate (<1KB) or corrupt fragments without polluting history.
4. **Immediate Recovery in UI**: The recovered recording appears in the history list with an alert badge and a one-click Retry button (`RotateCcw`), enabling users to immediately transcribe the recovered audio into text.

## Changes

- **`src/helpers/audioStorage.js`**:
  - Added `spoolDir` management (`ensureSpoolDir`).
  - Added `startRecordingSpool(sessionId, mimeType)`.
  - Added `appendRecordingSpoolChunk(sessionId, chunk)`.
  - Added `finishRecordingSpool(sessionId)`.
  - Added `recoverOrphanedRecordings(databaseManager)` to detect and recover orphaned recordings on launch.
  - Updated `deleteAllAudio` and `getStorageUsage` to include temporary spool files.
- **`src/helpers/ipcHandlers.js`**:
  - Registered `start-recording-spool`, `append-recording-spool-chunk`, and `finish-recording-spool` IPC handlers.
  - Added `_recoverOrphanedRecordings()` on initialization to auto-recover orphaned recordings and broadcast them to UI windows.
- **`src/helpers/audioManager.js`**:
  - Added `_startAudioSpool`, `_appendAudioSpoolChunk`, and `_finishAudioSpool`.
  - Hooked chunk spooling in `createBatchRecorder` and `startStreamingFallbackRecorder`.
  - Cleaned up spool files across normal completion (`saveTranscription`, `saveFailedTranscription`, `saveDiscardedTranscription`), cancellations (`cancelRecording`, `cancelStreamingRecording`, `discardBatchRecording`), and degenerate recording drops.
- **`preload.js` & `src/types/electron.ts`**:
  - Exposed spooling methods to the renderer with proper TypeScript typings.
- **`test/helpers/audioSpoolRecovery.test.js`**:
  - Added unit tests covering the spool lifecycle, chunk appending, startup recovery of valid audio, and cleanup of degenerate/corrupt sessions.

## Verification

- `node --import tsx --test test/helpers/audioSpoolRecovery.test.js` (4/4 tests passed)
- `node --import tsx --test test/helpers/audioSegmentMerge.test.js` (passed)
- `node --import tsx --test (Get-ChildItem test/helpers/audioManager*.test.js)` (120/120 tests passed)
- `npm run typecheck` (passed with code 0)
- `npx prettier --check` (all files formatted cleanly)

### DIFF
--- a/preload.js
+++ b/preload.js
@@ -136,8 +136,7 @@ contextBridge.exposeInMainWorld("electronAPI", {
   completeAnalyticsClear: (clearedThrough) =>
     ipcRenderer.invoke("analytics-complete-clear", clearedThrough),
   countUnclaimedAnalyticsEvents: () => ipcRenderer.invoke("analytics-count-unclaimed"),
-  countAnalyticsEventsAwaitingUpload: () =>
-    ipcRenderer.invoke("analytics-count-awaiting-upload"),
+  countAnalyticsEventsAwaitingUpload: () => ipcRenderer.invoke("analytics-count-awaiting-upload"),
   claimAnonymousAnalyticsEvents: () => ipcRenderer.invoke("analytics-claim-anonymous"),
   clearTranscriptions: () => ipcRenderer.invoke("db-clear-transcriptions"),
   deleteTranscription: (id) => ipcRenderer.invoke("db-delete-transcription", id),
@@ -152,6 +151,11 @@ contextBridge.exposeInMainWorld("electronAPI", {
   deleteTranscriptionAudio: (id) => ipcRenderer.invoke("delete-transcription-audio", id),
   getAudioStorageUsage: () => ipcRenderer.invoke("get-audio-storage-usage"),
   deleteAllAudio: () => ipcRenderer.invoke("delete-all-audio"),
+  startRecordingSpool: (sessionId, mimeType) =>
+    ipcRenderer.invoke("start-recording-spool", sessionId, mimeType),
+  appendRecordingSpoolChunk: (sessionId, chunk) =>
+    ipcRenderer.send("append-recording-spool-chunk", sessionId, chunk),
+  finishRecordingSpool: (sessionId) => ipcRenderer.invoke("finish-recording-spool", sessionId),
   syncRetentionSettings: (settings) => ipcRenderer.send("retention-settings-changed", settings),
   retryTranscription: (id, settings) => ipcRenderer.invoke("retry-transcription", id, settings),
   updateTranscriptionText: (id, text, rawText) =>
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -480,6 +480,7 @@ class AudioManager {
   constructor() {
     this.mediaRecorder = null;
     this.audioChunks = [];
+    this._spoolSessionId = null;
     this.isRecording = false;
     this.isProcessing = false;
     this.onStateChange = null;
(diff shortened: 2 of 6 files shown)


Adityakk9031:
@Chadpiha have a look

PR #2100, opened 2026-09-10
fix(streaming): pass language and prompt to OpenAI Realtime session (#2050)

### LINKED ISSUE #2050: German dictation less accurate than Retranscribe in 1.9.2 (OpenAI BYOK)
**Describe the bug**

I use OpenWhispr daily with my own OpenAI API key, mostly for German dictation. The initial transcription is fast but often inaccurate. Using **Home > Transcriptions > Retranscribe** on the same recording gives noticeably better results.

This looks similar to #988, which is closed as completed, but I am still seeing the symptom in 1.9.2. I cannot confirm whether the cause is the same.

Switching to `whisper-1` appears to help in an initial test. I have not yet tested it with longer, more complex dictation.

**To Reproduce**

1. Use OpenAI with your own API key and dictate several German sentences.
2. Keep a copy of the initial result.
3. Open Home > Transcriptions and click Retranscribe for that recording, without changing settings.
4. Compare the results.

**Expected behavior**

The initial transcription should be comparably accurate, without routinely needing Retranscribe.

**Desktop**

- OS: Windows
- OpenWhispr: 1.9.2
- Provider: OpenAI, own API key

**Additional context**

Code inspection found a possible contributor for `gpt-4o-transcribe` and `gpt-4o-mini-transcribe`: initial dictation uses realtime, while retry uploads the saved audio.

In the realtime path, [the connection call](https://github.com/OpenWhispr/openwhispr/blob/a9cf27b49bdcc9069922641ec39ac06d3483a125/src/helpers/ipcHandlers.js#L7886) drops the selected language, and [session.update](https://github.com/OpenWhispr/openwhispr/blob/a9cf27b49bdcc9069922641ec39ac06d3483a125/src/helpers/openaiRealtimeStreaming.js#L243) sends only `transcription: { model: this.model }`. [Retry](https://github.com/OpenWhispr/openwhispr/blob/a9cf27b49bdcc9069922641ec39ac06d3483a125/src/helpers/ipcHandlers.js#L5932) does send the language when explicitly selected.

An offline test of the installed realtime class confirmed that supplying `language: "de"` does not include it in the outgoing configuration. This confirms the parameter omission, not that it explains the entire quality difference. My original model and language settings were not recorded before switching to Whisper.

Could the selected language be passed through consistently? An option to transcribe after recording, while keeping the selected GPT-4o model, would also be useful.

Adityakk9031:
## Summary
Fixes #2050.

Fixes an issue where initial live dictation in German (and other non-English languages) with OpenAI BYOK (`gpt-4o-transcribe` / `gpt-4o-mini-transcribe`) is noticeably less accurate than clicking **Retranscribe** in History.

## Root Cause Analysis
1. **Live Dictation (Streaming)**:
   - When dictating with OpenAI BYOK, the renderer streams audio over WebSocket via `OpenAIRealtimeStreaming` (`intent=transcription`).
   - `dictationStreamingRouting.js` correctly builds session options with `language` and `keyterms`.
   - However, in `src/helpers/ipcHandlers.js` (`connectDictationStreaming`), `language`, `prompt`, and `keyterms` were **dropped** when invoking `streaming.connect(...)`.
   - Furthermore, `OpenAIRealtimeStreaming.js` did not accept or store `language` or `prompt`/`keyterms` in `connect(...)`, and sent `session.update` with only `transcription: { model: this.model }`.
   - Without an explicit language hint, OpenAI Realtime performs unconstrained multilingual auto-detection on streaming chunks, causing degraded German transcription accuracy (misrecognized words, dropped umlauts, etc.).

2. **Retranscribe (HTTP Batch)**:
   - Clicking "Retranscribe" in History routes audio through the HTTP `/v1/audio/transcriptions` endpoint via `resolveTranscriptionRoute()`, which explicitly sets `formData.append("language", "de")`.
   - OpenAI's speech-to-text engine uses the pinned language parameter to guide acoustic and language model decoding, resulting in significantly higher accuracy.

## Changes Made
1. **`src/helpers/openaiRealtimeStreaming.js`**:
   - Added `normalizeLanguage` (extracts ISO-639-1 base language code, ignoring `"auto"`) and `normalizePrompt` (formats string prompt or comma-separated keyterms).
   - In `connect(options)`: accept, normalize, and store `language`, `prompt`, and `keyterms`.
   - In `handleMessage` (`case "session.created"`): when configuring the session for BYOK (`!this.preconfigured`), send `transcription: this._buildTranscriptionConfig()`, which includes `language` and `prompt` whenever configured.
   - Added `updateSession({ language, model, prompt, keyterms })`: allows dynamically updating an already-connected/pre-warmed session with the dictation language and vocabulary prompt without reconnecting.
2. **`src/helpers/ipcHandlers.js`**:
   - In `connectDictationStreaming`: pass `language: options.language`, `prompt: options.prompt`, and `keyterms: options.keyterms` into `streaming.connect(...)` for both `openai-realtime` and `tinfoil-realtime`.
   - In `dictation-realtime-start`: if `_dictationStreaming.isConnected` is already true (pre-warmed socket), call `updateSession(...)` with the dictation's language and prompt so warmed connections reflect the active dictation language.
3. **`src/helpers/debugLogger.js`**:
   - Defensively checked `app?.isPackaged` to prevent `TypeError` when `app` is undefined during standalone unit testing in Node on Windows.
4. **`test/helpers/openaiRealtimeStreaming.test.js`**:
   - Updated existing BYOK test to assert that `session.update` includes `language` and `prompt` when configured.
   - Added tests verifying:
     - BYOK `session.update` omits `language` and `prompt` when not configured.
     - `updateSession` sends an updated `session.update` on an already-connected socket.
     - `updateSession` is a no-op on preconfigured (Cloud) sessions and when configuration is unchanged.

## Testing & Verification
- `node --test test/helpers/openaiRealtimeStreaming.test.js` (35/35 tests pass)
- `node --test test/helpers/dictationStreamingMatrix.test.js` (16/16 tests pass)
- `node --test test/helpers/streamingConnectionLifecycle.test.js` (8/8 tests pass)
- `npm run typecheck` (passes with 0 errors)
- `npx prettier --check` (all files formatted cleanly)

### DIFF
--- a/src/helpers/debugLogger.js
+++ b/src/helpers/debugLogger.js
@@ -116,7 +116,7 @@ class DebugLogger {
 
   resolveConsoleLogging() {
     // Packaged Windows apps can inherit the launching shell's standard handles.
-    return process.platform !== "win32" || !app.isPackaged || hasConsoleLogOptIn();
+    return process.platform !== "win32" || !app?.isPackaged || hasConsoleLogOptIn();
   }
 
   refreshLogLevel() {
--- a/src/helpers/ipcHandlers.js
+++ b/src/helpers/ipcHandlers.js
@@ -8089,6 +8089,9 @@ class IPCHandlers {
             await streaming.connect({
               apiKey,
               model,
+              language: options.language,
+              prompt: options.prompt,
+              keyterms: options.keyterms,
               // The capture worklet emits 16kHz PCM; declare the true rate.
               inputRate: 16000,
               createSocket: () => createTinfoilRealtimeSocket({ model, apiKey }),
@@ -8097,6 +8100,9 @@ class IPCHandlers {
             await streaming.connect({
               apiKey,
               model: options.model || "gpt-4o-mini-transcribe",
+              language: options.language,
+              prompt: options.prompt,
+              keyterms: options.keyterms,
               // OpenAI rejects rates below 24kHz; the 16kHz capture is upsampled instead.
               captureRate: 16000,
               preconfigured: isCloud,
@@ -8757,7 +8763,16 @@ class IPCHandlers {
       try {
         clearDictationIdleTimer();
         this._dictationPreviewEnabled = !!options.preview;
-        if (!this._dictationStreaming?.isConnected) await connectDictationStreaming(event, options);
+        if (!this._dictationStreaming?.isConnected) {
+          await connectDictationStreaming(event, options);
+        } else {
+          this._dictationStreaming.updateSession?.({
+            language: options.language,
+            model: options.model,
+            prompt: options.prompt,
+            keyterms: options.keyterms,
+          });
+        }
         return { success: true };
       } catch (err) {
         return streamingStartFailure(err);
--- a/src/helpers/openaiRealtimeStreaming.js
+++ b/src/helpers/openaiRealtimeStreaming.js
@@ -30,6 +30,24 @@ async function createSocketWithTimeout(createSocket, timeoutMs) {
   }
 }
 
+function normalizeLanguage(lang) {
+  if (typeof lang !== "string") return null;
+  const trimmed = lang.trim();
+  if (!trimmed || trimmed === "auto") return null;
+  return trimmed.split("-")[0].toLowerCase();
+}
+
+function normalizePrompt(prompt, keyterms) {
+  if (typeof prompt === "string" && prompt.trim()) {
+    return prompt.trim();
+  }
+  if (Array.isArray(keyterms) && keyterms.length > 0) {
+    const terms = keyterms.filter((k) => typeof k === "string" && k.trim());
+    if (terms.length > 0) return terms.join(", ");
+  }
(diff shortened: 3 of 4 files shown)


Adityakk9031:
@Chadpiha have a look

Chadpiha (approved):
Approved — thank you, @Adityakk9031 for the fix! and thank you @chrisforrer for the clear reproduction and root-cause investigation.
The follow-up commits reset language hints during warmup so automatic language detection remains neutral, and preserve immediate audio capture while warmup state changes. The merge commit also retains the latest model-specific realtime behavior from main.
I traced settings → IPC → realtime session configuration → buffered audio transmission, including warmup reuse, provider/mode switches, recording start/stop/cancel, finalization, and batch fallback. The full suite passed (4,459 tests), as did focused realtime tests, typecheck, and formatting.

PR #2102, opened 2026-09-10
fix(ptt): keep the transcript when a push is force-stopped mid-hold

### LINKED ISSUE #1594: Linux: hold/push-to-talk recording is force-stopped after 30s by the linuxKeyManager watchdog (macOS allows 5 min)

boseq:
Fixes #1594
Refs #2047

> [!IMPORTANT]
> **Supersedes #2096, which is now closed.** That PR raised the Linux watchdog from 30 s to 5 minutes; removing the watchdog outright subsumes it, so the truncation in #1594 is fixed here on its own. This branch still carries #2096's commit, but the repo squash-merges, so what lands on `main` is the net end state only — none of the constructor options, static exports or defensive `?.` that #2096 added.

## Problem

#2096 fixes the 30-second truncation. It does not fix the part that loses data.

A push-to-talk dictation that reaches its safety ceiling ends **while the trigger keys are still physically down** — that is what "held too long" means. The stop path was byte-for-byte the path a real release takes, so auto-paste went ahead and injected its shortcut into those held modifiers:

```js
// src/helpers/clipboard.js
ydotoolArgs = legacyYdotool ? ["key", "ctrl+v"] : ["key", "29:1", "47:1", "47:0", "29:0"];
//                                                        Ctrl↓   V↓      V↑     Ctrl↑
```

With physical Ctrl still down, nothing lands in the target app. The transcript is then dropped with no error. A user who dictated for four minutes loses all of it and is told nothing.

Raising the ceiling doesn't address this — at five minutes the failure is identical, just rarer. And it is not Linux-specific: Windows and **macOS compound (modifier+key) PTT** share the same `MAX_PUSH_DURATION_MS = 300000` ceiling and the same silent loss. macOS Globe/Fn, right-modifier and mouse-button PTT keep their own state in `main.js` closures with *no ceiling at all* — they never force-stop, so they are outside this fix by construction. That is pre-existing, but worth knowing, since Globe is the macOS default.

## Fix

**A stop that isn't a physical release is now a distinct terminal state.** `windowManager` reports it on one `dictation-force-stopped` channel; the renderer keeps the transcript in the clipboard and surfaces it through the existing dictation-error pill (Retry + View Transcript) instead of pasting it. Both forced paths carry a reason — the safety ceiling (`timeout`) and a mid-push hotkey or activation-mode change (`reset`).

Half of this seam already existed and was dead: `forceStopMacCompoundPush` has always sent `compound-ptt-force-stopped` with a reason, and **nothing anywhere listened for it** — it is absent from `preload.js` and from every renderer file. That channel is folded into the shared one and actually consumed now.

**macOS no longer hides the dictation panel on a forced stop while a recording is still being transcribed.** The recovery pill is a toast *inside that window*, and transcription finishes seconds after the stop, so hiding it rendered the pill invisibly and the fix was inert on that platform. Both the ordinary macOS release (`handleMacPushModifierUp`) and the Windows/Linux path already kept the panel up here; the forced path was the only one that didn't.

**The `linuxKeyManager` watchdog is removed.** This is a prerequisite, not cleanup: it enforced a second ceiling by *synthesizing a key-up*, which arrives as an ordinary release and is indistinguishable from the user letting go. It is also armed before `emit("key-down")` reaches `startWindowsPushToTalk`, so with both timers at 300000 ms it always fires first — it would have silently bypassed this fix on Linux entirely.

It was redundant regardless. `startWindowsPushToTalk` (shared by Windows and Linux) already arms `MAX_PUSH_DURATION_MS` in the layer that owns push state, which is why `windowsKeyManager` has never needed a watchdog. Its only reachable effect app-wide was ending a Linux push early: in tap mode key-up is ignored, and `dispatchNativeKeyUp` returns early for the `voiceAgent`/`translation`/`meeting` slots. Removing it also drops the constructor options, static exports and defensive `?.` that #2096 added — all of which existed only to serve that mechanism.

## Scope and judgement calls

- **The pill picks between two strings instead of always asserting the clipboard holds the text.** `keepInClipboard` swallowed every failure and the message asserted "Your transcription is on the clipboard" regardless, so a rejected or throwing write left the sentence false. It now reports whether the write landed. In practice that branch is near-unreachable today: `ClipboardManager.writeClipboard` returns `{ success: true }` unconditionally and the Wayland path swallows backend failures fire-and-forget, so only a rejecting IPC call reaches it. The guard is still correct and becomes live the moment `writeClipboard` reports honestly — that read-back is worth a follow-up, not this PR. The transcript action on the pill is the recovery path either way.
- **The clipboard write ignores `keepTranscriptionInClipboard`,** but only on the forced path where a paste was actually prevented. That setting governs whether ordinary dictations linger in the clipboard; this is error recovery, and the codebase already treats clipboard-as-fallback as acceptable regardless of it (`allowClipboardFallback` in `safePaste`). With auto-paste disabled the branch does not fire at all — nothing would have been injected, so there is no loss to report and no reason to touch a clipboard the user asked to leave alone.
- **`reset` is Windows/Linux only.** `resetWindowsPushState` touches `winPushState`; the macOS compound path has no equivalent reset.
- **`recordCleanupFailure` is not reported on the forced path.** It runs only after a successful paste, matching every other non-pasting branch — but here the transcript *is* delivered (to the clipboard), so a cleanup-model failure goes unmentioned. Left consistent rather than special-cased; happy to change it.
- **The latch is cleared at the top of `performStartRecording`,** before the start is validated, so a start that is later rejected still clears it. I could not construct a reachable path — both native listeners latch key-down so auto-repeat cannot re-fire it while held, and `isDictationProcessing()`/`canStartDictation` gate the rest — so it is left co-located with the sibling per-recording flags.
- **Wake-word dictations (`result.assistantConversation`) are deliberately not covered.** Their answer is delivered by `assistantResponseDelivery`, which captures its own target and revalidates before pasting, and delivery happens seconds later once the LLM has streamed. Bringing it into this fix would mean changing a subsystem this bug was never reported against. Worth a separate look.
- **#2047 is `Refs`, not `Fixes`.** This closes the data loss the reporter hit, by the remedy they proposed. The mechanism they diagnosed survives: `resources/linux-fast-paste.c` queries no modifier state on any backend (XTest, uinput, portal keysyms), while `resources/windows-fast-paste.c` has had `ReleaseModifiers()` since #438. `resources/linux-key-listener.c:348` fires `emit_key_up()` on a *partial* chord release — with `Ctrl+Shift+K`, letting go of Shift alone qualifies while Ctrl is still down — which dispatches `reason: "release"`, so this latch never sets and the paste proceeds into a held modifier. Follow-up: #2113 — port `ReleaseModifiers()` to Linux and macOS.
- `force-stop-dictation` can still pass `reason: "manual"`, but that handler is not exposed in `preload.js` and has no caller, so it is currently unreachable. The renderer lists the reasons it acts on rather than negating `"release"`, so wiring it up later cannot silently suppress a deliberate stop.

## Testing

- **`test/helpers/useAudioRecordingForceStop.test.js`** (new) — the renderer behaviour, on the existing `test/lib/rendererTestHarness`: a forced stop (both reasons) writes the clipboard, raises the error pill and injects **no** paste; an ordinary dictation still pastes; a forced stop with auto-paste disabled is inert; the latch does not leak into the next dictation; an unrecognised reason does not latch; and a clipboard write that fails — whether it reports failure or throws — is not described as a success.
- **`test/helpers/pushToTalkForceStop.test.js`** — a physical release reports no forced stop; the Windows/Linux and macOS ceilings each report one **before** the stop they cause, both driven through the real timer rather than by calling the force-stop directly; a forced stop leaves the window up; a settings-driven reset reports `reset`, not a release.
- **`test/helpers/linuxKeyManager.test.js`** (rewritten) — pins that a held key is *never* released by the listener at 30 s, 5 min or an hour. Rebuilt on the sibling `globeKeyManager.test.js` harness so it drives the real `setKeys`/spawn path, which is what lets the production escape hatches go away.

Each behaviour the fix introduces is mutation-checked against the **full** suite: deleting the renderer branch, neutering the IPC latch, widening the latch to accept any reason, removing the latch reset, dropping the `autoPasteEnabled` guard, restoring the macOS unconditional hide, deleting the kept `hideDictationPanel()` from the not-recording branch, dropping the `"timeout"` argument at either ceiling, and reverting the `reset` reason each fail tests. Sources were byte-compared against backups afterwards.

Full suite `3803 pass / 0 fail`; `lint`, `typecheck`, `i18n:check` and `prettier --check` all clean. i18n keys added to all 10 locales (the repo has `ja` as well as the 9 CLAUDE.md lists).

### Verified on real hardware

**Linux (Ubuntu 26.04 VM, real evdev, real `linux-key-listener`, real `linuxKeyManager.js`).** A 35-second held key, same hardware, old code vs new:

| | events observed |
|---|---|
| Old (30 s watchdog, as shipped in 1.9.2) | `key-down@7s`, **`key-up@37s`**, `key-up@42s` |
| This PR | `key-down@7s`, `key-up@42s` |

The old build synthesizes a release exactly 30 s after key-down and the real release lands 5 s later against an already-stopped recording — the "29.7 s segment followed ~2 s later by a short one" #1594 reported from stored audio durations. This PR leaves one release: the real one.

**macOS.** Real Electron + the real `preload.js` contextBridge + real IPC (the layer every unit test stubs): `onDictationForceStopped` is genuinely exposed, payloads arrive in order with `reason` intact, and the returned disposer really unsubscribes — the hook's cleanup depends on it. Driving the real `windowManager` end to end, a forced stop reaches the renderer with its reason while a physical release stays silent on that channel but still stops dictation. The app also boots clean.

Still not covered: a human holding a real hotkey through the full 5-minute ceiling in the packaged app with live transcription. The 5-minute timer itself is pinned in unit tests; what is unverified is the transcription-and-pill sequence that follows it.

### Known, not addressed

Five locales (de, es, it, pt, ru) say "the key" rather than each file's established word for *hotkey*, and PTT hotkeys are often compound. Carried over unchanged from the first draft of these strings; left alone rather than risk agreement errors in languages I can't verify. Same for the French "raccourci … enfoncé" collocation.


Chadpiha (approved):
Approved.

Removing the watchdog is the right call rather than raising it. `startWindowsPushToTalk` is shared with Linux and has armed `MAX_PUSH_DURATION_MS = 300000` all along, so the watchdog was a second, shorter ceiling on top of the real one — and because it ends a push by synthesizing a `key-up`, keeping it would have made a forced stop indistinguishable from a release and bypassed this fix on Linux entirely.

Verified locally on the updated branch:

- Force-stop path traced end to end — no paste is injected, `_restoreClipboardAfterDelay` never arms so the transcript survives in the clipboard, and History is written unconditionally before any delivery branch.
- Mutation-tested: 16 of 17 targeted reverts are caught, including the watchdog re-add and both ceiling reasons.
- `npm test` 4082 tests / 3885 pass / 0 fail; lint, `format:check` and `i18n:check` clean. The 4 lint warnings are pre-existing in a file this PR doesn't touch.
- The ENODEV / SYN_DROPPED stuck-key cases are still capped at the same 300 s by `windowManager`, and now end with `reason: "timeout"` — so the transcript is preserved rather than lost. Strictly better than before.

Two notes, neither blocking:

- `Refs #2047` rather than `Fixes` is correct — `linux-key-listener.c:348` fires `emit_key_up()` on a partial chord release, which dispatches `reason: "release"`, so the latch never sets and the paste still lands in a held modifier. Follow-up to port `ReleaseModifiers()` from `windows-fast-paste.c` to Linux and macOS.
- Squash-merging, so please keep the `Co-authored-by: Henry Su <henrysu4707@gmail.com>` trailer — #2096 diagnosed the watchdog first and this supersedes it.

PR #2103, opened 2026-09-10
fix(transcription): normalize underscore, casing, and whitespace in base language resolution

hsusul:
## Problem

In `src/utils/languageSupport.ts` and `src/helpers/transcriptionRoute.ts`, language resolution relied on splitting solely by hyphen:

```ts
export function getBaseLanguageCode(language: string | null | undefined): string | undefined {
  if (!language || language === "auto") return undefined;
  return language.split("-")[0];
}
```

And in `src/helpers/transcriptionRoute.ts`:

```ts
const language =
  request?.effectiveLanguage ??
  (!s.preferredLanguage || s.preferredLanguage === "auto"
    ? undefined
    : s.preferredLanguage.split("-")[0]);
```

When system locales, environment variables, or user settings format languages with underscores (e.g. `zh_CN`, `pt_BR`, `de_DE`, `es_ES`, `fr_FR`—the standard POSIX/Linux format and common in Windows exports), with whitespace (`"  es_ES  "`), or uppercase (`"EN-US"`):
1. `getBaseLanguageCode("pt_BR")` and `getBaseLanguageCode("zh_CN")` return `"pt_BR"` and `"zh_CN"`, failing to extract the base language code (`"pt"`, `"zh"`).
2. In `transcriptionRoute.ts`, `XAI_STT_LANGUAGES.has(language)` check fails because `XAI_STT_LANGUAGES` only contains 2-letter codes (`"pt"`, `"zh"`, etc.). As a result, xAI drops the language parameter entirely and disables language-specific text normalization (ITN).
3. `getBaseLanguageCode` passes `"pt_BR"` or `"zh_CN"` directly to cloud STT APIs (OpenAI, Groq, Mistral, Corti, Gemini), which require standard 2-letter ISO 639-1 language codes, resulting in 400 Bad Request errors or degraded transcription.
4. If `request?.effectiveLanguage` is provided to `resolveTranscriptionRoute`, it was passed through raw without extracting the primary subtag or trimming.

## Fix

1. **Robust base language code resolution in `src/utils/languageSupport.ts`**:
   - Trim whitespace and normalize underscores to hyphens before splitting.
   - Lowercase the extracted base language subtag so mixed/uppercase inputs (`"EN-US"`, `"ZH_CN"`) produce valid ISO 639-1 codes (`"en"`, `"zh"`).
   - Add `with { type: "json" }` attribute to `languageRegistry.json` import so the utility can be cleanly loaded across renderer and Node/main-process ESM contexts.
2. **Standardize language resolution in `src/helpers/transcriptionRoute.ts`**:
   - Use `getBaseLanguageCode(request?.effectiveLanguage ?? s.preferredLanguage)` so both effective language overrides and preferred language settings are uniformly normalized.
   - Retains proper language passing for xAI, OpenAI, Groq, Mistral, Corti, Gemini, self-hosted, and managed enterprise routes.

## Testing

- Added `test/utils/languageSupport.test.js`:
  - Verified simple codes (`en`, `es`, `zh`, `pt`).
  - Verified hyphenated tags (`en-US`, `zh-CN`, `pt-BR`, `es-419`).
  - Verified underscore-separated locales (`zh_CN`, `pt_BR`, `de_DE`, `es_ES`, `fr_FR`, `en_GB`).
  - Verified case normalization (`EN`, `EN-US`, `ZH_CN`, `Pt-Br`).
  - Verified whitespace trimming (`"  en  "`, `"  zh_CN  "`, `"\tes-ES\n"`).
  - Verified empty, null, undefined, auto, and non-string inputs return `undefined`.
- Extended `test/helpers/transcriptionRoute.test.js`:
  - Verified underscore and padded locale normalization for route language and xAI support (`pt_BR` -> `pt` retained for xAI; `  zh_CN  ` -> `zh` for Groq; `auto` -> `undefined`).
  - Verified all 19 tests pass under both `node --test` and `node --import tsx --test`.

### DIFF
--- a/src/helpers/transcriptionRoute.ts
+++ b/src/helpers/transcriptionRoute.ts
@@ -27,6 +27,7 @@ import {
   type TranscriptionProviderBaseUrl,
 } from "../services/transcriptionBaseUrl.ts";
 import type { ManagedEnterpriseRequestContext } from "../types/enterpriseIdentity.ts";
+import { getBaseLanguageCode } from "../utils/languageSupport.ts";
 
 const BYOK_FILE_SIZE_LIMIT = 25 * 1024 * 1024;
 
@@ -212,11 +213,7 @@ export function resolveTranscriptionRoute({
 }: TranscriptionRouteInput): TranscriptionRoute {
   const s = settings || {};
   const managed = policy?.status === "managed";
-  const language =
-    request?.effectiveLanguage ??
-    (!s.preferredLanguage || s.preferredLanguage === "auto"
-      ? undefined
-      : s.preferredLanguage.split("-")[0]);
+  const language = getBaseLanguageCode(request?.effectiveLanguage ?? s.preferredLanguage);
 
   // Managed enterprise STT outranks every personal setting; the resolution is
   // already policy-checked where it was computed (enterpriseIdentityStore /
--- a/src/utils/languageSupport.ts
+++ b/src/utils/languageSupport.ts
@@ -1,4 +1,4 @@
-import registry from "../config/languageRegistry.json";
+import registry from "../config/languageRegistry.json" with { type: "json" };
 
 function buildLanguageSet(key: "whisper" | "assemblyai"): Set<string> {
   const set = new Set<string>();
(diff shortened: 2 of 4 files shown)


gabrielste1n:
Thanks for this, @hsusul. We're closing it because the inputs it normalizes can't reach `getBaseLanguageCode`.

`preferredLanguage` is only ever written from the registry-backed language pickers (Settings and onboarding) and the `zh` to `zh-CN` migration, and `src/config/languageRegistry.json` has no underscore, uppercase or padded codes; nothing reads a system locale or environment variable into this setting. Both producers of `effectiveLanguage` already pass base codes. So `pt_BR`, `AUTO` and padded values have no way in, and we keep helpers trusting inputs the app itself generates.

The change would also pull `languageSupport.ts` and the registry JSON into the main process through `transcriptionRoute.ts`, which needs the codebase's only `with { type: "json" }` import attribute. That isn't something we want without a concrete need. The tests are correct for the inputs they feed; nothing produces those inputs. If a real path that stores a non-registry code turns up, please open an issue with the repro.

PR #2104, opened 2026-09-10
fix(audio): guard nullish and empty labels in microphone detection and resolution

hsusul:
## Problem

In `src/utils/audioDeviceUtils.ts`:

```ts
export function isBuiltInMicrophone(label: string): boolean {
  const lowerLabel = label.toLowerCase();
```

And in `src/helpers/microphoneSelection.js`:

```js
export function normalizeMicrophoneLabel(label = "") {
  return String(label)...
```

1. **Uncaught TypeError on devices without labels**:
   `isBuiltInMicrophone(candidate.label)` is called across `src/helpers/microphoneSelection.js` (line 83) during built-in microphone resolution and in `src/components/ui/MicrophoneSettings.tsx` (line 65). When audio permissions have not yet been granted, or when virtual/USB/headless audio devices enumerate with `label: ""` or without a `label` property (`undefined` or `null`), passing `candidate.label` throws an unhandled:
   `TypeError: Cannot read properties of undefined (reading 'toLowerCase\)`
   This crashes microphone resolution and aborts settings initialization.
2. **String coercion of `null`**:
   `normalizeMicrophoneLabel(null)` defaulted only for `undefined` via ES6 default parameters, coercing `null` via `String(null)` to `"null"`, which compares as a literal device name rather than an empty string.

## Fix

1. **Nullish & Empty Safety in `src/utils/audioDeviceUtils.ts`**:
   - Guard `isBuiltInMicrophone(label?: string | null)`: return `false` immediately if `label` is not a non-empty string.
2. **Safe Label Normalization in `src/helpers/microphoneSelection.js`**:
   - Check `typeof label !== "string" || !label` and return `""` immediately in `normalizeMicrophoneLabel`.
   - Add explicit module extensions on internal imports (`.ts` and `.js`) so the module loads cleanly across both bundled and Node test environments.

## Testing

- Updated `test/utils/audioDeviceUtils.test.js`:
  - Added test case verifying `isBuiltInMicrophone` returns `false` and does not throw for `undefined`, `null`, `""`, `"   "`, and non-string values.
- Extended `test/helpers/microphoneSelection.test.js`:
  - Verified built-in mode safely skips devices with `null`, `undefined`, or empty labels without throwing.
  - Verified built-in mode finds the real built-in microphone in the presence of unlabeled devices.
  - Verified `normalizeMicrophoneLabel` safely returns `""` for `null`, `undefined`, `""`, and non-string inputs.
  - Verified all 12 tests pass under both `node --test` and `node --import tsx --test`.

### DIFF
--- a/src/helpers/microphoneSelection.js
+++ b/src/helpers/microphoneSelection.js
@@ -1,5 +1,5 @@
-import { isBuiltInMicrophone } from "../utils/audioDeviceUtils";
-import { resolveMicDeviceSelection } from "./micDeviceSelection";
+import { isBuiltInMicrophone } from "../utils/audioDeviceUtils.ts";
+import { resolveMicDeviceSelection } from "./micDeviceSelection.js";
 
 export const MICROPHONE_SELECTION_MODES = ["system", "built-in", "specific"];
 
@@ -14,6 +14,9 @@ export function getMicrophoneSelectionMode(settings = {}) {
 }
 
 export function normalizeMicrophoneLabel(label = "") {
+  if (typeof label !== "string" || !label) {
+    return "";
+  }
   return String(label)
     .normalize("NFKC")
     .replace(/^\s*(?:default|communications)\s*-\s*/i, "")
--- a/src/utils/audioDeviceUtils.ts
+++ b/src/utils/audioDeviceUtils.ts
@@ -7,7 +7,10 @@
  * Determines if a microphone device is a built-in device based on its label.
  * Works across macOS, Windows, and Linux platforms.
  */
-export function isBuiltInMicrophone(label: string): boolean {
+export function isBuiltInMicrophone(label?: string | null): boolean {
+  if (typeof label !== "string" || !label.trim()) {
+    return false;
+  }
   const lowerLabel = label.toLowerCase();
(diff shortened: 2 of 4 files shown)


gabrielste1n:
Thanks for the tests and the write-up, @hsusul. We're closing this one: the inputs it guards against can't reach these functions.

Every caller of `isBuiltInMicrophone` and `normalizeMicrophoneLabel` gets its devices from `navigator.mediaDevices.enumerateDevices()` (`src/components/ui/MicrophoneSettings.tsx` and `resolvePreferredMicrophone` in `src/helpers/microphoneSelection.js`), and `MediaDeviceInfo.label` is a non-nullable string: `""` before permission is granted, never `undefined` or `null`. Both functions already handle `""` without throwing, and the native default name is validated in `src/helpers/systemDefaultMicrophone.js` before it gets anywhere near them. We keep internal helpers free of guards for values the browser already guarantees, which is the same policy behind the earlier guard-only PRs.

The branch would also need a rebase, since #2180 changed the same file and its tests. If you ever see a device enumerate without a label in a real session, a debug log in a new issue would be very welcome.

Issue #1764, opened 2026-08-21
OpenRouter speech-to-text models (e.g. openai/gpt-4o-transcribe) never appear in the model picker — /api/v1/models queried without output_modalities=transcription


carlos-reynosa:
**Describe the bug**

When OpenRouter is added as a provider, only chat/text models appear in the Speech-to-Text model picker. Speech-to-text models such as `openai/gpt-4o-transcribe`, `openai/whisper-1`, `openai/gpt-4o-mini-transcribe`, `google/chirp-3`, and `deepgram/nova-3` are missing, so OpenRouter cannot be used for transcription at all.

The root cause is that OpenWhispr queries `https://openrouter.ai/api/v1/models` to populate the picker. OpenRouter intentionally excludes STT models from that default endpoint — they are only returned when filtering by output modality:

```
GET https://openrouter.ai/api/v1/models?output_modalities=transcription
```

Per OpenRouter's docs: "STT model IDs don't show up in the default `/api/v1/models` catalog. That's expected, because transcription is an output modality you filter for."

**To Reproduce**

1. Go to Settings → AI Models → Speech-to-Text
2. Add OpenRouter as the provider with a valid API key
3. Open the Speech-to-Text model picker
4. Search for `gpt-4o-transcribe` (or any transcription model)
5. See that only chat/text models are listed — no transcription models appear

**Expected behavior**

The Speech-to-Text model picker should list OpenRouter's STT models (querying `?output_modalities=transcription` when populating the STT picker). For example, `openai/gpt-4o-transcribe` should be selectable and produce a working transcription.

**Screenshots**

N/A

**Desktop (please complete the following information):**

- OS: Linux (Ubuntu)
- Browser: N/A (desktop app)
- Version: 1.8.3

**Smartphone (please complete the following information):**

- Device: N/A
- OS: N/A
- Browser: N/A
- Version: N/A

**Additional context**

`openai/gpt-4o-transcribe` is a valid OpenRouter STT model (OpenAI's GPT-4o-based transcription, $2.50/$10 per 1M tokens). It is served via the dedicated endpoint `POST https://openrouter.ai/api/v1/audio/transcriptions` and should be selectable as a Speech-to-Text model, but it does not appear because of the missing filter above.

henmedia:
shouldn't OpenRouter be made available as an own provider tab rather that as custom model?

PR #2105, opened 2026-09-10
feat(transcription): add OpenRouter as a speech-to-text provider

henmedia:
## Why

OpenRouter is already a first-class provider for reasoning models, but the Speech-to-Text picker never listed it. Reaching `openai/gpt-transcribe` or `deepgram/nova-3` with an OpenRouter key meant selecting **Custom** and typing the base URL and the vendor-prefixed model id by hand.

#1764 reports OpenRouter's STT models as missing because `/api/v1/models` is queried without `output_modalities=transcription`. That is exactly right — but not in the place the report points at. The transcription **picker** builds its tabs from the static registry and fetches no catalog at all; OpenRouter simply had no entry there. The **connection test** does fetch the catalog, and there the missing filter is the whole bug. This PR fixes both halves.

## What

Little new plumbing was needed — the key slot (`OPENROUTER_API_KEY`, `openrouterApiKey`), the IPC accessors and the brand icon all exist for the reasoning side already.

- **Registry**: four curated STT models, `openai/gpt-transcribe` first so a new selection lands on it. None carry `streaming`, so OpenRouter stays out of the streaming-only scopes (meeting, Note Recording) — it brokers batch HTTP transcription and has no realtime socket. A test pins that.
- **Route**: OpenRouter joins Groq in a small base-URL table rather than a third ternary, and resolves to `https://openrouter.ai/api/v1/audio/transcriptions` with `keyRef: "openrouter"`.
- **Connection test**: a transcription-scope test now fetches `/v1/models?output_modalities=transcription`. Without it the catalog returns ~440 text models and no transcription ones, so every OpenRouter STT selection failed with `modelNotFound` — and since the onboarding step's Continue button is gated on a passing test, the provider could be selected but never committed. `scope` already travels in the test config, so no caller changed; the reasoning scope keeps the plain catalog, and Deepgram/AssemblyAI keep the `MODEL_LIST_UNVERIFIABLE` path they arrived with.
- **Model ids are a shortlist, not a whitelist**: every OpenRouter id is vendor-prefixed, so the slash is what tells a real selection apart from another provider's leftover. Ids outside the four listed pass through deliberately — at request time in `resolveByokModel`, and in the store's per-provider model memory, which otherwise resets a stored id on the next tab switch. OpenRouter's catalog moves faster than our releases.
- **Picker**: model rows carry the upstream vendor's icon (`openai/…`, `google/…`, `mistralai/…` already have aliases in `providerIcons`), falling back to OpenRouter's own mark for vendors we ship no icon for.
- **Existing Custom endpoints pointed at OpenRouter keep working unchanged**: `reconcileCloudProviderSelection` returns early for a stored `custom` provider, so nobody is migrated out from under their configured key slot. Pinned by test.
- **Docs**: the `openrouter.ai` row in `docs/network-allowlist.md` said "selected as a reasoning provider"; it is now reachable for transcription too.
- **i18n**: four description keys across all eleven locales.

### Relationship to the providers that landed since this branch opened

Rebased onto `main` after #2076, #2077, Deepgram/AssemblyAI and the Arabic locale merged.

`gpt-transcribe` on the OpenAI tile (#2076) and this PR are independent: that one is OpenAI-direct, this one is the same model through an OpenRouter key.

Deepgram is more interesting. As a direct provider it is realtime-only — `STREAMING_ONLY_PROVIDERS` makes `resolveTranscriptionRoute` refuse file uploads and retries for it. Brokered through OpenRouter, `deepgram/nova-3` is plain batch HTTP, so it covers uploads and retries that the direct provider cannot. The two complement each other rather than overlap.

## Verified

Run end to end on Windows 11 against the live API, through the packaged onboarding flow with a real OpenRouter key:

| Step | Result |
|---|---|
| Provider dropdown | OpenRouter listed, with its icon |
| Default selection | `GPT Transcribe` preselected; all four models offered |
| Connection test | passes for `openai/gpt-transcribe` (failed with `modelNotFound` before the connection-test fix — this is how that bug was found) |
| Continue | commits the selection and advances the flow |

`typecheck`, `lint` (both roots), `prettier --check` on every touched file, `i18n:check` and `build:renderer` are clean.

Six new tests, all passing:

| Test | Pins |
|---|---|
| `openrouter routes to its own base and keeps vendor-prefixed model ids` | endpoint, `keyRef`, size cap; an unlisted id survives; a bare id degrades to the default |
| `a Custom endpoint pointed at OpenRouter stays on the custom key slot` | no silent migration for existing hand-configured users |
| `OpenRouter remembers ids outside the shipped shortlist` | full round-trip through `switchCloudTranscriptionProvider` |
| `no OpenRouter model advertises streaming` | the batch-only fence |
| `a transcription test asks OpenRouter for its speech-to-text catalog` | the modality filter, scoped to transcription and to OpenRouter only |
| `an OpenRouter transcription model passes the availability check` | the whole test path, including the URL actually fetched |

`openrouter` also joins `PREFIX_VALIDATED_PROVIDERS`, so the two existing registry/resolver agreement guards now cover it.

Full suite on this branch: 3762 tests, 3355 pass, **198 fail**. Unmodified `main` on the same machine: 3756 tests, 3349 pass, **198 fail** — the same 44 files, byte-identical list, all Electron- and GPU-dependent (`Cannot read properties of undefined (reading 'isPackaged')`), failing because `npm ci` skips install scripts locally. Net effect: +6 tests, +6 passing, no change in failures.

## Notes for review

- The eleven locale strings are my own, including ar, ja, ru, zh-CN and zh-TW. They follow the terseness of the neighbouring entries but have not been checked by native speakers.
- The four listed models are a judgement call from OpenRouter's 20 transcription models, chosen for coverage rather than completeness, and sized to match how short every other provider's list is. Happy to trim or extend.
- A dynamic catalog fetch for the picker would suit OpenRouter's breadth better, but it needs caching, offline behaviour and error states, so it seemed worth keeping separate from making the provider reachable at all.
- **Out of scope, noticed while testing**: the onboarding *assistant* step reads `cloudProviders`, which has no OpenRouter entry, so OpenRouter cannot be picked there — even though the Settings reasoning selector has had an OpenRouter tab all along. That predates this PR and belongs to the LLM side, so I left it alone; happy to file it separately.

Refs #1764


henmedia:
Rebased onto `main` — this now sits on top of #2076, #2077 and the Deepgram/AssemblyAI providers, and the Arabic locale is covered. No conflicts.

Two things that may help triage:

**The connection-test fix is the actual #1764 bug.** The report blames the model picker, but that builds its tabs from the static registry and fetches no catalog at all. The connection test does fetch one, and `/v1/models` returns OpenRouter's ~440 text models and none of its transcription ones — so every OpenRouter STT selection failed with `modelNotFound`, and since the onboarding step gates Continue on a passing test, the provider could be selected but never committed. Fixed by filtering on `output_modalities=transcription`, for transcription-scope tests only; Deepgram and AssemblyAI keep the `MODEL_LIST_UNVERIFIABLE` path they arrived with.

**Deepgram doesn't overlap with this.** As a direct provider it is realtime-only — `STREAMING_ONLY_PROVIDERS` makes the resolver refuse file uploads and retries. Brokered through OpenRouter, `deepgram/nova-3` is plain batch HTTP, so it covers exactly what the direct provider cannot.

Verified end to end on Windows 11 through the packaged onboarding flow against the live API: provider listed, `GPT Transcribe` preselected, connection test green, selection committed. Full suite matches unmodified `main` on the same machine (same 44 failing files, all Electron/GPU-dependent locally); +6 tests, +6 passing.

Workflows are still sitting at `action_required` — first contribution from a fork, so they need a maintainer to start them. Happy to rebase again if anything lands in the meantime.

ab-qf:
Now with microsoft/mai-transcribe-2, this would be very interesting.

henmedia:
Thanks — I tried exactly that, and it does not work from this app today, so I left it out of the list.

Dictation POSTs OpenAI-style multipart to `https://openrouter.ai/api/v1/audio/transcriptions`. With `microsoft/mai-transcribe-2` that comes back as:

```
{"error":{"message":"Provider returned 400","code":400}}
```

The request was as minimal as it gets — two fields:

```
file:  audio.webm  (audio/webm;codecs=opus, 10.2 s, 164 KB)
model: microsoft/mai-transcribe-2
```

No `language`, no `prompt`. The same key, the same endpoint and the same recording work with `openai/gpt-transcribe`, so this is specific to the model rather than to my setup. The failed call doesn't appear under Activity either, so the upstream message isn't visible there.

From outside I can't tell which of these it is:

- the Azure endpoint behind it doesn't accept WebM/Opus — OpenRouter's examples for this model use `wav` and `mp3` throughout, and the guide notes Azure forwards fields as-is, so an unsupported value surfaces as a provider error; or
- it doesn't serve the multipart path at all and wants the JSON `input_audio` body.

Either way it is out of reach for a client that records WebM and posts multipart, so the shortlist stays at the four models I could actually transcribe with. Worth revisiting though — at $0.10/hour against $0.0045/minute it would be the cheapest entry by a wide margin, and it is the strongest multilingual one on paper. If anyone gets it to answer through multipart, I'm glad to add it.

PR #2116, opened 2026-09-11
fix(macos): use the arm64-native ONNX Runtime slice for sherpa-onnx on Apple silicon (~3x faster Parakeet/Orukeet)

Zeljaa:
## Summary

On Apple silicon, every Parakeet-family model in OpenWhispr (Orukeet, Parakeet TDT v3, Nemotron, Cohere) runs ~3x slower than it should. The sherpa-onnx macOS archive we download bundles `libonnxruntime.1.27.0.dylib` as a universal2 build, and that build's arm64 slice runs INT8 kernels far slower than the arm64-only build of the **same ONNX Runtime version** from the same maintainer ([csukuangfj/onnxruntime-libs v1.27.0](https://github.com/csukuangfj/onnxruntime-libs/releases/tag/v1.27.0), the zip sherpa-onnx's own `cmake/onnxruntime-osx-arm64.cmake` pins). sherpa-onnx's `osx-arm64` archive is lipo-thinned from the universal2 build and is equally slow, so switching archives does not help.

On a macOS host, `scripts/download-sherpa-onnx.js` now downloads that zip, verifies its SHA-256, and rebuilds the bundled dylib with the arm64 slice from it. The x86_64 slice is kept byte-for-byte, so the file stays universal2 and Intel Macs are unchanged. Linux and Windows paths are untouched. The install marker records the new slice so existing dev installs re-extract instead of keeping the slow library.

## Measurements

Apple M5, OpenWhispr `main` @ 4b4f290c, 11 s JFK clip, real `ParakeetManager.transcribeLocalParakeet` under Electron, warm median of 10 calls, default thread count:

| Model | Shipped | This PR |
|---|---:|---:|
| `orukeet-v0.1.0` | 605 ms | 173 ms |
| `parakeet-tdt-0.6b-v3` | 665 ms | 230 ms |

Transcripts identical in every run. Encoder-only cross-check with the ONNX Runtime C++ API (Orukeet `encoder.int8.onnx`, 4 threads): universal2 dylib 583 ms, sherpa `osx-arm64` archive 601 ms, arm64-only zip 152 ms, pip `onnxruntime` 1.27.0 arm64 153 ms.

## Notes

- Runs for both macOS targets because they install the same file, so an `--all` run cannot overwrite the fix. Non-macOS hosts have no `lipo` and keep the upstream library.
- If a future sherpa-onnx bump ships a different ONNX Runtime version, the script fails asking for `MACOS_ARM64_ONNXRUNTIME` to be updated, rather than splicing mismatched versions.
- The arm64 zip library's deployment target is macOS 15.5, same as the bundled one.

## Verification

- `--current --force` and `--force`: library is `x86_64 arm64`; arm64 slice hash equals the zip's library; x86_64 slice `cmp`-identical to upstream; marker carries `onnxRuntime: "arm64-1.27.0"`; rerun without `--force` reports "Already exists".
- `test/scripts/downloadSherpaOnnx.test.js` 10/10. `npm test`: 3970 pass, one failure (`settingsStore imports without a bare localStorage global`) that fails identically on unmodified `main`. `npm run format:check` clean.

The long-term fix is for sherpa-onnx to bundle a properly built ONNX Runtime in its macOS archives; this step can be removed once that lands.

### DIFF
--- a/scripts/download-sherpa-onnx.js
+++ b/scripts/download-sherpa-onnx.js
@@ -2,6 +2,7 @@
 const fs = require("fs");
 const path = require("path");
 const { execFileSync } = require("child_process");
+const crypto = require("crypto");
 const {
   cleanupFiles,
   downloadFile,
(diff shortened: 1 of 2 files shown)


gabrielste1n (approved):
Verified end to end on Apple silicon: the pinned zip hash matches, both slices keep the 15.5 deployment target, install name, dependencies and exports (including the CoreML provider) are unchanged, the x86_64 slice is byte-identical to upstream outside the signature, and all three shipped sherpa binaries load the rebuilt library. Driving the real offline WS server with parakeet-tdt-0.6b-v3 went from a 579 ms to a 211 ms warm median with identical transcripts, tokens and log-probs. Full test suite passes locally. Thanks for the thorough fix and measurements.

PR #1969, opened 2026-08-31
fix(ai): handle multipart responses and thinking parts in Gemini inference provider (#1968)

hsusul:
## Description

Fixes Gemini inference provider response extraction to correctly join multi-part text responses and skip internal thought parts (from Gemini thinking).

### Root Cause
In `src/services/ai/inferenceProviders/gemini.ts`, `geminiProvider` directly accessed index 0 of the parts array (`candidate.content.parts[0].text`).
This caused:
1. Truncated responses when text was returned across multiple parts (`parts[0]` was kept and subsequent parts were dropped).
2. Internal thought blocks leaking when Gemini returned thinking parts first (`parts[0]` containing `thought: true`).
3. Spurious "Gemini returned empty response" errors when `parts[0]` was a non-text block while valid text existed in subsequent parts.

### Fix
Created `extractGeminiText` in `src/helpers/geminiResponse.js` to filter out thought parts (`!part.thought`) and join all valid text parts, returning trimmed text or `null` if no valid text parts exist. Updated `geminiProvider` to use `extractGeminiText`.

## Verification
- Added comprehensive unit tests in `test/helpers/geminiResponse.test.js` testing single text parts, multiple text parts, thought-part skipping, empty parts, and malformed candidates.
- Ran unit tests: `node --test test/helpers/geminiResponse.test.js` (7/7 passed)
- Ran linter: `npm run lint` (clean)
- Ran i18n check: `npm run i18n:check` (clean)

Closes #1968

PR #2130, opened 2026-09-11
fix(cleanup): fail truncated cleanup replies instead of pasting them

boseq:
## Summary

- Refs #2091. The reported symptom (Gemini 3 splitting its reply across `parts` while only `parts[0]` was read) was fixed in #1969 and shipped in v1.10.0. This PR closes the remaining path to the same symptom: a cleanup reply cut off at the output token limit was pasted as if it were complete, because providers reject truncation only when `requireCompleteOutput` is set, and only selection edits set it.
- `resolveReasoningRoute` now builds one cleanup config — shared by the cleanup route and the translation chain's cleanup step — that sets the flag, and the history-retry cleanup in `ControlPanel` sets it too. A truncated cleanup then keeps the raw transcript: the cleanup route pastes it, the translation chain translates it, and both raise the "Cleanup failed" toast.
- That toast's title is the provider's error message, so the truncation and empty-reply errors now carry i18n keys (`hooks.audioRecording.errorDescriptions.cleanupTruncated` and `cleanupEmptyReply`), added to all 11 locales. Renderer providers share `truncatedOutputError()` / `emptyOutputError()` from `chatRequestBody.ts`, including `ReasoningService.callChatCompletionsApi` for Groq, LAN, Corti and OpenCode; the IPC-bridged providers (Anthropic, enterprise, local llama) attach the key in the main process and hand it back with their error result, where `anthropic.ts` and `local.ts` re-attach it the way `enterprise.ts` already did.

## Behavior changes

- A truncated cleanup reply now fails instead of pasting partial text. That is the point of the fix.
- Retrying cleanup on a history row used to save the truncated text. With the flag it saves nothing, so both history-retry paths now report the failure (title "Cleanup failed", description from the error's key) instead of leaving the success toast as the only feedback.
- An empty OpenAI or Tinfoil cleanup reply now shows the cleanup-failed toast (the raw transcript is still pasted) instead of silently pasting it. This matches Gemini and Anthropic, which already threw.
- Gemini's "reached token limit before generating response" (the cap hit before any text arrived) now carries the truncation key, so its toast is translated as well.

## Verified in a real build

An isolated dev build against a throwaway profile, pushing a fixed 660-character transcript through the real cleanup pipeline:

| Run | Base code | This branch |
| --- | --- | --- |
| Mock model that always stops at its limit | pasted 83 of 660 characters, cut mid-sentence, no warning | full transcript, plus the "Model output was truncated" toast |
| Real Gemini 3 Flash capped at 64 output tokens | pasted 293 characters ending "On Thursday, we have the design review" | full transcript, plus the toast |
| Real Gemini 3 Flash, normal limits | — | cleaned normally (660 to 643 characters), no toast |

## Not in this PR

- OpenWhispr Cloud cleanup: truncation there is handled server-side, and its providers are Groq/OpenAI/OpenRouter/Parasail, with no Gemini.
- OpenAI, Anthropic, Groq/OpenRouter/Tinfoil and enterprise output caps: still sized from the input, so a heavy-thinking model on those providers now fails visibly instead of truncating. Gemini's cap is removed in this PR (second commit); the other providers are a separate change.
- Note actions (`actionProcessingStore`) still save model output without this flag, so a truncated note can be saved silently.
- The history-retry cleanup call still doesn't pin `temperature: 0` or `inferenceScope` the way dictation cleanup does. That drift predates this PR.

## Test plan

- [x] `node --import tsx --test` across the cleanup and provider suites: 84 pass, including new coverage for the truncation and empty-reply keys (`openaiProviderResponse.test.js`, which now also asserts the chat endpoint is really reached), the translation chain reporting a truncated cleanup (`audioManagerCleanupFallback.test.js`), and both cleanup configs requiring complete output (`cleanupRouteConfig.test.js`, renamed from `cleanupRouteTemperature.test.js`)
- [x] `npm run typecheck`, Prettier and ESLint clean on the changed files
- [x] `npm test` locally: 4136 pass, 125 fail. Every failure is the pre-existing `ERR_DLOPEN_FAILED` on `better_sqlite3.node` in this checkout (a Windows DLL left by a cross-build) — the failure count equals the dlopen count — and none are in the touched areas
- [x] Manual: isolated dev build with the mock model and with real Gemini, as in the table above


gabrielste1n:
Rebased onto `main` and pushed to this branch (maintainer edits are enabled), so please reset your local branch before adding commits: `git fetch && git reset --hard origin/fix/2091-cleanup-truncated-output`.

**Conflict resolution against #2177 / #2142 / #2155**

- `emptyResponseError` from #2177 is now the single empty-reply path: it returns the keyed `truncatedOutputError` / `emptyOutputError`, so OpenAI, Tinfoil and the chat-completions transport don't need a parallel `requireCompleteOutput` branch. `truncatedOutputError` takes an optional message so Gemini's cap-before-text case and the "ran out of output tokens" case reuse it.
- Local llama follows the #2142 convention: `llamaServer` rejects with `code: "OUTPUT_TRUNCATED"`, `modelManagerBridge` preserves that code (it rewrapped everything except `CONTEXT_TOO_LARGE`, so a key set in the main process never reached the renderer), and `buildLocalInferenceError` maps it to the cleanup key. Covered by a bridge test and a `localInferenceError` test.
- `ipcHandlers` uses two module constants for the Anthropic/enterprise keys instead of repeated literals.

**Second commit: no output cap on Gemini**

Gemini bills thinking against `maxOutputTokens`, so the text-sized cap (2000 floor, 8192 ceiling) was being spent on thinking before the cleaned dictation arrived. That is the root cause behind #2091; failing the reply makes it visible, this makes it rare. The provider now sends no `maxOutputTokens` at all: the model's own limit and the request timeout bound the reply, and a caller's pinned budget no longer needs the #2142 special case. The unused Gemini token limits are dropped. Two request-capturing tests in `geminiProviderResponse.test.js` replace the #2142 cap test.

Typecheck, Prettier and ESLint are clean; 67 tests across the touched suites pass.

gabrielste1n (approved):
Rebased onto main, reconciled with #2177/#2142, Gemini cap removed. Typecheck, lint and 67 tests pass locally; quality gates green in CI.

PR #2131, opened 2026-09-11
fix(chat): render GFM tables in assistant markdown

greatcoat:
**Problem.** An assistant answer that contains a markdown table renders in chat as one run-on paragraph of literal pipe characters, separator row included. Headings, bold and lists in the same answer render fine.

**Fix.** Pass `remark-gfm` to `react-markdown` and map the table elements explicitly in the renderer's existing style vocabulary, inside a horizontal-scroll wrapper. Column alignment is forwarded through `style`. Three small commits ride along: a test that pins the GFM extras and the unchanged URL sanitiser, the removal of three Tailwind classes that emit no CSS, and a review fix that aligns header cells with `text-start` instead of `text-left`.

## Problem

Ask the chat agent for anything it answers with a table. The stored message is a well-formed GFM table with correct line breaks, so the model and persistence are fine; rendering that exact stored message through `react-markdown` gives zero `<table>` elements and 68 literal pipes across two tables.

**This is pre-existing, not a 1.10.0 regression.** `remark-gfm` has never been in this repository's dependencies at any point in its history. Tables have never rendered in any released version.

## Root cause

`react-markdown` follows CommonMark, and CommonMark has no table syntax, so a table is parsed as an ordinary paragraph and its single newlines collapse into spaces. Everything that did render is CommonMark. GFM tables need the `remark-gfm` plugin, which `react-markdown`'s own documentation names as the way to get them.

## Fix

- `remark-gfm@^4.0.1` (same organisation and maintainers as `react-markdown`, MIT, 22 KB; its 18 packages are siblings of the 39 unified/micromark packages already installed).
- `table`/`thead`/`tr`/`th`/`td` mapped in `MarkdownRenderer.tsx` with the file's existing `border-current/…` vocabulary, wrapped in `overflow-x-auto` so a wide table scrolls instead of stretching the bubble. `th` and `td` forward `style`, which is how `react-markdown` carries the `:--`/`:-:`/`--:` alignment. `th` uses `text-start`, not `text-left`: the wrapper is `dir="auto"` and every other spacing class in the file is logical, so an RTL answer with the usual `|---|---|` separator keeps its header cells on the same edge as its body cells.
- The wrapper's `prose prose-sm max-w-none` classes are removed. `@tailwindcss/typography` is not installed, so Tailwind 4 emits nothing for them (proved by compiling `src/index.css` with `@tailwindcss/node`: `prose=false, prose-sm=false`). That absence is why the table elements are mapped by hand; installing the plugin now would restyle all twenty already-mapped elements in both consumers.

Both consumers of the renderer are affected identically: `ChatMessage.tsx` (sidebar chat) and `AssistantPanel.tsx` (voice assistant panel).

**Enabling the plugin also turns on** autolinks, strikethrough, task lists and footnotes. That is intended; the first three are pinned by a test. `react-markdown`'s URL sanitisation is unchanged: `javascript:` and `data:` hrefs are stripped and raw HTML stays escaped, also pinned.

**Deliberately untouched:**
- The `components` map is still recreated per render (see the memo note in `AssistantPanel.tsx`); hoisting it is a separate improvement.
- `del` and the task-list checkbox render with defaults, unstyled.
- Footnotes render react-markdown's default `Footnotes` section; the fragment links go through the existing `a` mapping's `target="_blank"`. Rare in chat, no data exposure; recorded, not restyled.
- Chat re-parses the whole answer on every streamed chunk. The plugin makes each parse ~4.5x more expensive, which in absolute terms is 1.71 ms per chunk and 355 ms across a whole 4 KB answer, inside a frame budget. A flush throttle in `useChatStreaming.ts` is a cheap independent improvement and follows as its own PR.
- The notes editor (TipTap) has no table support either; different library, different fix, and the note-generation prompts already say "Do NOT use tables". Not addressed here.

## Tests

`test/components/markdownRenderer.test.js`, written first and watched fail against the shipped component:

- RED on main:
  ```
  not ok 1 - a GFM table renders as a real table, not literal pipes
    error: 'renders a table element'
  not ok 2 - column alignment from the separator row is preserved
    error: 'left-aligned column'
  not ok 3 - a wide table scrolls instead of stretching its container
    error: 'table sits in a scrollable wrapper'
  not ok 6 - GFM extras the plugin enables render as elements, not literal syntax
    error: 'strikethrough renders as del'
  ```
  The `prose` guard and the `text-start` pin were each also watched fail before their one-line fix.
- GREEN on this branch: 7/7.
- Covers: a real `<table>` with one header and two body rows and no surviving pipe; left/centre/right alignment via `style`; the scroll wrapper; header cells carry `text-start`; bold inside a cell; non-table markdown unchanged and no `prose` class; strikethrough, autolink and two task-list checkboxes; `javascript:`/`data:` hrefs stripped and raw HTML escaped on both versions.

Full suite on Node 24: 4301 tests, 4081 pass, 0 fail, 219 skipped (platform-gated tests, unchanged). `npm run quality-check` exits 0.

Streaming was simulated token by token: raw pipes show only until the separator row completes, then it resolves into a table and stays one.

## Provenance

Found during 1.10.0 release testing, 2026-09-10. The `prose` wrapper dates from 2d1f52ed (release notes rendering, 2026-01).


Chadpiha (approved):
Approved. Reviewed the latest changes end to end, including table rendering, alignment, horizontal scrolling, retained scroll position, and footnote navigation.
The footnote fix preserves accessible targets, scopes IDs per message, and keeps external-link behavior intact. No remaining actionable findings.

PR #2134, opened 2026-09-11
fix(settings): relaunch the app after Reset app data

### LINKED ISSUE #2132: Reset app data leaves the database closed: new transcriptions aren't saved until the app is restarted
**Describe the bug**
After **Settings → System → Reset app data**, the window reloads into onboarding but the main process keeps running with the transcription database closed and its file deleted. Dictation still transcribes and pastes, but nothing is saved to history (and history can't load) until OpenWhispr is fully quit and relaunched. Dictations made in between are lost.

**To Reproduce**
1. Use OpenWhispr normally.
2. Settings → System → **Reset app data** → confirm.
3. The window reloads into onboarding. Finish it (e.g. *Continue without an account* and any provider).
4. Dictate a few times. Text is transcribed and pasted as usual.
5. Open history: none of the new dictations are there.
6. Quit fully and relaunch: history is empty; the dictations from step 4 were never stored.

**Expected behavior**
After a reset the app works end to end: a fresh database is created and new dictations are saved. The confirm dialog already says "The app will reload after reset", but only the renderer reloads.

**Root cause (1.10.0)**
- `cleanup-app` (`src/helpers/ipcHandlers.js`) calls `this.databaseManager?.db?.close()` and unlinks `transcriptions.db` / `-wal` / `-shm`, then returns. It never reopens the database or relaunches the app, and it leaves the closed handle on `databaseManager.db`.
- `DatabaseManager.initDatabase()` (`src/helpers/database.js`) only runs from the constructor at startup.
- The Reset button (`SettingsPage.tsx`, `settingsPage.developer.resetAppData`) follows up with `window.location.reload()`, which reloads the window, not the main process.
- `saveTranscription()` / `getTranscriptions()` only guard `if (!this.db)`. The handle is closed but not null, so they reach `prepare()` and throw on the closed connection.

Observed on macOS after a reset: the running app process held no `.db` file open (`lsof`), `transcriptions.db` was missing from userData, and new dictations never reached history until a restart.

Also seen in the same run: `~/.cache/openwhispr/qdrant-data` was deleted while the Qdrant sidecar process kept running, so local semantic search was broken until restart too.

Delete account's device-erase step (`eraseDeviceData` in `SettingsPage.tsx`) calls the same `cleanup-app` handler.

**Suggested fix**
Relaunch after a successful reset (`app.relaunch(); app.exit(0)`) instead of reloading only the window. That recreates the database and restarts every stopped service with fresh state, and matches the dialog copy. Reopening the DB in place (`initDatabase()`) would also need the stopped sidecars restarted, so relaunching looks simpler.

**Desktop**
- OS: macOS (the code path is platform-independent)
- Version: 1.10.0

boseq:
## Summary

Fixes #2132. **Reset app data** (and Delete account with device erase) runs `cleanup-app`, which closes the transcription database and stops local services in the main process. The renderer then only reloaded its window. Nothing reopened the database, so dictations still pasted but were never saved to history, and semantic search stayed down, until the user quit and relaunched.

- A new `relaunch-app` IPC (`relaunchApp()` in preload) restarts the app and quits through the normal `before-quit` path, so sidecars shut down the same way as Quit from the tray. Both callers use it instead of `window.location.reload()`. Delete account without device erase still just reloads.
- `getRelaunchOptions()` (pure, in `autoStartPolicy.js`) keeps the relaunch from replaying how this process was launched:
  - It drops `--hidden`, so a process started at login on Windows/Linux doesn't come back in the tray with no window.
  - It drops a cold-start `openwhispr://` deep link, which startup would otherwise handle again. A sign-in link would restore the session the reset just cleared.
- On AppImage, `app.relaunch()` never comes back, even with `execPath: $APPIMAGE`. Electron's relaunch helper runs from inside the FUSE mount, which goes away when the app exits. So the handler instead spawns a detached `/bin/sh` that waits for the app to exit (releasing the single-instance lock) and then starts the AppImage file.
- A downloaded update isn't installed on the reset's quit (`UpdateManager.deferInstallOnQuit()`). electron-updater would otherwise replace the executable the relaunch starts; on AppImage it deletes `$APPIMAGE` and can move the new build to a new file name. The update waits for a later quit.
- Under `npm run dev` it only quits: `concurrently -k` stops Vite as soon as Electron exits, so a relaunched dev instance would have no renderer.
- The dialogs now say the app will restart rather than reload, in all 11 locales.

The macOS and Linux build checks fail on fork PRs, this one included, until #2135 lands.

## Test plan

- [x] `node --import tsx --test test/helpers/autoStartPolicy.test.js test/helpers/deviceCleanupContract.test.js test/helpers/updater.test.js`: 27/27, including three new relaunch cases and one updater case
- [x] `tsc --noEmit` (src), `npm run i18n:check`, eslint and prettier on the changed files
- [x] `npm test`: the only failures are the pre-existing `better_sqlite3` `ERR_DLOPEN_FAILED` ones from a broken local native module, none in touched code. One run also hit a timing flake in `textEditMonitorWindowBounds.test.js`, which this PR doesn't touch; it passes when run alone, on this branch and on main
- [x] AppImage, tested in an arm64 Ubuntu 26.04 VM with a minimal Electron 41.10.5 AppImage running this handler's relaunch branch, launched with `--hidden` and a stale sign-in link:
  - The app relaunches with only `--no-sandbox` and leaves no mount behind.
  - `app.relaunch()` never came back, with or without `execPath`.
  - The same build extracted to a folder relaunched through `app.relaunch()`.
- [x] macOS, real app: an unpackaged production-mode run of this branch on a throwaway profile, driven over CDP:
  - After Reset, saving a transcription fails with "The database connection is not open", and it still fails after a window reload. That's the old behavior.
  - `relaunchApp()` restarts the app with the same profile and arguments. The new process opens a fresh database, and saving and history work again.
- [ ] Packaged macOS: Settings → System → Reset app data. The app quits and reopens into onboarding; new dictations show up in history and survive a restart
- [x] Windows 11 VM, real app: this branch's app code swapped into a local 1.10.0 Windows x64 build (same dependencies and native modules), driven over CDP:
  - After Reset, saving fails with "The database connection is not open", and still fails after a window reload. `relaunchApp()` starts a new process with the same arguments, and saving and history work again.
  - Started with `--hidden`, as a login launch is, only the dictation window exists. After Reset, the relaunched process has no `--hidden` and the control panel window is back.
- [ ] Delete account with and without device erase: relaunch vs. window reload
- [ ] `npm run dev`: reset quits the app instead of leaving an orphaned instance


gabrielste1n:
Thanks for the thorough fix and the test plan. I traced both flows end to end per platform and pushed one follow-up commit to your branch (ff6b65c1) before merging:

- **Windows portable build**: `app.relaunch()` restarted the exe inside the `%TEMP%` unpack dir, which the NSIS portable stub deletes the moment the app exits (`portable.nsi`: `ExecWait` then `RMDir /r`), so the app never came back. The handler now spawns a detached PowerShell waiter that waits for the stub (the app's parent) and starts `$PORTABLE_EXECUTABLE_FILE`, mirroring your AppImage approach.
- **macOS with a downloaded update**: `deferInstallOnQuit()` only affects the NSIS/AppImage quit handler. `MacUpdater` hands a finished download to Squirrel.Mac at download time, and Squirrel installs it on any quit regardless of the flag, so the reset relaunch raced ShipIt's bundle swap. When Squirrel already holds an update (`squirrelDownloadedUpdate`), the handler now lets `installUpdate()` own the restart, falling back to the plain relaunch if that does not quit.
- Both waiters are built by `getRelaunchWaiter()` in `autoStartPolicy.js` so the sh and PowerShell command forms are unit-tested; dropped the redundant `2>/dev/null` (stdio is already ignored).
- Trimmed comments that repeated the same rationale across the handler, `autoStartPolicy.js`, `CLAUDE.md` and the tests, and made the es/pt reset copy use "app" consistently.

Verified: typecheck, lint, prettier, i18n check, and the full suite (4129 passing) on the branch and on a trial merge with current main.

PR #2137, opened 2026-09-11
fix(downloads): say Installing in the download tray once extraction starts

greatcoat:
## Problem

The background download tray's header strip is hardcoded to **"Download in progress"**. A local model that has finished transferring then sits behind a full bar, a 100% label and that same header while extraction runs, so it reads as a stalled download. The settings pane already says "Installing <model>" at the same moment, which makes the tray look wrong rather than merely quiet.

The per-row `installing` flag was already tracked and already used to disable the row's cancel button. Only the header ignored it.

## Fix

- The header reads **"Installing"** followed by an ellipsis cycling `""` → `"."` → `".."` → `"..."` every 400ms. Once byte progress stops, that motion is the only remaining signal that work is live.
- It switches only when **every** row has reached extraction, so a tray holding one installing model and one still downloading keeps saying "Download in progress".
- The dots are `aria-hidden`: the tray is an `aria-live` region, and an announced dot would re-read the whole strip every frame. The phase change itself is still announced once.
- The interval is skipped under `prefers-reduced-motion`, leaving a static "Installing".
- Drops the trailing `...` from the Arabic `onboarding.rehaul.local.installing` string, which no other locale carries and which would otherwise double up against the animated dots.

The two pure decisions live in `localDownloadState.ts` next to `mergeHydratedDownloads`, so they are unit-testable like the rest of that module.

## Tests

- `localModelSetupStep.test.js` renders the real tray, drives a Parakeet download to 100%, asserts the header still says downloading, then emits the `installing` event and asserts it flips. Reverting the component change makes it fail with the reported symptom (`downloadInProgress` where `installing` was expected).
- `localDownloadState.test.js` pins the mixed-tray rule (an installing row beside a downloading or errored row does not claim installation, and an empty tray never does) and the exact four-frame dot cycle.
- Full suite: 4076 pass, 219 skipped, 1 fail — `settingsStore imports without a bare localStorage global`, the documented Node 25 artefact (it asserts a bare `localStorage` global is absent, and Node 25 defines one). It fails the same way on untouched `origin/main`; CI runs Node 24 and does not see it.
- `prettier --check`, `eslint` and `tsc --noEmit` clean on the touched files; `check-i18n.js` passes.

Not verified in a running build — the change is a label and a timer, covered by the render test above.

### DIFF
--- a/src/components/onboarding/BackgroundModelDownloadTray.tsx
+++ b/src/components/onboarding/BackgroundModelDownloadTray.tsx
@@ -460,8 +460,14 @@ export default function BackgroundModelDownloadTray({
                   {downloadDisplay(download).name}
                 </span>
               </span>
+              {/* Extraction pins the percentage at 100, so the number stops
+                  carrying information exactly when the bar stops moving. Naming
+                  the phase per row — as the setup step and the settings picker
+                  already do — keeps that readable whatever the other rows do. */}
               <span className="shrink-0 font-medium tabular-nums text-muted-foreground">
-                {Math.round(download.percentage)}%
+                {download.installing
+                  ? t("onboarding.rehaul.local.installing")
+                  : `${Math.round(download.percentage)}%`}
               </span>
             </div>
             {download.error ? (
--- a/src/components/onboarding/BackgroundModelDownloadTray.tsx
+++ b/src/components/onboarding/BackgroundModelDownloadTray.tsx
@@ -23,7 +23,7 @@ import type {
   ParakeetDownloadProgressData,
   WhisperDownloadProgressData,
 } from "../../types/electron";
-import { mergeHydratedDownloads } from "./localDownloadState";
+import { ellipsisFrame, isTrayInstalling, mergeHydratedDownloads } from "./localDownloadState";
 import { ONBOARDING_SESSION_KEY, isRequiredModelsOnboardingStepActive } from "./flow";
 import { getPlatform } from "../../utils/platform";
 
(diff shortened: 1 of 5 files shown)


Chadpiha (approved):
Approved. Good catch, and the reasoning in the description holds up.

## What I verified

- **Premise confirmed.** `parakeet.js:398` is the only `type: "installing"` emit in the
  codebase and it carries `percentage: 100` explicitly, so the bar really does sit full
  and the label really does read 100%, with no further events until `complete`/`error`.
  Extraction is a 680MB–2.7GB single-threaded bzip2 stream behind a 5-minute tar timeout
  and up to 2 retries, so the dead air is real, not theoretical.
- Reverting the header change does fail the new test, with exactly the symptom described.
- **The Arabic claim checks out across all 11 locales** — `ar` was the only one carrying a
  trailing ellipsis. Worth adding to the commit message: that key is shared with
  `ProviderSetupStep.tsx:1070` and `RequiredModelDownloadStep.tsx:142`, and dropping the
  `...` is right for those too — their ternary partner `downloadingShort` was already
  ellipsis-free in `ar`, so this fixes an inconsistency rather than introducing one.
- RTL is correct, and it's correct *because* the dots sit inside the label span with no
  bidi isolate. Might be worth a note there so nobody later "fixes" it with a `dir` or `<bdi>`.
- Full suite green on Node 24, 0 failures. Your single failure is the Node 25 artifact; CI
  won't see it.

## One follow-up commit added (b9658477)

The all-rows rule means the switch misses the journey it's aimed at. Local onboarding
always runs both steps (`SETUP_ROUTES.local`) and Proceed is enabled mid-download, so the
tray normally holds a dictation model *and* an assistant LLM. LLMs are 0.8–19.8GB against
Parakeet's 680MB, so the LLM row is still transferring while Parakeet extracts — and it
pins the strip on "Download in progress" at exactly the moment the Parakeet row shows the
full bar and 100%.

So I moved the signal to the row, matching what `ProviderSetupStep` and
`RequiredModelDownloadStep` already do; it now reads correctly regardless of neighbours.
Also made `isTrayInstalling` ignore errored rows — an error row has already stopped, so
waiting on it left the strip claiming a download that was no longer running (reachable
whenever the assistant LLM fails while the dictation model is still extracting).

Separate issues, not yours: the JS extraction fallback has no watchdog, and Windows always
takes it since `systemTar.js:44` rejects `.tar.bz2`; and `DownloadProgressBar.tsx:44`
hardcodes English `Installing ${modelName}`.

PR #2139, opened 2026-09-11
fix(assistant): keep the vision override inert until the user picks a model

greatcoat:
## Problem

Josh (managed workspace, app 1.10.0, macOS) reported post-release: Settings → Language Models → **Voice Assistant** shows OpenWhispr Cloud as Active, yet under **Separate vision model** (ON) the full BYOK picker renders — OpenAI selected, API key empty, **GPT-6 Astra badged Active** — and using the assistant "errors out".

Evidence on his machine (debug log `debug-2026-09-11T11-54-48-586Z.log`, renderer localStorage, `workspace-policy.json`):

- stored `dictationAgentMode = openwhispr`, `useDictationAgentVisionModel = true`, `voiceAgentScreenContext = true`; **no** `dictationAgentVisionProvider` / `dictationAgentVisionModel` key exists at all, and no OpenAI key is stored anywhere.
- the assistant run: transcript produced, `reasoningProcessingDurationMs: 8`, no `Cloud reason request`, no paste, no error line. Two plain dictations in the same session reached Cloud fine.
- his policy is `managed: true` and permits every LLM mode and BYOK provider (locks only, nothing forced).
- the API's `/api/reason` and `/api/agent/stream` accept a screenshot (`screenContext`) and route it to the `assistantVision` chain server-side; the admin console has no vision control. Neither layer can produce a failure with no network call.

What the user sees is an assistant bubble `Error: OpenAI API key not configured` (`useChatStreaming.ts` catch); the log stays silent because every diagnostic on that path is `debug`-level.

## Root cause

**The workspace-policy overlay invents a provider and model for the never-configured vision override, and the panel treats that invented model as the user's choice.**

1. `bb881d10` (#2115, 2026-09-10) pinned `dictationAgentVisionMode` to the constant `"providers"` and made the picker render on the toggle alone, with no check of the assistant's own mode.
2. `selectPolicyEffectiveSettings` loops over every `INFERENCE_SCOPES` entry. With the vision scope now permanently in `providers` mode and its stored provider empty, `resolveEffectivePolicySelection` substitutes `allowedByokProviders[0]` (`openai`) and, because the provider "changed", writes that provider's default model (`gpt-6-astra`) into the effective settings — display only, never persisted, but exactly what the tab badges Active.
3. `useChatStreaming` calls `getSettings()`, which is the **overlaid** view. `resolveDictationAgentVisionInference` decides "the user chose a target" by `!!settings.dictationAgentVisionModel`, so the invented model flips the override to active; `resolveModeReachability` never checks credentials; `resolveAgentImageTarget` picks the override over Cloud; `ReasoningService.getApiKey("openai")` throws before any request.

In 1.9.2 the same overlay code produced `mode: "openwhispr"` for that scope, the model-writing branch never fired, and the override stayed inert. Only managed users hit this, which is why nobody else did.

## Fix

- `src/config/inferenceScopes.ts` — `dictationAgentVision` is marked `optional: true`: inert until the user picks its own model.
- `src/stores/settingsStore.ts` — the overlay skips an optional scope while it has no stored model, and when policy moves a chosen override's provider it clears the model instead of writing that provider's default, so the override goes inert and the picker asks again (tested both ways).
- `src/components/chat/useChatStreaming.ts` — a failed assistant request now logs at error level with scope, mode, provider, model, whether a screenshot rode along, and the message.
- `CHANGELOG.md` — Unreleased → Fixed entry.

**Product note:** the override stays available under a Cloud assistant, as #2115 shipped it. A first cut of this PR hid it on Cloud; Josh reversed that on review (a Cloud user may legitimately point screenshot commands at their own key), so that rule is gone from the tab and the resolver. With the overlay no longer inventing a choice, an unconfigured switch on Cloud is simply inert. A workspace-policy field to govern the override (allowed / hidden / forced off) is a separate follow-up; today an admin can only limit the BYOK provider list, forbid BYOK entirely, or turn screen context off.

**Deliberately untouched**
- `resolveModeReachability` still ignores API keys (the known gap #2106 recorded). A user who *explicitly* picks a BYOK vision model without pasting a key still gets the accurate "API key not configured" bubble; a credential-aware reachability check is a separate change.
- The overlay's clamp for *base* scopes (first allowed provider + default model) is unchanged — that is the intended behaviour when a policy forces a mode switch; only the optional override is treated as "clear, don't repoint".
- `DictationAgentSettings.tsx` is untouched relative to `main`: the tab's gates are exactly #2115's.
- No migration for stale `dictationAgentVision{Provider,Model}` left by ≤1.9.2 profiles that used the override in BYOK and then switched it to Cloud: after this PR they only matter off-Cloud, where the user did choose them.
- The dead `dictationAgentVisionMode` localStorage key is left as is.

## Tests

Written first and watched fail on the unfixed tree (Node 24):

- `test/stores/policyEffectiveVisionOverride.test.js` (new, renderer harness) — reproduces the profile: managed BYOK-permitting policy, assistant on Cloud, override on, nothing chosen.
  - "the override keeps no provider or model" — RED: `actual: 'openai', expected: ''`
  - "a screenshot command stays on the cloud with its screenshot" — RED: `actual: 'dictationAgentVision', expected: 'dictationAgent'`
  - "a policy that moves a configured override's provider clears its model" — after the review follow-up: provider lands on the allowed one, model is empty, and the panel stays on the assistant's own (allowed) provider instead of a keyless default.

After the fix: the three files (`dictationAgentInference`, `policyEffectiveVisionOverride`, `dictationRouting`) run 86/86. Full suite on the final tree, merged with `main` at 4335972b (Node 24): 4450 tests, 4230 pass, 0 fail, 219 skipped (the Electron-ABI DB skips). Under parallel load, `audioActivityDetector`'s darwin native-listener test (which compiles the Swift mic listener with a 120 s limit) and the two known load-flaky `textEditMonitor` tests can time out; each passes alone and all pass in this quiet run. `npm run quality-check` equivalents: prettier and eslint clean on every touched file; `tsc --noEmit` shows only the 4 pre-existing missing-optional-dep errors present on `main`.

Not verified visually: the settings tab is unchanged, but the disappearance of the phantom "Active" badge on a managed account was checked through the store test only, not by eye.

## Review follow-ups (second and third commits)

An independent review pass (rebuilt the base commit under the new tests to confirm RED, ran `quality-check` and the non-DB suite) found no Critical issues and four Important ones. Addressed: a configured-but-policy-moved override was still handed an invented default model (now cleared → inert); no changelog entry (added); the error log lacked the fields that discriminate this bug class (model + screenshot attached, added). The fourth, duplication of a cloud-only rule between tab and resolver, became moot when Josh reversed that rule (third commit). Left as noted: reachability still ignores keys; a policy in `loading`/`error` status briefly judges the raw mode (pre-existing for every policy gate); the ≤1.9.2 stale-vision-mode migration.

## Provenance

Josh's report + debug log, 2026-09-11 13:20 UTC. Origin of the vision override: `22c52d4d` (v1.8.2). Regression: `bb881d10` (#2115) with the panel moved onto the Voice Assistant scope by `918b0754` (#2106). RCA filed in Titan: `wiki/areas/engineering/assistant-vision-override-managed-policy-rca.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Voice Assistant on OpenWhispr Cloud with a stray "Separate vision model" switch.** On a managed workspace, an override that was switched on but never given a model was shown with an invented provider and model, and every screenshot command went to that provider without a key, failing with "OpenAI API key not configured". Workspace policy no longer invents a target for an override the user never configured, and a chosen override that policy moves to another provider is cleared rather than repointed, so it asks to be picked again. Screenshot commands now stay on OpenWhispr Cloud while the assistant runs there; the separate BYOK vision model is offered, and applies, only when the assistant runs elsewhere. A failed assistant request is now written to the debug log.
+
 ## [1.10.0] - 2026-09-11
 
 The desktop app gets a new look: an inset content container with a top bar and ⌘K search, a redesigned Home and note editor, the Yowza brand font and Nucleo icons, and one brand-blue glass surface for every primary action, onboarding included. Notes start recording the moment you create them, offer an AI summary when a recording ends, and gain Detailed Notes and Follow-up email as built-in actions. Orukeet arrives as the recommended local speech-to-text model and local models run up to three times faster on Apple silicon. Managed enterprise speech-to-text lands for workspaces on Azure OpenAI or AI Foundry, transcription gains streaming from Deepgram, AssemblyAI and Gemini Live plus Cohere Transcribe as a local engine, gpt-transcribe becomes the OpenAI default, onboarding was rebuilt and reordered, Insights gains an opt-in leaderboard and backfilled history, updates install themselves, Arabic joins as the eleventh UI language, and Astra and Fable 5.1 join the reasoning models.
--- a/src/components/chat/useChatStreaming.ts
+++ b/src/components/chat/useChatStreaming.ts
@@ -5,6 +5,7 @@ import { isEnterpriseProvider } from "../../models/ModelRegistry";
 import { providerSupportsImages } from "../../services/ai/inferenceProviders";
 import { getSettings, useSettingsStore } from "../../stores/settingsStore";
 import { resolveChatStreamingInference } from "../../helpers/dictationAgentInference.js";
+import logger from "../../utils/logger";
 import {
   isAgentAllowed,
   isLlmSelectionAllowed,
@@ -565,6 +566,16 @@ export function useChatStreaming({
             )
           );
         } else {
+          logger.error(
+            "Assistant request failed",
+            {
+              scope: llmConfig.scope,
+              mode: llmMode,
+              provider: llmConfig.provider,
+              error: (error as Error).message,
+            },
+            "reasoning"
+          );
           announceResponse();
           setMessages((prev) =>
             prev.map((m) =>
--- a/src/components/chat/useChatStreaming.ts
+++ b/src/components/chat/useChatStreaming.ts
@@ -572,6 +572,8 @@ export function useChatStreaming({
               scope: llmConfig.scope,
               mode: llmMode,
               provider: llmConfig.provider,
+              model: llmConfig.model,
+              attachScreenContext,
               error: (error as Error).message,
             },
             "reasoning"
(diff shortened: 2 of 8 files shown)


Chadpiha (approved):
Approved — reviewed policy handling, model selection, and assistant screenshot routing end to end. Unconfigured or policy-blocked vision overrides remain inactive, explicit selections continue to work, and the phantom Active badge is resolved. No actionable regressions found.

Validated the regression tests against both the pre-fix and updated code. Full suite: 4,232 passed, zero failures, 219 skipped, one todo. TypeScript, changed-file lint, and formatting passed.

Issue #2086, opened 2026-09-09
1.9.2 - Local-only transcription unexpectedly consumes OpenWhispr Cloud word allowance


Shad0wWulf:
## Describe the bug

I'm using OpenWhispr 1.9.2 and have explicitly configured the application to use local/on-device models.

Despite this, my Free account has consumed **2,112 / 2,000 OpenWhispr Cloud words this week** and now reports **"Weekly limit reached"**.

This appears to be more than just a cosmetic/accounting problem.

The initial transcription appears to have followed the OpenWhispr Cloud path and failed because the weekly cloud allowance had been exhausted. I then pressed Re-transcribe without changing any settings, and the same recording succeeded using the configured local model.

This makes it appear that the initial transcription path may be attempting to use OpenWhispr Cloud despite Local being explicitly selected, while the Retry/re-transcription path correctly uses the local model.

I have not intentionally used OpenWhispr Cloud transcription.

I cannot reproduce without buying more tokens.

### Speech-to-Text

All relevant STT modes are explicitly set to **Local**:

- Dictation: **Local**
  - NVIDIA Parakeet Unified EN 0.6B active
- Note Recording: **Local**
- Audio Upload: **Local**

The UI describes this mode as:

> On-device models. Fully private.

### Language Models

- Dictation Cleanup: **Local**
  - Qwen3.5 2B active
- Voice Assistant: **Local**
- Translation: **Disabled**
- Note Formatting: **Local**
- Chat: **Local**

### Privacy

- Cloud backup: **Off**
- Usage analytics: **Off**

Screenshots are attached showing these settings.

## Actual behaviour

The Personal Plan screen reports:

    Free
    2,112 / 2,000 words this week
    Limit reached

The Free plan specifically describes this allowance under:

    OPENWHISPR CLOUD
    2,000 words/week transcription

The application also displays:

    Weekly limit reached
    Upgrade to Pro for unlimited transcriptions.

After reaching the limit, I observed:

1. Started a normal dictation while Local was selected.
2. The transcription failed after the OpenWhispr Cloud allowance had been exhausted.
3. I used Retry on the failed transcription.
4. The retry successfully transcribed using the configured local model.

Unfortunately, I did not capture a screenshot of the original failure before retrying it.

## Expected behaviour

[...]

PR #2144, opened 2026-09-11
fix(onboarding): sign in without restarting onboarding, and reopen BYOK steps on the saved setup

### LINKED ISSUE #2128: Clicking login button resets app state to initial onboarding wizard and wipes custom API keys
Describe the bug
When running OpenWhispr in an unauthenticated state (not logged in), after completing initial setup and configuring custom API endpoints and BYOK API keys, clicking any login or sign-in button immediately redirects the application back to the initial onboarding setup wizard. Going through this wizard wipes all previously configured custom API keys, custom base URLs, and associated settings, requiring everything to be reconfigured from scratch.

To Reproduce
1. Launch OpenWhispr without signing in to an account.
2. Complete initial setup and configure custom API keys and endpoint URLs in settings.
3. Verify that speech transcription works properly.
4. Click on any login or sign-in button in the application.
5. See the app immediately redirect to the initial setup guide / onboarding flow.
6. Observe that all custom API keys and configurations are cleared or overwritten.

Expected behavior
Clicking login should initiate an authentication flow or display a login prompt without redirecting back to the onboarding wizard, and without wiping existing locally stored API configurations and keys.

Desktop (please complete the following information):
- OS: Windows 10
- Browser: N/A (Desktop App)
- Version: Latest release

Additional context
This causes severe friction for users using their own API keys and custom reverse proxy endpoints, as an accidental click on login forces a complete reconfiguration of the app.

boseq:
## Summary

Fixes #2128. It had two halves, and this PR covers both:

1. **Clicking sign-in restarted the whole wizard.** A signed-out user who clicks sign-in is now logged out instead, onto the sign-in screen the app already shows a signed-out account.
2. **The restarted wizard showed API keys and endpoints as wiped.** Nothing was actually deleted: the BYOK steps just opened blank. They now reopen on the saved setup, which still matters for the remaining restart path (logging out from inside onboarding).

### Sign in without restarting onboarding

Every signed-out sign-in button called `useStartOnboarding`, which reset onboarding progress and reloaded the window. A guest who clicked any of these went back through the whole wizard, and re-walking it rewrote the setup they had configured — the Cloud tile alone rewrites 13 transcription keys and 5 LLM scopes:

- the OpenWhispr Cloud tile in Speech-to-Text, Note Recording, Audio Upload or Language Models
- Account → Create Free Account
- Plans → Create Free Account and Get Started
- Upload → Create account

The app already has the screen those should lead to. `AppRouter` renders `ReauthenticationScreen` whenever onboarding is finished, the user is signed out and the guest flag is unset, and Settings → Sign out has always landed there with every setting intact.

- **Signing in logs the guest out.** `requestSignIn` removes `authenticationSkipped` and `skipAuth` and reloads — the same two steps `handleSignOut` takes. No onboarding state is written or cleared, so the reload lands on that screen rather than the wizard, and signing in returns the user to the control panel as they left it. `useStartOnboarding` is removed with its last callers.
- **No new UI.** The re-auth screen, the auth flow inside it and the toast below are existing components, unchanged. The only new strings are the toast's three, in all 11 locales.
- **Backing out is one click.** "Continue without an account" on that screen writes the guest flag back and returns to the control panel, setup untouched.
- **Signing in switches nothing.** A post-sign-in cloud switch is what overrode Local for #2086, so no setting changes after sign-in.
- **Cloud nudge.** After signing in, a user whose dictation runs on their own setup (local, API keys or self-hosted) gets a toast saying OpenWhispr Cloud is available. Its Open Settings action goes to Speech-to-Text → Dictation.
  - The toast waits for the account's policy, and skips accounts whose policy forbids Cloud or whose dictation is already on Cloud — read policy-effective, so a managed user the policy clamps onto Cloud is not told otherwise.
  - Sign-in always crosses a reload — this one, and again for Google, Microsoft, Apple and SSO (`applySessionTokenAndRefresh`) — so the marker the toast reads lives in localStorage and expires after 15 minutes.

**Trade-off worth naming.** While the re-auth screen is up the window is compact and `setOnboardingActive(true)` suppresses global hotkeys and the dictation pill. That is what Sign out already does to an account user, but it is newly reachable by a guest from a Settings tile: a guest who abandons sign-in by quitting instead of clicking "Continue without an account" reopens there. Signing in also returns them to the control panel home rather than the Settings tab they clicked from; the nudge's Open Settings covers that.

### Reopen the BYOK steps on the saved provider setup

`resetOnboardingProgress` deletes the session draft, and `ByokProviderStep` seeded the provider, model, endpoint URL and model ID only from that draft. Users with their own keys and custom or self-hosted endpoints got an empty form, and only the key came back from the store.

- **New `src/components/onboarding/savedByokConfig.ts`.** It resolves the setup already saved for each BYOK step, in the same order as `resolveTranscriptionRoute`: Settings self-hosted server, then custom endpoint, then hosted provider. It returns nothing for local, OpenWhispr Cloud, enterprise or a first run. Drafts still never carry a key.
- **Seeding.** `ByokProviderStep` opens on the session draft first, then on saved settings, the same order `LocalModelSetupStep` uses. A blank draft left by an older build falls back to saved settings.
- **A stored key belongs to one endpoint.** The seed carries the endpoint its key was saved for, and the step offers that key only while the URL in the field still resolves to it, under the same scheme rule the commit applies. Nothing saved, a different server typed, or a late-arriving secret after an edit all leave the field empty, so a bearer token saved for one endpoint is never sent to another. A stranded prefill is dropped by byte-identity only; a typed key is never cleared.
- **Mode switching.** Switching between hosted and self-hosted keeps both halves of the form and only swaps the shared key field. Keys that load after mount fill an empty field without replacing anything typed.
- **Endpoints without a key are saved as the Settings self-hosted server.** The step writes `remoteTranscriptionUrl`, `remoteTranscriptionModel` and `remoteTranscriptionType`, as `migrateProviderSettings` does, and leaves the Custom URL, key and model alone.
  - Settings → Speech-to-Text and upload show the server.
  - Re-confirming it no longer overwrites an unused Custom endpoint or deletes its key.
  - Clearing a reopened Custom key and saving takes this path too, so the stored key stays; it is never sent.
- **Endpoints with a key go to the Custom slot** and clear `remoteTranscriptionUrl`, because only the Custom route authenticates. The step now switches the provider before setting the model: `switchCloudTranscriptionProvider` filed the model under the outgoing provider, which turned the Model ID into `whisper-1`.
- **Policy gate.** `fieldsReady` requires the provider to still be in the policy-filtered list, so a policy that loads after mount can't leave a removed provider committable.

It does not also require `cloudTranscriptionMode === "byok"` before prefilling:

- `reconcileTranscriptionRouting` repairs a Cloud-mode mismatch on every store load, and every onboarding restart reloads the window.
- Every in-session writer keeps the two fields in sync.
- The prefill only applies once the user has picked BYOK or Self-hosted, and nothing is saved until Proceed.

### Follow-ups (not in this PR)

- A custom endpoint that uses a key still opens under the key-less Self-Hosted tab in Settings, because `deriveTranscriptionMode` maps byok + custom to self-hosted. Upload readiness also checks only `remoteTranscriptionUrl`.
- A Settings self-hosted server saved without a model reopens with Model ID blank, and the onboarding form requires it.
- After any self-hosted save, onboarding's copy to all scopes sets Note Recording to Self-Hosted, which Note Recording can't use yet (`meetingTranscriptionRouting.js`). This predates this PR.
- #2132 (Reset app data leaves the database closed).

## Test plan

Sign-in:

- [x] `test/utils/requestSignIn.test.js` (new): signing in drops both guest flags, reloads once, and writes no onboarding state, so the reload lands on the re-auth branch rather than the wizard. It also pins that nothing the user configured changes, carrying over the #2086 check from the removed `useStartOnboarding` test.
- [x] `test/hooks/signInCloudNudge.test.js` (new) mounts the nudge hook: it nudges once after the account's policy settles and its action opens Speech-to-Text; no nudge for an abandoned sign-in, an ordinary launch, dictation already on Cloud, or a policy that forbids Cloud.
- [x] `test/utils/signInCloudNudge.test.js` (new): the decision for each mode, policy status and prompt age.
- [x] Mutation check: leaving `authenticationSkipped` in place fails the new test.
- [x] Ran a dev build against an isolated profile as a guest using Groq. Settings → Speech-to-Text → OpenWhispr Cloud opens the re-auth screen; `onboardingCompleted` stays `true`, no `onboardingSessionV2` or `onboardingCurrentStep` is written, both guest flags are gone and every routing key is unchanged. "Continue without an account" returns to the control panel with the setup intact.
- [ ] Real sign-in with email, Google and SSO on a CI build. Not done; it needs a real account. The toast itself was rendered in an earlier round with the signed-in state faked by a temporary local patch.
- [ ] Upload → Create account clicked in the app. It needs a BYOK file over the size limit; the change there is the same one-line handler swap.

Saved provider setup:

- [x] `test/components/savedByokConfig.test.js` (new): the resolver's order and exclusions, and the draft shape.
- [x] `test/components/byokSavedConfig.test.js` (new) mounts `ByokProviderStep` and covers:
  - seeding for hosted, custom endpoint, Settings self-hosted and assistant
  - draft precedence
  - reopening from the saved draft (this test fails on the pre-fix code)
  - mode switching and keys that load late
  - policy removing a provider (API-key provider and Corti)
  - key-less vs keyed saves, resolved through `resolveTranscriptionRoute` for dictation and upload
  - re-confirming a Settings server keeps an unused Custom endpoint and key, and clearing a reopened key keeps it stored
  - nothing saved plus a leftover custom key commits key-less; retyping the endpoint takes the prefilled key back out; a draft naming another server reopens key-less; the same for the assistant step
- [x] Reproduced #2128 on the installed 1.10.0: in the restarted wizard, endpoint URL and model ID were blank.
- [x] Ran a dev build against an isolated profile and a stand-in OpenAI-compatible endpoint:
  - after restarting onboarding (through the Cloud tile, before the sign-in change), the self-hosted dictation and assistant steps reopened with the saved URL and model
  - finishing the wizard left the saved settings unchanged
  - Settings → Speech-to-Text showed the endpoint

Whole branch:

- [x] `npm test`: 4,714 tests, 0 failures. `npm run quality-check` (Prettier, ESLint, `tsc --noEmit`) exits 0.
- [x] Four fresh-eyes review passes over the whole diff; the last returned nothing above nit level. Three fixes came out of it (`3b2d5c0b`, `2c9d73ae`, `7e425cb8`), each with a test proven to fail against the pre-fix source.
- [ ] Real-world runs of the CI builds on macOS and Windows.

https://claude.ai/code/session_01NcBZG5nae85TZKoGWCWAFp

### DIFF
--- a/src/components/onboarding/ProviderSetupStep.tsx
+++ b/src/components/onboarding/ProviderSetupStep.tsx
@@ -45,11 +45,7 @@ import {
   rememberPendingLocalModel,
 } from "./pendingLocalModels";
 import { isLocalStageDownloadActive } from "./localDownloadState";
-import {
-  isBlankByokDraft,
-  resolveSavedByokConfig,
-  selfHostedRemoteTranscriptionUrl,
-} from "./savedByokConfig";
+import { isBlankByokDraft, resolveSavedByokConfig } from "./savedByokConfig";
 
 export function SetupStageStepper({ stepId }: { stepId: OnboardingStepId }) {
   const { t } = useTranslation();
@@ -490,19 +486,23 @@ export function ByokProviderStep({
         store.setChatAgentModel(draftCustomModel);
         store.setChatAgentMode("self-hosted");
         store.setChatAgentProvider("custom");
-      } else {
+      } else if (draftApiKey.trim()) {
+        // Only the Custom route sends the key; a Settings server left in place would win.
         store.setCloudTranscriptionBaseUrl(committedBaseUrl);
         store.setCustomTranscriptionApiKey(draftApiKey);
         // Switch before setting the model: a switch files the current model under the
         // outgoing provider and loads the incoming one's, replacing what was typed here.
         store.switchCloudTranscriptionProvider("dictation", "custom");
         store.setCloudTranscriptionModel(draftCustomModel);
-        const remoteUrl = selfHostedRemoteTranscriptionUrl(committedBaseUrl, draftApiKey);
-        store.setRemoteTranscriptionUrl(remoteUrl);
-        if (remoteUrl) {
-          store.setRemoteTranscriptionModel(draftCustomModel);
-          store.setRemoteTranscriptionType("openai-compatible");
-        }
+        store.setRemoteTranscriptionUrl("");
+        store.setCloudTranscriptionMode("byok");
+      } else {
+        // Saved as the Settings self-hosted server, so the Custom URL, key and model are
+        // kept. Custom is still the provider: byok + custom derives the self-hosted mode.
+        store.switchCloudTranscriptionProvider("dictation", "custom");
+        store.setRemoteTranscriptionUrl(committedBaseUrl);
+        store.setRemoteTranscriptionModel(draftCustomModel);
+        store.setRemoteTranscriptionType("openai-compatible");
         store.setCloudTranscriptionMode("byok");
       }
     } else if (assistant) {
--- a/src/components/onboarding/ProviderSetupStep.tsx
+++ b/src/components/onboarding/ProviderSetupStep.tsx
@@ -45,6 +45,12 @@ import {
   rememberPendingLocalModel,
 } from "./pendingLocalModels";
 import { isLocalStageDownloadActive } from "./localDownloadState";
+import {
+  isBlankByokDraft,
+  resolveSavedByokConfig,
+  selfHostedRemoteTranscriptionUrl,
+  type SavedByokConfig,
+} from "./savedByokConfig";
 
 export function SetupStageStepper({ stepId }: { stepId: OnboardingStepId }) {
   const { t } = useTranslation();
@@ -321,30 +327,40 @@ export function ByokProviderStep({
       ),
     [assistant, policy, scope]
   );
-  const initialProvider =
-    providers.find((provider) => provider.id === resumeState?.selectedProvider)?.id ?? "";
-  const initialProviderModels =
-    providers.find((provider) => provider.id === initialProvider)?.models ?? [];
-  const initialModel = initialProviderModels.some(
-    (model) => model.id === resumeState?.selectedModel
+  // The session's draft wins; without one (onboarding was restarted, or an older build
+  // left it blank) the step reopens on what the user already saved, the same order
+  // LocalModelSetupStep uses.
+  const [seed] = useState<SavedByokConfig | null>(() =>
+    resumeState && !isBlankByokDraft(resumeState)
+      ? { draft: resumeState, keyless: false }
+      : resolveSavedByokConfig(stepId, store)
+  );
+  const initialProviderData = providers.find(
+    (provider) => provider.id === seed?.draft.selectedProvider
+  );
+  const initialProvider = initialProviderData?.id ?? "";
+  const initialModel = initialProviderData?.models?.some(
+    (model) => model.id === seed?.draft.selectedModel
   )
-    ? (resumeState?.selectedModel ?? "")
-    : (initialProviderModels[0]?.id ?? "");
+    ? (seed?.draft.selectedModel ?? "")
+    : pickDefaultModelId(initialProviderData);
   const initiallySelfHosted = selfHostedRequested && selfHostedAllowed;
+  // Hosted and self-hosted share the key field, so each mode shows its own saved key.
+  // The Settings self-hosted server has none; a leftover custom key shown there would
+  // move the save onto the keyed route.
+  const credentialFor = (selfHostedMode: boolean, providerId: string) => {
+    if (!selfHostedMode) return providerCredential(providerId, store).value;
+    if (assistant) return store.chatAgentCustomApiKey;
+    return seed?.keyless ? "" : store.customTranscriptionApiKey;
+  };
   const [selfHosted, setSelfHosted] = useState(initiallySelfHosted);
-  const [selectedProvider, setSelectedProvider] = useState(
-    initiallySelfHosted ? "" : initialProvider
-  );
-  const [selectedModel, setSelectedModel] = useState(initiallySelfHosted ? "" : initialModel);
+  const [selectedProvider, setSelectedProvider] = useState(initialProvider);
+  const [selectedModel, setSelectedModel] = useState(initialModel);
   const [draftApiKey, setDraftApiKey] = useState(() =>
-    initiallySelfHosted
-      ? assistant
-        ? store.chatAgentCustomApiKey
-        : store.customTranscriptionApiKey
-      : providerCredential(initialProvider, store).value
+    credentialFor(initiallySelfHosted, initialProvider)
   );
-  const [draftBaseUrl, setDraftBaseUrl] = useState(resumeState?.baseUrl ?? "");
-  const [draftCustomModel, setDraftCustomModel] = useState(resumeState?.customModel ?? "");
+  const [draftBaseUrl, setDraftBaseUrl] = useState(seed?.draft.baseUrl ?? "");
+  const [draftCustomModel, setDraftCustomModel] = useState(seed?.draft.customModel ?? "");
   const [draftCortiClientId, setDraftCortiClientId] = useState(
     initialProvider === "corti" ? store.cortiClientId : ""
   );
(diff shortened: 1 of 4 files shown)


boseq:
## Screenshots: sign-in and the Cloud nudge

Captured from a dev build on a throwaway profile, signed out and using Groq. Before is `main` (`1c123e87`); after is this branch (`166537bf`).

### Clicking the OpenWhispr Cloud tile in Settings → Speech-to-Text

On `main`, the tile restarts onboarding: the control panel shrinks into the compact onboarding window, and onboarding progress is reset. With this PR, a sign-in dialog opens over Settings. The window doesn't reload and no settings change.

| Before (`main`) | After (this PR) |
| --- | --- |
| Before: the Cloud tile restarts onboarding in the compact window | After: the sign-in dialog opens over Settings |

### Nudge after signing in

After sign-in, a user whose dictation runs on their own setup sees this toast once. Open Settings goes to Speech-to-Text → Dictation.

Toast: You're signed in, with an Open Settings button

A real sign-in would create a production account, so for this capture the signed-in state was faked with a local patch that isn't committed, and the policy was set to unmanaged. The toast is the unmodified component.

<details>
<summary>Dark mode</summary>

| Before (`main`) | After (this PR) |
| --- | --- |
| Before, dark: the Cloud tile restarts onboarding | After, dark: the sign-in dialog opens over Settings |

Toast in dark mode: You're signed in, with an Open Settings button

</details>

boseq:
The sign-in half is rebuilt in `1c680a18`: the dialog is gone.

A guest who clicks sign-in is now simply logged out — `requestSignIn` drops `authenticationSkipped` and `skipAuth` and reloads, the same two steps `handleSignOut` takes — so the app shows the re-authentication screen `AppRouter` already shows a signed-out account. No new surface, no onboarding reset, nothing switched, and "Continue without an account" backs out in one click. The Cloud nudge stays, as a hook instead of a component that existed only to host the dialog.

The saved-BYOK half of the PR is unchanged. Screenshots of the new flow are in the comment below.

boseq:
## Screenshots: signing in without restarting onboarding

Captured from a dev build on a throwaway profile — a guest, signed out, dictation on Groq with a saved endpoint. Before is `main` (`1c123e87`), after is this branch (`1c680a18`).

### Clicking the OpenWhispr Cloud tile in Settings → Speech-to-Text

These two are identical, which is the point: the sign-in screen is the one the app already had. What differs is underneath. On `main` the click wipes `onboardingCompleted` and writes a fresh `onboardingSessionV2`, so this is the wizard's first step. With this PR no onboarding key is touched — `onboardingCompleted` stays `"true"`, only the guest flag is dropped — so this is the re-authentication screen `AppRouter` already shows a signed-out account.

The sign-in screen after clicking the Cloud tile, on main and on this PR

### Where each one leads

"Continue without an account" is the quickest way to show that. On `main` you are dropped into the setup wizard and have to walk it again — which is what rewrites the provider setup #2128 is about. With this PR you are back in the app, untouched.

After Continue without an account: the setup wizard on main, the control panel on this PR

### After signing in

Toast: You're signed in, with an Open Settings button

The toast only informs — signing in never switches a setting. A real sign-in would create a production account, so this was rendered with the signed-in state faked by a temporary local patch that is not committed, and the policy set to unmanaged. The toast component itself is unchanged.

<details>
<summary>Dark mode</summary>

Dark: the sign-in screen after clicking the Cloud tile, on main and on this PR

Dark: after Continue without an account, the wizard on main and the control panel on this PR

Dark: toast saying You're signed in, with an Open Settings button

</details>

boseq:
Pushed three fixes from a review pass over the whole diff (`3b2d5c0b`, `2c9d73ae`, `7e425cb8`). All three are the same defect, found and narrowed three times.

**A stored API key could be sent to an endpoint it was not saved for.** The self-hosted BYOK step prefills the key you already have for a saved endpoint. The gate on that prefill asked "was a keyed endpoint ever saved" — which is the wrong question, in three separate ways:

- `resolveSavedByokConfig` returns nothing for local, OpenWhispr Cloud, enterprise or a first run, and the seed read that absence as "keyed", so a leftover `customTranscriptionApiKey` from an abandoned endpoint was offered for a freshly typed one. The flag is now the positive `usesCustomKey`.
- The assistant branch of `credentialFor` never consulted the flag at all. Latent while the secret hydrated asynchronously and left the field empty — but this branch's late-arrival effect fills it once the secret lands, so a user who has since moved to a hosted provider got their old agent key typed into the next endpoint they entered.
- Even gated, the key was tied to *whether* something was saved, not to *which endpoint*. With a saved keyed endpoint, replacing the URL kept the old token in the field, and confirming sent it to the new server.

The rule is now that a stored key belongs to one endpoint: it is offered only while the URL in the field still resolves to the endpoint it was saved for, under the same scheme rule `commitAndProceed` applies. When the URL changes, a stranded prefill is dropped by byte-identity against the stored value, so nothing typed is ever cleared. A saved keyed endpoint reopened with its URL untouched still shows its key, on both steps.

Also in `2c9d73ae`: the Cloud nudge reads `transcriptionMode` through `selectPolicyEffectiveSettings`, so a managed user whose policy already clamps dictation onto Cloud is not told their dictation still runs on their own setup.

**No screenshots this round** — none of this changes a layout or a control. The only thing visible is a key field being empty where it used to be pre-filled, in the specific saved-config cases above, which a screenshot would not show any more clearly than the description.

Tests: 4,714 passing, 0 failures; `quality-check` clean. Each new test was run against the pre-fix source and confirmed to fail there.

Known follow-up, deliberately not bundled here: the nudge is keyed to the user's dictation mode, not to the tile they clicked. It fires from any of the Cloud entry points, but its copy and its Open Settings action always refer to dictation — so a guest who clicks the tile under AI Models, Note Recording or Audio Upload gets a toast about the wrong scope, and a guest whose dictation is already on Cloud gets none at all. Making it scope-aware needs a scope threaded through all five `requestSignIn` call sites plus new copy in 11 locales, which does not belong in a credential fix.

*(Edited: an earlier version of this paragraph said those tiles show no toast. That was wrong — it describes only the already-on-Cloud case.)*

Chadpiha (approved):
Approving. I went through both halves end to end and pushed three small commits onto the branch along the way (5cfc2830, d79b5208, 2d5671aa); please pull before you touch it again.

Sign-in half. Traced requestSignIn through AppRouter: with onboarding completed the wizard branch is unreachable, so dropping the guest flags lands on the existing ReauthenticationScreen. Email, Google, Microsoft, Apple and SSO all return to the control panel with no settings write anywhere on the path, and "Continue without an account" restores the guest flag. All five former useStartOnboarding call sites are gone with no leftovers.

BYOK half. resolveSavedByokConfig mirrors resolveTranscriptionRoute's precedence exactly. Running the new byokSavedConfig tests against the merge-base fails 32 of 43 cases for behavioural reasons, and all 12 mutations I tried against the seed, key gating, commit order and nudge logic were caught by a named test.

What I changed on the branch

Reopen on the saved card whichever setup tile was clicked; hosted commit clears the Custom URL; endpoint identity via new URL().href; nudge waits through a policy load error; resetOnboardingProgress drops the sign-in marker; withHttpsScheme hoisted; ToastActionButton shared.
Reverted a wipe I had introduced: the hosted commit no longer clears remoteTranscriptionUrl. That key is shared with the Upload tab's Self-hosted panel, and the post-step fan-out already derives providers, so the wipe only cost Upload its server.
Tidied the card-sync effect and moved the BYOK step mount helper and managed-policy fixture into shared test modules.

Gates: 4543 pass / 0 fail, quality-check and i18n:check clean, all 11 locales translated with "OpenWhispr Cloud" untouched.

Follow-ups, @boseq could you please fix these in another PR.

- The keyed-custom commit still has to clear remoteTranscriptionUrl, and the fan-out then derives self-hosted with an empty URL, so Settings shows a blank Self-hosted panel and uploads are blocked until the user re-enters a server. Root cause is deriveTranscriptionMode mapping byok + custom to self-hosted, which you already list as a follow-up. Worth an issue.
- Closing the re-auth panel with Cmd+W hides it to the tray with _onboardingActive still set, so the dictation hotkey is silently dead until the panel is reopened. Pre-existing for signed-out account users, but guests can now reach it from a Settings tile. Releasing the gate when the mounted route is re-auth would close it.
- Minor: the nudge's Open Settings action does nothing if Settings is already open, and a late policy recovery can fire the toast up to 15 minutes after sign-in.

Thanks for the thorough test plan; it made this straightforward to verify.

PR #2146, opened 2026-09-11
fix(ai): send x-opencode-session on OpenCode Go requests

### LINKED ISSUE #2145: OpenCode Go requests fail because OpenWhispr omits x-opencode-session
**Describe the bug**

OpenWhispr cannot use an OpenCode Go subscription because requests to the OpenCode Go API do not include the required `x-opencode-session` header.

The request fails with:

```text
Test failed: Error from provider (Console Go): Request is missing x-opencode-session and cannot be routed efficiently. Please see https://opencode.ai/docs/go/#where-can-i-use-it
```

OpenCode Go's documentation says clients must send a stable session ID in `x-opencode-session` for each conversation.

**To Reproduce**

1. Subscribe to OpenCode Go and create an API key.
2. Configure OpenWhispr with the OpenCode Go OpenAI-compatible endpoint:
   `https://opencode.ai/zen/go/v1`
3. Enter the OpenCode Go API key and select an available Go model.
4. Run an OpenWhispr AI operation or provider test.
5. Observe the `MissingSessionID` error.

**Expected behavior**

OpenWhispr should generate or reuse a stable session identifier for each conversation and send it on requests as:

```http
x-opencode-session: <stable-session-id>
```

The same ID should be reused for requests belonging to the same conversation, while separate conversations should receive different IDs.

**Actual behavior**

OpenCode Go rejects the request with HTTP 400 because `x-opencode-session` is not present.

**Screenshots**

Image

- OS: MacOS 26.6.2

**Additional context**

This is related to [#2041](https://github.com/OpenWhispr/openwhispr/issues/2041), which concerns OpenCode Go endpoint routing, and [PR #2095](https://github.com/OpenWhispr/openwhispr/pull/2095). However, the endpoint-routing change does not address the missing session header.

Related OpenCode report: [anomalyco/opencode#47763](https://github.com/anomalyco/opencode/issues/47763).

OpenCode Go documentation: https://opencode.ai/docs/go/#where-can-i-use-it

Please do not include API keys in logs or screenshots.

L4XB:
Closes #2145.

## What was wrong

OpenCode Go routes each conversation by a stable `x-opencode-session` id and rejects requests without one (`HTTP 400`: "Request is missing x-opencode-session and cannot be routed efficiently"). OpenWhispr's OpenAI-compatible provider never sent that header, so a custom provider pointed at `https://opencode.ai/zen/go/v1` failed every AI operation and the provider test, as reported.

## Change

In the OpenAI provider, a base URL whose host is `opencode.ai` (or a subdomain) is recognised by host, the same way the Mistral/DeepSeek/Cerebras dialects are, and every POST of that call carries `x-opencode-session: <uuid>`. One cleanup or formatting call is one conversation, so the id is created once per call and reused across the endpoint fallback (`/responses` → `/chat/completions`), the parameter fallback, and retries; the next call gets a fresh id, which matches the "same conversation, same id; separate conversations, different ids" rule in the OpenCode Go docs. Other endpoints do not get the header.

This is independent of #2095 (endpoint routing for OpenCode Go); it only adds the header, so both can land in either order.

## Tests

New `test/services/openaiOpenCodeSession.test.js` (same fetch-mock style as the endpoint retry test):

- a call against `https://opencode.ai/zen/go/v1` sends a UUID `x-opencode-session` on both the `/responses` attempt and the `/chat/completions` fallback, and the two ids are identical;
- a second call uses a different id;
- another custom endpoint sends no such header.

```
node --import tsx --test test/services/openaiOpenCodeSession.test.js test/services/openaiEndpointRetry.test.js   # 4 passed
npm test   # 3565 pass; the 144 failures are identical on main (realtime streaming, Hyprland, GNOME portal helpers) and unrelated
eslint on the two files   # clean
```

Note for anyone running the suite locally: on Node 24.2 the named exports of the TypeScript provider are not visible to the test runner and every provider test fails with "Cannot read properties of undefined (reading 'call')"; Node 24.21 (what `.nvmrc` → `24` resolves to on CI) runs them fine.

### DIFF
--- a/src/services/ReasoningService.ts
+++ b/src/services/ReasoningService.ts
@@ -33,6 +33,7 @@ import {
 } from "./ai/chatRequestBody";
 import { getModelFamilyConstraints } from "./ai/modelFamilyConstraints";
 import { detectEndpointDialect } from "./ai/thinkingSuppressionDialects";
+import { openCodeSessionHeaders } from "./ai/openCodeSession";
 import { createStreamingThinkFilter } from "./ai/streamingThinkFilter";
 import { extractApiErrorMessage } from "./ai/apiErrorMessage";
 import { clearTinfoilClientCache } from "./ai/tinfoilClient";
@@ -319,6 +320,9 @@ class ReasoningService extends BaseReasoningService {
       requestBody: JSON.stringify(requestBody).substring(0, 200),
     });
 
+    // Minted before the retry loop so every attempt of this call is one conversation.
+    const openCodeHeaders = openCodeSessionHeaders(endpoint);
+
     const requestGeneration = this.requestCancellationGeneration;
     const response = await withRetry(async () => {
       if (requestGeneration !== this.requestCancellationGeneration) {
@@ -331,6 +335,7 @@ class ReasoningService extends BaseReasoningService {
       try {
         const headers: Record<string, string> = {
           "Content-Type": "application/json",
+          ...openCodeHeaders,
         };
         if (apiKey) {
           headers["Authorization"] = `Bearer ${apiKey}`;
@@ -621,6 +626,7 @@ class ReasoningService extends BaseReasoningService {
 
     const headers: Record<string, string> = {
       "Content-Type": "application/json",
+      ...openCodeSessionHeaders(endpoint),
     };
     if (apiKey) {
       headers["Authorization"] = `Bearer ${apiKey}`;
--- a/src/services/ai/inferenceProviders/openai.ts
+++ b/src/services/ai/inferenceProviders/openai.ts
@@ -14,6 +14,7 @@ import { detectEndpointDialect } from "../thinkingSuppressionDialects";
 import { getLlmRequestTimeoutSeconds } from "../../../helpers/llmRequestTimeout.js";
 import { extractApiErrorMessage } from "../apiErrorMessage";
 import { wrapCleanupTranscript } from "../../../config/prompts";
+import { openCodeSessionHeaders } from "../openCodeSession";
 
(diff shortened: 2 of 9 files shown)


Chadpiha (approved):
Approving, thanks for this @L4XB! Right diagnosis, correct mechanism. Verified against a mock that enforces the real contract (400 whenever the header is absent): cleanup completes, one id spans the endpoint fallback, param fallback and retries, and other hosts are untouched.

I pushed three commits, two of them needed for CI:

Test isolation + formatting. format:check was failing on the new test file. The three tests also shared one base, and openai.ts caches the winning endpoint per base in a module-level Map, so test 1 only passed because it ran first. Each test now has its own base.
Shared helper. The session decision moved to src/services/ai/openCodeSession.ts, with the host match extracted to matchesHost in urlUtils — detectEndpointDialect uses it too, so one implementation instead of three. No behaviour change.
The header extended to the transports it was missing. custom is a cloud chat provider, so the Voice Assistant panel and every chat surface stream through providers.ts, and the self-hosted lane through ReasoningService's own fetches — all three hit the same base and still got 400 MissingSessionID. createOpenAI already takes a headers option. Each site now has a test capturing the headers actually put on the wire.
Final state: typecheck and format:check clean, 4304 tests / 0 failures, and all four call sites mutation-tested.

One correction for the record: GET /zen/go/v1/models returns 200 live with no key and no session header, so the onboarding connection test and the model picker were never affected. The failing "provider test" is the Prompt Studio button, which your original change fixes.

Left alone, non-blocking: a fresh id per call means no prompt caching (a stable id per inferenceScope would fix it); the gate is host-wide, so Zen gets a header it doesn't need; 

Thanks for including a test that actually fails without the production change — made this quick to verify.

PR #2147, opened 2026-09-11
fix: walk GitHub release pages for prefix tags

### LINKED ISSUE #2005: Bug: Fresh Windows builds silently skip windows-key-listener download, breaking Hold/Push-to-Talk

stantheman0128:
`tagPrefix` release lookup only requested `/releases?per_page=50`. `windows-key-listener-v1.0.0` is now on page 2, so a fresh Windows `prebuild:win` prints a warning, exits 0, and Hold / Push-to-Talk never gets `windows-key-listener.exe`. The same helper feeds fast-paste, text-monitor, meeting-aec, mic-listener, and system-audio downloads.

This keeps prefix search (latest matching tag) and follows GitHub `Link: rel=next` up to 20 pages of 100. Exact `tag` and `/releases/latest` paths are unchanged.

Fixes #2005

## Test plan

- [x] `node --test test/helpers/download-utils.test.js` (4 passed)
- [x] First page of 50 app tags, second page holds `windows-key-listener-v1.0.0`
- [x] Live `fetchLatestRelease(..., { tagPrefix: 'windows-key-listener-v' })` on this Win11 box returns `windows-key-listener-v1.0.0` and the zip asset

## Evidence

HEAD (first 50 releases, 2026-09-12):

```
page 1: no windows-key-listener-* tags
page 2: windows-key-listener-v1.0.0
```

After the helper change:

```
{
  tag: windows-key-listener-v1.0.0,
  assets: [windows-key-listener-win32-x64.zip]
}
```

I did not pin every script to an exact tag. That would hide the next helper that ages off page 1.

## AI assistance

Drafted with Cursor. I checked the live release pages on this machine and wrote the pagination plus tests.

---

**Update:** maintainers added commits on top of this PR:
- The test moved to `test/scripts/fetchLatestRelease.test.js` (3 tests over a mocked `https.get`). `test/helpers/download-utils.test.js` no longer exists.
- `afterPack` now fails a Windows build without `windows-key-listener.exe`, tested in `test/scripts/afterPackWindowsKeyListener.test.js`.
- CI compiles the listener instead of downloading it first.

More detail in my comment below.

### DIFF
--- a/scripts/lib/download-utils.js
+++ b/scripts/lib/download-utils.js
@@ -7,14 +7,16 @@ const REQUEST_TIMEOUT = 30000;
 const MAX_RETRIES = 3;
 const RETRY_DELAY = 2000;
 const MAX_REDIRECTS = 5;
+const RELEASE_PAGE_SIZE = 100;
+const MAX_RELEASE_PAGES = 20;
 
 /**
- * Fetch JSON from a URL with proper error handling.
+ * Fetch JSON and response headers from a URL.
  * @param {string} url - URL to fetch
  * @param {number} [redirectCount=0] - Current redirect count (internal use)
- * @returns {Promise<object>} - Parsed JSON response
+ * @returns {Promise<{json: object, headers: object}>}
  */
-function fetchJson(url, redirectCount = 0) {
+function fetchJsonResponse(url, redirectCount = 0) {
   return new Promise((resolve, reject) => {
     if (redirectCount > MAX_REDIRECTS) {
       reject(new Error("Too many redirects"));
@@ -46,7 +48,7 @@ function fetchJson(url, redirectCount = 0) {
             return;
           }
           const redirectUrl = location.startsWith("/") ? new URL(location, url).href : location;
-          fetchJson(redirectUrl, redirectCount + 1)
+          fetchJsonResponse(redirectUrl, redirectCount + 1)
             .then(resolve)
             .catch(reject);
           return;
@@ -61,7 +63,7 @@ function fetchJson(url, redirectCount = 0) {
         res.on("data", (chunk) => (data += chunk));
         res.on("end", () => {
           try {
-            resolve(JSON.parse(data));
+            resolve({ json: JSON.parse(data), headers: res.headers });
           } catch (e) {
             reject(new Error(`Failed to parse JSON: ${e.message}`));
           }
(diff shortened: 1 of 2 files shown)


boseq:
Thanks @stantheman0128, this does it. I checked it on a Windows 11 VM with no compiler on PATH: `main` can't find `windows-key-listener-v1.0.0` and exits 0, and with this branch the binary downloads.

I pushed three commits on top:

- 630d5ab6 makes `afterPack` fail a Windows build that's missing `windows-key-listener.exe`, so a failed download can't quietly ship a build where Hold and the default Control+Super never fire.
- b2191558 tests the pagination through a mocked `https.get` instead of the `fetchPage` option, so the tests read the real `link` header and fail against the old single-page lookup. It also drops a few branches that never run and moves the test to `test/scripts/`.
- 386e0a71 removes the CI step that downloaded the listener before `prebuild:win`. Now that the download works again, it would make CI skip compiling `windows-key-listener.c` on a cache miss and ship the prebuilt release binary instead.

Closing #2006 in favour of this one. It fixes the same bug by pinning exact tags, and @B0-KA tracked down the cause in #2005. The runtime side (Hold and modifier-only hotkeys doing nothing when the listener is missing) is in #2239.

stantheman0128:
Thanks for running it on a bare Win11 VM. That is the case that was failing.

The three commits look right:
- afterPack should fail the Windows build if `windows-key-listener.exe` is missing
- pagination tests should go through the real `Link` header, not a `fetchPage` hook
- CI should not download the listener before `prebuild:win`, or a cache miss ships the prebuilt binary instead of compiling `windows-key-listener.c`

Closing #2006 in favor of this is correct. Pinning today's tags and walking `rel=next` are not substitutes.

I'll leave the Hold / modifier-only empty state to #2239.

boseq (approved):

PR #2155, opened 2026-09-11
fix(llama): size the local context to the request instead of capping it at 16384

### LINKED ISSUE #1203: macOS watchdog kernel panic when local LLM polishing begins (M4 Air, 16 GB, v1.7.5)
**Describe the bug**

OpenWhispr's local transcription completes successfully, but when local LLM polishing/cleanup begins, the entire Mac becomes unresponsive and eventually requires or undergoes a reboot. This happened with two very different local models, so it does not appear to be caused solely by model size:

- `gpt-oss-20b-mxfp4`
- `Llama-3.2-3B-Instruct-Q4_K_M.gguf` (approximately 2.02 GB)

The Llama model is correctly Q4_K_M quantized. The same failure occurs at the same pipeline stage with both models.

macOS generated a kernel panic after the latest incident:

```
watchdog timeout: no checkins from watchdogd in 91 seconds
```

The panic report said:

- Compressor: 18% of compressed-pages limit (OK)
- Segments: 16% of limit (OK)
- 9 swapfiles with swap space reported as OK
- Panicked task: `kernel_task`

This suggests the incident was not ordinary memory exhaustion. I did not find an explicit GPU Restart or Metal fault string in the panic report.

macOS also generated repeated resource diagnostics specifically for OpenWhispr:

- 2,147.54 MB of file-backed writes over 1,800 seconds
- 8,590.00 MB over 338 seconds (about 25.41 MB/s)
- 2,147.50 MB over 994 seconds

The 8.59 GB event's heaviest OpenWhispr stack was repeatedly blocked in `write()` through Electron/Node. OpenWhispr's own recorded footprint in that report was only approximately 113-153 MB. The Llama GGUF finished being written shortly before the later freeze.

**To reproduce**

1. Run OpenWhispr 1.7.5 on an M4 MacBook Air with 16 GB unified memory.
2. Configure local Whisper `base` for transcription.
3. Configure local dictation cleanup/polishing with either:
   - `gpt-oss-20b-mxfp4`, or
   - `Llama-3.2-3B-Instruct-Q4_K_M.gguf`.
4. Dictate a short passage.
5. Observe that transcription finishes successfully.
6. When local polishing begins, the entire computer freezes.
7. After the latest occurrence, macOS recorded a watchdog kernel panic.

**Expected behavior**

The quantized 3B cleanup model should polish the transcript without making macOS unresponsive. If local inference cannot start safely, OpenWhispr should fail gracefully, stop the model process, and show an error.

**Desktop**

- Device: MacBook Air (Mac16,12)
- Chip: Apple M4, 10 cores
- Memory: 16 GB unified memory
- OS: macOS 26.1 (25B78)
- OpenWhispr: 1.7.5 (arm64)
- Local transcription: Whisper `base`
- Local cleanup model: `Llama-3.2-3B-Instruct-Q4_K_M.gguf`
- Llama model size: 2,019,377,696 bytes

**Additional context**

The issue is tightly correlated with the local polishing stage:

- Opening OpenWhispr does not cause the freeze.
- Local transcription works and produces text.
- The freeze begins only when cleanup/polishing starts.
- Changing from a 20B model to a correctly quantized 3B model does not prevent it.
- Plain local transcription without local polishing is stable.

Relevant local reports are available under `/Library/Logs/DiagnosticReports/`:

- `OpenWhispr_2026-07-16-135347_*.diag`
- `OpenWhispr_2026-07-16-135926_*.diag`
- `OpenWhispr_2026-07-16-230643_*.diag`
- `panic-full-2026-07-16-232329.0002.panic`

I have not attached the raw reports yet because they may contain private machine identifiers. I can provide redacted excerpts or reports if needed.

Please let me know which additional logs or debug-mode reproduction data would be most useful. I am avoiding another local-polishing reproduction for now because it caused a system-level watchdog panic.

### LINKED ISSUE #2142: Generate Notes fails on meetings over ~80 min with any local model: SERVER_CONTEXT_SIZE hard-caps llama-server at 16384 tokens
## Describe the bug

Generate Notes fails on any meeting longer than roughly 80 minutes when Note Formatting is set to a bundled local model. The app surfaces a raw `llama-server` JSON error to the user:

```
Inference failed: llama-server returned status 400: {"error":{"code":400,
"message":"request (19566 tokens) exceeds the available context size (16384 tokens),
try increasing it","type":"exceed_context_size_error","n_prompt_tokens":19566,"n_ctx":16384}}
```

This is **not** a model limitation and **not** a user misconfiguration. `n_ctx: 16384` is our own constant.

## Root cause

`SERVER_CONTEXT_SIZE = 16384` in `src/helpers/modelManagerBridge.js:21` caps every llama-server start:

```js
serverOptions(modelInfo) {
  return {
    contextSize: Math.min(SERVER_CONTEXT_SIZE, modelInfo.model.contextLength || SERVER_CONTEXT_SIZE),
    ...
```

It reaches `--ctx-size` in `src/helpers/llamaServer.js:165-166`.

**All 35 bundled local models declare a context length of 32,768 or more** (11 at 262,144, 11 at 131,072, 10 at 32,768, 3 at 128,000). `Math.min` therefore always returns 16,384, so the cap is the binding constraint for every local model, never the model itself. Llama 3.1 8B, the model in the customer report, natively supports 131,072.

Nothing on the note-generation path bounds or chunks the prompt. `src/stores/actionProcessingStore.ts:165` sends note content + meeting context + the entire transcript to `reasoningService.processText` with no length check, and `src/components/notes/PersonalNotesView.tsx:744-750` concatenates the full transcript unconditionally.

### Why the cap exists

Added in 466145fe ("bound llama-server context size to prevent OOM system freeze", #1236) to fix #1203, where an unbounded context allocated a ~15 GB KV cache and hard-froze 16 GB Macs into a watchdog kernel panic. **The cap is protecting against a real and severe failure and must not simply be raised to a bigger constant.**

### History

`SERVER_CONTEXT_SIZE` has been touched by exactly one commit in the entire repo history and has held the value 16384 since 2026-07-18. It ships in every release from **v1.7.6** onward (13 tags).

## Reproduction

Confirmed against the bundled `llama-server-darwin-arm64` (b9763) shipped in OpenWhispr.app, using the app's exact spawn arguments and environment:

```
LLAMA_ARG_FIT=off llama-server-darwin-arm64 --model <gguf> --host 127.0.0.1 \
  --port 8231 --threads 4 --ctx-size 16384 --jinja --n-gpu-layers 99
```

Posting a 19,571-token chat completion reproduces the customer's error byte for byte:

```
{"error":{"code":400,"message":"request (19571 tokens) exceeds the available context size (16384 tokens), try increasing it","type":"exceed_context_size_error","n_prompt_tokens":19571,"n_ctx":16384}}
```

The **identical request succeeds** when the same server is started with `--ctx-size 32768`:

```
usage: {'completion_tokens': 50, 'prompt_tokens': 19571, 'total_tokens': 19621}
```

The cap is the sole cause.

## Blast radius

* **Platform independent.** The cap lives in `serverOptions()`, above the per-platform branch in `_doStart`. macOS, Windows and Linux are equally affected, despite both customer reports arriving from macOS.
* **Every bundled local model**, not a subset.
* **Effective ceiling is roughly 80 minutes of speech**, measured at about 1.33 tokens per word against a usable transcript budget of roughly 15,000 tokens after the system prompt.
* **Unaffected:** cloud routes, dictation cleanup, selection editing (capped at 6,000 characters), and title generation (capped at 2,000 characters).
* **Sibling failure on the self-hosted route:** #2008 reports Generate Notes returning the transcript instead of notes on longer transcripts via Ollama. That is consistent with the same missing capability, except that route silently truncates the prompt (dropping the system prompt) instead of erroring. One missing notion of a context budget, two different symptoms.

## Secondary defects found while investigating

1. **The raw error reaches the user.** `src/stores/actionProcessingStore.ts:204` surfaces `err.message` verbatim, so a customer sees llama.cpp JSON. Even with the limit unchanged, this needs a translated, actionable message.
2. **`ReasoningConfig.contextSize` is a dead field on the local path.** It is declared in `src/services/BaseReasoningService.ts:11` and set in `src/helpers/audioManager.js:2721` (`Math.max(config?.contextSize || 0, 16384)`), but `serverOptions()` never reads a caller value and `src/services/localReasoningBridge.js` never forwards it. The selection-edit floor looks like it controls something and does not. Either honour it or remove it.

## Suggested direction

Not a bigger constant. Three parts, roughly in priority order:

1. **Size the context to the request, bounded by memory.** Restart llama-server when a request needs more context than the running instance has, at `min(needed + headroom, model.contextLength, memorySafeCeiling)`. The bundled binary already ships `--fit` and `--fit-ctx` for exactly this, though we force `LLAMA_ARG_FIT=off` in `_buildEnv` because probing costs about 70s at startup, and `--fit` only adjusts arguments we leave unset.
2. **Preflight and fail gracefully.** Estimate the prompt before sending. If it cannot fit even at the ceiling, show a real message naming the model and the option, not llama.cpp JSON.
3. **Map-reduce long transcripts.** Summarise in chunks, then summarise the summaries. This is the only path that makes a 3 hour meeting work locally on a 16 GB machine, and it also removes the silent-truncation class behind #2008.

## Customer impact

Two paying customers, one month apart, both blocked on the core feature:

* 2026-08-07, 17,784 tokens (Linear CUS-90). The 2026-08-25 reply told this customer *"We've since increased the token count in our recent updates"* and recommended updating the app. **That was incorrect** — the constant has never changed — so the customer was sent back to a build with the identical limit.
* 2026-09-03, 19,566 tokens (Linear CUS-170). A barrister who explicitly declined the cloud workaround: *"I want to run it locally, not using the cloud."* Local is a professional confidentiality requirement for this segment, not a preference, so "use cloud" is not an available answer.

Both reports reached Linear and neither reached this repo, which is why the bug has been open for five weeks with no engineering ticket.

Chadpiha:
## Summary

Generate Notes failed on any meeting over roughly 80 minutes whenever Note Formatting used a bundled local model, and showed the user a raw llama.cpp JSON body. `SERVER_CONTEXT_SIZE = 16384` capped every llama-server start; all 35 bundled models declare 32768 or more, so the `Math.min` against the model's own context never did anything.

The constant was not gratuitous — it fixed #1203, where an unbounded context allocated a ~15 GB Metal-wired KV cache and froze a 16 GB Mac into a watchdog kernel panic. **So this does not raise it, it derives it.**

Fixes #2142.

## Verification

Reproduced byte-for-byte against the bundled `llama-server-darwin-arm64` (b9763) with Qwen3.5-9B-Q4_K_M and the app's exact spawn args. Identical 20,514-token request:

| `--ctx-size` | result |
|---|---|
| 16384 | HTTP 400 `exceed_context_size_error` |
| 32768 | HTTP 200 |

End-to-end through the app's own path after the fix, same machine, same model, same prompt:

```
Resolved llama-server context ceiling {
  model: 'qwen3.5-9b-q4_k_m', architecture: 'qwen35',
  kvBytesPerToken: 135168, ceiling: 118272,
  reason: 'memory-bound', cacheRamMiB: 2048
}
server ctx = 32768
notes preview: "# Meeting Notes: Billing Service Migration Plan Review..."
```

And the refusal path, with the ceiling forced down to a small machine's:

> This is too long for Qwen3.5 9B to read on this computer. It needs about 22058 tokens of context, but only 16384 fit in memory. Try a cloud model, or run the action on a shorter selection.

## Why the ceiling comes from the GGUF

KV cost is exactly linear in context but varies ~12x across the registry and is **anti-correlated with file size**, so no RAM-tier or model-size heuristic is sound:

| model | file | KV/token |
|---|---|---|
| Qwen3.5-9B-Q4_K_M | 5.9 GB | **32 KiB** (measured: 512 MiB @ 16384, 2048 MiB @ 65536) |
| Llama-3.2-3B-Q4_K_M | 2.0 GB | **112 KiB** (reproduces the ~15 GB at 131072 from #1203) |
| Gemma-3-12B-Q4_K_M | 7.3 GB | **384 KiB** |

`ggufMetadata.js` reads it out of the model's own header (~150 lines, header only, bounded, `null` on any failure, no new dependency). `llamaContextPolicy.js` turns it into a ceiling:

```
budget   = totalMemory x 0.50 - 2 GiB
kvBudget = budget - weights - draft - 384 MiB - (recurrent ? 512 MiB : 0)
ceiling  = clamp(kvBudget / (kvPerToken + 4096), 16384, min(trainedCtx, gpuCap))
```

On the #1203 machine and model that yields **32256 tokens at a 5.6 GiB footprint**, against the ~15 GiB that panicked — and against the ~9.6 GiB llama.cpp's own `--fit` would permit there.

**Every failure path returns the historical 16384**, so the worst case of this whole mechanism is today's behaviour. An 8 GB Mac running a 3B model gets exactly what it gets now.

## How a request is sized

The estimate only picks the size the server *starts* at. `POST /tokenize` then prices the prompt exactly — about 13 ms for 85 KB, chat template included via `/apply-template` — and `/props` reports the window actually in force rather than the one we asked for. If the exact count does not fit, the server grows once; if even the ceiling is too small, the request is refused before it is sent rather than burning minutes of prefill to earn a 400.

Contexts snap to a ladder (`16384, 32768, 65536, 131072, 262144`) so near-identical requests cannot each cost a restart, and `start()` grows the context but never shrinks it, so a short dictation right after a long note reuses the window.

**Short requests keep today's behaviour exactly: no measurement, no restart, no extra round trip.** That guard is the first test in `modelManagerContextSizing.test.js`, because every dictation goes through this same path.

## Also in here

- **`--cache-ram` is bounded.** llama-server otherwise reserves 8192 MiB of host RAM for the prompt cache on top of the KV cache, which would undo the budget. Capped, not disabled (`0` forces a full re-prefill every request).
- **`ReasoningConfig.contextSize` is no longer dead.** It was declared, written by selection editing, and then rebuilt away in `localReasoningBridge`, so it never reached anything — and could not have worked anyway, since `start()` refused to restart on options-only changes. Now forwarded as a floor, with a test pinning the wiring.
- **Vulkan retries at a smaller context before dropping to CPU.** The ceiling is computed against system RAM, but a discrete GPU's memory is not probed, so a big-RAM/small-card machine can now ask for more VRAM than exists. That is a risk this change introduces; halving the context beats a silent 10-30x CPU fallback. The rung is omitted when there is nothing to step down to, so existing failure paths are unchanged.
- **Separate commit — silent output truncation.** Note formatting passed no `maxTokens`, so it inherited the generic 2048-token default (~1,500 words) and long summaries were cut off and saved with no error. Now given a real budget, which fixes the cloud path too since the value travels in the shared config. `requireCompleteOutput` stays off deliberately, with a test saying so: a selection edit sets it because a partial replacement corrupts the user's own text, but a clipped summary is still worth keeping.
- **`translationCoverage` now scans `messageKey` literals.** They are passed to `t()` as variables and were invisible to the existing scan; this retroactively guards the 29 existing sites.

## Two corrections to the issue

1. **#2008 is not a sibling failure.** It is a GLM-5.3 always-reasoning / prompt-adherence problem on the Ollama route, sharing no code path with `SERVER_CONTEXT_SIZE`. Posted on the issue; **#2008 should not be closed by this PR.**
2. The `~70s` cost attributed to `--fit` in the comment at `llamaServer.js` did not reproduce — it measured 4.4 s here. `--fit` is still not the right lever (it commits the whole device budget and its default 1024 MiB margin is thinner than this design's on a 16 GB Mac), but the comment's number is wrong.

## Known limits

- A summary that genuinely exceeds 4096 output tokens can still be clipped silently. Closing that needs a way to warn without discarding the note, which this path does not have.
- Local **chat** streaming bypasses `runInference` entirely (`ReasoningService` starts the server and talks to the port from the renderer), so it keeps the baseline 16384 and can still surface a raw 400. Natural follow-up: a `contextSize` hint on the `llama-server-start` IPC.
- Map-reduce chunking for transcripts beyond the ceiling is deliberately out of scope.

## Test plan

- `npm test` — 3934 passing, 0 failing, 196 skipped (DB tests skip locally).
- `npm run typecheck`, `npm run lint` (0 errors), `npm run i18n:check`, `prettier --check` on all touched files.
- New: `llamaContextPolicy` (27 cases, including a named #1203 regression asserting the footprint stays under 40% of RAM), `ggufMetadata` (headers built in memory — no model download), `llamaServerContext`, `modelManagerContextSizing`, `localInferenceError`, `actionProcessingNoteOutput`.
- Manual, macOS: long meeting note formats successfully and the debug log shows the chosen ceiling and one restart; a short dictation immediately after causes no restart; the refusal path shows the translated message with no JSON.

## Still owed

Manual pass on Windows/Vulkan and Linux — the step-down rung and the `discreteGpuUnverified` cap are unit-tested but have not run on real hardware.


Chadpiha:
Tested locally on macOS arm64 (48 GB M5 Pro) with Qwen3.5-9B-Q4_K_M set as the Note Formatting model.

Grow path — 85k-char meeting transcript:

Ceiling resolved from the GGUF: architecture: qwen35, kvBytesPerToken: 135168, ceiling: 118272, reason: memory-bound
Server started at --ctx-size 65536, startupTimeMs: 3017
Note generated: 25,777 prompt tokens in, 457 output tokens of the 4096 budget, truncated = 0
--cache-ram 2048 reached the server and the prompt cache grew to 1187 MiB under that bound

Short-request invariant — a dictation immediately afterwards skipped the preflight and reused the grown window: no restart, no /tokenize, 2.16 s end to end.

The restart costs ~3 s. The 59 s total for the note is dominated by prompt eval (38 s / 25,777 tokens), which is inherent to a prompt that size rather than anything this PR adds.

One observation: the 3-chars/token estimate over-counted by 11% (28,698 vs 25,777 actual), crossing a rung boundary and starting the server at 65536 where 32768 would have sufficed. Safe — the ceiling bounds it — but the window only grows within a session, so that doubling persists.

Not covered by this pass: the trim and refusal paths (this machine's ceiling is too high to reach them without forcing it), and the Windows/Vulkan and Linux passes still listed as owed.

Issue #2136, opened 2026-09-11
Linux: Live dictation bubble can be hovered only once to cancel


szotsaki:
**Describe the bug**
Live dictation bubble can be hovered only once to cancel

**To Reproduce**

1. Under KDE Wayland, set up a custom URL for dictation. This may not be necessary but it does change the live bubble format (since streaming is not available)
2. Press Ctrl + Super to start dictation, and observe that the dictation bubble appeared in the lower right corner.
3. Hover with your mouse over the bubble as if you were to cancel it. The X appears, but do not click on it.
4. Move your mouse away and wait until the X disappears.
5. Hover it again to click X, but this time X doesn't appear.

There is nothing written in the debug log during this time.

**Expected behavior**
The closing X appears for each time, and the dictation can be cancelled. 

- OS: Linux with KDE Wayland
- Version: 1.10.0

PR #2156, opened 2026-09-11
fix: keep retry bubble visible during recording

### LINKED ISSUE #2141: Linux: after dictation "Retry", the floating bubble becomes invisible
**Describe the bug**
Linux: after a failed dictation when clicking on "Retry", the floating bubble becomes invisible.

**To Reproduce**

1. Set the live dictation floating bubble to appear only when the dictation happens.
1. Under Linux KDE Wayland environment, start dictation by pressing the assigned hotkeys, but do not emit any sound
2. Press the hotkey again to stop the dictation. Thus, nothing was recorded.
3. Observe that a bubble pops up, telling that was a transcription error, and there's a retry button in it. Press that button.
4. Now continue dictating, but this time the floating bubble is invisible.
5. Press the hotkey again to finish the dictation and see the results. It's a proof that the dictation was happening in the background, even when the bubble was not visible.

**Expected behavior**
During a retry, the floating bubble is visible.

OS: Linux with KDE Wayland
Version: 1.10.0

dajiaohuang:
## Summary

Fixes #2141 by keeping the error-to-pill handoff from auto-hiding the native window while recording or processing is active. This preserves the floating bubble when Retry starts a new recording.

## Validation

- 
ode --import tsx --test test/helpers/dictationErrorPillHandoff.test.js
- 
pm run lint
- 
pm run typecheck
- git diff --check

The KDE Wayland reproduction was not available on this host; the regression is covered at the handoff boundary.

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -310,6 +310,7 @@ export default function App() {
     toastCount,
     isCommandMenuOpen,
     isCompactPill: windowFitsCompactPill,
+    hasLiveActivity: isVisuallyProcessing || isRecording,
     assistantOpen: assistant.open,
     assistantMounted: assistant.mounted,
     assistantOpenRef,
--- a/src/hooks/useMainWindowSizeOwner.js
+++ b/src/hooks/useMainWindowSizeOwner.js
@@ -23,6 +23,7 @@ export function useMainWindowSizeOwner({
   toastCount,
   isCommandMenuOpen,
   isCompactPill,
+  hasLiveActivity,
   assistantOpen,
   assistantMounted,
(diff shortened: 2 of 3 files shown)


boseq:
Thanks @dajiaohuang, and sorry for the overlap here. Closing in favour of #2161, which lands the same fix for #2141 alongside a fix for #2136 (the Linux hover/click-through bug on the same pill), with a regression test and an end-to-end reproduction on a Linux VM.

Your diagnosis was right and it is the approach #2161 takes — you are credited in its description. Happy to reopen this if you would rather land yours first.

PR #2159, opened 2026-09-11
feat: New note button, pill and tray quick actions, and settings polish

boseq:
## Summary

- **Settings — org-managed notice.** The "managed by your organization" banner sat inset (`mx-4` inside the content padding), 40px down, with no gap before the section. It now renders through a `notice` slot on the settings shell: the shared `InfoBox` at the full content width, starting just below the floating close button in both regular and compact layouts, with the sections' own `mb-6` rhythm before the first section, and a shield icon.
- **Settings — focused close button.** Radix focuses the first tabbable element when a dialog opens; in the settings shell that is the X, which sits before the sidebar, and its ring used `focus:`, so every open showed a focused X. The dialog now takes the initial focus itself (focus trap and Esc unchanged; the first Tab reaches the X), and the X's ring only shows for keyboard focus. Scoped to `SidebarModal` (Settings is its only consumer); the shared `DialogContent` is untouched.
- **Notes — "Enhanced" is now "AI Summary".** Matches the "Generate AI Summary" action that produces it, in all locales.
- **Quick actions — pill and tray.** The dictation pill's right-click menu gains *Start meeting recording* (idle only, and only when policy allows meeting transcription). The tray menu leads with the pill's quick actions: *Start listening* / *Stop listening*, *Ask assistant*, *Start meeting recording*.
  - **Two of the three need nothing from the renderer.** *Start listening* goes straight to the same dictation start every other input path uses, and *Start meeting recording* calls `windowManager.startManualMeeting()` — the one gated entry for manual meetings, exactly as the meeting hotkey has always called it. The recording those open is policy-gated where it actually begins, in the control panel, so routing them through the dictation renderer bought nothing.
  - **Only *Ask assistant* asks the renderer**, because only the renderer can open the assistant panel and only it knows the policy and panel state the pill's menu gates the item on. `trayActionPolicy.js` is a pure seam deciding what it does with the request: run, or refuse with the reason. A refusal surfaces the pill first, so the message is visible even when the pill was hidden.
  - **A policy that never loaded is not an org restriction.** Policy fails closed while it is loading or after a failed fetch, so an offline user was being told their organization had blocked something it never blocked. The resolver takes `policyResolved` and says the settings have not loaded yet instead; a real restriction still names itself (`common.policyAgentRestricted`).
  - `startManualMeeting()` returns early when a recording is live and brings that note forward instead of opening an empty one over it. The engine's recording session carries the note id, which is the only signal that actually means "recording" (`_meetingModeActive` can stick, and the stored meeting note id goes stale after a streaming stop).
  - The tray's listen item is a real toggle, rebuilt when dictation starts or stops.
- **Windows — no paste into our own window.** A dictation started from the tray captures our own menu-owner window as the paste target, and the fast-paste helper force-activates that handle before Ctrl+V, which would pull the foreground onto OpenWhispr, drop the keystroke, and then restore the previous clipboard over the dictated text. `windowsPasteTarget.js` is a pure, unit-tested seam (the repo's `dockPolicy` / `meetingMicGate` pattern, for logic no non-Windows machine can exercise) that drops a target which is ours but is not one of our `BrowserWindow`s. Real windows — the control panel's note editor — stay valid targets, so dictating into a note still pastes there.
- **Notes topbar — New note split button.** `+ New note | ▾`, with the chevron offering *Note*, *Assistant chat*, and *Team space*.
  - **Placement follows the platform.** On Windows and Linux it sits just right of the search bar, with the window controls still at the far end. macOS has no window controls in this bar, so the button takes the far right.
  - **Matched to the search bar beside it**, which previously shared none of its styling: the same translucent fill that lifts on hover, under the same hairline, with the same keyboard focus ring. The search bar gains the border it lacked. Those overrides are scoped to this instance, so the note editor's Share control — which shares the same split-button definition — keeps the pill it already had.
  - The main segment is the notes page's existing New note: it creates a note in the current space or folder and starts recording. *Assistant chat* opens a new chat in the Chat tab, and is hidden when policy turns the assistant off (the Chat tab is hidden then too). *Team space* opens the existing team-space dialog, for users who can create one.
  - The chat and team-space items keep the focus they hand to what they open; a new note and a dismissed menu return focus to the chevron, as Radix does by default.
  - The topbar exposes an actions slot that `PersonalNotesView` portals into, so the handlers and the notes-onboarding state stay in the view; the button is absent during notes onboarding and in the side-panel layout.
  - The trailing grid column no longer shrinks below its content, so the button and the window controls are never clipped; the search yields at narrow widths.
  - The "can create a team space" rule moves from `SpacesTree` into `useCanCreateTeamSpace`, and the Share control's split-pill classes move to `ui/splitButton`, so both call sites share one definition.

## A note on scope

An earlier revision of this branch mirrored the renderer's gating into the main process for all three tray entries: a `start-meeting` push channel, a `tray-action-refused` channel, a renderer-readiness handshake, and the meeting hotkey rerouted through the renderer for symmetry. That last change was the source of two regressions — the hotkey went silent whenever the dictation window was not mounted, which a signed-in launch does while policy resolves.

It was also unnecessary: the meeting recording is already policy-gated at the control panel's recording mount, so the round trip bought a slightly earlier refusal message rather than any enforcement. All of it has been removed (net −284 lines). The meeting hotkey is byte-for-byte what it was before this branch, and the only surviving renderer round trip is the one that genuinely needs it.

## Test plan

- [x] `npm run typecheck`, `npm run lint` (0 errors), `npm run i18n:check`
- [x] Full suite green, with no `ERR_DLOPEN_FAILED` (those failures on some machines are a broken local `better-sqlite3`, not this PR)
- [x] Tests added: the Windows paste-target seam; the tray action policy (org restriction, busy pill, and an unresolved policy not blaming the org); the tray menu's three entries and where each one goes; the manual-meeting guard during a live recording, with and without a note; the assistant request taking no focus and refusing under hotkey capture; the pill menu's meeting item; `NewNoteMenu`'s items (including the assistant hidden by policy); `useCanCreateTeamSpace`
- [x] `useTrayQuickActions` is mounted for real in its test (not SSR-rendered), so the wiring — not just the pure resolver — is covered: a swapped or forgotten input fails loudly
- [x] macOS dev build driven over CDP: Settings opened with ⌘, and by click keeps focus on the dialog with no ring on the X; org notice full width in regular and compact layouts; pill menu shows *Start meeting recording* and fits its window; *AI Summary* toggle on a summarized note; New note and its menu in the topbar; the button is absent in the side-panel layout and on other pages
- [x] Placement on both platforms: macOS — New note flush to the far right (12px gap to the window edge); Windows 11 VM — the button sits right of the search at x 881–1014 with the window controls after it at 1073–1177
- [x] New note and the search bar measured in the running app: identical fill, border and radius
- [ ] Tray *Ask assistant* under an org policy that blocks the assistant: the refusal explains itself and no panel opens
- [ ] Tray *Start meeting recording* while a meeting is already recording brings the live note forward
- [ ] Windows: dictation started from the tray pastes into the app you were using (or leaves the text recoverable)
- [ ] *Team space* with a signed-in account that can manage a workspace

https://claude.ai/code/session_013hZY5WHN4aNYZA2fAowZrj

### DIFF
--- a/main.js
+++ b/main.js
@@ -1161,10 +1161,11 @@ async function startApp() {
 
   // Set up meeting mode hotkey
   const isMeetingPress = createHotkeyRepeatGate();
+  // Through the renderer, like the tray's item: it owns the meeting policy the
+  // main process cannot see, so both entries refuse the same way.
   const meetingHotkeyCallback = () => {
     if (!isMeetingPress()) return;
-    debugLogger.info("Meeting hotkey triggered", {}, "meeting");
-    windowManager.startManualMeeting();
+    windowManager.sendStartMeeting();
   };
 
   const savedMeetingKey = environmentManager.getMeetingKey?.() || "";
@@ -1728,7 +1729,7 @@ async function startApp() {
       } else if (hotkeyManager.slotHasHotkey("translation", key)) {
         windowManager.sendToggleTranslation();
       } else if (hotkeyManager.slotHasHotkey("meeting", key)) {
-        windowManager.startManualMeeting();
+        windowManager.sendStartMeeting();
       }
     };
 
--- a/main.js
+++ b/main.js
@@ -1331,6 +1331,9 @@ async function startApp() {
 
   trayManager.setWindows(windowManager.mainWindow, windowManager.controlPanelWindow);
   trayManager.setWindowManager(windowManager);
+  // The tray's listen item is a toggle, so it has to rebuild when dictation
+  // starts or stops.
+  windowManager.onDictationStateChanged = () => trayManager.updateTrayMenu();
   trayManager.setCreateControlPanelCallback(() => windowManager.createControlPanelWindow());
   await trayManager.createTray();
 
--- a/main.js
+++ b/main.js
@@ -1163,11 +1163,8 @@ async function startApp() {
   const isMeetingPress = createHotkeyRepeatGate();
   const meetingHotkeyCallback = () => {
     if (!isMeetingPress()) return;
-    if (hotkeyManager.isInListeningMode()) return;
-    // Fail closed during onboarding, like every other hotkey slot.
-    if (!windowManager.isMeetingInputAllowed()) return;
     debugLogger.info("Meeting hotkey triggered", {}, "meeting");
-    meetingDetectionEngine?.startManualMeeting();
+    windowManager.startManualMeeting();
   };
 
   const savedMeetingKey = environmentManager.getMeetingKey?.() || "";
@@ -1728,9 +1725,7 @@ async function startApp() {
       } else if (hotkeyManager.slotHasHotkey("translation", key)) {
         windowManager.sendToggleTranslation();
       } else if (hotkeyManager.slotHasHotkey("meeting", key)) {
-        if (!hotkeyManager.isInListeningMode() && windowManager.isMeetingInputAllowed()) {
-          meetingDetectionEngine?.startManualMeeting();
-        }
+        windowManager.startManualMeeting();
       }
     };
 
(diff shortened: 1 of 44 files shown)


boseq:
Superseded by https://github.com/OpenWhispr/openwhispr/pull/2163: identical branch and commits, but hosted in this repository so the follow-up fixes can stack on it as a PR here (GitHub can only target a base branch that lives in the same repo). Closing this one.

PR #2160, opened 2026-09-11
fix(dictation): cut the Windows mic cold start on the first hotkey press

Chadpiha:
Windows users report 1–2s between pressing the dictation hotkey and the recording actually starting. Regression introduced in v1.9.0 (#1670), still present on v1.10.0.

## Root cause

`getAudioConstraints()` has a `cachedMicDeviceId` fast path that is unreachable on Windows. `resolveSystemDefaultMicDevice()` filtered `physicalInputs` by `deviceId !== "default"`, which leaves in Chromium's Windows-only `communications` role alias — it carries the same label as the real device, so `exactMatches`/`containedMatches` always saw 2 candidates, required exactly 1, and fell through to the uncacheable Chromium `default` id.

With nothing cached, every mic open re-ran `enumerateDevices()` plus `getSystemDefaultMicrophone()`, which shells out to `powershell.exe` with `Add-Type -TypeDefinition <C# source>` (runtime C# compilation, 3s timeout, only a 30s cache). `prepareMicCapture` waits on that before `getUserMedia`, and `startRecording()`'s `take()` waits on the whole thing — so `isRecording`, the visible start of the recording, sits behind it.

macOS/Linux have no `communications` device, so they resolved `native-exact`, cached, and never felt this.

## Changes

1. `microphoneSelection.js` — exclude both Chromium aliases (`default`, `communications`) from `physicalInputs`, so Windows resolves `native-exact` and the result is cacheable.
2. `useAudioRecording.js` — pre-resolve the device in the mount effect. `cacheMicrophoneDeviceId()` already existed but was only reachable from `warmupStreamingConnection()`, i.e. never for local/non-streaming users, so without this the first press of every session still paid the full cost.

## Tests

Both written test-first and watched fail. Every existing case in `microphoneSelection.test.js` used a macOS-shaped device list, which is why this was never caught.

- Windows device topology now resolves `native-exact` / cacheable (was `chromium-default-unmatched` / not cacheable).
- New `useAudioRecordingMicPrecache.test.js` pins that the mount effect pre-caches with streaming off.

## Not covered

A user whose OS default name matches 0 or 2+ enumerated labels (two identical USB mics) still resolves to the uncacheable Chromium default and re-runs the lookup every press past the 30s TTL. Closing that means not awaiting a cold lookup on the hot path, or replacing `Add-Type` with a prebuilt native helper.

Separately: `"Recording start timing"` starts its clock inside `startRecording()`, after every wait above — so Windows logs look clean even when the user waited seconds. Worth fixing before asking reporters for logs.

### DIFF
--- a/src/helpers/audioManager.js
+++ b/src/helpers/audioManager.js
@@ -559,13 +559,7 @@ class AudioManager {
 
     // Invalidate the pinned mic device when the OS adds/removes/suspends inputs.
     // Otherwise wake-after-idle keeps requesting a stale deviceId that yields silence.
-    this._onDeviceChange = () => {
-      this.cachedMicDeviceId = null;
-      this._micWarmedAt = 0;
-      this.rejectedMicDeviceId = null;
-      this.cancelPreparedMicCapture();
-      this.micStreamHold.drop();
-    };
+    this._onDeviceChange = () => this._handleDeviceChange();
     navigator.mediaDevices?.addEventListener?.("devicechange", this._onDeviceChange);
     this.recordingStartTime = null;
     this.reasoningAvailabilityCache = { value: false, expiresAt: 0 };
@@ -583,6 +577,7 @@ class AudioManager {
     this.streamingTextDebounce = null;
     this.cachedMicDeviceId = null;
     this.rejectedMicDeviceId = null;
+    this._refreshSystemDefaultOnNextResolve = false;
     this.persistentAudioContext = null;
     this.workletModuleLoaded = false;
     this.workletBlobUrl = null;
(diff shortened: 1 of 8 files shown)


gabrielste1n:
Thank you for this. The root-cause analysis was exactly right, and your note that the timing log starts inside `startRecording()` matched what a Windows user asked us independently this week.

Since you opened it, #2178 landed with the caching and device-change parts: the default-mic lookup now runs once at startup and is kept until a devicechange refresh, callers share one in-flight lookup, and the renderer refreshes in the background when devices change. That also covers the case you listed under "Not covered", since the lookup no longer repeats on the hotkey path at all.

The two pieces #2178 did not have, the `communications` alias filter and the mount pre-cache, are exactly right. We ported them with your tests to #2180 with you as the commit author, so the credit follows the code. Closing this one in favour of that PR. Thanks again.

PR #2161, opened 2026-09-11
fix(linux): keep the dictation pill hoverable after the first hover, and visible after Retry

### LINKED ISSUE #2136: Linux: Live dictation bubble can be hovered only once to cancel

### LINKED ISSUE #2141: Linux: after dictation "Retry", the floating bubble becomes invisible

boseq:
## Summary

Two Linux reports against the floating dictation pill, both reproduced end to end on an Ubuntu 26.04 GNOME Wayland VM running the released 1.10.0 package. They look similar but have separate root causes, so they are fixed separately here.

### #2136 — the cancel X only appears on the first hover

`setMainWindowInteractivity` returned the pill to click-through on mouse-leave with `setIgnoreMouseEvents(true, { forward: true })`. Electron only honours `forward` on macOS; on Linux it sets a 1×1 X11 input region and forwards nothing, so the window never received another `mouseenter`. The cancel X appeared once and never came back, leaving no way to cancel a recording from the pill.

Only macOS can forward hover through a click-through window, so Linux now keeps normal hit-testing like Windows already does. This is the same Electron limitation that stranded the meeting card's buttons in #1456, and the agent companion pill already guards against it.

Why the *first* hover works at all: Chromium resets the X11 input region whenever the window's bounds change (`DesktopWindowTreeHostLinux::OnBoundsChanged` → `UpdateFrameHints` → `SetInputRegion(nullopt)`), so the resize behind the error card silently re-enables hit-testing for exactly one hover.

### #2141 — the pill disappears after Retry

Retry starts the next dictation, and a successful start dismisses the error card. Leaving the error footprint releases the pill visibility handoff, whose `shouldAutoHide` only asked whether "Auto-hide when idle" was enabled — so it hid the entire pill window while the new recording was running. The recording continued with nothing on screen until the user pressed the hotkey again. The same happens when the hotkey, rather than Retry, starts a dictation while the card is up.

The handoff now also requires that nothing is recording or processing, which is the idle condition `App.jsx`'s own auto-hide effect already applies.

## Test plan

- `node --import tsx --test test/helpers/mainWindowInteractivity.test.js test/hooks/dictationErrorAutoHide.test.js test/helpers/pillVisibilityHandoff.test.js test/helpers/notificationInteractivity.test.js test/helpers/windowManagerMeetingNotification.test.js` — 38 passing.
- Both new tests were confirmed to fail on the unpatched tree and pass with the fix; the macOS, Windows, destroyed-window and idle-auto-hide cases pass either way.
- `npm run lint`, `npm run typecheck`.

### Reproduced and verified in a VM

Ubuntu 26.04, GNOME Wayland (the app forces XWayland), released 1.10.0 `.deb`, driven with real `/dev/uinput` pointer events so X11 hit-testing genuinely applies, with the Electron main process instrumented to log every `setIgnoreMouseEvents` / `hide` / `setBounds` call.

Before, #2136: hover #1 logged `setIgnoreMouseEvents(false)` and revealed the cancel button; after mouse-leave returned the window to click-through, hovers #2 and #3 produced **no native calls at all** and no cancel X.

Before, #2141: after the Retry click the main process logged `hide()` and reported the window not visible for the rest of the recording, while a PipeWire capture stream stayed live — the recording ran invisibly. Nothing was written to the debug log, matching the report.

After, running a Linux package built from this branch in the same VM:

- #2136: hover #1, #2 and #3 each reveal an interactive cancel button (`present: true`, `ariaHidden: null`, `disabled: false`), and every native call in the run is `setIgnoreMouseEvents(ignore: false)` — the window is never put back into click-through.
- #2141: after the Retry click the main process logs only the resize back to the pill box, with **no `hide` call**, and reports the window visible 1.5s and 5s later while the capture stream is live. Auto-hide still fires when genuinely idle — before the dictation and again once the retried recording stops — so the setting keeps working.

## Credit

The #2141 fix takes the same approach as #2156 by @dajiaohuang, which identified the same root cause — the dictation-error pill handoff must not hand the window back to auto-hide while a dictation is still running — and which was reviewed before this implementation was written.

This PR additionally fixes #2136, adds a regression test for #2141 that fails on the unpatched tree, and validates both fixes end to end against a Linux package built from this branch. #2156 is closed in favour of this one.

Closes #2136
Closes #2141

https://claude.ai/code/session_01Rs6g9RUbpHn5Hg8gBRPdWx

### DIFF
--- a/preload.js
+++ b/preload.js
@@ -541,6 +541,11 @@ contextBridge.exposeInMainWorld("electronAPI", {
   ackMainWindowResizeMask: (token) => ipcRenderer.send("main-window-resize-mask-ready", token),
   setMainWindowInteractivity: (interactive) =>
     ipcRenderer.invoke("set-main-window-interactivity", interactive),
+  getMainWindowPointerPosition: () => ipcRenderer.invoke("get-main-window-pointer-position"),
+  onMainWindowVisibilityChanged: registerListener(
+    "main-window-visibility-changed",
+    (callback) => (_event, visible) => callback(visible)
+  ),
   setNotificationInteractivity: (interactive) =>
     ipcRenderer.invoke("set-notification-interactivity", interactive),
   resizeMainWindow: (sizeKey) => ipcRenderer.invoke("resize-main-window", sizeKey),
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -310,6 +310,7 @@ export default function App() {
     toastCount,
     isCommandMenuOpen,
     isCompactPill: windowFitsCompactPill,
+    isDictationActive: isRecording || isVisuallyProcessing,
     assistantOpen: assistant.open,
     assistantMounted: assistant.mounted,
     assistantOpenRef,
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -5,6 +5,7 @@ import { useToast } from "./components/ui/useToast";
 import { useHotkey } from "./hooks/useHotkey";
 import { formatHotkeyListLabel } from "./utils/hotkeys";
 import { useWindowDrag } from "./hooks/useWindowDrag";
+import { useLinuxPillInteractivity } from "./hooks/useLinuxPillInteractivity";
 import { useAudioRecording } from "./hooks/useAudioRecording";
 import { useAssistantPanel } from "./hooks/useAssistantPanel";
 import { useOnboardingAssistantDemo } from "./hooks/useOnboardingAssistantDemo";
@@ -60,6 +61,7 @@ export default function App() {
   const [isHovered, setIsHovered] = useState(false);
   const [isCommandMenuOpen, setIsCommandMenuOpen] = useState(false);
   const buttonRef = useRef(null);
+  const pillPresenceRef = useRef(null);
   const { toast, dismiss, toastCount, dictationErrorActionCount, dismissByPresentation } =
     useToast();
   const { t } = useTranslation();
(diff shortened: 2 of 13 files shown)


Chadpiha (approved):
Reviewed end to end with three subagents, including my two follow-up commits:
- 3617e85 — Preserve click-through around the dictation pill.
- 8e0b71c — Restore pill hover with native input regions.

The final implementation addresses repeated hover/cancel and Retry visibility while preserving click-through padding, overlay interactions, and idle auto-hide. No blocking issues found.

Approved, thanks @boseq

PR #2162, opened 2026-09-11
fix(meetings): keep the recording timer across note switches and offer the AI summary when a meeting auto-ends

boseq:
Fixes two meeting-recording bugs. This was originally stacked on #2163; now that PR has landed on `main` as 48ec6308, this is rebased onto `main` and the diff is the two original commits plus one review follow-up.

## The two bugs

**1. Switching notes during a recording reset the timer to 00:00.** The elapsed seconds lived in `NoteRecordControl` as a count of `setInterval` ticks, and `PersonalNotesView` keys `NoteEditor` by note id — so leaving the recording note and coming back remounted the control and restarted the count from zero. The session's start timestamp now lives in the meeting store (`recordingStartedAt`) and the header derives its elapsed time from that, which also stops the timer drifting while the window is throttled.

**2. When a meeting auto-ended, no "Generate AI Summary" appeared.** Two separate causes:

- The offer was gated on the Transcript tab. Every detected or quick-action meeting opens its note on the Notes tab, and nothing switches tabs when a recording ends, so the offer was hidden exactly when a meeting ended on its own. The condition moved into a pure `shouldOfferMeetingSummary` helper that no longer looks at the open tab.
- The auto-end card now carries a second **Generate AI Summary** action whenever the note has a transcript and no summary yet: main decides that from the note itself, and the action routes through the existing note-navigation queue so the control panel is surfaced and the note opened before the notes view runs its existing summary action.

## Review follow-up (third commit)

- The "Recording stopped automatically" toast stays. It only ever fired on the paths that end a recording without a restart card (the stop did not persist a result, main rejected the completion, or the card failed to show), so removing it left those paths silent for anyone not looking at the note.
- The summary request is handled by one effect in `PersonalNotesView`, beside the existing `meetingRecordingRequest` effect, instead of two new `NoteEditor` props. The onboarding early return moves below the hooks it would otherwise skip.
- `MEETING_AUTO_END_ACTIONS` is defined once in `meetingAutoEndLifecycle` and imported by the engine.

## Notes for review

- The IPC changes are additive: `MeetingAutoEndAction` gains `"summary"`, the card payload gains an optional `canSummarize`, and the note-navigation payload gains an optional `generateSummary`. Existing callers are untouched.
- `AUTO_END_NOTIFICATION_WINDOW_SIZE` grows from 128 to 144px tall for the second action, and the localized-overflow guard now measures both action labels across all 11 locales (Arabic included, which it previously skipped).
- Tests added or extended: the timer's remount behaviour, the pure summary-offer predicate, `canSummarize` surviving the window-manager hop, the engine's summary response, and the widened size guard. Typecheck, lint, `i18n:check` and the full suite pass against `main`.

## Screenshot

The auto-end card carrying the new action, rendered from the real component, model and translations. The dashed outline is the actual 480x144 notification window: English light and dark, German (the longest label), Arabic (RTL), French and Japanese.

Auto-end card with the Generate AI Summary action


boseq:
Superseded by https://github.com/boseq/openwhispr/pull/7, which stacks this work directly on `feat/ui-polish-quick-actions` (#2159) instead of `main`. That branch lives only in the fork, and GitHub can only retarget a PR's base to a branch in the same repository, so the stacked PR has to live there too. Closing this one to avoid two PRs carrying the same commits.

gabrielste1n:
Thanks @boseq — this nails the timer reset, and offering the summary from the auto-end card is a real improvement.

Since #2163 landed on `main` as 48ec6308, I retargeted this PR to `main`, rebased your two commits onto it (unchanged apart from the rebase), and pushed a third commit with the review follow-ups now described in the PR body: the auto-ended toast is back on the paths where no card appears, the summary request is handled in `PersonalNotesView` rather than through `NoteEditor` props, and the action set is defined once. Typecheck, lint, `i18n:check` and the full suite pass against `main`. Marking it ready for review.

gabrielste1n (approved):
Traced end to end against main: timer derives from the store's start stamp and survives note switches, the summary offer reaches both tabs and the auto-end card, and the no-card toast is preserved. Typecheck, lint, i18n and the full suite pass on main.

PR #2163, opened 2026-09-11
feat: New note button, pill and tray quick actions, and settings polish

boseq:
## Summary

- **Settings — org-managed notice.** The "managed by your organization" banner sat inset (`mx-4` inside the content padding) with no gap before the first section. It now renders through a `notice` slot on the settings shell as the shared `InfoBox`, at the full content width, just below the floating close button in both regular and compact layouts, with a shield icon.
- **Settings — focused close button.** Radix focuses the first tabbable element when a dialog opens; in the settings shell that is the X, so every open showed a focused X. The dialog now takes the initial focus itself (focus trap and Esc unchanged; the first Tab reaches the X), and the X's ring only shows for keyboard focus. Scoped to `SidebarModal`, whose only consumer is Settings.
- **Notes — "Enhanced" is now "AI Summary".** Matches the "Generate AI Summary" action that produces it, in all locales.
- **Quick actions — pill and tray.** The dictation pill's right-click menu gains *Start meeting recording* (idle only, and only when policy allows meeting transcription). The tray menu leads with *Start listening* / *Stop listening* (a real toggle, rebuilt when dictation starts or stops), *Ask assistant*, and *Start meeting recording*.
  - **Each entry starts where it runs.** Listen goes to the existing `sendStartDictation()` / `sendStopDictation()`. A meeting goes to `windowManager.startManualMeeting()`, now the one entry for manual meetings: both meeting-hotkey sites in `main.js`, the pill menu (through a new `start-manual-meeting` IPC) and the tray all use it. It fails closed during onboarding and hotkey capture, and returns early when a recording is live, bringing that note forward instead of opening an empty one over it; the engine's recording session now carries the note id for that.
  - **Only *Ask assistant* asks the renderer**, since only the dictation renderer can open the panel. `trayActionPolicy.js` decides `run` or `refuse`, with policy outranking a busy pill. A refusal surfaces the pill and explains itself: `common.policyAgentRestricted`, `common.policyUnresolved` while policy is still loading or failed to load, or "finish the current dictation first".
  - **What the tray does not gate.** A listen click that hits a main-process gate (transcription still processing, onboarding, hotkey capture) is a quiet no-op, as is ordinary for a tray menu. The meeting item is shown regardless of org policy: like the meeting hotkey, main creates the meeting note and opens it, and the recording itself is refused at the control panel's recording mount with the restricted-by-organization message.
- **Windows — no paste into our own window.** A dictation started from the tray captures our own menu-owner window as the paste target, and the fast-paste helper would force-activate it before Ctrl+V, pulling the foreground onto OpenWhispr, dropping the keystroke, and restoring the clipboard over the dictated text. `windowsPasteTarget.js` is a pure, unit-tested seam that drops a target which is ours but is not one of our `BrowserWindow`s. Real windows (the control panel's note editor) stay valid targets. `getWinTargetHwnd` becomes `getWinTarget`, which carries the window's identity.
- **Notes topbar — New note split button.** `+ New note | ▾`, with the chevron offering *Note*, *Assistant chat* and *Team space*.
  - On Windows and Linux it sits right of the search bar with the window controls at the far end; macOS has no window controls in this bar, so the button takes the far right. The trailing grid column no longer shrinks below its content, so nothing is clipped. The search bar gains a hairline border so the two controls read as one family.
  - The main segment is the notes page's existing New note: it creates a note in the current space or folder and starts recording. *Assistant chat* opens a new chat in the Chat tab and is hidden when policy turns the assistant off. *Team space* opens the existing team-space dialog for users who can create one. The chat and team-space items keep the focus they hand to what they open; a new note or a dismissed menu returns focus to the chevron.
  - The topbar exposes an actions slot that `PersonalNotesView` portals into, so the handlers and the notes-onboarding state stay in the view; the button is absent during notes onboarding and in the side-panel layout.
  - The "can create a team space" rule moves from `SpacesTree` into `useCanCreateTeamSpace`, and the Share control's split-pill classes move to `ui/splitButton`, so both call sites share one definition.

## Test plan

- [x] `npm run typecheck`, `npm run lint` (0 errors), `npm run i18n:check`, prettier
- [x] Full suite green at 794ab051: 4319 tests, 0 failures; also 4487 tests, 0 failures with current `main` merged in
- [x] Tests added: the Windows paste-target seam; the tray action policy (refusal keys, unresolved policy, policy outranking busy); `useTrayQuickActions` mounted for real; the tray toggle and routing; `startManualMeeting` failing closed and the live-recording guard; the assistant request taking no focus and refusing under hotkey capture; the pill menu's meeting item; `NewNoteMenu`'s items, including the assistant hidden by policy; `useCanCreateTeamSpace`
- [x] macOS dev build driven over CDP: Settings opened with ⌘, and by click keeps focus on the dialog with no ring on the X; org notice full width in regular and compact layouts; pill menu shows *Start meeting recording* and fits its window; *AI Summary* toggle on a summarized note; New note and its menu in the topbar; the button is absent in the side-panel layout and on other pages
- [x] Placement on both platforms: macOS — New note flush to the far right; Windows 11 VM — the button right of the search with the window controls after it
- [ ] *Start meeting recording* while a meeting is already recording brings the live note forward (unit-tested; not yet checked by hand)
- [ ] Windows: dictation started from the tray pastes into the app you were using (unit-tested seam; not yet checked by hand)
- [ ] *Team space* with a signed-in account that can manage a workspace

---

Moved from #2159 — same branch, same commits.


gabrielste1n (approved):
Reviewed end to end at 794ab051 in an isolated worktree. Typecheck, lint, prettier and i18n:check pass; 4319 tests with 0 failures at head, 4487 with 0 failures after merging current main (clean merge). Traced the tray, meeting-start, Windows paste-target, settings focus and New note paths; no defects found. Updated the description to match the final architecture (tray entries start where they run; only Ask assistant asks the renderer).

PR #2171, opened 2026-09-13
perf(chat): flush streamed content to state at most every 32 ms

greatcoat:
**Problem.** Every streamed token writes chat state, and both chat surfaces re-parse the whole answer through react-markdown on each write, so parse cost scales with token count rather than answer length.

**Fix.** Buffer streamed text and flush it to state at most every 32 ms. Every exit from the stream loop flushes synchronously first, so completion, explicit cancel and unmount all leave exactly the text they leave today; the error path cancels the timer instead.

## Problem

Measured on 1.10.0 with the GFM table fix (#2131): 1.71 ms per chunk, 355 ms of parsing across one 4 KB answer. Not a visible defect, but pure waste that grows with every plugin added to the renderer.

## Root cause

`useChatStreaming.ts` calls `setMessages` inside the `for await` chunk loop for every `content` chunk. React commits after each awaited iteration, so each token is a full re-render of the markdown tree in `ChatMessage` and `AssistantPanel`.

## Fix

Three closures inside `sendToAI`, in the file's existing `setTimeout`/`clearTimeout` idiom: schedule one flush per interval on each token, flush synchronously before a tool step and on every exit from the loop, cancel on error. The timer, the buffer and the assistant message id are all per-request closures, so a cancel-then-resend inside one interval cannot write into the next reply. No new module; the existing `useDebouncedCallback` is trailing-edge and would never flush a stream that does not pause.

The one observable difference: on an explicit cancel or an unmount, the last few tokens that arrived inside the current interval appear a frame later than before. The final on-screen and persisted text is identical.

**Deliberately untouched:** the renderer's `components` map is still recreated per render; the `tool_calls`/`tool_result` writes are unbuffered (they are rare and must be immediate).

## Tests

- `useChatStreamingFlush.test.js`, fail-first, driving the real hook through the repo's Vite SSR harness: RED on main `expected at most 3 content writes for 50 tokens, got 50`; GREEN here. Plus a guard for the catch-path cancel, shown failing with `error text shown, got "tok0 "` when the cancel line is removed. The guard's fixture errors one tick after the token in `pull()`; erroring in `start()` discards the queued token and never reaches the hook.
- `useChatStreamingCancellation.test.js` passing: it is the regression net for every exit path (empty completion, cancel before a token, cancel during setup, unmount mid-stream, raw-stream cancel). One edit: its unmount-mid-stream test waits on real time for the partial token instead of microtask ticks, which a timer-based flush made insufficient. Its assertions are unchanged.
- Full suite on Node 24: 4446 tests, 4226 pass, 0 fail, 219 skipped (platform-gated tests, unchanged). `npm run quality-check` exits 0.

## Provenance

Follow-up from the chat-table fix (#2131), found during 1.10.0 release testing.


Chadpiha (approved):
Approved. Reviewed the streaming flow through chat and the Assistant panel, including cancellation, unmount, errors, tool transitions, persistence, and response delivery. No blocking issues found.
Added regression coverage for progressive flushing, cancellation/unmount with buffered text, error cleanup, and tool metadata. All 16 focused tests and quality checks pass. Mutation checks confirmed the new tests fail when the guarded behavior is removed.

PR #2172, opened 2026-09-13
fix(notes): summarise long recordings in parts when the local model cannot fit them

### LINKED ISSUE #2142: Generate Notes fails on meetings over ~80 min with any local model: SERVER_CONTEXT_SIZE hard-caps llama-server at 16384 tokens

greatcoat:
## Summary

#2155 replaced the flat 16,384-token cap with a per-machine, per-model ceiling and turned every overflow into a translated refusal. That closes #2142 on paper, but not on the machine class that reported it. Run through the landed policy (`resolveContextCeiling`: RAM × 0.5 − 2 GiB, minus weights and compute), a **16 GB Mac** gets:

| model | 8 GB | 16 GB | 24 GB | 32 GB |
|---|---|---|---|---|
| Qwen3.5 9B (recommended, default) | 16384 | **16384** | 27648 | 58880 |
| Llama 3.1 8B (the CUS-170 customer's model) | 16384 | **16384** | 37376 | 69120 |
| Gemma 4 E4B (recommended) | 16384 | **16384** | 29184 | 53760 |
| Mistral Nemo 12B (recommended) | 16384 | **16384** | 16384 | 38912 |
| gpt-oss 20B (recommended) | 16384 | **16384** | 16384 | 30208 |
| Llama 3.2 3B (recommended) | 16384 | 32256 | 68608 | 104448 |
| Qwen3.5 4B | 16384 | 19456 | 51200 | 82944 |
| LFM2.5 1.2B (recommended) | 16384 | 32768 | 32768 | 32768 |

Five of the six recommended models stay at exactly the old constant on 16 GB. The two customers in #2142 sent 17,784- and 19,566-token prompts; after #2155 they get a polite sentence, not their notes. The ceiling is right where it is (#1203 was a kernel panic on this exact machine class), so this PR does the third part of the direction in #2142: **stop sending the whole transcript at once.**

Refs #2142. Follows #2155.

## What it does

- **Anything that fits today still runs exactly as today.** One request, the same prompt, plus one IPC read of the ceiling. That guard is the first test.
- **Route check first.** Only when the action is not in cloud mode and the resolved note-formatting mode is `local`. Cloud, BYOK, enterprise and self-hosted are byte-for-byte unchanged. (#2008 is a different bug, per the correction on #2142.)
- **Budget read.** New IPC `get-local-context-budget(modelId)` returns the ceiling `contextCeiling()` already computes (the GGUF read is memoised there) plus the model's display name. If it fails, the action runs the single request, so this can never be worse than #2155.
- **Fit test** with the same chars-per-token rule the main process uses. Fits → single request. If that request still comes back `CONTEXT_TOO_LARGE` (the exact tokenizer disagreed with the estimate), the action falls through to parts instead of failing.
- **Parts.** The transcript (or a plain note's body) is packed into parts along line boundaries at 85% of the per-part budget. Every part carries the meeting-context block, so speaker attribution holds, and a fixed part-notes prompt: detailed working notes for *this part only*, keep specifics and owners, no title, no guessing identities. Manual notes are held for the final pass; they are the user's emphasis, not material to summarise five times.
- **Merge.** The user's real action prompt (dictionary suffix included) with a short addendum saying the material is ordered working notes from N parts of one recording. User content: manual notes, meeting context, then `## Notes from part i of N` in order. If that does not fit, neighbouring part-notes are consolidated for up to three rounds; beyond that the existing translated `contextTooLargeGeneric` refusal names the model.
- **A part the exact tokenizer still refuses** is halved along lines, to a depth of three.
- **Progress.** The overlay shows "Part 2 of 5" under the action name (11 locales). A chunked run on a laptop takes minutes, and a static spinner reads as a hang. Cancelling between requests stops issuing them.

## Verification

- `npm test` on Node 24: 4460 tests, 4239 passing, 0 failing, 220 skipped (the DB-backed tests that skip under system node).
- `npm run quality-check` (eslint, prettier, tsc): 0 errors (2 pre-existing warnings in files this PR does not touch). `npm run i18n:check`: clean.
- New tests: `noteChunking` (7: packing never splits a line unless the line alone exceeds the budget, joined parts reproduce the input, CJK sliced by tokens, the estimate agrees with `llamaContextPolicy.estimateTokens` on Latin/CJK/mixed fixtures), `localContextBudgetIpc` (2), `actionProcessingChunked` (7: fits → one unchanged call; cloud never reads the budget; parts carry the context and `part i of N` and not the manual notes, the final pass carries the action prompt, the manual notes and every part's notes in order and never the raw transcript; failed budget read → single call; single-call `CONTEXT_TOO_LARGE` → parts; cancel after part 1 → no further calls, no save; progress runs 1/N → N/N).
- **Real model, real binary, this 16 GB Mac (the #1203 machine class):** driven through the real renderer store (Vite SSR harness), the real `modelManagerBridge` and the bundled `llama-server-darwin-arm64`, with Llama 3.2 3B (the #1203 model) downloaded into the app's models directory. Ceiling resolved from the GGUF: 33,792 (memory-bound). A 227 KB, 4,481-line real meeting transcript (about 60k tokens) with two facts planted at 5% and 95%:
  - 7 parts + 1 merge, 8 calls, 393 s wall, progress 1/8 → 8/8, no server restart (12k parts fit the baseline window), no part headings or raw transcript lines in the output.
  - The first pass, before the part prompt was structured and parts were capped, showed why both were needed: given a 20k-token part the 3B wrote five generic bullets and dropped every number. With the structured prompt, every planted number and date in part 1 came through the part step.
  - **Known limit of the smallest model:** at the merge the 3B kept the late planted fact (recency) but compressed away the early one, with the completeness instruction in either position. The merge input was 11.8 KB against a 4,096-token allowance, so this is the model choosing brevity, not the window. Larger local models (the default is a 9B) should do better; a per-section merge is the natural follow-up if they do not. The alternative today is a refusal, so this is still strictly better for the user.

## Known limits

- Local **chat** streaming still bypasses `runInference` and keeps the 16,384 baseline (named in #2155).
- A summary whose *output* exceeds 4,096 tokens can still be clipped (named in #2155).
- The memory policy in `llamaContextPolicy` is untouched. Whether 0.5 × RAM − 2 GiB is too tight for a 9B on 16 GB is a #1203 argument with its own measurements, not this PR.
- The token estimator is duplicated in `noteChunking.js` on purpose: the renderer cannot import CommonJS `llamaContextPolicy.js`. A test pins the two together.
- Windows/Linux not run on real hardware; nothing here is platform-specific.


Chadpiha:
## Review: outstanding findings after the fix commits

A real-model run of bfdd18c3 (Qwen3.5 9B, memory forced to 16 GiB so the ceiling resolved to 16,384, a 19.4k-token diarised transcript, Detailed Notes) took **32 calls and 23.7 minutes**: both planned parts were halved to depth 3 because the 9B wrote more than the 2,048-token part budget under the "be thorough" prompt, the merge was refused as too large (16 leaf outputs ≈ 19k tokens) and consolidated, and **1 of 6 planted facts** survived in the saved note although all 6 were present in the leaf outputs. The overlay showed "Part 1 of 3" for 11.6 minutes.

The three commits d6e10403, 010c5426 and cea5ed49 address the blockers (per-part allowance sized to the merge room, clipped parts kept instead of halved, merge saved rather than refused when clipped, plain notes never condensed, empty part reply fails the note, cancel control, typed and translated busy error). Everything below is what is still open.

### Needs a real-model rerun before merge
- **Word target is unverified.** Parts now ask for about half their token allowance in words and keep a clipped reply. Whether the 9B stays inside 2,048 tokens on a ~11k-token part, and whether the planted facts now survive the merge, has not been measured since the change. Nothing logs a clipped part today (`llamaServer.js` never logs `finish_reason`).

### Behaviour gaps
- **Cancel is soft.** `cancelAction` discards the result, but the in-flight llama request keeps generating and holds `localReasoningBridge.isProcessing`. An immediate rerun, or dictation cleanup, gets the new "still working on another request" toast about the user's own cancelled request. A real abort needs an IPC path to `req.destroy()`.
- **Chat with a different local model kills an in-flight run.** `llamaServerStart` from `ReasoningService.ts` stops the server when the chat model differs from the notes model; the part in flight dies with an untranslated "Inference failed" toast. Pre-existing single-model server design, now routine because runs take minutes.
- **Models that ignore `enable_thinking: false`** (Gemma 4, GPT-OSS; `llamaServer.js` notes non-Qwen templates ignore it) can return a clipped, unterminated `<think>` block that `stripThinkingTags` reduces to "", which now surfaces as "Model returned no text". That string is untranslated, matching the existing whole-note check.
- **Windows/Linux Vulkan.** `get-local-context-budget` returns the theoretical ceiling, not the loaded window. When the reduced-context rung loaded 16,384 but the ceiling reports more, a merge above the loaded window costs a failed larger start plus a reload. Not reproduced on hardware.
- **`get-local-context-budget` memoises a fallback ceiling** for a missing or partial GGUF (no `checkModelValid` before `contextCeiling`), for the process lifetime. Not reachable from this PR's renderer path, but the IPC is public.
- **Progress stays at "Part N of N"** through consolidation rounds. Consolidation is now a rare fallback, so this is cosmetic.

### Tests
- "cloud mode never reads the budget" only exercises the no-refusal path; a cloud-mode run with a `CONTEXT_TOO_LARGE` first call would read the budget. Production safety rests on the IPC returning "not found" for a cloud model id.
- The newline-cost mutation in `noteChunking.js` (`+ 1` → `+ 0`) still survives.
- Two new store tests flaked once under heavy parallel load; the file's shared-globals pattern (`globalThis.__processTextCalls`) is pre-existing.

### PR description
The body still describes the removed route-check, budget-read and fit-test design, says "actionProcessingChunked (7)" where there are now 31, and says a failed budget read "runs the single request" where it now rethrows the refusal. With the new allowance sizing, consolidation is a fallback for tokenizer disagreement rather than the normal path, and plain notes are no longer chunked.

boseq (approved):
Tested on macOS (Qwen3.5 9B with memory forced to 16 GiB, and LFM2.5 1.2B): meetings that main refuses now get summarised in parts and saved, with progress and Cancel in the overlay.

I pushed three follow-ups: Cancel now aborts the local request in flight (so a rerun isn't refused as busy), part replies can use the final pass's room instead of stopping at 2,048 tokens (4 → 5 of 6 planted facts kept on the 9B), and an empty reply shows a translated message. The description is updated to match the current design.

PR #2173, opened 2026-09-13
feat(telemetry): report whether the desktop is the Windows portable build

greatcoat:
## Why

276 Windows users were still on v1.9.1 on 2026-09-09, eleven days after v1.9.2
fixed the OAuth regression that stops them signing in at all. The leading
hypothesis is that they are running the **portable** build:
`src/updater.js:263` calls `autoUpdater.quitAndInstall()`, which on Windows
runs the downloaded NSIS installer, so the update succeeds into Program Files
while the user keeps double-clicking the old portable exe in Downloads.

Nothing in our telemetry records which artifact a user installed, and
`PORTABLE_EXECUTABLE_DIR` was read nowhere in the codebase, so the app could
not tell either. This makes the hypothesis measurable.

## What

- `src/helpers/portableBuild.js` - the only reader of `PORTABLE_EXECUTABLE_DIR`.
- `x-openwhispr-portable: 1|0` attached at the three sites that already set
  `x-openwhispr-version` (`policyRequestHeaders.js`, the `cloudApiRequest.js`
  public branch, `enterpriseIdentityManager.js`).

## Privacy

Only a boolean leaves the machine. `PORTABLE_EXECUTABLE_DIR` is a path under
the user's profile and contains their Windows account name; a test asserts it
never reaches a header value.

## Server side

Paired with openwhispr-api `feat/portable-build-activity-column`, which adds
`user_daily_activity.portable_build`. That side is independent of this one:
until this ships in a release the column simply stays NULL, which reads as
"unknown", never as "not portable".

## Tests

`npm run quality-check` and `npm test` pass on Node 24.


greatcoat:
Closing without merge. Josh decided on 2026-09-14 not to pursue the portable-build check; he no longer sees value in it. The branch stays in place so this can be reopened if that changes. Nothing was applied to production.

PR #2189, opened 2026-09-14
fix(assistant): paste caret-bound answers as plain text unless the app keeps markdown

greatcoat:
## Problem

When a voice-command answer pastes at the caret, it arrives with markdown intact: `**bold**` lands as literal asterisks in a plain document. Nothing told the model to write prose and nothing in the delivery path touched the text. At the same time, plain text is the wrong answer for markdown editors, AI prompt boxes, standard-markdown chat tools and code editors. Separately, `pasteAtCapturedTarget` returned four refusal codes and logged none of them, so a declined paste was undiagnosable.

## Fix

- **Decide at capture.** `captureSelectedText` judges the app under the caret against `src/helpers/markdownTargets.js` and returns `acceptsMarkdown`. The list is researched (evidence per app in the design spec's Appendix A) and matched by whole token on the identity each platform already exposes: `.app` name and executable on macOS, exe name on Windows, WM_CLASS on Linux. Plain text is the default for anything unlisted or unreadable. Slack, Teams, WhatsApp and Telegram are deliberately excluded; a real paste into Slack and Teams confirmed every marker stays literal.
- **Instruct.** For a plain-text caret, `plainTextResponse` appends `PLAIN_TEXT_RESPONSE_SUFFIX` last in the system prompt. It is conditional rather than an edit to `DEFAULT_CHAT_AGENT_PROMPT`, so panel answers keep rendered markdown, custom chat prompts are covered, and `retiredPrompts.js` is untouched.
- **Strip.** `markdownToPlainText` (new and newline-preserving; deliberately not `stripMarkdownPreview`, which collapses newlines) runs inside `deliverAssistantResponse` for plain-text paste deliveries and their clipboard fallback. Markdown-friendly targets and clipboard-only deliveries stay verbatim. A strip that leaves nothing falls back to the raw answer. History stores what the model wrote.
- **Log.** Each refusal in `pasteAtCapturedTarget` emits one `info` line with the code, session state and probe verdict. A success emits one `debug` line with the target kind, the markdown verdict and the matched app identity. Return values are unchanged.

Design: `docs/superpowers/specs/2026-09-10-agent-paste-plain-text-design.md`. Plan: `docs/superpowers/plans/2026-09-10-agent-paste-plain-text.md`.

Out of scope, on purpose: rich-text clipboard, per-site detection for web apps in a browser (needs the window title; the natural follow-up once #1952 makes browser targets reachable), a user setting, and #1952 itself. Chromium and Electron targets do not receive agent paste today; once #1952 lands the allowlist becomes load-bearing, because most markdown-friendly apps are Electron.

## Tests

- `test/helpers/markdownToPlainText.test.js`: one case per strip rule, plus must-not-alter cases (`2 * 3`, snake_case, `__init__`, markdown inside inline code, `#hashtag`, `x > y`, `\*`) and a two-paragraph answer.
- `test/helpers/markdownTargets.test.js`: every platform identity shape for listed apps matches; plain-text apps, own-dialect chat tools, browsers, `Xcode` and the vetoes never match; no generic tokens.
- `test/helpers/selectionManager.test.js`: the verdict on the capture result and session for macOS, Windows, Linux and unreadable pids; one test per refusal code; the success line carries the verdict.
- `test/helpers/audioManagerAssistantDirective.test.js` and `assistantResponseDelivery.test.js`: the verdict rides with the session id; stripped versus verbatim per delivery; empty-strip fallback.
- `test/helpers/agentPlainTextSuffix.test.js`: the suffix appends once at the end, and the panel prompt never carries it.
- `retiredPrompts.test.js` is unchanged and green, so the prompt-hash protocol was not triggered.
- Full suite on Node 24: 4481 tests, 0 failures (219 local Electron-ABI skips). `npm run quality-check` exits 0.

**Manual, macOS dev build against production** (branch head before the final review's fix commit):

| Target | Result |
|---|---|
| TextEdit, plain-text caret, "three reasons to drink more water" | Three plain lines, no markers; log `acceptsMarkdown: false` |
| Gemini desktop app prompt box, "a markdown table of three fruits" | Table pasted with pipes intact; log `acceptsMarkdown: true` |
| TextEdit, focus moved to Finder before the answer arrived | Not pasted; log `paste declined`, `code: target_changed`; plain-text answer on the clipboard |


boseq (approved):
Tested this on macOS against main and it does what it says. On main, a voice answer pasted into TextEdit keeps `##`, `**bold**`, backticks and table pipes. On this branch it pastes as clean lines, the plain-prose instruction is the last thing in the system prompt, and history still keeps the model's markdown. A paste declined because focus moved now logs `target_changed`, and the plain-text answer goes to the clipboard.

I pushed four small commits on top:

- **Strip fixes:**
  - A code span inside a quoted path or a link address pasted invisible U+E000 characters.
  - "Choose **A** or **B**." pasted as "Choose A** or **B.", because the emphasis patterns were greedy.
  - A line starting "+ 5 points" became "- 5 points".
- **Tests:**
  - New tests check that the plain-prose instruction is only asked for on plain-text paste deliveries and that it actually reaches the prompt.
  - The strip rules are now tested once, in `markdownToPlainText.test.js`. The same cases also ran three times through the delivery tests.
- **Removed:**
  - The success debug log. On macOS its `targetSignature` was always null, since pid targets carry no app name.
  - The session fields only that log read.
  - The `xcode` veto, which whole-token matching made unreachable.
  - The repeated `=== true` coercions.
  - The prompt-wording tests.

Re-tested after those commits: `npm run quality-check` and the full test suite pass. In TextEdit, "Choose A or B" and "+ 5 points" paste correctly, and no hidden characters get through.

One thing I left alone because it's a product call: after a declined paste, the clipboard holds the plain text, but the panel's Copy button still copies the raw markdown.

Not tested: Windows and Linux (only unit tests cover the exe name and WM_CLASS matching), and pasting into an app that keeps markdown. On macOS the Electron editors and Sublime never get caret delivery.

boseq:
Follow-up testing on Windows and Linux, with 28b8fde swapped into the 1.10.2 release in VMs:

- **Windows 11:** with the caret in Notepad, the answer pasted as clean plain text. "Choose A or B" and "+ 5 points" came through as written, the table became tab-separated rows, and the prompt carried the plain-prose instruction. An exe named `obsidian.exe` (a test text box, since no listed app was installed) matched on its exe name and got the markdown verbatim, with no instruction added.
- **Linux (Ubuntu, GNOME Wayland):** the matching is right. Via AT-SPI plus `ps comm`, gnome-text-editor reads as plain text and a process named `obsidian` as markdown. Via WM_CLASS, `code`, `md.obsidian.obsidian` and `org.kde.kate` read as markdown and `gnome-text-editor` as plain text.

PR #2192, opened 2026-09-14
fix(meetings): remove the post-meeting notification

greatcoat:
## Intent

Remove the post-meeting notification that says the meeting ended and offers 30 seconds to restart. This is a follow-up to #2162 after the product decision to remove the notification entirely.

## Change

- Keep automatic meeting stop and transcript persistence unchanged.
- Stop requesting the post-meeting card after auto-end.
- Remove the renderer restart listener and fallback ended/restart-failed toasts.
- Add a regression test proving auto-end stops recording without requesting or rendering a post-meeting notification.

## Validation

- npm test on Node 24.20.0: 4,343 passed, 0 failed, 219 skipped, 1 todo
- npm run build:renderer
- npm run typecheck
- npm run format:check (passes with pre-existing warnings only)
- 116 focused meeting lifecycle tests passed


Chadpiha (approved):
Approved!
Reviewed the auto-end flow end to end. The post-meeting notification and restart path are no longer triggered, while session-scoped stopping, transcript persistence, teardown, meeting detection, and existing warning behavior remain intact.

PR #2200, opened 2026-09-14
fix(llama): don't return reasoning_content as the answer when thinking is on

### LINKED ISSUE #2187: Getting all logs in the active app
**Describe the bug**
i am using local mode after finishing transcription it paste the whole logs -
```
Thinking Process:

1.  **Analyze the Request:**
    *   **Task:** Clean a raw speech transcript provided between `<transcript>` tags.
    *   **Input:** `<transcript>Hi there, how are you? I am checking Open Whisper. Let's see how much it can Like record it and paste the what I am saying today is gonna spooja but I choose to stay in home and study Uh do some DSA then uh we'll do some Open sour contribution I will try to Like In my personal projects also</transcript>`
    *   **Constraints:**
        *   Output *only* the cleaned transcript.
        *   No preamble, labels, quotes, tags, commentary, or answers.
        *   Language: Auto-detect (English in this case).
        *   Specific rules: Remove filler words (um, uh, er, like, you know) unless genuine meaning. Fix grammar/spelling/punctuation. Remove false starts. Keep technical terms, proper nouns, and jargon exactly as spoken (e.g., "OpenWhispr" if mentioned, though here it's "Open Whisper").
        *   Specific corrections: Self-corrections ("wait no", "I meant", "scratch that") keep only corrected version. Spoken punctuation convert to symbols or break.
        *   Specific words: "OpenWhispr" (keep exact spelling if mentioned).
        *   Input text contains "Open Whisper" (likely a typo for Open Whisper Chat or similar, but I must keep it as spoken).
        *   Input text contains "DSA" (Digital Systems Analysis? Data Science? Just keep as is).
        *   Input text contains "spooja" (likely "spoke" or "spoke" or "spooja" as a name? Context suggests "spooja" might be a typo for "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or "spoke" or " 
```

**To Reproduce**
press ctrl or your hotkey say what you want after finishing the trancription it will show logs in the active app 

**Expected behavior**
it should only paste what i am saying 

- OS: [windows 11]
- Version [
Current version

You're on the latest version

1.10.0
Latest
]

boseq:
## Summary

Fixes #2187. With a local Qwen3.5 model and **Disable thinking output** switched off, the app used the model's raw reasoning ("Thinking Process: 1. **Analyze the Request:** …") as its result.

**Cause.** With thinking on, the model reasons into `reasoning_content`. When it spends its whole token budget reasoning, `content` comes back empty. `LlamaServerManager.inference()` then fell back to `reasoning_content` and returned the reasoning as the answer.

**Fix.** Keep that fallback only when the request suppressed thinking. #809 added it for llama-server builds that put a suppressed-thinking *answer* in `reasoning_content`, and that behavior is unchanged. With thinking on, the result is `content` alone, which matches how the LAN and cloud providers already read replies.

**Scope on current `main`.** The reporter was on 1.10.0. #2130, which shipped in 1.10.1, already rejects the reported dictation case (a cut-off reply) and pastes the raw transcript with the "ran out of room" toast. This PR fixes the shared source for the paths #2130 doesn't cover:

| Caller | `main` today, thinking on | With this PR |
| --- | --- | --- |
| Note formatting | Reasoning saved as the enhanced note, with no error. Verified: 14,547 chars | Nothing saved. Existing "Model returned no text" error |
| Dictation translation step, live and history retry | Reasoning replaces the translation | Keeps the previous text. Fallback toast on live dictation; history retry only logs |
| Prompt Studio tests (cleanup, translation, agent) | Reasoning shown as the output | No output shown |
| Dictation cleanup (including the translation chain's cleanup step) and history-retry cleanup, all `requireCompleteOutput` | Cut-off reply rejected by #2130. A reply that *finishes* with empty `content` still returned the reasoning | Empty result. The raw transcript is kept, silently |
| Selection edit, also `requireCompleteOutput` | Cut-off reply rejected. A finished empty reply returned the reasoning, which the completion-marker check rejects | Same outcome |

The fix doesn't depend on platform.

## Test plan

- [x] Unit tests in `test/helpers/llamaServerContext.test.js`, using the existing stub server:
  - Thinking on: reasoning is not returned for a cut-off reply, or for a finished reply with empty `content` from a strict caller. The request carries no `chat_template_kwargs`.
  - Thinking suppressed, or the flag omitted: the #809 fallback still returns the misrouted answer. The suppressed request carries `{ enable_thinking: false }`.
  - `content` wins in both modes.
- [x] Mutation checks, each of which makes a test fail:
  - `main`'s unpatched `llamaServer.js` (the strict case also fails on its own)
  - always sending the kwarg
  - removing the #809 fallback
  - keeping the fallback only when `disableThinking === true`
- [x] `node --import tsx --test` on `llamaServerContext`, `modelManagerContextSizing`, `localReasoningBridgeChain`, `stripThinking` and `localInferenceError`: 57/57 pass. Prettier is clean on both files; eslint is clean on the test file (`llamaServer.js` is in eslint's ignore list).
- [x] E2E on macOS: dev build on an isolated profile, real Qwen3.5 2B on the bundled llama-server, driven through the app's own pipeline.
  - Thinking on: an A/B against `main`'s `llamaServer.js` confirmed the note formatting, Prompt Studio cleanup and dictation cleanup rows above.
  - Thinking suppressed: dictation cleanup (620 ms) and note formatting (1.8 s) still produce cleaned output.
- [x] Reproduced the original report in the installed 1.10.0 app (Prompt Studio, Qwen3.5 2B). The output matched the issue almost verbatim with the toggle off and came out clean with it on.

## Follow-ups (not in this PR)

- **Prompt Studio** renders nothing for an empty result, from any provider (`{testResult && …}`).
- **"Model returned no text"** (note formatting, from #2177) has no translation key.
- **Local dictation cleanup and history-retry cleanup** keep the raw transcript silently on an empty reply, as OpenWhispr Cloud and enterprise do. Bring-your-own-key and self-hosted providers raise `cleanupEmptyReply` instead.
- **gpt-oss-20b:** its template ignores `enable_thinking`, so with thinking suppressed its analysis may still arrive in `reasoning_content`. A cut-off analysis could then reach callers without `requireCompleteOutput`. Unverified, and it predates this PR. Needs a live check before choosing a `finish_reason` guard.
- **Product decision:** a small local model with thinking on rarely finishes inside cleanup's token budget. Always suppress thinking for local cleanup, or hide the toggle for that scope?

https://claude.ai/code/session_01VrvzpDuarBxf1fLMJjuwfg


Chadpiha (approved):
Reviewed the local inference pipeline and affected dictation, translation, selection-editing, notes, and Prompt Studio paths. The fix is focused, prevents reasoning from replacing the answer, and preserves existing fallback/error handling.

Approved. Real-model desktop behavior was not independently retested. Thanks @boseq !

PR #2201, opened 2026-09-14
fix(notes): Prevent blank summaries for non-substantive recordings

Chadpiha:
Recordings containing only greetings, filler, or recording checks can produce a blank response from Generate AI Summary: the detailed prompt removes the content and omits every section. The Cloud API correctly rejects that result, leaving the user with an enhancement error.

Update Generate Notes and Detailed Notes to return a brief factual sentence when neither the transcript nor manual notes contain substantive content. The instruction overrides section and bullet-count requirements and respects the requested output language. Existing untouched defaults migrate automatically; user-edited prompts remain intact. The blank-response safeguard stays enabled.

Validation:
- 24 focused tests passed, including fresh installation, prompt migration, migration idempotency, customization preservation, and blank-result rejection.
- Live synthetic Parasail comparison reproduced the blank response with the old prompt. All eight updated-prompt checks returned text, covering three repetitions of the failing case, Generate Notes, a substantive meeting, manual notes, longer filler-only input, and Spanish output.
- Typecheck and changed-file formatting passed. Repository lint completed with only existing warnings.


gabrielste1n:
Thanks Chad, the prompt change and migration look right. I pushed one test cleanup on top (dbf00971): the new test now uses the shared `test/helpers/harness/db.js` database helper instead of its own electron stub, and it iterates each action's `previousPrompts` rather than a fixture copy of the 1.10.1 text, so the 1.10.0 Detailed Notes upgrade path is covered too. Merging once CI is green.

gabrielste1n (approved):
Reviewed the prompt change and the seeding migration; fresh installs, untouched-default upgrades, and user-edited prompts are all covered by the database test.

PR #2202, opened 2026-09-14
feat(insights): show layout-shaped skeletons while Insights loads

boseq:
## Summary

Both Insights tabs showed a spinner in a short card while loading, then jumped to a much taller layout. Now the page's fixed copy renders immediately, and only the data waits behind skeletons shaped like it. Data fills in a page that is already there, so it looks better and feels faster.

The rule is simple: fixed copy and icons are real from the first frame, and anything that comes from the response or the current selection is a placeholder.

- **Your Usage**: there is no separate skeleton. The real cards and heatmap render while the summary loads.
  - Real while loading: card labels and icons, "All time", "Speaking activity", month and weekday labels, today's marker, the legend, and the data-retention banner.
  - Placeholders: the four values, the WPM-coverage and longest-streak details, and cell colors.
  - The heatmap keeps the same element tree, so arriving data only recolors cells. Its cells are `inert` while loading, so no "0 words" tooltip or label is read for a day that hasn't been counted.
- **Leaderboard**:
  - `LeaderboardSkeleton` shows "Updates hourly", "Top performers", "Rank" and "Member", plus the podium's gold, silver and bronze places with their crown and medal rank badges.
  - The scope, member count, period, names, values and both selects are placeholders.
  - `LeaderboardPodium` now shares its frame and place markup with `LeaderboardPodiumSkeleton`, so the skeleton doesn't copy it.
  - It replaces the three spinners (access loading, participation loading, board loading) and serves as the tab's lazy-chunk fallback. The fallback is for signed-in users only, so signed-out users never see a board before the sign-in preview.
  - A loading board's real header also holds the member count and period picker, so nothing shifts when the board arrives.
- **Shared `Skeleton`**:
  - It renders a block `<span>`, so it can sit inline in a `<p>` or heading and keep that line's exact height.
  - It uses `bg-foreground/6`, the heatmap's empty-cell tone. `bg-muted` is darker than `card` in dark mode, so skeletons on cards were nearly invisible, including the existing ones in Settings. Light mode is effectively unchanged.

Measured against the loaded pages, these boxes match within 0.2px: the metric cards and each value and detail line, the activity card and heatmap grid, and the board section, header, podium cards, table head and rows. A screen-reader "Loading..." status stands in for the old visible spinner text. Refreshes still keep the previous data on screen.

## Screenshots

These are the real `InsightsView`, `LeaderboardSection` and `Skeleton` rendered from the base commit (before) and this branch (after), at 2x. IPC and `LeaderboardService` are stubbed so a request either stays pending (loading) or resolves with sample data (loaded).

**Your Usage: loading**

| Before | After |
| --- | --- |
| Your Usage loading, before | Your Usage loading, after |

**Your Usage: loading, then loaded** (nothing moves; values and cell colors fill in)

| Loading | Loaded |
| --- | --- |
| Your Usage loading | Your Usage loaded |

**Leaderboard: while access loads**

| Before | After |
| --- | --- |
| Leaderboard access loading, before | Leaderboard access loading, after |

**Leaderboard: board loading, then loaded** (the real header already holds the count and period picker)

| Loading | Loaded |
| --- | --- |
| Leaderboard board loading | Leaderboard board loaded |

**Dark mode, loading**

| Your Usage | Leaderboard |
| --- | --- |
| Your Usage loading, dark | Leaderboard loading, dark |

**Shared `Skeleton` on a Settings panel, dark (the API keys loading rows)**

| Before (`bg-muted`) | After (`bg-foreground/6`) |
| --- | --- |
| Settings skeleton, dark, before | Settings skeleton, dark, after |

## Test plan

- [x] `npm run quality-check` passes (ESLint, Prettier, `tsc --noEmit`)
- [x] `npm test` passes: 4625 tests, 0 failures
- [x] New render test: the loading board shows its fixed copy and keeps the podium's accents. The existing podium test still passes after the extraction
- [x] Rendered both tabs in the loading and loaded states, in light and dark, and compared element boxes between loading and loaded
- [x] In the app, open Insights → Your Usage: labels appear at once, then values and heatmap colors fill in with no layout jump
- [x] Signed in, open Leaderboard: header and podium copy stay put while the board loads, then the board, a join card or a setup card
- [x] Signed out, open Leaderboard: sign-in preview, no skeleton flash
- [x] Dark mode: placeholders are visible on cards (Insights and Settings → Account)

https://claude.ai/code/session_01SSoE6WfHPrZS7H5oQYwZCE


Chadpiha (approved):
Thanks @boseq ! Reviewed the loading states, affected user journeys, and shared Skeleton consumers. No actionable regressions found.
The layout checks confirmed matching loading/loaded dimensions for the usage cards, heatmap, and podium. Pending heatmap cells are excluded from the accessibility tree, and dark-mode placeholders remain visible. Existing authentication, retry, empty-state, and refresh behavior is preserved.

LGTM. Please resolve the current mergeability issue and rerun CI before merging.

PR #2203, opened 2026-09-15
fix(models): resolve openrouter-prefixed model metadata

fthozdemir:
OpenRouter returns model IDs with a vendor prefix, such as google/gemini-3.5-flash-lite, while the cloud model registry stores the upstream ID as gemini-3.5-flash-lite. This caused the vision capability lookup to return no model and reject screenshots as unsupported.

Pass the provider to getCloudModel and retry the lookup without the vendor prefix only for OpenRouter. Exact model IDs are still checked first, and other providers keep their existing behavior.


gabrielste1n (approved):
Verified locally on 48655914: typecheck, lint, prettier, and the full test suite (4626 tests, 0 failures) pass.

Pushed one cleanup commit on top: getOpenAiApiConfig already stripped the OpenRouter vendor prefix on its own, so it now goes through the provider-aware getCloudModel lookup instead of a second copy, and a registry test covers the prefixed lookup.

PR #2221, opened 2026-09-16
fix(dictation): enforce Groq prompt limit in UTF-8 bytes

greatcoat:
Groq rejects prompts above 896 UTF-8 bytes, while the existing safeguard counts JavaScript characters. For example, 448 accented letters plus one ASCII letter occupy 897 bytes despite being only 449 characters. Multipart line-ending expansion can also exceed the limit after the current check.

Apply a Groq-specific byte cap to the final composed prompt, including custom endpoints pointing at Groq. Use shared, parsed exact-host matching so misleading paths, queries, userinfo and lookalike hosts cannot select the Groq budget. Normalize Unicode and line endings before measuring, preserve whole code points/CRLF pairs and the existing comma-boundary preference, and log byte counts without prompt contents. Other providers retain their current character budgets.

Check an echo of the actual sent Groq prompt using a full normalized match. Ordinary speech sharing its vocabulary must still be delivered: when a long dictionary entry leaves only `OpenWhispr, on my way` in the request, `On my way.` now completes normally. The full echoed prompt remains rejected even with changed case, punctuation or spacing. Existing full-dictionary heuristics keep their behavior.

Validation (Node 24):

- Original byte-cap regressions produced 28 failures against the original code; URL-classification regressions exposed 16 failures before correction.
- Six normal-speech cases failed before the echo-check correction. All 227 focused AudioManager, prompt, echo, routing and recovery tests now pass. The new cases exercise the full `processAudio` path for direct and custom Groq, verifying completion for ordinary speech and no-audio handling for actual echoes.
- `npm run quality-check` and `npm run build:renderer` pass; six existing lint warnings and existing bundler notices remain.
- Prior independent byte-cap validation checked 636 synthetic cases for byte limits, Unicode/newline boundaries, intact prefixes, idempotence and identical desktop/API SDK serialized output. All 636 matched in Electron 41.10.5 / Chromium 146 on macOS. The byte limiter is unchanged by the echo-check correction.
- Package manifests and lockfiles are unchanged. Physical platform and customer recovery testing remain outstanding.

Related: #1663 also addresses dictionary echoes. This patch checks the directly captured sent Groq prompt; the existing broad dictionary matcher is unchanged, and the new exact matcher reuses its normalization.

Companion API fix: https://github.com/OpenWhispr/openwhispr-api/pull/210. Direct BYOK Groq requests require this desktop fix; deploying the API change alone does not cover that path.

Source: [CUS-240](https://linear.app/openwhispr/issue/CUS-240/groq-transcription-rejects-prompt-as-897-chars-limit-896-emmanuel-de).


xlurie:
One finding on the exact check, from #1636 (I filed it). It can meet a one-term prompt on Groq: in your "shortened Groq prompt" test, make the long phrase the second entry and the request carries only `OpenWhispr`, so a dictated "OpenWhispr." is rejected as an echo where `main` delivers it. The several-terms condition from your #2314 follow-up closes that and loses nothing there: an untrimmed one-term prompt is the whole join, which `matchesDictionaryPrompt` already catches on equality. Keyed on that condition rather than on `isGroqEndpoint`, the check also reaches whisper-1 and custom or self-hosted endpoints, where a full echo of the trimmed prompt still passes at 0.7 against the full join (the #1636 mock). That widening can go here with the rebase or into the follow-up, your call. Per #851: this catches a full echo of the sent prompt; tail fragments of token-expensive lists stay uncaught either way.

PR #2228, opened 2026-09-16
fix(ui): give every control panel page the same 1024px content column

clarkmaan:
## Summary

Home centers its content in a `max-w-5xl` (1024px) column, but every other control panel page picked its own width, so the content jumped around when switching tabs. This PR gives all of them the same column.

| Page | Before | After |
| --- | --- | --- |
| Home | List at 1024px, but the three banners above it (required models, past due, GPU) were `max-w-3xl` and sat 32px narrower | Banners share the list's column and gutter exactly |
| Insights | 1024px | Unchanged, now uses the shared constant |
| Chat | Full width | Message column and composer capped inside the content pane; conversation list stays flush left |
| Notes | Editor full width, folder overview 672px | Editor header, conflict banner, and body capped; folder overview moves to the shared column so toggling note ↔ folder no longer snaps the width |
| Upload | 448px card with 320px inner caps | Fills the shared column |
| Dictionary / Snippets | Full width | Capped and centered |
| Integrations | 512px column | Fills the shared column |

## How

- New `PAGE_CONTENT_WIDTH_CLASS` in `src/components/ui/pageWidth.ts` (`mx-auto w-full max-w-5xl`), following the existing exported-class-constant pattern in `ui/glass.ts` and `ui/splitButton.ts`.
- Every page applies it with its horizontal padding on the **same element**, so caps and gutters line up across pages.
- `ChatMessages` gets an optional `contentClassName` applied to the inner message column. The scroll container stays full width, so `EmbeddedChat` and `OverviewAskSection` are unaffected.
- `NoteEditor` wraps the scroll container's children in the column with `h-full`, so the percentage-height chain that the transcript virtualizer, stick-to-bottom, and `EditorContent` scroller depend on is preserved.

## Screenshots (1600px-wide window, dev profile seeded with sample notes, a chat, and dictations)

| Page | Before | After |
| --- | --- | --- |
| Home | home before | home after |
| Insights | insights before | insights after |
| Chat | chat before | chat after |
| Notes | notes before | notes after |
| Upload | upload before | upload after |
| Dictionary | dictionary before | dictionary after |
| Integrations | integrations before | integrations after |

## Review notes

Two rounds of multi-agent review with adversarial verification ran on the diff. Confirmed findings, all fixed here:

- Home banners padded inside the column while the list padded outside it, leaving a 32px mismatch → HistoryView collapsed to one wrapper with padding inside.
- Chat composer was 12px narrower than the message column → composer wrapped in the same padded shell as the messages.
- Upload's column class was a visual no-op behind 320px inner caps → caps removed, gutter moved onto the column.
- Folder overview stayed at 672px → moved to the shared column.
- **Height-chain regression:** the first version of the NoteEditor wrapper had `height: auto`, which would have defeated the transcript virtualizer, stick-to-bottom during live meetings, and click-to-focus below short notes → wrapper now carries `h-full`.

## Verification

- `npm run typecheck`, eslint, and prettier pass on all touched files.
- Screenshots above were captured from the dev build at 1600px via CDP, before and after, on the same seeded profile.


boseq:
## Screenshots: visual changes from the follow-up commits

These cover only what `0d1a6f3a`, `4e5282c9` and `7e9ad0bd` change visually. Captured from this PR before the follow-ups (`5b0f1848`) and at the current head (`7e9ad0bd`), using a harness page that renders the real components with fake data. The required-models banner is stubbed to report a missing model. Light theme, 2x.

### Home: side padding 16px → 24px

The list, its day headers and the banners above it now use the same 24px padding as Insights, Integrations and Upload. Shown in a 990px pane, about what the default 1200px window gives. The other three banners (update required, past due, GPU) use the same wrapper as the one shown.

Home with the required-models banner: before and after, content 8px further in on each side

### Dictionary and Snippets: side padding 20px → 24px

Dictionary tab: before and after, content 4px further in on each side

<details>
<summary>Snippets tab</summary>

Snippets tab: before and after, content 4px further in on each side

</details>

### Notes: body text sits 3px left of the title once the note scrolls

Otherwise the note layout is unchanged. The editor's scroll area spans the pane again, so on panes wider than the 1024px column its 6px scrollbar takes width from the pane and the centered text shifts left by half of it. Shown in an 1180px pane.

Note editor, top left: before and after, body text 3px further left relative to the title

Not pictured: the note and transcript scrollbar moved from the column edge to the pane edge. It only draws on hover, and neither headless Chrome nor a hidden Electron window paints it into a capture. Wheel scrolling over the side margins and clicking there to place the caret are behavior fixes with no visual difference.

boseq (approved):

Issue #1891, opened 2026-08-27
windows-system-audio-helper captures digital silence while `probe` returns `ok:true`, which makes the Chromium fallback unreachable (regression after #1474)


KishenG:
**Describe the bug**

After #1474, `windows-system-audio-helper.exe` spawns as intended on my machine. It captures nothing. Every sample it emits is zero, and the app has no way to notice, because it selects its capture strategy from the helper's `probe` verdict and never from whether audio actually arrives.

The result is worse than the bug I originally reported in #1471. Before the fix the helper did not spawn, `available` was false, and meeting capture used the Chromium loopback fallback. Now the helper spawns and reports healthy, so `strategy` is `wasapi-loopback` and the fallback is never reached. Meeting transcripts silently contain only my own microphone. No error is raised anywhere in the UI or the log.

There are two separate defects here. The second is the one that matters.

1. The helper emits silence on this hardware.
2. `probe` returns `ok:true` on the same machine and in the same session, so it cannot gate the fallback.

**To Reproduce**

1. Start a meeting recording with audio playing to the default output device.
2. Observe the debug log: `"systemAudioStrategy": "wasapi-loopback"`.
3. Observe repeated `Skipping silent meeting chunk` entries with `"source":"system"` and `"peak":"0.0000"`, while `"source":"mic"` chunks transcribe normally.
4. Run the helper standalone to confirm the app is not at fault:

```
windows-system-audio-helper.exe start --exclude-pid <any pid> --sample-rate 16000 > out.pcm
```

**Expected behavior**

If the helper's stream is silent while audio is rendering, the app should fall back to Chromium loopback and say so. A capability probe that cannot fail should not be the only gate on a fallback path.

**Actual behavior, measured**

I sampled the WASAPI endpoint meter and the helper's PCM output over the same 8 seconds, with the helper run standalone, outside OpenWhispr, using the arguments the app itself passes.

| Measurement | Result |
|---|---|
| Output endpoint peak, `Headphones (2- WF-1000XM4)` | 0.3859 |
| Helper PCM emitted | 282560 bytes, 141280 samples |
| Helper PCM peak amplitude | 1 of 32767 |

A well-formed stream of the correct length at the correct rate, containing silence.

In the same session, `probe` returned:

```json
{"ok":true,"supportsSystemAudio":true,"supportsNativeCapture":true,"source":"wasapi-process-loopback"}
```

[...]

PR #2233, opened 2026-09-17
fix(meetings): keep the Windows silent-capture check armed for the whole call

### LINKED ISSUE #1265: No sound recorded within MS Teams meeting - Affected 1.7.4 and above - Windows

boseq:
## Credits

This one was found and narrowed by other people first:

- **@jaszczurovsky** reported #1265 and did the version bisect (1.7.2 and 1.7.3 fine, 1.7.4–1.7.6 broken), across Bluetooth and built-in devices, with exclusive mode disabled, on every local model — which pointed straight at #960.
- **@TaiFeng** and **@jsholcomb** found the workarounds (switch Teams' speaker to a virtual device; toggle Enhancements on the playback device). Both force Teams onto a different device or effects chain, which is what made the audio capturable again.
- **@wadamek65** and **@KishenG** confirmed it on their machines.
- **@amarpreet2209** opened #1866 and established the useful half of the diagnosis: the endpoint-loopback path hears Teams when process loopback does not.
- **@stantheman0128** reviewed #1866, and wrote #1474 after **@KishenG** reported #1471 and #1891 — the earlier rounds of "the helper is silent and nothing notices".
- **@gabrielste1n** built the native capture in #960 and the `capture_silent` fallback in #1935 that this PR extends.

## What was wrong

Windows hides Teams call audio from WASAPI process loopback. Measured in a Windows 11 VM against a live Teams call, all at the same moment:

| Capture | Peak |
| --- | --- |
| The shipped helper (exclude our own tree) | 0.0000 for 15 s |
| Process loopback aimed at Teams' own audio process | 0.0000 |
| Process loopback aimed at the Teams tree root | 0.0000 |
| Endpoint loopback | 0.38–0.65 |

That holds across 15 configurations — s16 at 24k/48k, float32, mono 16k, with and without `AUTOCONVERTPCM`, event-driven and polling. The same 15 configurations captured a control sound at 0.25–0.36 in the same session, so the tap works; Teams' stream simply is not in it. Teams' session is ordinary shared-mode (visible and metered via `IAudioSessionManager2`, owned by `ms-teams.exe`, not inside our excluded tree), and restarting the whole audio stack so Teams reopened its stream changed nothing.

#1935 already handles "capture is silent while speakers are playing" by handing the session to Chromium loopback. It didn't fire here, for two reasons:

1. `resources/windows-system-audio-helper.c` disarmed the check permanently on the first audible sample from **any** application.
2. `degradeMeetingSystemAudioToLoopback` refused the handover once `meetingSystemAudioHeard` was set.

So one notification sound, one browser tab, or Teams' own join chime early in the meeting left the rest of the call recorded mic-only, with no warning — the 45 s "no meeting audio" notice is gated on the same flag.

Reproduced on v1.7.6 and on v1.10.2:

- v1.10.2, Teams the only sound: `capture_silent` at +5 s → "Renderer loopback took over system audio capture" → the remote side transcribes. Working as designed.
- v1.10.2, a Windows sound played as the recording started: that sound was transcribed (`"[Music]"`), then every system chunk logged `peak 0.0000` for ~2 minutes while Teams was audible at 0.28–0.58 on the endpoint. No `capture_silent`, no handover, no warning.

## The fix

- **`resources/windows-system-audio-helper.c`** — the comparison now runs every second for the whole session instead of stopping at the first audible sample. A tick that captured audio resets the confirmation count; only consecutive ticks where the endpoints are playing and our capture heard nothing count toward `capture_silent`. The warning is still emitted once, and a machine without a meter still never reports.
- **`src/helpers/ipcHandlers.js`** — `degradeMeetingSystemAudioToLoopback` keeps its one-shot latch but no longer consults `meetingSystemAudioHeard`.

No changes to the event, the preload bridge, the renderer takeover or any IPC contract. The renderer already guards the takeover on session id, recording state and an existing stream, so a mid-call handover after audio was flowing is safe.

## Why not #1866's approach

It makes endpoint loopback the default for everyone, which only hears the **default** output device — the exact limitation #960 was built to remove and #1471 complained about. Its stated cause (Teams using WASAPI exclusive mode) doesn't hold: endpoint loopback can't capture exclusive-mode streams either, yet it captures Teams, and @jaszczurovsky had exclusive mode disabled. This keeps the native capture where it works and switches only when audio is provably missing.

## Verification

- `npm test` — 4439 pass, 0 fail. `npm run format:check` and `npm run typecheck` clean.
- New `test/helpers/meetingSystemAudioDegrade.test.js` executes the real degrade closure and pins the handover after audible chunks; it fails against the old `|| meetingSystemAudioHeard` gate.
- `test/helpers/meetingStreamingWiring.test.js` updated — it pinned the old gate by source text.
- The C change compiles clean with `-Wall -Wextra` (MinGW cross-compile locally; MSVC in CI on dispatch).
- **Verified against a live Teams call**, Windows 11 VM, patched helper plus the un-gated guard swapped into an installed 1.10.2. Same scenario that fails today — a Windows sound playing as the recording starts, then Teams only:

  | Time | Log |
  | --- | --- |
  | 11:20:46 | recording starts, `systemAudioStrategy: "wasapi-loopback"` |
  | 11:20:59 | the sound is captured and transcribed (`"(upbeat music)"`) — the point at which the old build disarmed for good |
  | 11:21:04.839 | `Windows system audio warning {"code":"capture_silent"}` |
  | 11:21:04.840 | `Windows system audio helper captured only silence, switching to renderer loopback` |
  | 11:21:05.729 | `Renderer loopback took over system audio capture` |
  | 11:21:11 onward | 14 × `Local meeting transcription candidate {"source":"system"}` carrying the remote speaker's words |

  The shipped v1.1.0 helper, run back to back in the same call, emitted no warning at all across 30 s while its capture read 0.0000 and the endpoint read 0.30–0.71. Standalone, the rebuilt helper emitted `capture_silent` ~5 s after the other sound stopped.

  Caveats on that verification: the binary was MinGW-built (CI builds with MSVC), on ARM64 Windows running the x64 app emulated, with a single virtual audio endpoint.

## Before release

Merging this runs `build-windows-system-audio-helper.yml` (the push trigger watches that `.c` file) and publishes `windows-system-audio-helper-v1.2.0`, which then wins the prefix lookup. Worth confirming that release exists and is the newest one with that prefix before the next Windows build, since `fetchLatestRelease` resolves by publish date rather than semver.

## Review follow-ups

Two independent review passes found no correctness bug in the capture change itself, and these, all fixed here:

- **The rebuilt helper would not have reached anyone.** `resources/windows-system-audio-helper.c` is not compiled at package time; `prebuild:win` downloads the newest published release tagged `windows-system-audio-helper-v*`. The push trigger publishes whatever `inputs.version` defaults to, which was still `1.0.0` (July), so the rebuild would have lost to the existing `v1.1.0` (August) and Windows builds would have kept shipping the pre-fix binary while the changelog claimed otherwise. The workflow default is now `1.2.0`.
- **A failed takeover leaves nothing capturing the call.** It now publishes the existing, translated `systemAudioStopped` interruption (with its Stop action) instead of a raw English string that would have bypassed i18n, and tells auto-end the system channel is gone — otherwise auto-end still believed it had one and could end the recording after a minute of quiet microphone.
- The watchdog's restart hook is detached *before* the helper is stopped, so a restart already in flight cannot bring it back behind the renderer's takeover.
- Docs: the changelog entry moved to `### Meetings`, "no warning" became the accurate "no warning beyond a generic 'audio has gone quiet' three minutes later", the Teams finding is scoped to what was measured rather than "every machine", and neither doc now implies this rescues a meeting held alongside continuously playing audio — see the gap below.

Knowingly left alone: if the helper never captures anything at all and the takeover also fails, the user sees both the interruption notice and the 45-second "no meeting audio detected" notice. Two notices about one condition, in a path where both statements are true.

## Known gap

The confirmation streak resets on any second in which we captured audio, so the warning needs five consecutive seconds of nothing. A meeting held alongside continuously playing audio — music in a browser tab — never produces that gap, so Teams is still lost there. Covering it needs a level-ratio comparison rather than a silence test, which I would rather not add without measuring it; the binary-rename workaround (#1891) remains the escape hatch until then.

Fixes #1265


stantheman0128:
Thanks for the credit on #1474. That one was the stale capability cache / spawn side. This PR is the next latch: `capture_silent` dying after the first audible sample, then `meetingSystemAudioHeard` blocking the handover.

The Teams table (process loopback 0.0000, endpoint 0.38-0.65) matches what I saw on #1866. I agree not to make endpoint loopback the default.

Support keeping the check armed for the whole call, and dropping the `heard` gate. The new degrade tests pin the old `|| meetingSystemAudioHeard` path, which is the bit that would have shipped as a silent mic-only meeting.

Two notes, not merge blockers:

1. You already called the remaining gap: continuous other audio (browser tab) never makes a five-second silent streak. I would leave a level-ratio follow-up until someone measures it. The #1891 rename workaround stays the escape hatch.

2. Shipping the C change depends on `windows-system-audio-helper-v1.2.0` actually winning the prefix download. `prebuild:win` still uses `tagPrefix`, and older helper tags have already fallen off the first 50 releases (#2005). #2006 pins today's tags, which would keep installing v1.1.0 after this lands. #2147 walks `Link: rel=next` and keeps prefix-latest. Those two should not land as substitutes for each other. The 1.2.0 bump in the workflow only helps if the lookup can still see it.

Chadpiha (dismissed):
Reviewed the capture → fallback → transcription flow, including failure handling, stop/restart races, and auto-end behavior. The change addresses the early-audio case that previously disabled Teams silence detection. Before release, please confirm Windows packages include helper v1.2.0. Live Windows/Teams capture was not independently verified in this review; the documented continuous-background-audio limitation remains.

PR #2239, opened 2026-09-17
fix(windows): stop offering hotkeys the missing key listener can't fire

boseq:
## Summary

Follow-up to #2005 and #2147. On Windows, Hold and every modifier-only or right-side-modifier hotkey only fire through `windows-key-listener.exe`. When a build ends up without it, the app still offered all of them and none did anything, with no error. #2147 fixes the download and stops packaging without the binary. This makes the app handle a missing listener itself, for builds that still lack it (for example `npm run dev` on a machine with no compiler, when GitHub rate-limits the unauthenticated download).

Reproduced on a Windows 11 VM with the official 1.10.2 install and only that file removed:
- Holding F8 in Hold mode fired the global shortcut 65 times, and the callback dropped every one.
- Tapping `Control+Alt` did nothing.
- Settings still reported push-to-talk as supported and accepted `Control+Alt`.

**Fix.** `HotkeyManager` now knows whether the listener exists (`windowsKeyManager` is injected the same way it already is into `WindowManager`). It treats a missing listener like any other unsupported hotkey, so the existing fallbacks take over:

- `supportsPushToTalk()` returns false:
  - Settings and onboarding disable Hold with the existing reason.
  - `setActivationMode("push")` refuses.
  - A saved Hold is reset to Tap at startup. The startup check now reads the saved mode, because the cached one has already refused it.
- Modifier-only and right-side-modifier hotkeys are refused with a new message, translated into all 11 locales: `"Control+Alt" needs the Windows key listener, which this install is missing. Choose a shortcut with a regular key, like F8.`
- `getEffectiveDefaultHotkey()` returns F8, as it already does on X11 and GNOME:
  - Startup registers F8.
  - Onboarding and Reset to default stop offering `Control+Super` and Right Ctrl.
  - Hotkey suggestions skip modifier-only combos.

Nothing changes when the listener is present, or on macOS and Linux. `get-hotkey-mode-info` now asks `supportsPushToTalk()` on Windows, as it already does on macOS. The old Windows branch always returned true.

## Test plan

- [x] Two new tests in `test/helpers/hotkeyActivationMode.test.js` (Windows, listener missing vs present):
  - Hold, refused hotkeys, the effective default, and suggestions.
  - Startup landing on F8 when the saved hotkey is modifier-only.
- [x] In `test/components/onboardingHotkeyPresentation.test.js`, the existing Windows assertion now passes `Control+Super`, and a new one covers the F8 default. Onboarding itself wasn't run on the VM.
- [x] All three fail on `main`.
- [x] `npm run quality-check` exits 0, and `scripts/check-i18n.js` passes. The hotkey tests (`test/helpers/hotkey*.test.js`, `test/components/*otkey*.test.js`: 92, 4 skipped) and locale tests (21) have 0 failures.
- [x] Windows 11 VM (ARM64, x64 app emulated, 200%): this branch's JS swapped into the official 1.10.2 install, on a fresh profile with no real microphone.
  - With the listener: holding `Control+Alt` in Hold mode starts and stops dictation.
  - Without it, restarting with that saved Hold + `Control+Alt`:
    - F8 is registered.
    - Hold is reset to Tap and shown disabled with its reason.
    - `Control+Alt` is refused with the new message, and picking Hold reverts to Tap.
    - Tapping F8 toggles dictation.

Not covered:
- A saved modifier-only hotkey on a build without the listener falls back to F8 at startup, but nothing tells the user F8 is now active, and Settings keeps showing the saved key.
- A modifier-only Voice Assistant, Translation or Meeting hotkey fails at startup with only a log line, the same as any other registration failure for those slots.
- Linux push-to-talk availability is unchanged.


stantheman0128:
Support. This is the right follow-up to #2147.

#2147 only makes the download and the pack step honest. A machine that still has no listener (dev, or a rate-limited download) should stop advertising Hold and Control+Alt.

Treating a missing listener as unsupported, so the existing X11/GNOME fallbacks apply, is better than a new Windows-only mode.

The leftover you already listed is the one I would feel on a real box. Startup silently switches to F8, but Settings still shows the saved combo. Not a merge blocker. Someone who already has Hold + Control+Alt saved will think the old key still works.

The two new tests failing on main is the right bar.

Issue #2232, opened 2026-09-17
Notes editor: support pasting images and tables (Markdown structures the schema currently drops)


shrijayan:
**Is your feature request related to a problem? Please describe.**

Taking notes during a meeting means capturing things that aren't words — a screenshot of a shared screen, a diagram someone draws, a table of numbers, a link. The note body in the Notes page only accepts the structures its editor schema knows about, so anything else is silently dropped.

**Images are not supported at all.** `src/components/ui/RichTextEditor.tsx` builds its extension list from `StarterKit`, `TaskList`/`TaskItem`, `Placeholder`, `tiptap-markdown` and the optional mention extension. There is no image node in the schema, and no paste handler on the notes path that looks at `clipboardData.files` (the only `onPaste` in `NoteEditor.tsx` is `handleTitlePaste`, which reads `text/plain` for the title field). So pasting a screenshot into a note does nothing: no error, no placeholder, the paste just vanishes. Drag-and-drop of an image file onto the editor is dropped the same way.

**Markdown is supported only partially.** Plain-text Markdown does paste correctly (`Markdown.configure({ transformPastedText: true })`) and notes are persisted as Markdown via `getMarkdown()`, so this is closer than "unsupported" — but the gaps are the ones that matter for meeting notes:

- **Tables are not in the schema.** Only the read-only `MarkdownRenderer` used by the assistant/chat renders tables, so a table pasted into a note, or one produced by a notes-generation action, collapses into plain pipe-delimited lines. This is the same underlying gap reported in #1948.
- Rich content copied from a browser, Google Docs, Slack or Notion is clipped to whatever the schema supports: bold, italics, links, lists and headings survive, tables and images are dropped, and nothing signals that anything was lost.
- There is no way to view or edit a note's raw Markdown source, even though Markdown is the format notes are stored in and the format the markdown mirror writes to disk.

**Describe the solution you'd like**

Images:
- Paste an image from the clipboard straight into the note body (screenshot → Cmd/Ctrl+V)
- Drag and drop an image file onto the editor
- Actually persist it, in a way that survives the Markdown mirror — e.g. written next to the note's mirrored `.md` with a relative `![](…)` reference, or embedded as a data URI
- A visible placeholder while the image is being stored, rather than a silent no-op

Markdown:
- Table support in the editor (render and edit), so generated notes containing tables keep their structure instead of losing it
- When content the editor cannot represent is pasted, keep it as literal text rather than discarding it
[...]

PR #2255, opened 2026-09-18
feat(notes): tables in the note editor, plus a formatting toolbar

### LINKED ISSUE #1948: Generated notes containing tables lose formatting even if copied to another app
**Describe the bug**
Generated notes can include tables, but the formatting (vertical bars) are stripped, so even if you paste the notes into another program, you have to recreate the formatting manually or have an AI do it.

**To Reproduce**
Generate notes containing a table

**Expected behavior**
Any one or more of the following would be better:
- The vertical bars are kept (so the formatting can be reconstituted more easily)
- Tables are rendered as code/literal blocks (unless that too is unsupported),
- The model is told not to use whatever formatting methods the editor doesn't support, and what to use instead, or
- The app provides a way to copy the original LLM output

- OS: Windows 10
- Version 1.9.2

boseq:
## Summary

Closes #1948. Covers the tables part of #2232; images are not included.

Notes stay Markdown, and tables are stored as GFM pipe tables. Nothing changes in the database, sync, the API, the Markdown mirror, or the notes.openwhispr.com viewer, which already renders GFM tables.

Before this change the editor had no table node. Any table in a note was flattened into one run-on paragraph (`NameOwnerBudgetAna…`) the first time the note was edited on desktop. That covers tables written on mobile, through the API/MCP, by the assistant, or by a note action.

**Tables in the editor.** Tables render and are editable. The editor shows only what a GFM pipe table can store:
- a header first row
- one line of text per cell
- no merged cells
- no column alignment or widths

`RichTextEditorTable.ts` enforces this:
- **Cells:** each holds a single paragraph. Pasted cells are flattened while they're parsed, so lists, headings or several paragraphs in a cell become one line instead of splitting the table. Alignment and `<colgroup>` widths are not read.
- **Normalizer:** runs after each change. It makes the first row the header, splits merged cells into a plain grid (the text stays in the first slot), and turns line breaks inside a cell into spaces, keeping bold or italic.
- **Serializing:** each cell renders in its own serializer state and `|` is escaped, including inside code. Cells render mid-line, so text like `-5%` or `1. Plan` is not escaped. Saving a 50×6 table takes about 0.5 ms, about the same as 300 plain paragraphs.
- **`---`:** stays text when typed in a cell. The horizontal-rule shortcut would otherwise split the table. Other block shortcuts (`- `, `1. `, `# `) stay text there too.

**Keyboard**
- **Tab / Shift+Tab:** move between cells; Tab on the last cell adds a row.
- **Enter:** moves to the cell below and adds a row from the last row. On an empty last row it removes that row and leaves the table, like Enter on an empty list item.
- **Shift+Enter / Mod+Enter:** do nothing inside a cell.
- **Creating a table:** type a header row such as `| Item | Cost |` and press Enter. Undo returns to the typed line. A line containing a line break or an @mention isn't converted.

**Paste and copy**
- Tables pasted as HTML get a header row. Merged cells keep their columns aligned. The tests use generic HTML tables; pastes from Google Docs, Sheets, Excel and Numbers still need the real-app check below.
- A table on its own, or cells copied from one, pasted into a cell goes to prosemirror-tables (see Known limits). Anything else pasted into a cell is flattened into one line without touching other cells: multi-line text, several paragraphs, several tables, or a table with other content around it. Pasted over several selected cells, that line goes into each of them. A paste with nothing to insert, such as a screenshot or a pasted rule, does nothing inside a table. This applies to pastes only; drops follow ProseMirror's usual rules (see Known limits).
- A Markdown table pasted as text into a paragraph becomes a table.
- Copying several cells puts a GFM table on the text clipboard.
- Copying text inside one cell, or a single selected cell (triple-click), puts only that cell's text on the plain-text clipboard, as Markdown like a copied paragraph (bold stays `**bold**`). This includes tables nested in a list or quote. An empty cell copies as a space.

**Table menu.** A `⋯` button appears beside the table's header row while the editor is focused and the caret is in the table. It sits above the table when there's no room beside it. It offers:
- insert row above / below
- insert column left / right (flipped for RTL tables)
- delete row / column / table

"Insert row above" is disabled in the header row. "Delete row" and "Delete column" are disabled when they would remove every row or column; use "Delete table" instead.

The menu is Tiptap's BubbleMenu with the shared Radix dropdown:
- The dropdown is portaled into the bubble menu's own element, because the bubble menu hides when focus leaves it. It isn't portaled into the editor's scroller: EditorContent moves that element's children when a note closes, which crashed a dropdown that was still open.
- `DropdownMenuContent` gains an optional `container` prop for this.
- The menu's BubbleMenu props are stable, so typing doesn't dispatch extra transactions.

**Formatting toolbar.** So a table isn't the only thing you can insert from a menu, one toolbar offers:
- bold, italic, strikethrough, code
- text style: normal text, heading 1–3
- bulleted list, numbered list, checklist, quote
- insert table (3 columns, a header row and two empty rows; the caret lands in the first header cell)

It appears in two places:
- above selected text (Tiptap `BubbleMenu`)
- beside the caret on an empty line (Tiptap `FloatingMenu`), so it covers no text

It's hidden mid-line, on a selected image or rule, inside tables, which keep their `⋯` menu, and in code blocks, where the marks don't apply and a list would take the fence apart. On an empty note it takes the placeholder's place while the editor is focused. Buttons show their shortcut in the title (`Bold (Cmd+B)`) and highlight when active.
- Both menus share `RichTextEditorMenus.ts` with the table menu: the focus check, the dropdown portal container, and the refocus-on-close handler.
- The same file hides a menu when focus leaves the editor and its menus. Tiptap misses two cases: a click on a menu swallows the editor's next blur, and tabbing out of a menu never blurs the editor. Without it, the toolbar stayed up after you clicked Bold and moved on. This applies to the table menu too. The hide waits a microtask: Chromium fires `focusout` while it removes a focused element, and hiding from there broke Tiptap's own removal of the menu, which threw and skipped the note's save (Delete table from the ⋯ menu, Insert table from the keyboard).
- Buttons use a native `title`, like most icon buttons in the app, not `<Tooltip>`. The menus detach the toolbar without unmounting it, which left a hovered button's tooltip stuck on screen.
- For the same reason the text-style dropdown is controlled and closes with its menu. A dropdown left open when the note changed from elsewhere (dictation, a note action) stayed open over a scroll-locked page that swallowed the next click. The table menu had this before this PR; both are fixed here.
- Nine formatting glyphs (bold, italic, strikethrough, heading, the three lists, quote, table) are drawn in `src/components/icons/primitives.tsx`, beside the status marks already drawn there, rather than vendored: `sync-nucleo-icons.js` needs a Nucleo license on the machine that runs it. The generated `index.ts` re-exports them, so a later vendor pass can replace them without touching the toolbar.

**Other changes**
- The two wrapper system prompts no longer forbid tables. They wrap custom note actions and the built-in Generate Notes. They now say "Use a table only when the instructions ask for one", so default output should stay table-free. Detailed Notes and Follow-up email are standalone prompts and are untouched.
- Cmd+K note previews drop delimiter rows and cell pipes.
- New direct dependencies, pinned to 3.31.3 like the other Tiptap packages. Only `@tiptap/extension-table` and happy-dom (with its own dependencies) are new to the lockfile; the others were already installed through other packages.
  - `@tiptap/extension-table`
  - `@tiptap/extension-horizontal-rule`
  - `@tiptap/pm`
  - `@tiptap/extension-bubble-menu` and `@tiptap/extension-floating-menu`: `@tiptap/react/menus` imports both, but `@tiptap/react` lists them only as optional `^` dependencies.
  - plus `happy-dom` as a dev dependency for the editor tests.

**Known limits**
- Column alignment markers (`:-:`) become plain `---` once a table is edited on desktop.
- Body cells beyond the header's column count are dropped on the first desktop edit. An LLM table with an unescaped `|` inside a cell can lose its last cell this way.
- A table, or cells copied from one (even a single triple-clicked cell), pasted into a table replaces the caret's cell and the ones after it, adding rows or columns as needed. That's prosemirror-tables' spreadsheet-style paste, and it can be undone.
- Pasting a bare URL over selected cells turns each cell's existing text into a link, because Tiptap's link paste runs first.
- Text that Markdown paste reads as a horizontal rule (`---`, `***`), pasted into a cell, inserts nothing. Typing it works, and so does pasting it as plain text.
- Dragging paragraphs, a list or a heading into a cell can add columns or split the table, because ProseMirror fits the drop at the drop point. Undo restores it and no text is lost.
- Merged cells and line breaks in pasted cells don't survive, and you see that at paste time.
- A literal `<br>` inside a cell, which LLMs sometimes emit, shows as text and is saved as `&lt;br&gt;`. HTML is handled the same way everywhere else in notes (`html: false`).
- A typed header row becomes plain-text labels, so formatting typed in it (bold, links, code) is dropped.
- A code span in a cell that contains a literal `\|` saves as `\\|`. markdown-it, which reads these notes back, restores it, but GitHub and remark-gfm split the cell there. Escaping the backslash instead would be worse: markdown-it reads `\\\|` as an extra backslash, so every save would add one. CodeQL flags the same line (`js/incomplete-sanitization` on `escapeTableCellPipes`) for that reason and needs dismissing.
- The table menu and the formatting toolbar are mouse-only, and Tab stays in the table. The formatting actions have keyboard shortcuts, and typing a header row starts a table.
- On an empty line flush against the bottom of the note's visible area, the toolbar is nudged up a few pixels to stay visible and can touch the line above.
- The mobile app shows tables as raw pipes. That's a follow-up in openwhispr-mobile, which also still tells its note model not to use tables.

## Test plan

- [x] `node --import tsx --test test/components/richTextEditor.test.js test/utils/markdownTable.test.js`: 49 tests.
  - 37 run the editor's extension list in happy-dom (the mention extension, which only `RichTextEditor` adds, isn't in it).
  - 9 render the real `RichTextEditor` with its menus (React + Radix, loaded through Vite SSR) the way NoteEditor uses them: mentions enabled, the note in state, and switching notes remounting the editor. They emulate Chromium's removal of a focused element (blur first, then detach), which happy-dom doesn't do, and check that the note's saved value keeps up. They cover:
    - running Delete table from the menu, which fails with `RangeError: No cell found` without the menu's in-table guard, and saves the change
    - the menu's disabled states in a header-only table
    - switching notes while the table menu is open. It fails with a `removeChild` DOMException when the dropdown is portaled into the scroller, and with a destroyed-editor error when the menu's close handler focuses the editor without checking.
    - where the formatting toolbar shows (selection, empty line) and hides (mid-line, in a table), with one toolbar at a time
    - formatting through it: bold, bulleted list, heading 2, insert table from the keyboard (saved, and not announced as a toggle)
    - the toolbar and the table menu hiding when focus leaves them, after a click on the toolbar and when tabbing out
    - where the toolbar stays away: code blocks, empty list items, a selected rule, text selected in a table, and a selection the editor doesn't hold focus for
    - a dropdown closing with its menu when the note changes from elsewhere, leaving the page clickable
    - switching notes with the text-style dropdown open, from each toolbar
  - 3 are unit tests for the parsing and escaping helpers.

  The editor tests cover:
  - round trips
  - #1948
  - pipe escaping, including bold or italic after a pipe
  - no escapes on cell text starting with `-`, `1.`, `+` or `#`
  - alignment and widths not shown
  - Enter / Shift+Enter / Mod+Enter
  - the typed header row and its undo
  - block shortcuts in cells
  - insert table on an empty line, with text selected, over a selection spanning two lines, and undoing it
  - pasted tables: header-less, merged, ragged, nested, with lists, with widths
  - multi-line text, several paragraphs (text, HTML, plain-text paste), two tables, and a table plus text pasted into a cell
  - a table on its own pasted into a cell filling cells from the caret
  - several paragraphs pasted over selected cells going into each as one line
  - an empty paste, or one that flattens to nothing like a rule, changing nothing over selected text or selected cells
  - a line break dropped into a cell becoming a space
  - a drop below a table landing the same whether or not the caret was in a cell
  - Markdown paste
  - row and column commands
  - copying cells, a single cell (including an empty one), and text inside a cell, top-level or nested in a list or quote
  - undo
- [x] `npm run quality-check`, `npm run i18n:check`
- [x] Full `npm test`: 4753 tests, 0 failures.
- [x] Headless Chrome harness:
  - the menu is hidden until the editor is focused, including in a note that starts with a table
  - it sits beside the header row, above a wide table, and inside the scroller when a wide table starts the note
  - the dropdown anchors to the button and the menu stays while it's open
  - the disabled states are right in a header-only and a one-column table
  - selected header cells highlight in dark mode
  - focus returns to the editor after an action or Escape
  - the menu hides on blur and scrolls with the table
  - one transaction per keystroke
  - dark mode
  - formatting toolbar: above a selection, on the line beside the caret on an empty line (between lines, after a table, pressing Enter at the bottom of the note), active states, the heading dropdown anchoring and returning focus, insert table handing over to the table menu, the empty-note placeholder, no stuck hover titles, hiding after a click then a click outside and after tabbing out, a click in a toolbar gap keeping the caret, no errors, flicker or missed saves when a menu hides with focus inside it, dark mode
- [ ] Real app on macOS, Windows and Linux:
  - paste from Google Docs, Sheets, Excel, Numbers, Notion and a web page, into a paragraph and into a cell
  - custom action "put the decisions in a table"
  - an @mention in a cell
  - switch notes with the table menu open
  - formatting toolbar in the real note view
- [ ] Sync round trip: share link renders the table; editing on mobile and syncing back keeps it


boseq:
## What it looks like

Captured from `main` (`6d56d75e`) and this branch (`67ae1a69`) with a fake note, rendering the real `RichTextEditor` through the app's Vite config.

### Tables

Same note, same Markdown. On `main` the table collapses into one run-on paragraph (#1948); here it renders and stays editable.

Note with a table: main versus this PR

### The formatting toolbar

Above selected text, beside the caret on an empty line, and the table's own `⋯` menu. In the first panel the text style button is lit because the selection is a heading. The toolbar floats over the line above the selection, which is what a bubble menu does.

Formatting toolbar on selected text, on an empty line, and the table menu

<details>
<summary>Dark mode</summary>

Note with a table: main versus this PR, dark

Formatting toolbar and table menu, dark

</details>

Chadpiha (approved):
Reviewed the table and formatting-toolbar flows, including Markdown preservation, paste/copy, mentions, note switching, and existing non-table behavior.

My follow-up commit, 1b0811e9, fixes menus remaining actionable after editing is disabled. It closes the menus, prevents refocusing a read-only editor, guards delayed callbacks after editor destruction, and adds three regression tests covering disabling and re-enabling.

Thanks @boseq !

PR #2267, opened 2026-09-20
feat(macos): start the menu-bar icon near system controls

greatcoat:
Give OpenWhispr's macOS menu-bar icon a best-effort starting position near the system controls, then retain the position the user chooses. The tray registers a default starting position before it is created and uses one stable macOS identity. Windows and Linux keep their existing behavior.

The initial-position preference is an undocumented AppKit convention, so exact adjacency to Wi-Fi or Search is not guaranteed. First adoption uses a new identity, so an icon a user had moved will likely move to the new starting position once (its old position was saved under AppKit's automatic name); positions saved after that take priority.

References: Electron documents the macOS `guid` as what keeps the icon's position across relaunches ([tray.md](https://github.com/electron/electron/blob/v41.10.5/docs/api/tray.md#new-trayimage-guid)); Electron lowercases the GUID before using it as the status item's autosave name ([electron_api_tray.cc](https://github.com/electron/electron/blob/v41.10.5/shell/browser/api/electron_api_tray.cc)); Hammerspoon writes the same `NSStatusItem Preferred Position <autosaveName>` key just before assigning `autosaveName` ([libmenubar.m](https://github.com/Hammerspoon/hammerspoon/blob/23e387e2805a9890066366e0ac96c71b27f0cfd5/extensions/menubar/libmenubar.m#L442-L453)).

## Validation

- An Electron 41.10.5 probe on a 1440 px wide display put an untreated tray at x=979 and one with the registered position at x=1194.
- The changed TrayManager, run under real Electron 41.10.5 on macOS 26.5.2 in an isolated harness: the icon started further right, and a seeded saved position survived two relaunches, including the before-quit/app.exit path.
- @greatcoat launched the full macOS dev build on 21 September and reported that it works. The Command-drag/relaunch sequence and an official installer build are not recorded.
- The checks above ran before the review follow-up below, which merged main (the adm-zip bump only), removed an unreachable catch, rewrote two comments, deleted the plan and spec, and trimmed the tests. Local run on Node 24 at the current head: quality-check, the i18n check and `npm audit --audit-level=high` pass; the full suite has 4,689 passed, 0 failed, 17 skipped and 1 todo.

## Review follow-up

- Merged main, which brings in the adm-zip 0.6.1 bump from #2261, so the dependency audit now passes.
- Removed the plan and spec docs. They held local paths, an internal Slack link and status lines that would go stale at merge; the facts worth keeping from them are in this description. Windows feasibility is a separate investigation, not part of this PR.
- Removed the try/catch around `registerDefaults`. Electron throws there only for a non-dictionary argument, a null value, or an NSUserDefaults exception, and none of those can come from this constant.
- The code comments now say why the GUID must stay lowercase, what position 0 means, and that the position default is undocumented AppKit behavior.
- Trimmed `trayPlacement.test.js` to the behavior this PR adds and the behavior it keeps: macOS registers the starting position before creating the tray under the fixed GUID and still ignores double-clicks, and Windows and Linux create it without either. The GUID is spelled out in the test, so changing it fails the test.


boseq (approved):
Approving and merging at 10ff6259. I merged main to pick up the adm-zip bump from #2261, so the audit passes now, and pushed two cleanup commits. The plan and spec docs are gone, since they had local paths and an internal Slack link, and so is the try/catch around `registerDefaults`, because Electron can't throw there for a constant `{ key: 0 }`. The comments now say why the GUID has to stay lowercase and what position 0 means, and `trayPlacement.test.js` is down to the macOS case and the unchanged Windows/Linux case. Full suite passes locally (4,689 passed, 0 failed) and CI is green.

Squash-merging so the older commits with the removed docs stay out of main. Thanks @greatcoat!

PR #2272, opened 2026-09-21
feat(windows): preserve tray identity in signed production builds

greatcoat:
Windows currently creates its tray icon without an explicit identifier. This gives signed production builds one permanent GUID so Windows can associate the user's tray preferences with the same icon across launches and updates. Visibility outside the hidden-icons menu and icon order remain controlled by Windows and the user; first adoption may require arranging the icon once.

The GUID applies only on Windows, in the production channel, when the packaged metadata carries the signed-build marker. The default packaging config sets that marker and enforces signing; the unsigned config clears both. Installed and portable production builds share the identity; unsigned builds and other channels keep their existing constructor behavior.

macOS #2267 is merged, and this branch now includes it: macOS keeps its own GUID, and the Windows GUID tests sit next to its test in `trayPlacement.test.js`. No dependency, IPC channel, or workflow changes.

Validation at `1fd96a2c`:

- Tray tests pass, and they fail if the GUID or any of its checks (platform, channel, marker) is removed.
- Both builder configs resolve as intended: the signed config gives marker `true` with forced signing, the unsigned one `false` for both.
- Prettier passes on every changed file.

**Native Windows acceptance remains unverified.** No signed candidate artifact or Windows device was exercised. Before the first release that includes this, verify the running executable's signature/publisher, Windows 10/11 x64 installer and portable update pairs signed on different days, changed paths, relaunch (including after a force-kill) and reboot, and preservation of both visible and deliberately hidden choices. Unsigned PR CI cannot establish those signed-build results. The design doc (`docs/superpowers/specs/2026-09-21-windows-tray-identity-design.md`) has the acceptance procedure.


boseq (approved):
Approving at `8230e253`. @greatcoat I merged main (the adm-zip fix that was failing `quality`) and pushed one commit after reviewing:

- Removed the afterPack signing guard and its test. electron-builder already fails a marked build that doesn't get signed, so the guard only caught deliberate opt-outs like a `signExts` rule excluding the exe, and it depended on builder internals.
- The marker is a boolean now, and the `isPackaged` check is gone since only packaged metadata has the marker.
- The tray tests cover just the GUID cases. Each one fails if the GUID or the platform, channel or marker check is removed.
- The plan doc is gone. Its PowerShell signature check moved into the design doc, and acceptance now asks for the update pair to be signed on different days (Trusted Signing rotates the cert daily) plus a force-kill relaunch.
- CLAUDE.md has a short Windows note: never change the GUID, and never run the `win-unpacked` a failed signed build leaves behind.

The code looks good to me.

boseq (approved):
Merged main again to pick up #2267, which conflicted in `createTray`. The Windows GUID is now an `else if` between the macOS GUID branch and the plain tray, so macOS keeps its position setup and Linux and other Windows builds still get the one-argument tray.

The #2267 test runs `tray.js` in a VM without `process.env`, so its Windows case would have failed on the channel check. I gave its harness a channel and the packaged metadata and moved the Windows GUID tests into it, so both GUIDs are pinned in `trayPlacement.test.js` and `windowsTrayIdentity.test.js` is gone.

Still approving at `1fd96a2c`.

